PR #51, opened 2016-12-15
HDRenderLoop: Add UV2 control

sebastienlagarde:


sebastienlagarde:
Update of the PR, less shader variant to be more practical

PR #62, opened 2016-12-20
Update the implementation of the procedural sky model

EvgeniiG:


Assets/ScriptableRenderLoop/HDRenderLoop/Sky/ProceduralSky/ProceduralSkyParameters.cs
```diff
@@ -6,101 +6,112 @@ namespace UnityEngine.Experimental.ScriptableRenderLoop
...
+        // public Shader atmosphericShader = null;
+        // public Shader occlusionShader   = null;
+        public float worldScaleExponent    = 1.0f;
+        public ScatterDebugMode debugMode  = ScatterDebugMode.None;
 
         // Camera   m_currentCamera;
 
         // UnityEngine.Rendering.CommandBuffer m_occlusionCmdAfterShadows, m_occlusionCmdBeforeScreen;
 
+        void Awake()
```
JulienIgnace-Unity: Maybe I'm missing something but isn't that some kind of default gradient that should be initialized once in the renderer? And then user can provide their own as an asset?

EvgeniiG: This is the way it works in the demo. But perhaps we should make it more flexible. :-)

PR #66, opened 2016-12-29
C# Updates to match new RenderLoop native side refactor.

stramit:
Update to the RenderPipes to support the new paradigm on how to go about writing a renderpipe.

Renderloops now should only contain configuration information + a call out to the rendering logic (preferably living in a static function). A loop now executes within a 'rendering context' that passes along a sidecar configuration file called a 'DataStore'. Any transient information that is needed between frames should be stored in this datastore (things like the materials / rendertextures). When the renderloop is destroyed this sidecar data is automatically cleaned up. It can also be cleaned manually.

Currently only the BasicRenderLoop has been ported to this new model due to the other loops not having a separation of concerns between transient data and configuration. They need the loop owners to detangle this before porting to the new model can take place.  These existing loops still work, but they suffer from the same lifecycle issues they have had up until this point.


Assets/ScriptableRenderLoop/core/RenderingDataStore.cs
```diff
@@ -0,0 +1,45 @@
+using UnityEngine.Experimental.Rendering;
+
+namespace UnityEngine.ScriptableRenderPipeline
+{ 
+	public class RenderingDataStore : IScriptableRenderDataStore
+	{
```
phi-lira: I think in most cases this implementation of RenderingDataStorage is exactly what SRL users would need. It seems pretty standard to the point I would expect it should be in the engine code IMO (making the interface not necessary). I understand the flexibility this brings but I think most SRL users will end up either copying this file to their projects or replicating something very close to this. As a SRL user I would expect to tell how my storage is built/cleanup and not worry about lifecycle management.

Assets/ScriptableRenderLoop/core/RenderingDataStore.cs
```diff
@@ -0,0 +1,45 @@
...
+				m_NeedsBuild = true;
+			}
+		}
+
+		protected virtual void InternalCleanup()
+		{}
+
+		public IRenderPipeline owner { get; private set; }
+
+		public T GetRealOwner<T>() where T : IRenderPipeline
```
vlad-andreev: I'd either get rid of this function or at least rename it to GetOwner<T>().  The current name implies some sort of behind-the-scenes shenanigans with the owner.

Assets/ScriptableRenderLoop/core/RenderPipeline.cs
```diff
@@ -0,0 +1,63 @@
...
+
+				return m_CameraProvider;
+			}
+			set { m_CameraProvider = value; }
+		}
+
+		public override void Render(ScriptableRenderContext renderContext, IScriptableRenderDataStore dataStore)
+		{
+			if (dataStore == null)
+				throw new ArgumentException(string.Format("DataStore has been passed into pipe {0}", this));
```
vlad-andreev: shouldn't the exception text be something like "Null DataStore passed to pipeline {0}"?

Assets/ScriptableRenderLoop/core/RenderPipeline.cs
```diff
@@ -0,0 +1,63 @@
...
+
+			if (dataStore.owner == null)
+				throw new ArgumentException(string.Format("DataStore owner is not set to {0}", this));
```
vlad-andreev: the exception text here doesn't seem to match the condition...

Assets/ScriptableRenderLoop/HDRenderLoop/HDRenderLoop.cs
```diff
@@ -34,6 +35,30 @@ static void AddAdditionalLightData()
...
+            protected override void InternalBuild()
+            {
+                base.InternalBuild(); 
+                HDRenderLoop realOwner = GetRealOwner<HDRenderLoop>();
+                realOwner.Build();
+            }
+
+            protected override void InternalCleanup()
+            {
+                base.InternalBuild();
```
stramit: yes, fixed

Assets/ScriptableRenderLoop/fptl/FptlLighting.cs
```diff
@@ -16,9 +17,28 @@ static void CreateRenderLoopFPTL()
...
+            protected override void InternalBuild()
+            {
+                base.InternalBuild();
+                FptlLighting realOwner = GetRealOwner<FptlLighting>();
+                realOwner.Build();
+            }
+            
+            protected override void InternalCleanup()
+            {
+                base.InternalBuild();
```
stramit: yes fixed

Assets/ScriptableRenderLoop/core/RenderingDataStore.cs
stramit: yeah. Killing it. Reduces API surface also

Assets/ScriptableRenderLoop/core/RenderPipeline.cs
stramit: Fixed

Assets/ScriptableRenderLoop/core/RenderPipeline.cs
stramit: Fixed

Assets/ScriptableRenderLoop/core/RenderPipeline.cs
```diff
@@ -0,0 +1,63 @@
...
+
+			if (dataStore.owner != null && !ReferenceEquals(dataStore.owner, this))
```
vlad-andreev: kind of minor, but dataStore.owner != null is redundant, as the previous line would have thrown an exception (and even if it didn't, ReferenceEquals() handles null anyway)

PR #81, opened 2017-01-06
HDRenderPipeline: Fix bug where we don't apply reflection when there is no light

sebastienlagarde:


EvgeniiG:
LGTM. Thanks for the fix!

PR #108, opened 2017-01-20
Initial macOS Metal support for HDRenderPipeline + some iOS fixes in FTPL

joravainen:
- Tweaked all relevant #pragma target and #pragma only_renderers to support metal.
- Created initial API/Metal.hlsl (Straight copy from D3D11 for now. Needs review later.)
- Worked around an include path handling bug in the Unity side on couple of shaders (will change them back after fixing the bug)
- Tweaked project settings for Metal
- Fixed a hidden issue in lightlistbuild-bigtile.compute that was revealed on iOS


Assets/ScriptableRenderLoop/ShaderLibrary/API/Metal.hlsl
```diff
@@ -0,0 +1,84 @@
...
+#define SAMPLE_TEXTURECUBE_ARRAY_LOD(textureName, samplerName, coord3, index, lod) textureName.SampleLevel(samplerName, float4(coord3, index), lod)
+#define SAMPLE_TEXTURE3D(textureName, samplerName, coord3) textureName.Sample(samplerName, coord3)
+#define SAMPLE_TEXTURE2D_SHADOW(textureName, samplerName, coord3) textureName.SampleCmpLevelZero(samplerName, (coord3).xy, (coord3).z)
+#define SAMPLE_TEXTURE2D_ARRAY_SHADOW(textureName, samplerName, coord3, index) textureName.SampleCmpLevelZero(samplerName, float3((coord3).xy, index), (coord3).z)
+#define SAMPLE_TEXTURECUBE_SHADOW(textureName, samplerName, coord4) textureName.SampleCmpLevelZero(samplerName, (coord3).xyz, (coord3).w)
+
+#define TEXTURE2D_HALF TEXTURE2D
+#define TEXTURE2D_FLOAT TEXTURE2D
+#define SAMPLER2D_HALF SAMPLER2D
+#define SAMPLER2D_FLOAT SAMPLER2D
```
sebastienlagarde: You maybe don't want this. If I get it, those define have been added for metal, right ?

Assets/ScriptableRenderLoop/fptl/Internal-DeferredComputeShading.compute
```diff
@@ -25,7 +25,7 @@ float3 EvalIndirectSpecular(UnityLight light, UnityIndirect ind);
 #include "TiledLightingTemplate.hlsl"
 #include "TiledReflectionTemplate.hlsl"
 
-Texture2D _CameraDepthTexture;
+Texture2D_float _CameraDepthTexture;
```
sebastienlagarde: You will need to do the same thing for HDRenderLoop deferred shader, use TEXTURE2D_FLOAT / SAMPLER2D_FLOAT

Assets/ScriptableRenderLoop/ShaderLibrary/API/Metal.hlsl
joravainen: Yes, this is just the first quick bootstrap version of the file. As commented in the beginning of the file, all of this should be reviewed and tweaked to actually suite metal. These indeed will need to be removed.

PR #122, opened 2017-01-26
Refactor hdloop

stramit:
*HDRender loop now respects correct Loop Instance workflow 
*No common settings / sky settings on MB, now a singleton accessed via CommonSettingsSingleton / SkyParamatersSingleton. This can override the settings on the HDPipelineAsset
*Refactor SkyParamaters / SkyRenderer to return an instance for use / rendering (no need to cache type and use reflection / activator (yay!)
*Add config window to HDRenderLoop menu. Can be used to set sky / common settings override + create overrides in the project. 

----------------------
Still to do: Some of the shadow settings seem a little odd (one struct used for data storage AND runtime, needs to be tidied up).


Assets/ScriptableRenderLoop/HDRenderPipeline/Editor/HDRenderPipelineInspector.cs
```diff
@@ -269,6 +307,7 @@ public override void OnInspectorGUI()
                 return;
 
             DebugParametersUI(renderContext);
+            CommonSettingsUI(renderContext);
```
JulienIgnace-Unity: I'm not a fan of putting back the CommonSettings into the HDRenderPipeline editor. It's a per scene setting and the HDRenderpipeline is pretty much unique to the project so I'm not sure it makes sense to put it there. Besides, the parameters in the HDRenderPipe asset are pretty much only debug parameters and engine settings that should never be touched by an artist and are in no way related to a specific scene.
Is this supposed to be a shortcut for editing the CommonSettings asset? If so I'm not sure it adds something to just editing the asset itself.

stramit:
```
I'm not a fan of putting back the CommonSettings into the HDRenderPipeline editor. It's a per scene setting and the HDRenderpipeline is pretty much unique to the project so I'm not sure it makes sense to put it there.
```
I guess the question here is: Do we have default settings? If so then common settings can just use these defaults instead of needing them specified in the asset. It does make sense to me that the loop is able to specify defaults that are then overridden by a scene if needed. 

```
Several settings currently on HDRenderPipeline asset are related to the lightloop itself which ideally could be replace by another one. Mean all the option like tile/cluster etc... should have their own settings loaded if this particular lightloop is loaded and in this case it will be indeed global. (I will try to do that with current version)
```
Yes this is true, the 'Light Loop' itself should become an asset that is referenced by the render pipeline so it can be delegated to just like how the sky paramaters now work.

PR #170, opened 2017-02-21
Fix depth offset for POM

EvgeniiG:
Not sure whether I placed the 0.1 scaling factor in the best place.
Additionally, for height parity, changes need to be made to the tessellation shader (please see the TODO).


Assets/ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
```diff
@@ -398,6 +399,16 @@ void Resize(Camera camera)
 
         public void PushGlobalParams(HDCamera hdCamera, ScriptableRenderContext renderContext, SubsurfaceScatteringSettings sssParameters)
         {
+            var cmd = new CommandBuffer {name = "Push Global Parameters"};
+
+            cmd.SetGlobalVector("_ScreenSize",        hdCamera.screenSize);
```
sebastienlagarde: Caution: These settings are per pass frequency. It is true that's they will not vary a lot during a frame, but any offline render target could potentially change them (like shadow). I would advice to setup them at each view matrix manipulation or each pass. I will think more about it.

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
```diff
@@ -270,18 +272,20 @@ void ApplyPerPixelDisplacement(FragInputs input, float3 V, inout LayerTexCoord l
...
             ppdParam.uv = layerTexCoord.base.uvXY;
             viewDirTS = float3(V.z > 0.0 ? uvXY : -uvXY, V.z);
             numSteps = (int)lerp(_PPDMaxSamples, _PPDMinSamples, viewDirTS.z);
-            float2 offsetXY = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, maxHeight, ppdParam);
+            float2 offsetXY = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, maxHeight, ppdParam, height);
 
             layerTexCoord.base.uvXY += offsetXY;
             layerTexCoord.details.uvXY += offsetXY;
+
+            height = 0.0; // TODO
```
sebastienlagarde: I think the height will be the triplanar weight of all three hieght. i.e   height = dot(triplanarWeights, float3(heightZY, heightXZ, heightXY);

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
```diff
@@ -298,13 +302,21 @@ void ApplyPerPixelDisplacement(FragInputs input, float3 V, inout LayerTexCoord l
 
             float3 viewDirTS = isPlanar ? float3(uvXZ, V.y) : TransformWorldToTangent(V, worldToTangent);
             int numSteps = (int)lerp(_PPDMaxSamples, _PPDMinSamples, viewDirTS.z);
-            float2 offset = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, maxHeight, ppdParam);
+
+            // TEMP: 0.1 to achieve parity between tessellation and POM w.r.t. height for a single tile.
```
sebastienlagarde: don't do it here but when we get maxHeight above, so it will apply to triplanar too

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
```diff
@@ -298,13 +302,21 @@ void ApplyPerPixelDisplacement(FragInputs input, float3 V, inout LayerTexCoord l
...
+            // TODO: for tiling, the max. height of the tessellated height map should be reduced by NumTiles.
+            float2 offset = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, 0.1 * maxHeight, ppdParam, height);
 
             // Apply offset to all UVSet0 / planar
             layerTexCoord.base.uv += offset;
             layerTexCoord.details.uv += isPlanar ? offset : _UVDetailsMappingMask.x * offset; // Only apply offset if details map use UVSet0 _UVDetailsMappingMask.x will be 1 in this case, else 0
         }
+
+        // Since POM "pushes" geometry inwards (rather than extrude it), { height = height - 1 }.
+        return height * maxHeight - maxHeight;
```
sebastienlagarde: ah, I think I get why, it is because V is the View vector for lighting, i.E like L it start from the pixel world position to viewver, so we should not remove height, but rather invert V vector when we do worldpos -V * height

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
```diff
@@ -975,17 +988,20 @@ void ApplyPerPixelDisplacement(FragInputs input, float3 V, inout LayerTexCoord l
             // TODO: do we support object triplanar ? See ComputeLayerTexCoord
             float3 viewDirTS = isPlanar ? float3(-V.xz, V.y) : TransformWorldToTangent(V, worldToTangent);
             int numSteps = (int)lerp(_PPDMaxSamples, _PPDMinSamples, viewDirTS.z);
-            float2 offset = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, maxHeight, ppdParam);
+
+            // TEMP: 0.1 to achieve parity between tessellation and POM w.r.t. height for a single tile.
+            // TODO: for tiling, the max. height of the tessellated height map should be reduced by NumTiles.
+            float2 offset = ParallaxOcclusionMapping(lod, _PPDLodThreshold, numSteps, viewDirTS, 0.1 * maxHeight, ppdParam, height);
```
sebastienlagarde: same comment, move to getMaximum()

Assets/ScriptableRenderPipeline/ShaderLibrary/Common.hlsl
```diff
@@ -446,45 +456,48 @@ void UpdatePositionInput(float depthRaw, float depthVS, float3 positionWS, inout
...
 }
 
 // depthOffsetVS is always in the direction of the view vector (V)
-void ApplyDepthOffsetPositionInput(float3 V, float depthOffsetVS, inout PositionInputs posInput)
+void ApplyDepthOffsetPositionInput(float3 V, float depthOffsetVS, float4x4 viewProjMatrix, inout PositionInputs posInput)
 {
-    posInput.depthVS += depthOffsetVS;
-    // TODO: it is an approx, need a correct value where we use projection matrix to reproject the depth from VS
-    posInput.depthRaw = posInput.positionCS.z / posInput.depthVS;
+    posInput.depthVS -= depthOffsetVS;
```
sebastienlagarde: So here we add negate because we have done a negative height with height -1, right ? Thus why I suggest to go with -V * depthOffsetVS

Assets/ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
sebastienlagarde: For now, let it like that

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
EvgeniiG: I simply followed the convention: "depthOffsetVS is always in the direction of the view vector (V)".
Regarding the proposed change, I am not sure which function you are referring to, but if it's ApplyDepthOffsetPositionInput(), then this will not give correct results, since the world position will be displaced for the height value of 1 (which corresponds to the top plane of the height map).

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
EvgeniiG: Moved to ParallaxOcclusionMapping() instead. Can't apply it to 'maxHeight' everywhere, otherwise depth values are going to be wrong.

Assets/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/LitData.hlsl
EvgeniiG: Done.

PR #185, opened 2017-03-14
Fix single pass loop on PS4

EvgeniiG:
#1: Fix the detection of orthographic projection, and therefore the viewing direction when looking straight down.

#2: Fix flickering issues on PS4 for the single pass loop (+ initial fixes for FPTL).

Flickering is caused by the delay between values in the constant buffer (globals) and values in structured buffers becoming visible in the Deferred shader. It appears that the update latency (the number of frames between setting the value in C# and being able to access it in a shader) varies depending on the type of buffer. Therefore, moving globals into a structured buffer provides a workaround for the single pass loop.

Doing the same for FPTL does not appear to be sufficient. Could be that the additional compute pass introduces extra frame(s) of latency, or that globals from ShaderVariables.hlsl need to be moved as well.

Globals from Clustered appear to be set once for multiple shaders, which makes implementing a similar workarond even more complicated.

### DIFF
--- a/Assets/ScriptableRenderPipeline/HDRenderPipeline/Lighting/Resources/Deferred.shader
+++ b/Assets/ScriptableRenderPipeline/HDRenderPipeline/Lighting/Resources/Deferred.shader
@@ -48,7 +48,7 @@ Shader "Hidden/HDRenderPipeline/Deferred"
             #pragma shader_feature _ SSS_PRE_SCATTER_TEXTURING SSS_POST_SCATTER_TEXTURING
          // #endif
 
-            #pragma shader_feature SKY_LIGHTING
+            #pragma multi_compile _ SKY_LIGHTING
 
             #pragma multi_compile _ LIGHTING_DEBUG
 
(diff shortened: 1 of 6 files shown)


Assets/ScriptableRenderPipeline/HDRenderPipeline/Lighting/Resources/Deferred.shader
```diff
@@ -48,6 +48,8 @@ Shader "Hidden/HDRenderPipeline/Deferred"
             #pragma shader_feature _ SSS_PRE_SCATTER_TEXTURING SSS_POST_SCATTER_TEXTURING
          // #endif
 
+            #pragma shader_feature SKY_LIGHTING
```
JulienIgnace-Unity: I'm pretty sure this won't work in player as shader features are not changeable at runtime?

EvgeniiG: Good point, fixed. Thanks!

PR #189, opened 2017-03-21
Lowendmobile

phi-lira:
Merging Lowend mobile into master. 

Lowend mobile is rather isolated. Only shaded changes with other pipelines is on the material upgrader.


Assets/LowEndMobilePipeline/Shaders/LowEndMobilePipeline.shader
```diff
@@ -0,0 +1,282 @@
...
+                Emission(i, emissionColor);
+
+                // Compute direct contribution from main directional light.
+                // Only a single directional shadow caster is supported.
+                LightInput mainLight;
+                INITIALIZE_LIGHT(mainLight, 0);
+
+                half3 viewDir = i.viewDir.xyz;
+
+                half3 directColor = EvaluateOneLight(mainLight, diffuse, specularGloss, normal, i.posWS, viewDir);
```
sebastienlagarde: you seems to not test if the lighting is per vertex. Is it expected ? vertex lighting should output a color directly no ?

Assets/LowEndMobilePipeline/LowEndMobilePipeline.cs
```diff
@@ -0,0 +1,398 @@
...
+using UnityEngine.Rendering;
+
+namespace UnityEngine.Experimental.Rendering.LowendMobile
+{
+    public class LowEndMobilePipeline : RenderPipeline
+    {
+        private readonly LowEndMobilePipelineAsset m_Asset;
+
+        private static readonly int kMaxCascades = 4;
+        private static readonly int kMaxLights = 8;
```
phi-lira: You're right. For some reason I thought I was limiting that in while processing the vertex lights but just checked not and I'm not. I'll update it.

Assets/LowEndMobilePipeline/LowEndMobilePipeline.cs
```diff
@@ -0,0 +1,398 @@
...
+
+                // Render Shadow Map
+                bool shadowsRendered = RenderShadows(cull, context);
+
+                // Draw Opaques with support to one directional shadow cascade
+                // Setup camera matrices
+                context.SetupCameraProperties(camera);
+
+                var cmd = new CommandBuffer() { name = "Clear" };
+                cmd.ClearRenderTarget(true, true, Color.black);
```
phi-lira: It's true that the game will probably cover the whole screen, but on tiled GPUs, if the framebuffer is not invalidated or cleared the tile will need to be initialized. That offchip memory access is worse than clearing. Best is to invalidate the buffer but glInvalidateFramebuffer is not available in GL ES 2.0

Assets/LowEndMobilePipeline/Shaders/LowEndMobilePipeline.shader
phi-lira: You're right. If no pixel light is supported this will be wrong. I'll fix it.

PR #191, opened 2017-03-22
Add lod cross fade

sebastienlagarde:


tlaedre:
shouldn't this also be implemented in the shadow casting pass?

PR #220, opened 2017-04-11
Implement projector lights

EvgeniiG:


EvgeniiG:
Increased the number of area and projector lights on screen on 64.

PR #256, opened 2017-05-11
HDRenderPipeline: Add default material for HDRenderPipeline

sebastienlagarde:


sebastienlagarde:
This PR add default material and default shader + it fix an issue with the diffuse ligthing mode and SSS.

PR #269, opened 2017-05-23
Initial implementation of SSAO

keijiro:
This is a pull request for the initial implementation of SSAO for HDRenderPipeline.

This implementation is almost identical to the current implementation in the post-processing stack (v1). It modifies RT1:A (specular occlusion) and RT3:RGB (emissive buffer) after building GBuffer (more precisely, right after copying depth buffer). At the moment, the settings of SSAO are stored in the common settings (per-scene settings). This means that it can't be changed on a per-camera basis. The custom editor for the settings hasn't been ready yet because it needs an overhaul on CommonSettings.

The source code is stored in the "HDRenderPipeline/ScreenSpace" directory. The idea is that it's to store some screen-space effects that are applied during scene rendering (different from post-processing effects that are applied in the end of scene rendering).

Things not yet done / not yet under consideration:
- Is it possible to apply SSAO without influencing emissive color?
- Is simply applying SSAO to specular occlusion right?
- Per-camera settings?
- Async compute.


sebastienlagarde:
Hey Kejiro, thanks for the PR!, 
few answer before I do the review:

Settings of SSAO should be in their own SSAO settings. We can create them later. You can stay in common for now.

>The source code is stored in the "HDRenderPipeline/ScreenSpace" directory. 
SSAO / SSR are lighting, move it to 
HDRenderPipeline/Lighting/ScreenSpace please
thanks

Also please, can you output SSAO into a dedicated texture ? And not touch the Gbuffer ?
I want to sample it for various usage (see below). Just output the texture and I will deal with the integration in the other shader.

Extra question, can you do a depth only variant (to be use for forward only path) ? Can be done in a later PR. And yes, quality will suffer, and this is accepted.
Super bonus: We have object in deferred renderer that will be render in forward only (complex material) and will be in prepass depth but have no normal. In this case SSAO buffer should not use the normal either. The stencil is already tagged for this case, it should be 0. (Also note that SSAO don't need to be computed for the sky, sadly sky is also tagged as stencil 0...). Overhall it mean that we should do a SSAO pass with depth + normal based on stencil, and another one base on depth only again with stencil test.
and this output one texture.

Main reason:
- We need to be GBuffer agnostic. We don't know what is inside the Gbuffer (only lit material know)
- Note that in your SSAO shader you must use our FETCH_GBUFFER macro to get your normal map and bind all render target. You can't assume the position of the element. Shader Compiler will optimize out buffer not use.
- I want to be able to apply a small percentage on direct light (not PBR but looks better when apply like 20% on foliage). Just want to expose a control:  SSAO influence strenght for direct lighting
- I need a more complex formula for spec occlusoin derive from SSAO based on roughness and view vector (see below)

>Is it possible to apply SSAO without influencing emissive color?
Currently no. I was thinking that we could render emissive in forward in future but it have also a cost when you just have an emissive LED on a character :). It is not a problem. In practice (this is what we have done at frostbite) it work correctly as emissive is not so usual on regular objects with AO (think about glow sphere).

I will do the application of SSAO during indirect lighting pass instead (Coherency for forward and deferred path)

> Is simply applying SSAO to specular occlusion right?
no. I want to use the Triace Gotanda trick here =>

float specularOcclusion	= computeSpecularOcclusion(NdotV, SSAO,  roughness);

float computeSpecularOcclusion(
	float NdoV, 
	float SSAO, 
	float roughness)
{	return saturate(pow(NdoV + SSAO, exp2(-16.0f * roughness - 1.0f)) - 1.0f + SSAO);
}

This must be done in Lit material (EvalEnv fonction)

> Per-camera settings?
No. We just want to be able to enable/disable SSAO per camera

>Async compute
Can't be done yet, PS4 async compute PR not landed. But yes, utimately will be done during shadow pass in async.

I am open to any discussion :)

sebastienlagarde:
I will wait for an update of the PR, most remark about the code itself are in it
- use FETCH_BUFFER macro
- Don't blend into gbuffer, output a SSAO texture instead
Later
- Variant SSAO without normal
- SSAO in two stenciled pass

I think we will tag forward only object with another stencil value so we can avoid SSAO for the sky.

keijiro:
Thanks for the informative and instructive feedback! I've started working on update (it will take a little time).

PR #275, opened 2017-05-31
SSAO revisited for HDRP

seblagarde:
This is currently a no working version of SSAO. It have all the skeleton (settings, shader etc...) except that the shader doesn't work. I implemented most of the correction I asked to Keijiro.

Several things:
- In TilePassLopp.hlsl I comment the SSAO because I get black value when  SSAO is disable despite having a white 1x1 texture :(
 context.ambientOcclusion = 1; // LOAD_TEXTURE2D(_AmbientOcclusionTexture, posInput.unPositionSS).x;
- Don't know if original code from Keijiro was working, I may have break stuff but the downsampling is not working at all
- Also the AO is inverted in my case, so I need to call 1 - AO. This isn't expected and why I think I break the code somewhere, (haven't checked on the original implementation).
- AO get is not really what I may expect. Also it is not pure white where it should.

=> Short: Keijiro can you take a look at it and see why the SSAO is so bad. Thanks


sebastienlagarde:
Note that I have done the modification quickly and haven't investigate too much about problem as it is already late here (have done the update during a night).

keijiro:
I tried to fix some issues (downsampling, inverted AO, etc.). Here is the diff: https://github.com/keijiro/ScriptableRenderLoop/commit/48132ea267d364d82e0e7ae13137ff6da7d597fe

sebastienlagarde:
Thanks Keijiro, I have update the code. Now everything work as expected. Feel free to tweak the SSAO when you want. Cheers

PR #286, opened 2017-06-06
Update material framework

sebastienlagarde:


sebastienlagarde:
All material (Lit, Unlit) now derive from a base class:
RenderPipelineMaterial
previously namespace was use to material name, now it is a regular class
It allow to add more easily new material or replace current one
HDRenderPipeline with introspection will execute all material method
from found RenderPipelineMaterial
The unique deferred material that we suppose is still a special case as
several code depends on it

PR #294, opened 2017-06-12
Add the old SSS model back

EvgeniiG:


Assets/ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
```diff
@@ -211,6 +217,15 @@ public HDRenderPipeline(HDRenderPipelineAsset asset)
 
             m_FilterAndCombineSubsurfaceScattering = Utilities.CreateEngineMaterial("Hidden/HDRenderPipeline/CombineSubsurfaceScattering");
 
+            // Old SSS Model >>>
+            m_CameraFilteringBufferRT = new RenderTargetIdentifier(m_CameraFilteringBuffer);
+            m_FilterSubsurfaceScattering = Utilities.CreateEngineMaterial("Hidden/HDRenderPipeline/CombineSubsurfaceScattering");
+            m_FilterSubsurfaceScattering.DisableKeyword("SSS_FILTER_HORIZONTAL_AND_COMBINE");
+            m_FilterSubsurfaceScattering.SetFloat("_DstBlend", (float)BlendMode.Zero);
```
sebastienlagarde: Don't you need to set the blend mode for m_FilterAndCombineSubsurfaceScattering  too ? as it is now a parameter, may be hazardous to rely on default value.

EvgeniiG: I do set it. :-)

PR #362, opened 2017-08-01
Add XR Support to BasicRenderPipeline

robbiesri:
This PR uses the new XR APIs in ScriptableRenderContext and CullResults to add XR rendering to BasicRenderPipeline.  This is a simple example of how to add XR support. Porting of other Render Pipelines are likely to need more work.

There are three main changes:
- Usage of new SRP APIs to set up stereo properties (matrices, viewports, RT) and perform stereo rendering.
- Add usage of existing stereo shader infrastructure (from UnityCG.cginc) to BasicRenderPipelineShader.shader
- Add the option of rendering to an intermediate render texture, and then blitting to the final output.  This is to demonstrate the setup of intermediate XR render textures, which is different than setting up a monoscopic render texture.

On top of the SRP and shader changes, Single-Pass Stereo needs to be enabled via the Player Settings.  Any single pass mode should work on any platform that supports Single-Pass Stereo.

### DIFF
--- a/Assets/ScriptableRenderPipeline/BasicRenderPipeline/BasicRenderPipeline.cs
+++ b/Assets/ScriptableRenderPipeline/BasicRenderPipeline/BasicRenderPipeline.cs
@@ -58,7 +58,62 @@ public static class BasicRendering
 {
     // Main entry point for our scriptable render loop
 
-    public static void Render(ScriptableRenderContext context, IEnumerable<Camera> cameras, bool useIntermediateBlit)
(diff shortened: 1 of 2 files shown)


robbiesri:
I could also gate the VR code behind an asset option similar to what I did with UseIntermediateRenderTargetBlit in BasicRenderPipeline.  But the code isn't activated unless XR is active, so it should be fine.

robbiesri:
This is good to go, but Tim suggested I work on adding RenderPass support as an example in here as well.  Agreed, but I need to do engine-side work first.

PR #390, opened 2017-09-06
Added async compute to fptl light list build. 

rthompson-unity:
Also removed Compute Deferred Lighting from fptl and the associated shader that was not in use and causing compile errors on ps4.


mmikk:
Spoke with Rob. A few more things to remove/clean. But looking good.

rthompson-unity:
Made some more deletions as suggested by Morten following the removal of compute deferred lighting.

PR #398, opened 2017-09-13
Branch deferreddirshadow

JulienIgnace-Unity:
- Split directional shadow resolution in a separate pass to reduce VGPR pressure during lighting.
- Optimized cascade shaow with blend for VGPR (a compilation issue was causing VGPR pressure to double)


ScriptableRenderPipeline/HDRenderPipeline/Lighting/TilePass/TilePass.cs
```diff
@@ -2042,6 +2078,8 @@ public struct LightingPassOptions
 
                                 cmd.SetComputeBufferParam(deferredComputeShader, kernel, HDShaderIDs.g_vLightListGlobal, bUseClusteredForDeferred ? s_PerVoxelLightLists : s_LightList);
 
+                                cmd.SetComputeTextureParam(deferredComputeShader, kernel, HDShaderIDs._DeferredShadowTexture, deferredShadowTexture);
```
sebastienlagarde: Do a SetGlobalTexture for the forward shader case and for the pass that don't use computeEvaluation

ScriptableRenderPipeline/HDRenderPipeline/ShaderVariables.hlsl
```diff
@@ -192,6 +192,8 @@ CBUFFER_END
 TEXTURE2D_FLOAT(_MainDepthTexture);
 SAMPLER2D(sampler_MainDepthTexture);
 
+TEXTURE2D(_DeferredShadowTexture);
```
sebastienlagarde: Move this to tilepass.hlsl

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -1055,7 +1055,11 @@ void EvaluateBSDF_Directional(LightLoopContext lightLoopContext,
 
     [branch] if (lightData.shadowIndex >= 0)
     {
+#ifdef SURFACE_TYPE_TRANSPARENT
```
sebastienlagarde: Note: in case of deferred .shader, SURFACE_TYPE_TRANSPARENT will not be define. So it works, but it is not really clean from my point of view. Don't know if this can be more clear

PR #404, opened 2017-09-18
Branch optimaterialfeatureflag

JulienIgnace-Unity:
Material Feature flags are now also used to statically remove code that is not needed in case there are 2 or more different materials in the same tile (thus preventing us from statically knowing the materialID)

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
+++ b/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
@@ -89,6 +89,14 @@ CBUFFER_END
 // If a user do a lighting architecture without material classification, this can be remove
 #include "../../Lighting/TilePass/TilePass.cs.hlsl"
 
+static int g_FeatureFlags = 0xFFFFFFFF;
+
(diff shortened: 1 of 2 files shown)


ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -489,10 +499,10 @@ void DecodeFromGBuffer(
     // The material features system for material classification must allow compile time optimization (i.e everything should be static)
     // Note that as we store materialId for Aniso based on content of RT2 we need to add few extra condition.
     // The code is also call from MaterialFeatureFlagsFromGBuffer, so must work fully dynamic if featureFlags is 0xFFFFFFFF
-    int supportsStandard = (featureFlags & MATERIALFEATUREFLAGS_LIT_STANDARD) != 0;
-    int supportsSSS = (featureFlags & MATERIALFEATUREFLAGS_LIT_SSS) != 0;
-    int supportsAniso = (featureFlags & MATERIALFEATUREFLAGS_LIT_ANISO) != 0;
-    int supportClearCoat = (featureFlags & MATERIALFEATUREFLAGS_LIT_CLEAR_COAT) != 0;
+    int supportsStandard = (g_FeatureFlags & MATERIALFEATUREFLAGS_LIT_STANDARD) != 0;
```
sebastienlagarde: use HasMaterialFeatureFlag

PR #418, opened 2017-09-26
SSS: Added thickness map and adjusted settings for Perry head example

Invertex:


sebastienlagarde:
As explain in other PR we only merge in 2017.3

PR #419, opened 2017-09-27
SSS: Added thickness map and adjusted settings for Perry head example

Invertex:


sebastienlagarde:
Thanks for this update. We only take the change on 2017.3 as this branch is regularly merge to trunk.

EvgeniiG:
Thanks for the thickness map!

PR #427, opened 2017-09-28
Branch depth prepass

JulienIgnace-Unity:
Properly implemented depth prepass for the deferred renderer: Now the main GBuffer pass will use DepthEqual test and won't write depth anymore if a prepass is done. This allows alpha-tested geometry to benefit from HiZ on GCN hardware (any shader with a clip instruction can't use it).
In deferred renderer, depth prepass can optionnaly be limited to Alpha tested geometry.
Also added Debug window entries for the depth prepass and tile lighting control.


ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
```diff
@@ -1087,6 +1157,10 @@ void RenderDepthPrepass(CullResults cull, Camera camera, ScriptableRenderContext
             // Note that an object can't be in both list
             bool addForwardOnlyOpaqueDepthPrepass = !m_Asset.renderingSettings.ShouldUseForwardRenderingOnly() && !m_Asset.renderingSettings.useDepthPrepassWithDeferredRendering;
 
+            RenderQueueRange renderQueueRange = RenderQueueRange.opaque;
```
sebastienlagarde: move after test below

PR #453, opened 2017-10-06
LWPipe dead code removal

Chman:
On-going mission to remove all warnings from this repository. Note that the auto formatting tool also removed EOL spaces.


SampleScenes/OnTileDeferredTest/Common/MobileTestSceneManager.cs
```diff
@@ -25,7 +25,7 @@ public class MobileTestSceneManager : MonoBehaviour {
 	int detectSwipe() {
 		if (Input.touchCount == 1 && Input.GetTouch (0).phase == TouchPhase.Ended) {
 			Vector2 endPosition = Input.GetTouch (0).position;
-			Vector2 delta = endPosition - startPosition;
+			//Vector2 delta = endPosition - startPosition;
```
phi-lira: What's the reason to commit this line commented?

Chman: Dead code, `delta` isn't used anywhere and as such it's throwing a warning. I've commented it instead of removing it in case it's needed later in that block. This file is part of a test anyway.

PR #455, opened 2017-10-09
Use LOAD_TEXTURE2D to get depth from depth pyramid, fixed roughness t…

fredpointzero:
Removed the sampler for depth pyramid and usage of LOAD_TEXTURE2D_LOD to fetch the depth of pixels for refraction

Use correct mip map count of color pyramid when computing the mip level from roughness


ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -1498,6 +1497,13 @@ void EvaluateBSDF_Area(LightLoopContext lightLoopContext,
 // EvaluateBSDF_SSL
 // ----------------------------------------------------------------------------
 
+float PerceptualRoughnessToMipmapLevel(float perceptualRoughness, uint mipMapCount)
```
sebastienlagarde: why do you have this function here ?

sebastienlagarde: ah ok, move it to imagebasedLigthing   PerceptualRoughnessToMipmapLevel

and have PerceptualRoughnessToMipmapLevel call it

PR #466, opened 2017-10-11
Revamp of Parallax occlusoin mapping + depth offset

sebastienlagarde:
This PR update the behavior of Parallax occlusion mapping to correctly match tessellation behavior.

The important change is about the displacement mode. Now there is a displacement mode where users can select: vertex, pixel or tessellation displacement, they are all exclusive.

Tiling and object scale are correctly take into account. And there is an option to coupled them to the displacement or not. It also introduce the concept of primitive size as the Unity default plan is a size of 10x10 meter (Quad is 1x1) that need to be taking into account for POM.

POM and depth offset work with all mapping type (planar, triplanar) and with any kind of tiling of objects scaling. It match tessellation that also have been update to work correctly with tiling and scaling.

This PR also add a shader keyword for influence mask and remove object scale from interpolator

Note: POM doesn't work with LayeredShader for now.


ScriptableRenderPipeline/HDRenderPipeline/Material/LayeredLit/Editor/LayeredLitUI.cs
```diff
@@ -66,6 +66,8 @@ private class StylesLayer
             public readonly GUIContent heightOffset = new GUIContent("Height Offset", "Offset applied to the height before layering.");
             public readonly GUIContent heightTransition = new GUIContent("Height Transition", "Size in world units of the smooth transition between layers.");
 
+            public readonly GUIContent perPixelDisplacementLayersWarning = new GUIContent("For pixel displacement to work correctly, all layers with a heightmap must use the same UV mapping");
```
EvgeniiG: For per-pixel displacement

PR #475, opened 2017-10-12
light/probe lists reference implementation

vlad-andreev:
A reference implementation for using the light and reflection probe list API.  Implemented in FPTL to minimize potential impact, should be trivial to replicate in HDRP though.


mmikk:
I suppose we can just merge it into master only?

sebastienlagarde:
yes, that'ts the goal. Just wanted to confirm that it break really the version, so need to be careful on future merge :).

PR #497, opened 2017-10-16
Add test for scene view deferred draw mode blacklisting under SRP

Tak:
(Requires changes from unity `graphics/srp`)


Tak:
Adding to RenderPipeline/UX editor tests instead.

PR #498, opened 2017-10-16
Use the distribution of visible normals for runtime IBL prefiltering

EvgeniiG:


ScriptableRenderPipeline/Core/ShaderLibrary/ImageBasedLighting.hlsl
```diff
@@ -128,6 +128,56 @@ void SampleGGXDir(float2   u,
...
+
+    float3 localH = mul(viewH, viewToLocal);
+    NdotH = localH.z;
+
+    // Compute { localL = reflect(-localV, localH) }
+    float3 localL = -localV + 2 * VdotH * localH;
+    NdotL = localL.z;
+
+#if 0
+    H = mul(localH, localToWorld);
```
sebastienlagarde: garbage ?

PR #500, opened 2017-10-17
Feature/hd editor cli

fredpointzero:
* Added a utility class for HDRenderPipeline editor utilities
* Added a menu item to reset keywords of material, project wide
* Added a cli to reset keywords of materials, project wide


ScriptableRenderPipeline/HDRenderPipeline/Editor/HDEditorUtils.cs
```diff
@@ -32,5 +44,24 @@ public static string GetCorePath()
...
+            {
+                RemoveMaterialKeywords(material);
+                resetter(material);
+                EditorUtility.SetDirty(material);
+                return true;
+            }
+            return false;
+        }
+
+        public static void RemoveMaterialKeywords(Material material)
```
sebastienlagarde: Tojborn said me that we can do material.shaderKeywords = null instead of doing the foreach. Can you check ?

PR #506, opened 2017-10-18
Update fog and blend mode

sebastienlagarde:
- rename _LinearFogOoRange as _LinearFogOneOverRange
- add _BLENDMODE_PRESERVE_SPECULAR_LIGHTING keyword
- change _ENABLE_FOG to _ENABLE_FOG_ON_TRANSPARENT_
- update test scene
- fix issue with refractionMask not correct by default- rename (reverse)
- update test scene


ScriptableRenderPipeline/HDRenderPipeline/Material/Material.hlsl
```diff
@@ -25,7 +28,7 @@ float4 EvaluateAtmosphericScattering(PositionInputs posInput, float4 inputColor)
 {
     float4 result = inputColor;
```
JulienIgnace-Unity: I think this is going to break. Because user can check the fog box and then switch back to opaque and the keyword will still be there. Either we reset the keyword in the editor or we keep the _SURFACE_TYPE_TRANSPARENT test here.

PR #518, opened 2017-10-20
Refactored refraction mask as opacity mask

fredpointzero:
- Renamed refraction mask to opacity mask


ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Editor/LitUI.cs
```diff
@@ -411,7 +411,7 @@ protected void DoLayerGUI(Material material, int layerIndex)
 
             bool refractionEnable = (material.HasProperty(kRefractionMode) && refractionMode.floatValue > 0.0f) && (material.HasProperty(kPreRefractionPass) && preRefractionPass.floatValue == 0.0f);
 
-            m_MaterialEditor.TexturePropertySingleLine(refractionEnable ? Styles.baseColorRefractionMaskText : Styles.baseColorText, baseColorMap[layerIndex], baseColor[layerIndex]);
+            m_MaterialEditor.TexturePropertySingleLine(refractionEnable ? Styles.baseColorOpacityMaskText : Styles.baseColorText, baseColorMap[layerIndex], baseColor[layerIndex]);
```
sebastienlagarde: No need for a different text than usual here

SampleScenes/HDTest/GraphicTest/Common/Material/Mat_refraction_plane_Clear_ior101.mat
```diff
@@ -241,7 +241,7 @@ Material:
     - _ZTestMode: 8
     - _ZWrite: 0
     m_Colors:
-    - _BaseColor: {r: 1, g: 1, b: 1, a: 1}
+    - _BaseColor: {r: 1, g: 1, b: 1, a: 0}
```
sebastienlagarde: Don't think it is a good idea. What we would really need is a default depends in context. But as we have only one default value let's put 1. The trouble with 0 is that transparent with alpha or additive will disappear when switching surface type. Let's stick to the default 1 like vanilla unity

PR #553, opened 2017-10-31
Update outdated .meta files

EvgeniiG:


EvgeniiG:
Caused by an AssetPostprocessor. Solved.

PR #563, opened 2017-11-06
HDRenderPipeline: Add support of shadow mask (Disabled in 2017.3) - require C++ change

sebastienlagarde:
- Create clearColorAllBlack to have color with alpha channel with 0 for
clear.
- Add shadowmask and occlusion probe sampling functions
- Add various debug feature for shadow mask
- Refactor GBuffer manager
- Add SHADOWS_SHADOWMASK keyword to our various lit and debug shader
- In deferred, SHADOWS_SHADOWMASK add an extra GBuffer
- Create a BakeLightingData that contain bakeDiffuseLighting +
bakeShadowMask
- Update all lightLoop call with this new BakeLightingData struct, pass
it to EvaluateBSDF_XXX function
- Move #pragma multi_compile from .shader to pass that require it
(unsure if this save variant)


ScriptableRenderPipeline/HDRenderPipeline/Debug/DebugDisplay.cs
```diff
@@ -366,7 +366,7 @@ public enum DebugViewVarying
             VertexNormalWS,
             VertexColor,
             VertexColorAlpha,
-            // caution if you add something here, it must start below
+            // caution if you add something here, it must be use to start enum below
```
JulienIgnace-Unity: Suggestion: Instead of copying last value, add something like "Last" and then in the enum bellow : Depth = DebugViewVarying.Last,
This way, no need to copy anything.

ScriptableRenderPipeline/HDRenderPipeline/Debug/DebugDisplay.cs
```diff
@@ -376,14 +376,19 @@ public enum DebugViewGbuffer
             None = 0,
             Depth = DebugViewVarying.VertexColorAlpha + 1,
             BakeDiffuseLightingWithAlbedoPlusEmissive,
+            BakeShadowMask0,
+            BakeShadowMask1,
+            BakeShadowMask2,
+            BakeShadowMask3,
+            // caution if you add something here, it must be use to start enum below
```
JulienIgnace-Unity: Same here as above.

PR #585, opened 2017-11-11
Unity 2017.3

sebastienlagarde:
@phi-lira I have made a bad merge on Unity-2017.3. I have reverted my change, but want that you verify that none of your change are lost, I should have revert to the state of your last merge. And now we need to upgrade the change to master. Please check. Thanks (Particularly the test 037 that was having large name for windows


phi-lira:
I can't see any issue in the editor regarding the test 037 name. Actually there are tests with longer names in the project (e.g, 034).

phi-lira:
Ah Ok. I see it, it's actually the assets inside test 037. I'll fix it.

PR #588, opened 2017-11-17
Improve HLSL macros

EvgeniiG:


ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -118,77 +99,21 @@ bool IsBitSet(uint data, uint bitPos)
...
-int Mul24(int a, int b)
-{
-    return a * b;
-}
-
-uint Mul24(uint a, uint b)
-{
-    return a * b;
-}
+    TEMPLATE_2(Mul24, a, b, return a * b)
```
sebastienlagarde: I mitigated about the usage of TEMPLATE that you have here.
I have concern with TEMPLATE macro and intrinsic, doesn't think it is a good fit.
It make sense for something like Sqr because it doesn't rely on intrinsic, but for anything that rely on intrincis you need to be careful as it may not exist at all! in the case of Mul24 it make no sense for float for example.
for the various intrinsic I don't think we should do more than what exist. 

Also, regarding the name TEMPLATE itself, I don't think it should be call template.  Maybe something like DECLARE_ALL_TYPE_FUNCTIONS or something like that.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -375,31 +292,13 @@ static const float4x4 k_identity4x4 = {1.0, 0.0, 0.0, 0.0,
...
-float3 PositivePow(float3 base, float3 power)
-{
-    return pow(max(abs(base), float3(FLT_EPSILON, FLT_EPSILON, FLT_EPSILON)), power);
-}
-
-float4 PositivePow(float4 base, float4 power)
-{
-    return pow(max(abs(base), float4(FLT_EPSILON, FLT_EPSILON, FLT_EPSILON, FLT_EPSILON)), power);
-}
+TEMPLATE_2(PositivePow, base, power, return pow(max(abs(base), FLT_EPSILON), power))
```
sebastienlagarde: I am curious, doesn't this generate warning because FLT_EPSILON is not a float 4 or other ?
also once again it make no sense to have this function define for bool or for int...

PR #612, opened 2017-11-27
Feature/realtime cubemap

JulienIgnace-Unity:
First PR for the handling of real time reflection probes.
*THIS PR WILL NOT WORK WITHOUT THE CORRESPONDING C++ PR*
The code will still compile (and not break existing stuff) but realtime probes won't work.

- Introduces ReflectionProbeCache on top of texture cache to handle realtime GGX convolution and later BC6H compression
- Started splitting rendering parameter in Global and Frame parameters for future refactor.
- Modified HDRP main loop to render all available camera


ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -546,4 +546,11 @@ void LODDitheringTransition(uint2 unPositionSS, float ditherFactor)
     clip(ditherFactor - p);
 }
 
+
+// MACRO from Legacy Untiy
+// Transforms 2D UV by scale/bias property
+#define TRANSFORM_TEX(tex, name) ((tex.xy) * name##_ST.xy + name##_ST.zw)
```
sebastienlagarde: Move this in the new macro.hlsl file

sebastienlagarde: Adding @mmikk for the cache change, so Morten you are aware of it.

sebastienlagarde:
Adding @mmikk for the cache change, so Morten you are aware of it.

ScriptableRenderPipeline/HDRenderPipeline/FrameSettings.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+    // The settings here are per frame settings. They can be changed by the renderer based on its need each frame.
+    public class LightingSettings
```
sebastienlagarde: So, we have the namespace, but we should put them in a class FrameSettings, can be done in a later PR

ScriptableRenderPipeline/HDRenderPipeline/GlobalSettings.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+    // All the structures here represent global engine settings.
+    // It means that they are supposed to be setup once and not changed during the game.
+    // All of these will be serialized in the HDRenderPipelineInstance used for the project.
+    [Serializable]
```
sebastienlagarde: From naming point of view, even if we have the namespace, maybe we should have a class: GlobalSettings, and inside it, have TextureSettings, RenderSettings etc... Allow to gather all global settings in a class. Can be done in a later PR, depends how the abstraction is done later

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
```diff
@@ -1081,6 +1068,7 @@ void AccumulateDistortion(CullResults cullResults, Camera camera, ScriptableRend
                 int w = camera.pixelWidth;
                 int h = camera.pixelHeight;
 
+                cmd.ReleaseTemporaryRT(m_DistortionBuffer);
```
sebastienlagarde: I am wondering if we should do a wrapper here ? to always do release and GEt, because I fear we will miss it if we add another one (use funciton for shadow mask, same for GBuffer etc...)

ScriptableRenderPipeline/HDRenderPipeline/HDUtils.cs
```diff
@@ -58,5 +58,45 @@ public static void NextOverlayCoord(ref float x, ref float y, float overlayWidth
                 s_OverlayLineHeight = -1.0f;
             }
         }
+
+
+        public static Matrix4x4 ComputePixelCoordToWorldSpaceViewDirectionMatrix(float verticalFoV, Vector4 screenSize, Matrix4x4 worldToViewMatrix, bool renderToCubemap)
```
sebastienlagarde: Revisit the name :)
Maybe not need to add "Compute", use just PixelCoordToWorldSpaceViewDirectionMatrix, but still a bit long, why "direction" ?

ScriptableRenderPipeline/HDRenderPipeline/Lighting/ReflectionProbeCache.cs
```diff
@@ -0,0 +1,209 @@
...
+                m_TempRenderTexture.autoGenerateMips = false;
+                m_TempRenderTexture.Create();
+
+                m_ConvolutionTargetTexture = new RenderTexture(m_ProbeSize, m_ProbeSize, 1, RenderTextureFormat.ARGBHalf);
+                m_ConvolutionTargetTexture.dimension = TextureDimension.Cube;
+                m_ConvolutionTargetTexture.useMipMap = true;
+                m_ConvolutionTargetTexture.autoGenerateMips = false;
+                m_ConvolutionTargetTexture.Create();
+
+                m_ConvertTextureMaterial = CoreUtils.CreateEngineMaterial("Hidden/SRP/BlitCubeTextureFace");
```
sebastienlagarde: Add "Hidden/SRP/BlitCubeTextureFace" it to the resource system

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
fredpointzero: Is it better to have a Cleanup method when the frame is rendered, rather than releasing and getting the RT lazily?

TestbedPipelines/OnTileDeferredPipeline/OnTileDeferredRenderPipeline.cs
```diff
@@ -146,7 +146,7 @@ static void SetupDeferredRenderPipelineMaterials()
 				foreach (Material _material in _materials) {
 					if (_material == null)
 						continue;
-
+
```
sebastienlagarde: Plenty of space added in this files, please remove them!

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
sebastienlagarde: I like the idea of Fred of release RT at end of frame until we get thae RT manager

JulienIgnace-Unity: We plan on working on an abstraction for RenderTextures in SRP Core with Thomas that I think we should delay any change until then. These changes are just there to make things work with the current code.

TestbedPipelines/OnTileDeferredPipeline/OnTileDeferredRenderPipeline.cs
JulienIgnace-Unity: It's just space turned into tabs actually.

fredpointzero: Maybe we should have a commit hook to transform tab to space to avoid this?

PR #614, opened 2017-11-27
BC6H Support for ReflectionProbeCache

fredpointzero:
REQUIRES SPECIFIC C++ FEATURES

C# will compile but ReflectionProbeCache will use RGBAHalf format and not BC6H.


ScriptableRenderPipeline/Core/TextureCache.cs
```diff
@@ -334,12 +334,18 @@ public int ReserveSlice(Texture texture, out bool needUpdate)
...
         public void UpdateSlice(CommandBuffer cmd, int sliceIndex, Texture content, uint textureHash)
-        {
+                {
             // transfer new slice to sliceIndex from source texture
-            m_SliceArray[sliceIndex].sliceEntryHash = textureHash;
+            SetSliceHash(sliceIndex, textureHash);
             TransferToSlice(cmd, sliceIndex, content);
         }
 
+        public void SetSliceHash(int sliceIndex, uint updateCount)
```
JulienIgnace-Unity: Maybe rename updateCount to hash?

PR #618, opened 2017-11-29
[Refraction] Fix refraction sampling

fredpointzero:
The NDC for the refracted point were incorrect, so the color buffer was not properly sampled.


fredpointzero:
Fixes as well BUG-1205-002
=> Skybox are refracted with plane model

PR #619, opened 2017-11-29
Reintroduce EvaluateBSDF_Area()

EvgeniiG:


ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -441,8 +441,8 @@ struct PositionInputs
     float3 positionWS;  // World space position (could be camera-relative)
     float2 positionNDC; // Normalized screen UVs          : [0, 1) (with the half-pixel offset)
     uint2  positionSS;  // Screen space pixel coordinates : [0, NumPixels)
-    uint2  tileCoord;   // Screen space tile coordinates  : [0, NumTiles)
-    float  deviceDepth; // Depth from the depth buffer    : [0, 1]
+    uint2  tileCoord;   // Screen tile coordinates        : [0, NumTiles)
+    float  deviceDepth; // Depth from the depth buffer    : [0, 1] (typically reversed)
```
sebastienlagarde: No, we inverse it when we read/write to depth buffer, but this value is alway [0 close, 1 far] (unless we have done mistake in code :) )

PR #620, opened 2017-11-29
Optimize divergent execution of the anisotropic GGX

EvgeniiG:
If the tile has any pixels with anisotropic GGX, we evaluate the entire tile with anisotropic GGX rather than first evaluating pixels with anisotropic GGX and then pixels with isotropic GGX.

Before

; --- Statistics for Deferred.compute on GCN (Pitcairn) ---
; SGPRs: **92** out of 104 used
; VGPRs: 128 out of 256 used
; LDS: 0 out of 32768 bytes used
; 0 bytes scratch space used
; Instructions: **2676 ALU**, 183 Control Flow, 50 TFETCH

After

; --- Statistics for Deferred.compute on GCN (Pitcairn) ---
; SGPRs: **94** out of 104 used
; VGPRs: 128 out of 256 used
; LDS: 0 out of 32768 bytes used
; 0 bytes scratch space used
; Instructions: **2583 ALU**, 183 Control Flow, 50 TFETCH

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
+++ b/ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
@@ -401,6 +401,13 @@ BSDFData ConvertSurfaceDataToBSDFData(SurfaceData surfaceData)
     bsdfData.roughness = PerceptualRoughnessToRoughness(bsdfData.perceptualRoughness);
     bsdfData.materialId = surfaceData.materialId;
 
+    if (surfaceData.materialId != MATERIALID_LIT_ANISO)
+    {
+        // Notify the material classification system that we should not use
+        // the anisotropic GGX for forward rendering.
+        g_FeatureFlags &= ~MATERIALFEATUREFLAGS_LIT_ANISO;
+    }
+
     // IMPORTANT: In case of foward or gbuffer pass we must know what we are statically, so compiler can do compile time optimization
     if (bsdfData.materialId == MATERIALID_LIT_STANDARD)
     {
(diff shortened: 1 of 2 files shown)


ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -401,6 +401,13 @@ BSDFData ConvertSurfaceDataToBSDFData(SurfaceData surfaceData)
     bsdfData.roughness = PerceptualRoughnessToRoughness(bsdfData.perceptualRoughness);
     bsdfData.materialId = surfaceData.materialId;
 
+    if (surfaceData.materialId != MATERIALID_LIT_ANISO)
+    {
+        // Notify the material classification system that we should not use
+        // the anisotropic GGX for forward rendering.
+        g_FeatureFlags &= ~MATERIALFEATUREFLAGS_LIT_ANISO;
```
sebastienlagarde: I don't understand this comment, the material classification is not used in forward as the compiler know per pixel what it should use (there is no material tile concept).

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -627,6 +634,34 @@ void DecodeFromGBuffer(
             bsdfData.materialId = MATERIALID_LIT_CLEAR_COAT;
     }
 
+    // We avoid divergent evaluation of the GGX, as that nearly doubles the cost.
+    // If the tile has anisotropy, all the pixels within the tile are evaluated as anisotropic.
+    if (HasMaterialFeatureFlag(MATERIALFEATUREFLAGS_LIT_ANISO))
```
sebastienlagarde: I not sure to understand how this is working, what is happening when you have aniso and sss, or clear coat and aniso ? I don't think it work as then, there is no way to ensure that aniso is 0 as you will read it form the GBuffer and it can contain another information. I think the change only work when aniso and standard are use, right ?

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
sebastienlagarde: ok, I tihnk I got it from the implementation. too bad that material classification specific have side effect on forward, but I get it.

sebastienlagarde: You should just add a comment that in case of forward, g_FeatureFlags is set to UINT, and for optimieatoin purpose we need to remove MATERIALFEATUREFLAGS_LIT_ANISO in case of forward (even if material classification is not use).

PR #622, opened 2017-11-30
Motion Vector Support

Wadarass:
HDRenderpipeline changes to support Motion Vector Pass.
- Moved last frame positions to Texcoord4
-Also added proper handling of vertex animation in velocity pass.

Requires https://ono.unity3d.com/unity/unity/pull-request/57342/_/graphics/motion-vector-srp to be merged to trunk to work.


ScriptableRenderPipeline/HDRenderPipeline/ShaderPass/ShaderPassVelocity.hlsl
```diff
@@ -77,17 +77,46 @@ VaryingsPassToDS InterpolateWithBaryCoordsPassToDS(VaryingsPassToDS input0, Vary
 #define VARYINGS_NEED_PASS
 #include "VertMesh.hlsl"
 
+void PositionZBias(VaryingsToPS input)
```
sebastienlagarde: I don't think we should support this "Feature". In the original code of Unity we have:

            // this works around an issue with dynamic batching
            // potentially remove in 5.4 when we use instancing
#if defined(UNITY_REVERSED_Z)
            o.pos.z -= _MotionVectorDepthBias * o.pos.w;
#else
            o.pos.z += _MotionVectorDepthBias * o.pos.w;
#endif

We should check if this is still relevant.
At least we should let the comment and remove the code when instancing support is done ?
And if we let it, rename the function VecolityPositionZBias()

ScriptableRenderPipeline/HDRenderPipeline/ShaderPass/ShaderPassVelocity.hlsl
```diff
@@ -77,17 +77,46 @@ VaryingsPassToDS InterpolateWithBaryCoordsPassToDS(VaryingsPassToDS input0, Vary
...
+
     // It is not possible to correctly generate the motion vector for tesselated geometry as tessellation parameters can change
     // from one frame to another (adaptative, lod) + in Unity we only receive information for one non tesselated vertex.
     // So motion vetor will be based on interpolate previous position at vertex level instead.
-    varyingsType.vpass.positionCS = mul(_NonJitteredViewProjMatrix, mul(unity_ObjectToWorld, float4(inputMesh.positionOS, 1.0)));
-    varyingsType.vpass.previousPositionCS = mul(_PrevViewProjMatrix, mul(unity_MatrixPreviousM, unity_MotionVectorsParams.x ? float4(inputPass.previousPositionOS, 1.0) : float4(inputMesh.positionOS, 1.0)));
+	varyingsType.vpass.positionCS = mul(_NonJitteredViewProjMatrix, float4(varyingsType.vmesh.positionWS, 1.0));
+
+	//Need to apply any vertex animation to the previous worldspace position, if we want it to show up in the velocity buffer
+	float3 previousPositionWS = mul(unity_MatrixPreviousM, unity_MotionVectorsParams.x ? float4(inputPass.previousPositionOS, 1.0) : float4(inputMesh.positionOS, 1.0)).xyz;
```
sebastienlagarde: I think we should skip the whole block (of vertex animation) if  unity_MotionVectorsParams.x > 0, as it mean we disable the velocity calculation.

ScriptableRenderPipeline/HDRenderPipeline/ShaderPass/ShaderPassVelocity.hlsl
```diff
@@ -77,17 +77,46 @@ VaryingsPassToDS InterpolateWithBaryCoordsPassToDS(VaryingsPassToDS input0, Vary
...
+	#ifdef ATTRIBUTES_NEED_NORMAL
+		float3 normalWS = normalize(mul(inputMesh.normalOS, (float3x3)unity_MatrixPreviousMI));
```
sebastienlagarde: We should create a PreviousTransformObjectToWorldNormal function that do the transform as there is the UNITY_ASSUME_UNIFORM_SCALING that should be taken into account

sebastienlagarde:
Just adding some comment that don't need to be solve for this PR ( I can do it in a later PR) but to have in mind.
- Previous position calculation can depends on other stuff than just _Time (in the future), I think we should provide a funciton for PreviousVertexModification and let the Lit shader define inside how to process the previous position (in this case with _LastTime)
- When we are fully static, we shoud rely on camera motion blur even when there is displacement. the trouble with Lit is you can't know if dispplacmeent map is animated with a custom texture. I suggest that we add an extra checkbox on Lit.shader (in baseUnlit). When enabled, it mean that users want a veclocity pass (so the condition to enable a pass is this box and if  displacement is enabled (for tessellation or vertex) or if wind is enabled
- Ideally, motion vector should not be in TexCoord4 but in interpolator define by the renderPipeline (so the RenderPipeline can decide to put it anywhere it make sense for him. In our case, as we now extend to 8 texcoord with Kaspar comming PR, maybe we should at least setup it in TEXCOORD7 (?) until we can chose where to put it ?

ScriptableRenderPipeline/HDRenderPipeline/ShaderPass/ShaderPassVelocity.hlsl
sebastienlagarde: I see that on C++ side this is done.
Looks weird. We may need to have to validate that it is really an issue for HD, I don't see why this would happen (and HD currently rely on equal depth test for most passes, so we know for sure depth equal is working).

ScriptableRenderPipeline/HDRenderPipeline/ShaderPass/ShaderPassVelocity.hlsl
sebastienlagarde: side note: if we reintroduce bias, maybe we should reintroduce the force no motion for consistency with C++ ?  I don't like this option but...

Wadarass: force no motion I think is already there.  It's in the CalculateVelocity function.
return (positionCS.xy - previousPositionCS.xy) * unity_MotionVectorsParams.y; <---- force no motion.

Wadarass: Actually we still need the vertex animation even if unity_MotionVectorsParams.x > 0.  That's just deciding if it's a skinned mesh or not.  Still need to know if there was a vertex animation the previous frame.

sebastienlagarde: ok got it. Currently I was misunderstanding the name: HasLastPosition, which is really a bad name to say isSkinMesh... can you add comment about it somewhere, so other people don't make the mistake :). Thakns

PR #635, opened 2017-12-05
Hook up preliminary XR render paths

robbiesri:
XR HDRP will be forward only, so we need to make sure that when stereo is enabled, and the camera is going to perform stereo rendering, that we use forward-only.

To support this, we also need to make the choice to go forward-only per camera, as not all cameras might be rendering in stereo (or in a more general sense, not all cameras might want to be uniformly forward or deferred).  HDCamera now owns the state about whether the current camera is forward-only or not.

Additionally, I added an option to the RenderingDebugSettings to completely disable the stereo rendering path.

Reviewers: I've implemented some of this before in prototypes, so I re-used some of the functionality/ideas.  Some of the re-organization here fits in with some XR changes I have planned in the future (such as using HDCamera more, and initializing it a bit earlier).  Looking forward to your thoughts.


robbiesri:
I know I need to do a merge, will do tomorrow.

ScriptableRenderPipeline/HDRenderPipeline/Lighting/TilePass/TilePass.cs
```diff
@@ -485,7 +484,6 @@ int GetDeferredLightingMaterialIndex(int outputSplitLighting, int lightLoopTileP
                 // In HD, MSAA is only supported for forward only rendering, no MSAA in deferred mode (for code complexity reasons)
 
                 // If Deferred, enable Fptl. If we are forward renderer only and not using Fptl for forward opaque, disable Fptl
-                m_isFptlEnabled = !renderingSettings.ShouldUseForwardRenderingOnly() || tileSettings.enableFptlForForwardOpaque; // TODO: Disable if MSAA
                 m_isFptlEnabledForForwardOpaque = tileSettings.enableFptlForForwardOpaque; // TODO: Disable if MSAA
```
sebastienlagarde: Move m_isFptlEnabledForForwardOpaque to UpdateRenderingPathState just to keep parameter together.

sebastienlagarde:
Hey, 
so the change looks good to me. And you are right, here is the plan for the future.

We have HDAdditionalCameraData in HDRP, and we will add some Lightloop settings on it, so a users can define the rendering by camera (I think we will abstract the settings behind an enum, like (SimpleForward, fullDeferred, StereoForward etc...). And we will control MSAA from there.

ScriptableRenderPipeline/HDRenderPipeline/Lighting/TilePass/TilePass.cs
robbiesri: Sounds good, I was thinking of this too.  I'll make the change now.

robbiesri:
Ok @sebastienlagarde , I integrated your suggestion, along with merged in tip of master.

PR #641, opened 2017-12-06
Add support for forward sss

sebastienlagarde:
This PR add the support of forward SSS. 

For this we consider that SSS is not link to Lit shader anymore but is always Lambert SSS that you can combine with any Specular BRDF.

This PR move most of the SSS component to Material/SubsurfaceScattering folder

There is now an abstraction to encode/decode to SSS buffers that is use by both deferred and forward. These buffer are then read during SSSSS pass. There is a new SSS Manager Buffer to handle the C# side and allocate target in case of full forward.

We now use a single buffer for all SSS information (albedo + subsurface radius + profileID). The specular occlusion is store in RT2 instead of RT0 in the case of SSS as there is no problem to doing it.

PR have been tested on Jimenez and Disney SSS in both deferred and forward with pre/post scatter and post scatter mode with and without transmission.

There is also various misc small cleanup of the code:
- Changed the behavior of copy stencil to remove unnecessary operation. Now we only perform stencil operation for material classifiaction and Disney SSS

Performance:
This PR should improve performance of SSSSS as now it use only one buffer. Should be measured.
Also we should measure if there is any performance cost of binding 4 RT for forward opaque all the times.


ScriptableRenderPipeline/HDRenderPipeline/Lighting/LightLoop/LightLoop.cs
```diff
@@ -401,6 +401,7 @@ public void Allocate()
         static ComputeBuffer s_GlobalLightListAtomic = null;
         // clustered light list specific buffers and data end
 
+        bool m_isForwardRenderingOnly;
```
robbiesri: Oh this was a good idea.  I prefer this to the inferred state that I did in the forward-only mini-refactor.

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Resources/SubsurfaceScattering.shader
```diff
@@ -41,8 +41,7 @@ Shader "Hidden/HDRenderPipeline/SubsurfaceScattering"
 
             #include "ShaderLibrary/Common.hlsl"
             #include "../../../ShaderVariables.hlsl"
-            #define UNITY_MATERIAL_LIT // Needs to be defined before including Material.hlsl
-            #include "../../../Material/Material.hlsl"
+            #include "../../SubsurfaceScattering/SubsurfaceScattering.hlsl"
```
robbiesri: This isn't specific to this change, but does the Editor/Engine not have the functionality to specify shader include paths?

sebastienlagarde: It has, we need to do a pass on the include path stuff. But for now we keep all our include path relative.

ScriptableRenderPipeline/HDRenderPipeline/Lighting/LightLoop/materialflags.compute
```diff
@@ -50,7 +50,7 @@ void MATERIALFLAGSGEN(uint threadID : SV_GroupIndex, uint3 u3GroupID : SV_GroupI
 		int idx = i * NR_THREADS + threadID;
 		uint2 uCrd = min( uint2(viTilLL.x + (idx & 0xf), viTilLL.y + (idx >> 4)), uint2(iWidth - 1, iHeight - 1));
 
-		if (UnpackByte(LOAD_TEXTURE2D(_StencilTexture, uCrd).r) != STENCILLIGHTINGUSAGE_NO_LIGHTING) // This test is we are the sky/background or not
+		if (LOAD_TEXTURE2D(_StencilTexture, uCrd).r == 1.0) // This test if we are the sky/background or a forward opaque (i.e we don't require lighting)
```
EvgeniiG: Please use a define instead of a literal constant here.

ScriptableRenderPipeline/HDRenderPipeline/Material/LayeredLit/LayeredLit.shader
```diff
@@ -404,6 +404,11 @@ Shader "HDRenderPipeline/LayeredLit"
     // This shader support vertex modification
     #define HAVE_VERTEX_MODIFICATION
 
+    // If we use subsurface scattering, enable output split lighting (for forward pass)
+    #if defined(_MATID_SSS) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
EvgeniiG: Maybe it'd be good to add an #error if both SSS and transparent are enabled. Or forbid it in the GUI?

ScriptableRenderPipeline/HDRenderPipeline/Material/LayeredLit/LayeredLitTessellation.shader
```diff
@@ -416,6 +416,11 @@ Shader "HDRenderPipeline/LayeredLitTessellation"
     #define HAVE_VERTEX_MODIFICATION
     #define HAVE_TESSELLATION_MODIFICATION
 
+    // If we use subsurface scattering, enable output split lighting (for forward pass)
+    #if defined(_MATID_SSS) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
EvgeniiG: Same here.

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.hlsl
```diff
@@ -471,7 +424,11 @@ void EncodeIntoGBuffer( SurfaceData surfaceData,
     }
     else if (surfaceData.materialId == MATERIALID_LIT_SSS)
     {
-        outGBuffer2 = float4(surfaceData.subsurfaceRadius, surfaceData.thickness, 0.0, PackByte(surfaceData.subsurfaceProfile));
+        // Special case: For SSS we will store the profile id and the subsurface radius at the location of the specular occlusion (in alpha channel of GBuffer0)
```
EvgeniiG: Not too important, but you could make a function SwapSpecOcclusionWithSssRadiusAndProfileId(inout float4 gbuffer0, inout float4 gbuffer2), for the implementation of which you could move this comment, and then call it in both Encode- and DecodeFromGbuffer().

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Lit.shader
```diff
@@ -219,6 +219,11 @@ Shader "HDRenderPipeline/Lit"
     // This shader support vertex modification
     #define HAVE_VERTEX_MODIFICATION
 
+    // If we use subsurface scattering, enable output split lighting (for forward pass)
+    #if defined(_MATID_SSS) && !defined(_SURFACE_TYPE_TRANSPARENT)
```
EvgeniiG: Same thing about #error.

ScriptableRenderPipeline/HDRenderPipeline/Material/Lit/Resources/SubsurfaceScattering.compute
```diff
@@ -361,10 +361,18 @@ void SubsurfaceScattering(uint2 groupId       : SV_GroupID,
...
 
+    // TODO: Since we have moved to forward SSS, we don't support anymore a bsdfData.normalWS.
+    // Once we include normal+roughness rendering during the prepass, we will have a buffer to bind here and we will be able to reuse this part of the algorithm on demand.
+#if SSS_USE_TANGENT_PLANE
     // Compute the tangent frame in view space.
     float3 normalVS = mul((float3x3)viewMatrix, bsdfData.normalWS);
     float3 tangentX = GetLocalFrame(normalVS)[0] * unitsPerMm;
     float3 tangentY = GetLocalFrame(normalVS)[1] * unitsPerMm;
+#else
+    float3 normalVS = float3(0, 0, 0);
```
EvgeniiG: Let's add an #error here.

ScriptableRenderPipeline/HDRenderPipeline/Material/SubsurfaceScattering/SubsurfaceScatteringManager.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class SubsurfaceScatteringManager
+    {
+        // Currently we only support SSSBuffer with one buffer. If the shader code change, it may require to update the shader manager
+        public const int k_MaxSSSBuffer = 1;
+
+        readonly int m_SSSBuffer0;
+        readonly RenderTargetIdentifier m_SSSBuffer0RT;
+
+        public int sssBufferCount { get; set; }
```
EvgeniiG: What's the point of this if it's always set to 1? Am I missing something?

sebastienlagarde: done

PR #644, opened 2017-12-07
Branch in the transmission code

EvgeniiG:
Currently, transmission is scalarized. This means that if the shader is compiled with the SSS support, we always execute the transmission code.

This PR reverts this change.

Caution: TEST performance before merging


sebastienlagarde:
obsolete

PR #647, opened 2017-12-08
Separate hierarchy in testrunner for the different pipelines

RemyUnity:
Split the "TestScene" function and it's dependencies to have a separate hierarchy in the testrunner window for the different render pipelines graphic tests.

### DIFF
--- a/Tests/GraphicsTests/Framework/Editor/TestFramework.cs
+++ b/Tests/GraphicsTests/Framework/Editor/TestFramework.cs
@@ -23,12 +23,6 @@ namespace UnityEditor.Experimental.Rendering
             "RenderPipeline"
         };
 
-        private static readonly string[] s_PipelinePath =
-        {
-            "LightweightPipeline",
-            "HDRenderPipeline",
-        };
-
         // info that gets generated for use
         // in a dod way
         public struct TestInfo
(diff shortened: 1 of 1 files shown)


Tests/GraphicsTests/Framework/Editor/TestFramework.cs
```diff
@@ -70,9 +70,21 @@ public static IEnumerable scenes
...
+                            switch (pipelinePath)
+                            {
+                                case "LightweightPipeline":
+                                    name = "LTRP_" + name;
+                                    break;
+                                case "HDRenderPipeline":
+                                    name = "HDRP_" + name;
+                                    break;
+                            }
+
```
phi-lira: Do we still need to enumerate all pipelines scenes? It seems we could avoid this duplicated code by either removing it if we don't need it by defining it with GetScenesForPipeline()

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
```diff
@@ -81,6 +93,61 @@ public static IEnumerable scenes
...
+                foreach (var path in allPaths)
+                {
+                    var p = new FileInfo(path);
+                    var split = s_Path.Aggregate("", Path.Combine);
+                    split = string.Format("{0}{1}", split, Path.DirectorySeparatorChar);
+                    var splitPaths = p.FullName.Split(new[] { split }, StringSplitOptions.RemoveEmptyEntries);
+
+                    // Add prefix to the name for easier reading
+                    var name = p.Name;
+                    switch (_pipelinePath)
```
phi-lira: IMO it would be better to pass in the pipeline prefix as argument. This way if we add more pipelines to the test framework we can just add the pipeline property without having any changes to the GetScenesForPipeline

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
```diff
@@ -94,6 +161,17 @@ public void TearDown()
         }
 
         [UnityTest]
+        public IEnumerator HDRP_TestScene([ValueSource(typeof(CollectScenes), "HDScenes")]TestInfo testInfo)
+        {
+            return TestScene(testInfo);
+        }
+
+        [UnityTest]
+        public IEnumerator LTRP_TestScene([ValueSource(typeof(CollectScenes), "LTScenes")]TestInfo testInfo)
```
phi-lira: Minor thing. LW as abbreviation for Lightweight makes more sense.

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
RemyUnity: Well, I can even remove the prefix thing, it was first done before I found a way to have a separate "hierarchy" in testrunner. It's not really needed anymore now.

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
RemyUnity: No more used now, I will delete. I was kind of preservative about the code that was already there.

RemyUnity: Fixed

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
RemyUnity: Fixed

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
RemyUnity: Fixed

PR #650, opened 2017-12-08
Dispatch the SSS CS in 4x swizzled 16x16 groups per a 32x32 macrotile

EvgeniiG:


EvgeniiG:
It's a tiny bit faster.

PR #675, opened 2017-12-12
Fixed few  .SampleLevel call not using the abstractoin API

sebastienlagarde:


SampleScenes/Common/Shaders/CubemapColorMap.shader
```diff
@@ -9,7 +9,7 @@
 
     #include "UnityCG.cginc"
 
-    TextureCube _ColorMap;
+    TEXTURECUBE(_ColorMap);
```
robbiesri: TEXTURECUBE isn't defined in any of the current includes.  I think you meant to also add #include "ShaderLibrary/Common.hlsl"

I did try adding ShaderVariables.hlsl so I wouldn't have to use UnityObjectToClipPos before I realized that this was a SampleScene shader, which made it confusing to try to include ShaderVariables.hlsl (not even sure if I _wanted_ to do that).

PR #684, opened 2017-12-14
Set the pipeline asset once for a batch of tests.

RemyUnity:
Modify graphic test to set the pipeline asset once for a batch of tests. Can gain time when running tests.

### DIFF
--- a/Tests/GraphicsTests/Framework/Editor/TestFramework.cs
+++ b/Tests/GraphicsTests/Framework/Editor/TestFramework.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using NUnit.Framework;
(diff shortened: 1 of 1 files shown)


Tests/GraphicsTests/Framework/Editor/TestFramework.cs
```diff
@@ -86,37 +112,62 @@ public static IEnumerable GetScenesForPipeline(string _pipelinePath)
...
 
-        [UnityTest]
-        public IEnumerator LWRP_TestScene([ValueSource(typeof(CollectScenes), "LWScenes")]TestInfo testInfo)
+        public static RenderPipelineAsset beforeTestsRenderPipeAsset;
+        public static RenderPipelineAsset wantedTestsRenderPipeAsset;
+
+        public void SetupRenderPipeAsset()
         {
-            return TestScene(testInfo);
+            Debug.Log("Set " + _SRP_ID + " render pipeline.");
```
phi-lira: Is this intentional? Just asking because I sometimes use Debug.Log while developing and forget to remove them :D

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
```diff
@@ -39,37 +40,62 @@ public override string ToString()
...
             {
                 var absoluteScenesPath = s_Path.Aggregate(s_RootPath, Path.Combine);
 
                 var filesPath = Path.Combine(absoluteScenesPath, _pipelinePath);
 
                 // find all the scenes
-                var allPaths = System.IO.Directory.GetFiles(filesPath, "*.unity", System.IO.SearchOption.AllDirectories);
+                var allPaths = Directory.GetFiles(filesPath, "*.unity", SearchOption.AllDirectories);
+
+                // Convert to List for easy sorting in alphabetical ordre
```
phi-lira: ordre type here

Tests/GraphicsTests/Framework/Editor/TestFramework.cs
RemyUnity: I prefered to leave it, just so the user knows that the rederpipe get changed for the tests.

PR #685, opened 2017-12-14
Store light.range in lightData.size.x for box lights

EvgeniiG:


ScriptableRenderPipeline/HDRenderPipeline/Lighting/LightLoop/LightLoop.cs
```diff
@@ -834,10 +836,14 @@ public bool GetLightData(CommandBuffer cmd, ShadowSettings shadowSettings, Camer
             lightData.up = light.light.transform.up;
             lightData.right = light.light.transform.right;
 
-            lightData.size = new Vector2(additionalLightData.shapeLength, additionalLightData.shapeWidth);
+            lightDimensions.x = additionalLightData.shapeLength;
+            lightDimensions.x = additionalLightData.shapeWidth;
```
sebastienlagarde: there is a bug here with lightDimensions.y

EvgeniiG: Fixed.

PR #687, opened 2017-12-15
Lightweight pipeline upgrade to SRP shader library

phi-lira:
Lightweight Pipeline shader upgrade to use SRP shader library. 

 * Added GLCore, GLES3 and GLES2 shader files
 * Modified the shader library to compile for GLES2
      * Macros.hlsl -> ifdef uint versions as uint is redefined to int in GLES2
      * Common.hlsl -> Added new OpenGL apis. Ifdef some unsuported GLES2 functions
      * Random.hlsl and Packing.hlsl functions not supported on GLES2
* Entity Lighting -> Added dLDR HDR decode for lightmap, Added decode HDR for environment
* Upgraded all LW shaders to hlsl and added pragma to compile GLES2 with HLSLcc compiler.
* Structured lightweight constant buffers usage. 
       * Input.hlsl -> contains all LW pipeline PerCamera and PerFrame constants
       * InputBuiltin.hsls -> contains all Unity engine constants. V, P and inv in the pipeline use the same as unity as SetupCameraProperties already set them and I want to avoid setting duplicate constants.
       * InputSurface.hlsl -> contains all the Standard Shader material constants.


ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -127,21 +130,45 @@ float4 PackEmissiveRGBM(float3 rgb)
...
     return rgbmInput.rgb * pow(rgbmInput.a, 2.2f) * LIGHTMAP_RGBM_RANGE;
 }
 
-float3 SampleSingleLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), float2 uv, float4 transform, bool lightmapRGBM)
+half3 UnpackLightmapDoubleLDR(half4 encodedColor)
+{
+    return encodedColor.rgb * LIGHTMAP_DLDR_RANGE;
+}
+
+half3 DecodeHDRLightmap(half4 encodedIlluminance)
```
sebastienlagarde: Minor: I don't like the name "DecodeHDRLightmap", as currently they are limited range and precision. Maybe just DecodeLightmap is more appropriate.

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Core.hlsl
```diff
@@ -0,0 +1,135 @@
...
+    //Opengl => z clip range is [-near, far] -> should remap in theory but dont do it in practice to save some perf (range is close enough)
+    #define UNITY_Z_0_FAR_FROM_CLIPSPACE(coord) (coord)
+#endif
+
+half Pow4(half x)
+{
+    return x * x * x * x;
+}
+
+half LerpOneTo(half b, half t)
```
sebastienlagarde: define in CommonMaterial.hlsl, (note feel free to changed stuff to half in shader library, console doesn't care and it will benefit for various platform

sebastienlagarde: LerpWhiteTo

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Core.hlsl
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void AlphaDiscard(half alpha, half cutoff)
+{
+#ifdef _ALPHATEST_ON
+    clip(alpha - cutoff);
+#endif
+}
+
+half3 SafeNormalize(half3 inVec)
```
sebastienlagarde: can move this one to common.hlsl

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Core.hlsl
```diff
@@ -0,0 +1,135 @@
...
+{
+    half dp3 = max(1.e-4h, dot(inVec, inVec));
+    return inVec * rsqrt(dp3);
+}
+
+// Unpack normal as DXT5nm (1, y, 1, x) or BC5 (x, y, 0, 1)
+// Note neutral texture like "bump" is (0, 0, 1, 1) to work with both plain RGB normal and DXT5nm/BC5
+half3 UnpackNormalmapRGorAG(half4 packedNormal, half bumpScale)
```
sebastienlagarde: define in Packing.hlsl

sebastienlagarde: feel free to change float to half

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Particles.hlsl
```diff
@@ -0,0 +1,214 @@
+#include "Core.hlsl"
+#include "InputSurface.hlsl"
+
+#if defined (_COLORADDSUBDIFF_ON)
+half4 _ColorAddSubDiff;
+#endif
+
+#if defined(_COLORCOLOR_ON)
+half3 RGBtoHSV(half3 arg1)
```
sebastienlagarde: define in color.Hlsl: RgbToHsv same for next one

sebastienlagarde:
Good job, happy to see the transition :)

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,107 @@
...
+#define LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, index) textureName.Load(int4(unCoord2, index, 0))
+#define LOAD_TEXTURE2D_ARRAY_LOD(textureName, unCoord2, index, lod) textureName.Load(int4(unCoord2, index, lod))
+
+#if (SHADER_TARGET >= 45)
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  textureName.Gather(samplerName, coord2)
+#define GATHER_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index)     textureName.Gather(samplerName, float3(coord2, index))
+#define GATHER_TEXTURECUBE(textureName, samplerName, coord3)                textureName.Gather(samplerName, coord3)
+#define GATHER_TEXTURECUBE_ARRAY(textureName, samplerName, coord3, index)   textureName.Gather(samplerName, float4(coord3, index))
+#else
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  SAMPLE_TEXTURE2D(textureName, samplerName, coord2)
```
aleks01010101: This is not an equivalent of gather. Shouldn't we sample the texture four times with point filtering instead?
Or return float4(0,0,0,0) since this is not supported anyway.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,107 @@
+// This file assume SHADER_API_GLCORE is defined
```
aleks01010101: Perhaps do an #ifndef/#error/#endif instead of this comment?
And also, shouldn't we guard against double include?

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
+// This file assume SHADER_API_GLES is defined
```
aleks01010101: Same as GLcore.hlsl, perhaps guard + check if SHADER_API_GLES is defined and #error if it's not?

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+
+// Initialize arbitrary structure with zero values.
+// Do not exist on some platform, in this case we need to have a standard name that call a function that will initialize all parameters to 0
+#define ZERO_INITIALIZE(type, name) name = (type)0;
+#define ZERO_INITIALIZE_ARRAY(type, name, arraySize) { for (int arrayIndex = 0; arrayIndex < arraySize; arrayIndex++) { name[arrayIndex] = (type)0; } }
+
+// GLES2 might not have shadow hardware comparison support
+#if defined(UNITY_ENABLE_NATIVE_SHADOWS_LOOKUPS)
+#define SHADOW2D_TEXTURE_AND_SAMPLER sampler2DShadow
+#define SHADOWCUBE_TEXTURE_AND_SAMPLER samplerCUBEShadow
```
aleks01010101: As far as I know, ES2 doesn't support cubemap shadows even via an extension

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#define SHADOW2D_SAMPLE(textureName, samplerName, coord3) shadow2D(textureName, coord3)
+#define SHADOW3D_SAMPLE(textureName, samplerName, coord4) ((texCUBE(textureName,(coord4).xyz) < (coord4).w) ? 0.0 : 1.0)
```
aleks01010101: Given the comment above, this misses a `.r` before the comparison

aleks01010101: Btw, why is this called "SHADOW3D_SAMPLE" is it's sampling a cubemap?

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#endif
+
+// Texture util abstraction
+
+#define CALCULATE_TEXTURE2D_LOD(textureName, samplerName, coord2) #error calculate Level of Detail not supported in GLES2
+
+// Texture abstraction
+
+#define TEXTURE2D(textureName)                          sampler2D textureName
+#define TEXTURE2D_ARRAY(textureName)                    samplerCUBE textureName // No support to texture2DArray
```
aleks01010101: Wouldn't it be better to #error or to declare it and sample it as a 2D texture?

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#define TEXTURECUBE(textureName)                        samplerCUBE textureName
+#define TEXTURECUBE_ARRAY(textureName)                  samplerCUBE textureName // No supoport to textureCubeArray and can't emulate with texture2DArray
+#define TEXTURE3D(textureName)                          sampler3D textureName
+
+#define TEXTURE2D_SHADOW(textureName)                   SHADOW2D_TEXTURE_AND_SAMPLER textureName
+#define TEXTURE2D_ARRAY_SHADOW(textureName)             TEXTURECUBE_SHADOW(textureName) // No support to texture array
```
aleks01010101: Same as above, #error is probably better

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#define SAMPLER(samplerName)
+#define SAMPLER_CMP(samplerName)
+
+#define TEXTURE2D_ARGS(textureName, samplerName)                sampler2D textureName
+#define TEXTURE2D_ARRAY_ARGS(textureName, samplerName)          samplerCUBE textureName
+#define TEXTURECUBE_ARGS(textureName, samplerName)              samplerCUBE textureName
+#define TEXTURECUBE_ARRAY_ARGS(textureName, samplerName)        samplerCUBE textureName
+#define TEXTURE3D_ARGS(textureName, samplerName)                sampler3D textureName
+#define TEXTURE2D_SHADOW_ARGS(textureName, samplerName)         SHADOW2D_TEXTURE_AND_SAMPLER textureName
+#define TEXTURE2D_ARRAY_SHADOW_ARGS(textureName, samplerName)   SHADOWCUBE_TEXTURE_AND_SAMPLER textureName
```
aleks01010101: This is inconsistent with the stuff above, a comment/#error/non-array version should be used instead

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#if (SHADER_TARGET >= 30)
+    #define SAMPLE_TEXTURE2D_LOD(textureName, samplerName, coord2, lod) tex2Dlod(textureName, float4(coord2, 0, lod))
+#else
+    // No lod support. Very poor approximation with bias.
+    #define SAMPLE_TEXTURE2D_LOD(textureName, samplerName, coord2, lod) SAMPLE_TEXTURE2D_BIAS(textureName, samplerName, coord2, lod)
+#endif
+
+#define SAMPLE_TEXTURE2D_BIAS(textureName, samplerName, coord2, bias) tex2Dbias(textureName, float4(coord2, 0, bias))
+#define SAMPLE_TEXTURE2D_GRAD(textureName, samplerName, coord2, ddx, ddy) SAMPLE_TEXTURE2D(textureName, coord2)
+#define SAMPLE_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index) SAMPLE_TEXTURECUBE(textureName, samplerName, float3(coord2, index))
```
aleks01010101: And again, same as above

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,158 @@
...
+#define SAMPLE_DEPTH_TEXTURE_LOD(textureName, samplerName, coord2, lod) SAMPLE_TEXTURE2D_LOD(textureName, samplerName, coord2, lod).r
+
+#define LOAD_TEXTURE2D(textureName, unCoord2)                       ERROR_ON_UNSUPPORTED_FUNCTION(texelFetch)
+#define LOAD_TEXTURE2D_LOD(textureName, unCoord2, lod)              ERROR_ON_UNSUPPORTED_FUNCTION(texelFetchLod)
+#define LOAD_TEXTURE2D_MSAA(textureName, unCoord2, sampleIndex)     ERROR_ON_UNSUPPORTED_FUNCTION(texelFetchBias)
+#define LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, index)          ERROR_ON_UNSUPPORTED_FUNCTION(texelFetchArray)
+#define LOAD_TEXTURE2D_ARRAY_LOD(textureName, unCoord2, index, lod) ERROR_ON_UNSUPPORTED_FUNCTION(texelFetchLodArray)
+
+// Gather not supported. Fallabck to regular texture sampling.
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  SAMPLE_TEXTURE2D(textureName, samplerName, coord2)
```
aleks01010101: As in GLCore, this is an incorrect fallback. Should either sample the texture 4x times or return .rrrr

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES3.hlsl
```diff
@@ -0,0 +1,109 @@
+// This file assume SHADER_API_GLES3 is defined
```
aleks01010101: Same as GLCore.hlsl and GLES.hlsl

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES3.hlsl
```diff
@@ -0,0 +1,109 @@
...
+#define LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, index) textureName.Load(int4(unCoord2, index, 0))
+#define LOAD_TEXTURE2D_ARRAY_LOD(textureName, unCoord2, index, lod) textureName.Load(int4(unCoord2, index, lod))
+
+#if (SHADER_TARGET >= 45)
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  textureName.Gather(samplerName, coord2)
+#define GATHER_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index)     textureName.Gather(samplerName, float3(coord2, index))
+#define GATHER_TEXTURECUBE(textureName, samplerName, coord3)                textureName.Gather(samplerName, coord3)
+#define GATHER_TEXTURECUBE_ARRAY(textureName, samplerName, coord3, index)   textureName.Gather(samplerName, float4(coord3, index))
+#else
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  SAMPLE_TEXTURE2D(textureName, samplerName, coord2)
```
aleks01010101: Same as in GLCore and ES2, gather returns 4x red components for bilinear filtering. A better approximation would be .rrrr

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -71,6 +71,12 @@
 #include "API/Metal.hlsl"
 #elif defined(SHADER_API_VULKAN)
 #include "API/Vulkan.hlsl"
+#elif defined(SHADER_API_GLCORE)
+#include "API/GLCore.hlsl"
```
aleks01010101: Even with this, I would still suggest a guard in the actual .hlsl files.

sebastienlagarde: There is no guard as API files will only be include once from Common.hlsl
Common.hlsl is mandatory for any shaders write with ShaderLibrary. and API files should never be include directly.

aleks01010101: ok :)

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
phi-lira: If I understood correctly this should do it OES_depth_texture_cube_map extends OES_depth_texture to support it. I tried to play safe here with HLSLSupport.cginc that defines it.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
phi-lira: Good catch. Updated to SHADOWCUBE_SAMPLE

phi-lira:
@sebastienlagarde I've addressed your comments. @aleks01010101 I've addressed the gather issue as some other minor comments. I'll double check the fallback from array to cube in gles2 in next PR.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,109 @@
...
+#endif
+
+#define UNITY_UV_STARTS_AT_TOP 0
+#define UNITY_REVERSED_Z 0
+#define UNITY_GATHER_SUPPORTED 0
+#define UNITY_NEAR_CLIP_VALUE (-1.0)
+
+// This value will not go through any matrix projection convertion
+#define UNITY_RAW_FAR_CLIP_VALUE (1.0)
+#define FRONT_FACE_SEMATIC VFACE
```
EvgeniiG: Typo here: FRONT_FACE_SEMATIC -> FRONT_FACE_SEMANTIC.
Already fixed upstream.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,109 @@
...
+#define SAMPLE_DEPTH_TEXTURE(textureName, samplerName, coord2)                      SAMPLE_TEXTURE2D(textureName, samplerName, coord2).r
+#define SAMPLE_DEPTH_TEXTURE_LOD(textureName, samplerName, coord2, lod)             SAMPLE_TEXTURE2D_LOD(textureName, samplerName, coord2, lod).r
+
+#define LOAD_TEXTURE2D(textureName, unCoord2) textureName.Load(int3(unCoord2, 0))
+#define LOAD_TEXTURE2D_LOD(textureName, unCoord2, lod) textureName.Load(int3(unCoord2, lod))
+#define LOAD_TEXTURE2D_MSAA(textureName, unCoord2, sampleIndex) textureName.Load(unCoord2, sampleIndex)
+#define LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, index) textureName.Load(int4(unCoord2, index, 0))
+#define LOAD_TEXTURE2D_ARRAY_LOD(textureName, unCoord2, index, lod) textureName.Load(int4(unCoord2, index, lod))
+
+#if (SHADER_TARGET >= 45)
```
EvgeniiG: Gather is supported starting from SM 4.1:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb944003(v=vs.85).aspx
Not sure what Unity's equivalent is.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,109 @@
...
+// Texture util abstraction
+
+#define CALCULATE_TEXTURE2D_LOD(textureName, samplerName, coord2) textureName.CalculateLevelOfDetail(samplerName, coord2)
+
+// Texture abstraction
+
+#define TEXTURE2D(textureName)                  Texture2D textureName
+#define TEXTURE2D_ARRAY(textureName)            Texture2DArray textureName
+#define TEXTURECUBE(textureName)                TextureCube textureName
+#define TEXTURECUBE_ARRAY(textureName)          TextureCubeArray textureName
```
EvgeniiG: Cubemap arrays are only supported starting from SM 4.1:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb509695(v=vs.85).aspx
Not sure what Unity's equivalent is.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
```diff
@@ -0,0 +1,109 @@
...
+#define LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, index) textureName.Load(int4(unCoord2, index, 0))
+#define LOAD_TEXTURE2D_ARRAY_LOD(textureName, unCoord2, index, lod) textureName.Load(int4(unCoord2, index, lod))
+
+#if (SHADER_TARGET >= 45)
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  textureName.Gather(samplerName, coord2)
+#define GATHER_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index)     textureName.Gather(samplerName, float3(coord2, index))
+#define GATHER_TEXTURECUBE(textureName, samplerName, coord3)                textureName.Gather(samplerName, coord3)
+#define GATHER_TEXTURECUBE_ARRAY(textureName, samplerName, coord3, index)   textureName.Gather(samplerName, float4(coord3, index))
+#else
+#define GATHER_TEXTURE2D(textureName, samplerName, coord2)                  SAMPLE_TEXTURE2D(textureName, samplerName, coord2).rrrr
```
EvgeniiG: I don't think it's a good fallback.
In fact, I don't think there is a good fallback.
This could lead to hard to track bugs.
Best not make the functionality available at all.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
EvgeniiG: Agreed. Or not declare it at all. This is not a safe fallback.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -546,6 +574,13 @@ void ApplyDepthOffsetPositionInput(float3 V, float depthOffsetVS, float4x4 viewP
 // Misc utilities
 // ----------------------------------------------------------------------------
 
+// Normalize that account for vectors with zero length
+half3 SafeNormalize(half3 inVec)
```
EvgeniiG: Please add a specialization for float.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -546,6 +574,13 @@ void ApplyDepthOffsetPositionInput(float3 V, float depthOffsetVS, float4x4 viewP
...
+{
+    half dp3 = max(1.e-4h, dot(inVec, inVec));
```
EvgeniiG: Don't use magic numbers here:
```
float3 lenSq = dot(inVec, inVec);
return inVec * rsqrt(lenSq > 0 ? lenSq : 1);
```

EvgeniiG: Please use HALF_MIN or FLT_MIN here, depending on the parameter type.

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -102,6 +102,9 @@ float4 SampleProbeOcclusion(TEXTURE3D_ARGS(SHVolumeTexture, SHVolumeSampler), fl
 // RGBM lightmaps are currently always gamma encoded, so we use a constant of range^2.2 = 5^2.2
 #define LIGHTMAP_RGBM_RANGE 34.493242f
 
+// DLRD lightmaps are currently alwasy gamma encoded, so we use a constant of 2.0^2.2 = 4.59
```
EvgeniiG: Typo here.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -0,0 +1,159 @@
...
+// This value will not go through any matrix projection convertion
+#define UNITY_RAW_FAR_CLIP_VALUE (1.0)
+#define FRONT_FACE_SEMATIC VFACE
+#define FRONT_FACE_TYPE float
+#define IS_FRONT_VFACE(VAL, FRONT, BACK) ((VAL > 0.0) ? (FRONT) : (BACK))
+
+#define CBUFFER_START(name)
+#define CBUFFER_END
+
+#define uint int
```
EvgeniiG: This line alone already accounts for most of the #defines below, rendering them redundant.

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -127,21 +130,45 @@ float4 PackEmissiveRGBM(float3 rgb)
     return rgbm;
 }
 
-float3 UnpackLightmapRGBM(float4 rgbmInput)
+half3 UnpackLightmapRGBM(half4 rgbmInput)
```
EvgeniiG: For every function which takes a half, please add an overload which takes a float.
We will try to find a more elegant solution during your visit.

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Core.hlsl
```diff
@@ -0,0 +1,103 @@
...
+    //D3d without reversed z => z clip range is [0, far] -> nothing to do
+    #define UNITY_Z_0_FAR_FROM_CLIPSPACE(coord) (coord)
+#else
+    //Opengl => z clip range is [-near, far] -> should remap in theory but dont do it in practice to save some perf (range is close enough)
+    #define UNITY_Z_0_FAR_FROM_CLIPSPACE(coord) (coord)
+#endif
+
+half Pow4(half x)
+{
+    return x * x * x * x;
```
EvgeniiG: This is faster: ((x * x) * (x * x))

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -83,6 +89,19 @@
 // Common intrinsic (general implementation of intrinsic available on some platform)
 // ----------------------------------------------------------------------------
 
+// Error on GLES2 undefined functions
```
EvgeniiG: This should be in the platform-specific file, IMO.

ScriptableRenderPipeline/LightweightPipeline/Shaders/LightweightShaderLibrary/Core.hlsl
phi-lira: Updated just in case, but I think compiler will optimize this anyway.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
phi-lira: If that's ok for you I'll address this in a new PR. Can't remember now which ones but some functions when defining to #error they will prevent shaders from compiling at all. Seems like shader library call some of them, have to take a second look at it.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
phi-lira: I disagree. I don't think API files should have references to any shader library functions.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLCore.hlsl
phi-lira: Same. I'd like to address this and other API fallback issue on a separate PR.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
sebastienlagarde: yes, no generic function like LODDitheringTransition in API files, only intrinsic. Not sure that we have better way to handle this.

EvgeniiG: BitFieldExtract is an intrinsic.
The rest are already wrapped in `#if !defined(SHADER_API_GLES)` (below), the error line belongs down there.

EvgeniiG: Also note that both `uint` and bitfield manipulation is supported by GLES3.

EvgeniiG:
Will need to address the remaining concerns later, but should be OK for now.

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
aleks01010101: We'll need to check what HLSLcc produces here. The extension just adds depth cubemaps to gles2, but doesn't extend the glsl.

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -102,7 +102,7 @@ float4 SampleProbeOcclusion(TEXTURE3D_ARGS(SHVolumeTexture, SHVolumeSampler), fl
 // RGBM lightmaps are currently always gamma encoded, so we use a constant of range^2.2 = 5^2.2
 #define LIGHTMAP_RGBM_RANGE 34.493242f
 
-// DLRD lightmaps are currently alwasy gamma encoded, so we use a constant of 2.0^2.2 = 4.59
+// DLRD lightmaps are currently always gamma encoded, so we use a constant of 2.0^2.2 = 4.59
```
aleks01010101: Still a typo in the comment, DLRD -> DLDR :)

ScriptableRenderPipeline/Core/ShaderLibrary/API/GLES2.hlsl
```diff
@@ -63,13 +65,13 @@
 #define SHADOW2D_TEXTURE_AND_SAMPLER sampler2DShadow
 #define SHADOWCUBE_TEXTURE_AND_SAMPLER samplerCUBEShadow
 #define SHADOW2D_SAMPLE(textureName, samplerName, coord3) shadow2D(textureName, coord3)
-#define SHADOW3D_SAMPLE(textureName, samplerName, coord4) ((texCUBE(textureName,(coord4).xyz) < (coord4).w) ? 0.0 : 1.0)
+#define SHADOWCUBE_SAMPLE(textureName, samplerName, coord4) ((texCUBE(textureName,(coord4).xyz) < (coord4).w) ? 0.0 : 1.0)
```
aleks01010101: ...and this still misses a `.r`

mmikk:
In regards to availability of the texture cube array issue also notice this pound define:

UNITY_NO_CUBEMAP_ARRAY

It's pushed by the front-end compiler and we're already doing some things to compensate when it's set. See for instance the definition of SAMPLE_TEXTURECUBE_ARRAY_LOD_ABSTRACT which is using a texture 2D array when UNITY_NO_CUBEMAP_ARRAY is defined. The texture cache converts the cube maps to panorama slices on the fly when this is the case. But regardless UNITY_NO_CUBEMAP_ARRAY is set if the build target doesn't support cube array (this is done by our shader compiler front-end).

PR #688, opened 2017-12-16
Refactor FrameSettings system

sebastienlagarde:
Added Robs, Evgenii, for visibility (but needed to merge to continue to work :) ).
@julien, @fred provide feedback on C# if you see possible improvement (using struct instead of class etc...), work will be done in a next PR

This PR introduce a new Settings ssytem.

GlobalFrameSettings define settings that can't be change during runtime. It is equivalent to the GraphicsSettings of Unity (Tiers + shader variant removal).
This allow to allocate resource or not for a given feature.
FrameSettings control within a frame what is enable or not (enableShadow, enableStereo, enableDistortion...).

HDRenderPipelineAsset reference the current GlobalFrameSettings use, there is one per supported platform (Currently this feature is not implemented and only one GlobalFrameSettings is available).
A Camera with HDAdditionalData have one FrameSettings that configure how it will render. For example a camera use for reflection will disable distortion and postprocess.
Additionaly on a Camera there is another Framesettings call EffectiveFrameSettings that is created on the fly based on FrameSettings and allow modificaiton for debugging purpose at runtime without being serialized on disk.
The EffectiveFrameSettings is register in the debug windows at the creation of the camera.
A Camera with HDAdditionalData have a RenderPath that define if it use a "Default" FrameSettings, a preset of FrameSettings or a custom one.

HDRenderPipelineAsset contain a "Default" FrameSettings that can be reference by any camera with RenderPath.Defaut or when the camera don't have HDAdditionalData like the camera of the Editor.
It also contain a DefaultEffectiveFrameSettings

At each frame and for each camera, a temporary FrameSettings is aggregated from the GlobalFrameSettings for the platform and the EffectiveFrameSettings of the current camera.
this aggregation is then use everywhere and pass to various manager.

Addition of this PR:
- GlobalFrameSettings class
- FrameSettings class
- Update HDAdditionalData to contain Framesettings and registration to debug windows

Modified code:
- Move LightLoopSettings into itw own file
- Move m_ReflectionProbeCacheSize, m_CookieTexArraySize and m_CubeCookieTexArraySize into GlobalFrameSettings.GlobalLightLoopSettings
- Build fuction now take only Asset insted of multiple class like asset.resources, asset.shadowSettings
- LightLoop: Remove UpdateRenderingPathState and use NewFrame instead
- Reformat correcly some code
- Move async compute to FrameSettings.RenderSettings
- The new FrameSettings add various other framesettings control (like enableShadowMask), rename some and move some from DebugDisplay to it

Tested with Win64 StandAlone Player and pass graphicTest

Work for next PR:
- There is no inspector currently for the FrameSettings on a Camera. The inspector must show only what is necessary and also update the EffectiveFrameSettings when FrameSettings are dirty (like the HDRenderPipelineAsset do with DefaultFrameSettings)
- Settings in GlobalFrameSettings have currently no effect on resources but they affect the boolean result of the aggregation.
- Clean the allocation of buffer in LightLoop (Cluster buffer should always be allocated and fptl only if not disabled in GlobalFrameSettings)
- Debug windows is not refresh when adding, renaming or deleting camera


ScriptableRenderPipeline/HDRenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -1,24 +1,49 @@
...
+
+        public RenderingPath    renderingPath;
 
+        // To be able to turn on/off FrameSettings properties at runtime for debugging purpose without affecting the original one
+        // we create a runtime copy (m_effectiveFrameSettings that is used, and any parametrization is done on serialized frameSettings)
+        public FrameSettings frameSettings = new FrameSettings(); // Serialize frameSettings
+
+        // Not serialized, not visible
+        FrameSettings m_effectiveFrameSettings = new FrameSettings();
+        public FrameSettings GetEffectiveFrameSettings()
```
EvgeniiG: Maybe Active is a better name than Effective...

ScriptableRenderPipeline/HDRenderPipeline/Editor/HDRenderPipelineInspector.Styles.cs
```diff
@@ -14,35 +14,41 @@ sealed class Styles
...
-            // Rendering Settings
-            public readonly GUIContent renderingSettingsLabel = new GUIContent("Rendering Settings");
-            public readonly GUIContent useForwardRenderingOnly = new GUIContent("Use Forward Rendering Only");
-            public readonly GUIContent useDepthPrepassWithDeferredRendering = new GUIContent("Use Depth Prepass with Deferred rendering");
-            public readonly GUIContent renderAlphaTestOnlyInDeferredPrepass = new GUIContent("Alpha Test Only");
+            public readonly GUIContent globalFrameSettings = new GUIContent("Global Frame Settings");
 
-            // Texture Settings
-            public readonly GUIContent textureSettings = new GUIContent("Texture Settings");
+            // Shadow Settings
```
EvgeniiG: These are not Shadow Settings

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipelineAsset.cs
```diff
@@ -20,17 +20,35 @@ public RenderPipelineResources renderPipelineResources
...
+        public FrameSettings defaultFrameSettings = new FrameSettings(); // This are the defaultFrameSettings for all the camera and apply to sceneView, public to be visible in the inspector
+        // Not serialized, not visible, the settings effectively used
+        FrameSettings m_defaultEffectiveFrameSettings = new FrameSettings();
 
-        // Global Renderer Settings
-        public GlobalRenderingSettings globalRenderingSettings = new GlobalRenderingSettings();
-        public GlobalTextureSettings globalTextureSettings = new GlobalTextureSettings();
-        public SubsurfaceScatteringSettings sssSettings;
-        public LightLoopSettings lightLoopSettings = new LightLoopSettings();
+        public FrameSettings GetEffectiveDefaultFrameSettings()
```
EvgeniiG: Same point regarding the name.

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipelineAsset.cs
```diff
@@ -20,17 +20,35 @@ public RenderPipelineResources renderPipelineResources
...
-        public ShadowInitParameters shadowInitParams = new ShadowInitParameters();
+        public void OnValidate()
+        {
+            // Modification of defaultFrameSettings in the inspector will call OnValidate().
+            // We do a copy of the settings to those effectively used
+            defaultFrameSettings.CopyTo(m_defaultEffectiveFrameSettings);
+        }
+
+        // Store the various GlobalFrameSettings for each platform (for now only one)
+        public GlobalFrameSettings globalFrameSettings = new GlobalFrameSettings();
```
EvgeniiG: Naming needs to be improved.

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
```diff
@@ -866,33 +847,17 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
...
-                            // But resolving at each objects that write in distortion is not possible (need to sort transparent, render those that do not distort, then resolve, then etc...)
-                            // Instead we chose to apply distortion at the end after we cumulate distortion vector and desired blurriness.
-                            AccumulateDistortion(m_CullResults, camera, renderContext, cmd);
-                            RenderDistortion(cmd, m_Asset.renderPipelineResources);
+                        AccumulateDistortion(m_CullResults, camera, renderContext, cmd);
+                        RenderDistortion(cmd, m_Asset.renderPipelineResources);
 
-                            RenderPostProcesses(hdCamera, cmd, postProcessLayer);
-                        }
+                        RenderFinal(hdCamera, cmd, postProcessLayer);
```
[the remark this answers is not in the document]
sebastienlagarde: The thing is that if you disable postprocess, RenderPostProcesses() will call a blit function to the final render target ,thus why I rename  it. Thought ?

ScriptableRenderPipeline/HDRenderPipeline/RenderPipeline/GlobalFrameSettings.cs
```diff
@@ -0,0 +1,43 @@
...
+    // A Camera with HDAdditionalData have one FrameSettings that configure how it will render.For example a camera use for reflection will disable distortion and postprocess.
+    // Additionaly on a Camera there is another Framesettings call EffectiveFrameSettings that is created on the fly based on FrameSettings and allow modificaiton for debugging purpose at runtime without being serialized on disk.
+    // The EffectiveFrameSettings is register in the debug windows at the creation of the camera.
+    // A Camera with HDAdditionalData have a RenderPath that define if it use a "Default" FrameSettings, a preset of FrameSettings or a custom one.
+    // HDRenderPipelineAsset contain a "Default" FrameSettings that can be reference by any camera with RenderPath.Defaut or when the camera don't have HDAdditionalData like the camera of the Editor.
+    // It also contain a DefaultEffectiveFrameSettings
+
+   // GlobalFrameSettings represent settings that are immutable at runtime.
+   // There is a dedicated RenderRenderPipelineSettings for each platform
```
EvgeniiG: RenderRender... :-)

ScriptableRenderPipeline/HDRenderPipeline/RenderPipeline/GlobalFrameSettings.cs
```diff
@@ -0,0 +1,43 @@
...
+   [Serializable]
+    public class GlobalFrameSettings
```
EvgeniiG: I think it's a good name, maybe make it even longer: "GlobalFrameSettingsOverride".

ScriptableRenderPipeline/HDRenderPipeline/RenderPipeline/GlobalFrameSettings.cs
```diff
@@ -0,0 +1,43 @@
...
+        public class GlobalLightingSettings
+        {
+            public bool supportShadowMask = true;
+            public bool supportSSR = true;
+            public bool supportSSAO = true;
+            public bool supportSSSAndTransmission = true;
+        }
+
+        [Serializable]
+        public class GlobalRenderSettings
```
EvgeniiG: Rename "Render" -> "Rendering" for consistency with "Lighting".

ScriptableRenderPipeline/HDRenderPipeline/RenderPipelineResources/RenderPipelineSettings.cs
```diff
@@ -24,15 +24,12 @@ public class GlobalTextureSettings
     [Serializable]
     public class GlobalRenderingSettings
```
EvgeniiG: It's basically GlobalFrameSettings. We should avoid creating too many confusing structures.

EvgeniiG: Suggestion: define a `struct FrameSettings { /* no sub-structs */ }`
Then each camera has` FrameSettings m_FrameSettings;`
The HD pipe stores FrameSettings `m_GlobalFrameSettingsOverride;` Maybe add a `GetGlobalFrameSettingsOverride()` method as well.
Same deal with LightLoopSettings and GlobalLightLoopSettings.
Leave GlobalTextureSettings and GlobalRenderingSettings alone.

EvgeniiG: If not all Settings can be overriden, we can do the following:
`struct Settings { struct OverridableSettings { /* stuff */ }; /* more stuff */ }`;

ScriptableRenderPipeline/HDRenderPipeline/RenderPipeline/GlobalFrameSettings.cs
sebastienlagarde: fix

ScriptableRenderPipeline/HDRenderPipeline/Editor/HDRenderPipelineInspector.Styles.cs
sebastienlagarde: fix

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipeline.cs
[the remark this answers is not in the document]
sebastienlagarde: fix

ScriptableRenderPipeline/HDRenderPipeline/Camera/HDAdditionalCameraData.cs
sebastienlagarde: fix

ScriptableRenderPipeline/HDRenderPipeline/HDRenderPipelineAsset.cs
sebastienlagarde: fix

robbiesri:
Oops, sorry for being so late here.  The idea seems good, though the stereo plumbing has to be re-worked.

FrameSettings (or at least enableStereo) still has to be accessible in HDCamera, as it will still own the stereo state per camera.  The idea is that stereo is decided on via a per-camera basis (StereoTargetEyeMask) + the frame override that can disable stereo on all cameras.

robbiesri:
Oh, derp, this is already merged into master, so I'll work on re-adding the stereo support.  I don't totally understand why it was removed in the first place though...

sebastienlagarde:
>FrameSettings (or at least enableStereo) still has to be accessible in HDCamera, as it will still own the >stereo state per camera. The idea is that stereo is decided on via a per-camera basis >(StereoTargetEyeMask) + the frame override that can disable stereo on all cameras.

@robbiesri I think there is a misunderstanding here, FrameSettings is perCamera, so EnableStereo is per camera. Except that now the settings is in m_Framesettings in HDRenderPipeline instead of HDCamera. Look at InitializeFrameSettings. we do:

            aggregate.enableStereo = camera.cameraType == CameraType.Reflection ? false : frameSettings.enableStereo && UnityEngine.XR.XRSettings.isDeviceActive && (camera.stereoTargetEye == StereoTargetEyeMask.Both);
            // Force forward if we request stereo. TODO: We should not enforce that, users should be able to chose deferred
            aggregate.enableForwardRenderingOnly = aggregate.enableForwardRenderingOnly || aggregate.enableStereo;

And in Render() of HDRenderPipeline:


            foreach (var camera in cameras)
            {
                if (camera == null)
                    continue;

                // First, get aggregate of frame settings base on global settings, camera frame settings and debug settings
                var additionalCameraData = camera.GetComponent<HDAdditionalCameraData>();
                // Note: the scene view camera will never have additionalCameraData
                m_FrameSettings = FrameSettings.InitializeFrameSettings(    camera, m_Asset.GetRenderPipelineSettings(),
                                                                            (additionalCameraData && additionalCameraData.renderingPath != HDAdditionalCameraData.RenderingPath.Default) ? additionalCameraData.GetFrameSettings() : m_Asset.GetFrameSettings());


so accessin m_FrameSettings .EnableStereo allow you to access to the current state for this camera (and this include the control from the DebugWindows, + a glboal disabling if we don't support VR (that can be added to RednerPipelineSettings).

Note: Name have change since this PR.

sebastienlagarde:
Note: For your test, either you enable EnbleStereo on the Default frameSettings in HDRenderPipeline ASset (that we haven't expose in the inspector but we can), or on a camera, you add HDAdditonalCameraData (there is a script in interal->RenderPipeline->High DEfintion) for this, and you select a RenderPath-> Custom then you can enable Stereo.

robbiesri:
Gah, you are right!  I didn't load the diff correctly inside GitHub, because it hides a lot of diffs it considers large.  So I didn't see where the FrameSettings were getting built out.  This makes a lot more sense, thanks.

ScriptableRenderPipeline/HDRenderPipeline/Sky/SkyManager.cs
```diff
@@ -370,14 +370,14 @@ public void UpdateEnvironment(HDCamera camera, Light sunLight, CommandBuffer cmd
             }
         }
 
-        public void RenderSky(HDCamera camera, Light sunLight, RenderTargetIdentifier colorBuffer, RenderTargetIdentifier depthBuffer, CommandBuffer cmd, DebugDisplaySettings debugSettings)
+        public void RenderSky(HDCamera camera, Light sunLight, RenderTargetIdentifier colorBuffer, RenderTargetIdentifier depthBuffer, CommandBuffer cmd, FrameSettings m_frameSettings)
```
robbiesri: Did you mean to name this argument 'm_frameSettings'?

sebastienlagarde: Ah good catch. Will update the argument name

PR #692, opened 2017-12-18
Refactor of sky and fog to use interpolation volumes.

JulienIgnace-Unity:
This PR removes the Scene Setting system in favor of using interpolation volumes.
- Sky and Fog components have been updated to use this new system.
   Now users need to setup at least one volume and setup sky and fog with the new VisualEnvironment (to choose what type of sky/fog is used), HDRI/ProceduralSky (to setup this particular sky) and Linear/ExpFog (to configure the fog).
- As a consequence of using a position based interpolation system, users now need to specify explicitly what sky should be used for baking to ensure consistency when baking.
- SkyManager has been updated to be able to handle several sky rendering contexts: visual sky used for rendering the sky in the main view, baking sky for setting up the bake, and lighting override sky (not implemented yet)
- Updated all scenes in the HDRP repo to this new system.


ScriptableRenderPipeline/HDRenderPipeline/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1555,20 +1555,20 @@ public bool PrepareLightsForGPU(CommandBuffer cmd, ShadowSettings shadowSettings
 
                     if (GetEnvLightData(cmd, camera, probe))
                     {
-                        GetEnvLightVolumeDataAndBound(probe, lightVolumeType, worldToView);
+                    GetEnvLightVolumeDataAndBound(probe, lightVolumeType, worldToView);
```
sebastienlagarde: ident have been fucked here...

PR #695, opened 2017-12-18
Optimize loop: 1x FR & 2x QR => 4x FR

EvgeniiG:


sebastienlagarde:
image

Not sure it is better (left old, right, new)

sebastienlagarde:
Maybe at execution it is faster, but this need to be measured

EvgeniiG:
cos(psi) and sin(psi) are not scalarized (since the scalar unit does not support the use of trigonometry).
It's best if they are loaded directly from the constant buffer.

EvgeniiG:
Obsolete.

PR #705, opened 2017-12-20
Change ShaderLibrary to use REAL instead of float/half

sebastienlagarde:
This PR update the shaderLibrary to use REAL MACRO (and other variant REAL2...) instead of float.
After discussion betwen Tim, Felipe, Seb, Julien and Evgenii

#ifndef real
#define real float
#define real2 float2
#define real3 float3
#define real4 float4

#define real2x2 float2x2
#define real2x3 float2x3
#define real3x2 float3x2
#define real3x3 float3x3
#define real3x4 float3x4
#define real4x3 float4x3
#define real4x4 float4x4

#define REAL_MIN FLT_MIN
#define REAL_MAX FLT_MAX
#endif

Rule for shader library currently are:
- no uniform declaration
- use REAL or explicit half or float depends on usage (like Texcoord require float)
- no "h" or "f" at the end of the constant use 1.0 or real(1.0)

Now a RenderPipeline programmer need to define what is real before including the shader library else real default to float (or half for SHADER_API_MOBILE).
The reasons is that Lightweight pipeline need to have half version of various function

In case half or float is explicitly needed, it must be use instead of real (like for texture coordinate for example), this will be done case by case later.


Chman:
Does it have to be all caps? `real`, `real2`, `real3`, `real4` etc look better on the eye IMO.

uygar3d:
I know they're macros but I also think in this case it would be nice to make an exception and have real spelled out in lower case.

ScriptableRenderPipeline/Core/ShaderLibrary/Shadow/Resources/ShadowBlurMoments.compute
```diff
@@ -81,16 +81,16 @@
 #include "../ShadowMoments.hlsl"
```
uygar3d: Can you leave the floats as is in this file?

EvgeniiG:
After some thought, I agree with the lower case. :-)

sebastienlagarde:
All change do, go ahead for the review

ScriptableRenderPipeline/Core/ShaderLibrary/API/D3D11.hlsl
```diff
@@ -2,9 +2,9 @@
 
 #define UNITY_UV_STARTS_AT_TOP 1
 #define UNITY_REVERSED_Z 1
-#define UNITY_NEAR_CLIP_VALUE (1.0)
+#define UNITY_NEAR_CLIP_VALUE (real(1.0))
```
EvgeniiG: Type casting in constants is unnecessary

ScriptableRenderPipeline/Core/ShaderLibrary/API/D3D11.hlsl
```diff
@@ -2,9 +2,9 @@
...
 // This value will not go through any matrix projection conversion
-#define UNITY_RAW_FAR_CLIP_VALUE (0.0)
+#define UNITY_RAW_FAR_CLIP_VALUE (real(0.0))
 #define FRONT_FACE_SEMANTIC SV_IsFrontFace
 #define FRONT_FACE_TYPE bool
 #define IS_FRONT_VFACE(VAL, FRONT, BACK) ((VAL) ? (FRONT) : (BACK))
```
EvgeniiG: Unrelated: is VFACE a typo?

ScriptableRenderPipeline/Core/ShaderLibrary/API/D3D11.hlsl
```diff
@@ -66,22 +66,22 @@
 #define SAMPLE_TEXTURE2D_LOD(textureName, samplerName, coord2, lod)                  textureName.SampleLevel(samplerName, coord2, lod)
 #define SAMPLE_TEXTURE2D_BIAS(textureName, samplerName, coord2, bias)                textureName.SampleBias(samplerName, coord2, bias)
 #define SAMPLE_TEXTURE2D_GRAD(textureName, samplerName, coord2, dpdx, dpdy)          textureName.SampleGrad(samplerName, coord2, dpdx, dpdy)
-#define SAMPLE_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index)              textureName.Sample(samplerName, float3(coord2, index))
-#define SAMPLE_TEXTURE2D_ARRAY_LOD(textureName, samplerName, coord2, index, lod)     textureName.SampleLevel(samplerName, float3(coord2, index), lod)
-#define SAMPLE_TEXTURE2D_ARRAY_BIAS(textureName, samplerName, coord2, index, bias)   textureName.SampleBias(samplerName, float3(coord2, index), bias)
+#define SAMPLE_TEXTURE2D_ARRAY(textureName, samplerName, coord2, index)              textureName.Sample(samplerName, real3(coord2, index))
```
EvgeniiG: I don't think we want `real` inside the Macros.hlsl at all.
We cannot have several definitions of a single macro with the same name, and sampling textures with half coordinates does not make sense.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -34,15 +34,15 @@
 // VertDefault
 // FragDefault / FragForward / FragDeferred
 
-// constant floating number written as 1.0  (not 1, not 1.0f, not 1.0h)
+// constant floating number written as 1.0  (not 1, not 1.0, not 1.0h)
```
EvgeniiG: Fix the comment

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -34,15 +34,15 @@
...
 
 // uniform have _ as prefix + uppercase _LowercaseThenCamelCase
 
 // All uniforms should be in contant buffer (nothing in the global namespace).
 // The reason is that for compute shader we need to guarantee that the layout of CBs is consistent across kernels. Something that we can't control with the global namespace (uniforms get optimized out if not used, modifying the global CBuffer layout per kernel)
 
 // Structure definition that are share between C# and hlsl.
-// These structures need to be align on float4 to respect various packing rules from sahder language.
+// These structures need to be align on real4 to respect various packing rules from sahder language.
 // This mean that these structure need to be padded.
```
EvgeniiG: Fix the comment

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -53,12 +53,29 @@
 
 // headers from ShaderLibrary do not include "common.hlsl", this should be included in the .shader using it (or Material.hlsl)
 
-// Rules: When doing an array for constant buffer variables, we always use float4 to avoid any packing issue, particularly between compute shader and pixel shaders
+// Rules: When doing an array for constant buffer variables, we always use real4 to avoid any packing issue, particularly between compute shader and pixel shaders
```
EvgeniiG: Fix the comment

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -191,9 +208,9 @@ TEMPLATE_SWAP(Swap) // Define a Swap(a, b) function for all types
 #ifndef INTRINSIC_CUBEMAP_FACE_ID
 // TODO: implement this. Is the reference implementation of cubemapID provide by AMD the reverse of our ?
 /*
-float CubemapFaceID(float3 dir)
+real CubemapFaceID(real3 dir)
```
EvgeniiG: We have to be careful with intrinsics because I am not sure that all of them work with half.

EvgeniiG: Use float with intrinsics for now

Paulm-Unity:
I am a bit concerned with using REAL or real since they are actual valid data types in some programming languages, my first reaction was: 'What?? 'real' is now valid hlsl type?' can we have 'unity_real' instead to avoid confusion?

aras-p:
Hmm I'm not sure what problem does this approach solve?

I mean conceptually, can't everything "just work" if the shader library is written to use `half` for all things where FP16 is enough (colors, normalized vectors), or `float` for everything else?

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -511,27 +528,27 @@ float3 ComputeWorldSpacePosition(float2 positionNDC, float deviceDepth, float4x4
 
 struct PositionInputs
 {
-    float3 positionWS;  // World space position (could be camera-relative)
-    float2 positionNDC; // Normalized screen UVs          : [0, 1) (with the half-pixel offset)
+    real3 positionWS;  // World space position (could be camera-relative)
+    real2 positionNDC; // Normalized screen UVs          : [0, 1) (with the real-pixel offset)
```
EvgeniiG: Fix comment

aras-p: Hmm but that's exactly why I think "just use real everywhere" approach is bad. World space positions should _not_ use FP16, ever.

ScriptableRenderPipeline/Core/ShaderLibrary/Common.hlsl
```diff
@@ -583,46 +600,39 @@ void ApplyDepthOffsetPositionInput(float3 V, float depthOffsetVS, float4x4 viewP
...
 {
-    half dp3 = max(HALF_MIN, dot(inVec, inVec));
+    real dp3 = max(REAL_MIN, dot(inVec, inVec));
     return inVec * rsqrt(dp3);
 }
 
 // Generates a triangle in homogeneous clip space, s.t.
 // v0 = (-1, -1, 1), v1 = (3, -1, 1), v2 = (-1, 3, 1).
-float2 GetFullScreenTriangleTexCoord(uint vertexID)
+real2 GetFullScreenTriangleTexCoord(uint vertexID)
```
EvgeniiG: For these 3 functions, real doesn't make sense. but let's leave them for now.

ScriptableRenderPipeline/Core/ShaderLibrary/CommonLighting.hlsl
```diff
@@ -264,7 +264,7 @@ float3 GetViewReflectedNormal(float3 N, float3 V, out float NdotV)
     // For one, it does not solve the incorrect lookup direction problem, since the normal itself
     // is not modified. There is a more insidious issue, however. 'NdotV' is a constituent element
     // of the mathematical system describing the relationships between different vectors - and
-    // not just normal and view vectors, but also light vectors, half vectors, tangent vectors, etc.
+    // not just normal and view vectors, but also light vectors, real vectors, tangent vectors, etc.
```
EvgeniiG: Fix comment

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -59,32 +59,32 @@ float3 SampleSH9(float4 SHCoefficients[7], float3 N)
...
 {
-    float3 position = (transformToLocal == 1.0f) ? mul(WorldToTexture, float4(positionWS, 1.0)).xyz : positionWS;
-    float3 texCoord = (position - probeVolumeMin) * probeVolumeSizeInv.xyz;
+    real3 position = (transformToLocal == 1.0) ? mul(WorldToTexture, real4(positionWS, 1.0)).xyz : positionWS;
+    real3 texCoord = (position - probeVolumeMin) * probeVolumeSizeInv.xyz;
     // Each component is store in the same texture 3D. Each use one quater on the x axis
     // Here we get R component then increase by step size (0.25) to get other component. This assume 4 component
     // but last one is not used.
-    // Clamp to edge of the "internal" texture, as R is from half texel to size of R texture minus half texel.
+    // Clamp to edge of the "internal" texture, as R is from real texel to size of R texture minus real texel.
```
EvgeniiG: Fix comment

ScriptableRenderPipeline/Core/ShaderLibrary/API/D3D11.hlsl
sebastienlagarde: nope, name of the register

ScriptableRenderPipeline/Core/ShaderLibrary/API/D3D11.hlsl
sebastienlagarde: agree as we can have both half and float sampling, so we should have both in the API

ScriptableRenderPipeline/Core/ShaderLibrary/Color.hlsl
```diff
@@ -314,69 +314,69 @@ float3 LogCToLinear(float3 x)
...
+real4 FastTonemapInvert(real4 c)
 {
-    return float4(FastTonemapInvert(c.rgb), c.a);
+    return real4(FastTonemapInvert(c.rgb), c.a);
 }
 
 // 3D LUT grading
 // scaleOffset = (1 / lut_size, lut_size - 1)
-float3 ApplyLut3D(TEXTURE3D_ARGS(tex, samplerTex), float3 uvw, float2 scaleOffset)
+real3 ApplyLut3D(TEXTURE3D_ARGS(tex, samplerTex), real3 uvw, real2 scaleOffset)
```
EvgeniiG: Fix UV

ScriptableRenderPipeline/Core/ShaderLibrary/Color.hlsl
```diff
@@ -314,69 +314,69 @@ float3 LogCToLinear(float3 x)
...
+    real shift = floor(uvw.z);
     uvw.xy = uvw.xy * scaleOffset.y * scaleOffset.xx + scaleOffset.xx * 0.5;
     uvw.x += shift * scaleOffset.x;
     return SAMPLE_TEXTURE3D(tex, samplerTex, uvw).rgb;
 }
 
 // 2D LUT grading
 // scaleOffset = (1 / lut_width, 1 / lut_height, lut_height - 1)
-float3 ApplyLut2D(TEXTURE2D_ARGS(tex, samplerTex), float3 uvw, float3 scaleOffset)
+real3 ApplyLut2D(TEXTURE2D_ARGS(tex, samplerTex), real3 uvw, real3 scaleOffset)
```
EvgeniiG: Fix UV

ScriptableRenderPipeline/Core/ShaderLibrary/Color.hlsl
```diff
@@ -314,69 +314,69 @@ float3 LogCToLinear(float3 x)
...
+        SAMPLE_TEXTURE2D(tex, samplerTex, uvw.xy + real2(scaleOffset.y, 0.0)).rgb,
         uvw.z - shift
     );
     return uvw;
 }
 
 // Returns the default value for a given position on a 2D strip-format color lookup table
 // params = (lut_height, 0.5 / lut_width, 0.5 / lut_height, lut_height / lut_height - 1)
-float3 GetLutStripValue(float2 uv, float4 params)
+real3 GetLutStripValue(real2 uv, real4 params)
```
EvgeniiG: Fix UV

ScriptableRenderPipeline/Core/ShaderLibrary/CommonMaterial.hlsl
```diff
@@ -72,55 +72,55 @@ float3 BlendNormalWorldspaceRNM(float3 n1, float3 n2, float3 vtxNormal)
...
-    blendWeights = max(blendWeights, float3(0.0, 0.0, 0.0));
+    blendWeights = max(blendWeights, real3(0.0, 0.0, 0.0));
     blendWeights /= dot(blendWeights, 1.0);
 
     return blendWeights;
 }
 
 // Planar/Triplanar convention for Unity in world space
-void GetTriplanarCoordinate(float3 position, out float2 uvXZ, out float2 uvXY, out float2 uvZY)
+void GetTriplanarCoordinate(real3 position, out real2 uvXZ, out real2 uvXY, out real2 uvZY)
```
EvgeniiG: Fix UV?

ScriptableRenderPipeline/Core/ShaderLibrary/CommonMaterial.hlsl
```diff
@@ -72,55 +72,55 @@ float3 BlendNormalWorldspaceRNM(float3 n1, float3 n2, float3 vtxNormal)
...
 {
     // Caution: This must follow the same rule as what is use for SurfaceGradient triplanar
     // TODO: Currently the normal mapping looks wrong without SURFACE_GRADIENT option because we don't handle corretly the tangent space
-    uvXZ = float2(position.z, position.x);
-    uvXY = float2(position.x, position.y);
-    uvZY = float2(position.z, position.y);
+    uvXZ = real2(position.z, position.x);
```
EvgeniiG: Here as well?

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -150,24 +150,24 @@ float3 DecodeLightmap(float4 encodedIlluminance)
...
     // Take into account texture alpha if decodeInstructions.w is true(the alpha value affects the RGB channels)
-    float alpha = max(decodeInstructions.w * (encodedIrradiance.a - 1.0) + 1.0, 0.0);
+    real alpha = max(decodeInstructions.w * (encodedIrradiance.a - 1.0) + 1.0, 0.0);
 
     // If Linear mode is not supported we can skip exponent part
     return (decodeInstructions.x * pow(alpha, decodeInstructions.y)) * encodedIrradiance.rgb;
 }
 
-float3 SampleSingleLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), float2 uv, float4 transform, bool encodedLightmap)
+real3 SampleSingleLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), real2 uv, real4 transform, bool encodedLightmap)
```
EvgeniiG: Fix UV

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -177,10 +177,10 @@ float3 SampleSingleLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), float2
     return illuminance;
 }
 
-float3 SampleDirectionalLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), TEXTURE2D_ARGS(lightmapDirTex, lightmapDirSampler), float2 uv, float4 transform, float3 normalWS, bool encodedLightmap)
+real3 SampleDirectionalLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), TEXTURE2D_ARGS(lightmapDirTex, lightmapDirSampler), real2 uv, real4 transform, real3 normalWS, bool encodedLightmap)
```
EvgeniiG: Fix UV

ScriptableRenderPipeline/Core/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -177,10 +177,10 @@ float3 SampleSingleLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), float2
...
 {
     // In directional mode Enlighten bakes dominant light direction
-    // in a way, that using it for half Lambert and then dividing by a "rebalancing coefficient"
+    // in a way, that using it for real Lambert and then dividing by a "rebalancing coefficient"
```
EvgeniiG: For real, Lambert. :-)

ScriptableRenderPipeline/Core/ShaderLibrary/Macros.hlsl
```diff
@@ -35,18 +35,18 @@
 #define CENTIMETERS_PER_METER 100
 #define METERS_PER_CENTIMETER rcp(CENTIMETERS_PER_METER)
 
-#define FLT_EPS     5.960464478e-8  // 2^-24, machine epsilon: 1 + EPS = 1 (half of the ULP for 1)
+#define FLT_EPS     5.960464478e-8  // 2^-24, machine epsilon: 1 + EPS = 1 (real of the ULP for 1)
```
EvgeniiG: Fix comment

ScriptableRenderPipeline/Core/ShaderLibrary/NormalSurfaceGradient.hlsl
```diff
@@ -1,23 +1,23 @@
 // this produces an orthonormal basis of the tangent and bitangent WITHOUT vertex level tangent/bitangent for any UV including procedurally generated
 // method released with the demo for publication of "bump mapping unparametrized surfaces on the GPU"
 // http://mmikkelsen3d.blogspot.com/2011/07/derivative-maps.html
-void SurfaceGradientGenBasisTB(float3 nrmVertexNormal, float3 sigmaX, float3 sigmaY, float flipSign, float2 texST, out float3 vT, out float3 vB)
+void SurfaceGradientGenBasisTB(real3 nrmVertexNormal, real3 sigmaX, real3 sigmaY, real flipSign, real2 texST, out real3 vT, out real3 vB)
```
EvgeniiG: Maybe we should make all gradient domain stuff float, not sure what should be float other than ST

ScriptableRenderPipeline/Core/ShaderLibrary/Packing.hlsl
```diff
@@ -224,7 +224,7 @@ float3 UnpackFromR11G11B10f(uint rgb)
 
 /*
 // This is GCN intrinsic
-uint FindBiggestComponent(float4 q)
+uint FindBiggestComponent(real4 q)
 {
```
EvgeniiG: Wrong place for intrinsics and should probably be float

sebastienlagarde:
@aras
>Hmm I'm not sure what problem does this approach solve?
>I mean conceptually, can't everything "just work" if the shader library is written to use half for all things >where FP16 is enough (colors, normalized vectors), or float for everything else?

Regarding lighting, half vs float is always a problem. With PBR, physically based unit + HDR, using float is way better which is the usage of HD. For LW, it make more sense to use half because the  range will be lower and not use physical light unit anyway. For the lighting case we need to support both cases.
The other domain is color. All function related color can or not have impact (particularly with HDR color grading).

Outside of these cases not sure. Trigonometry (?).

The brute force conversion of this PR is just a first step to try to move stuff to half, as it is easier to move from half to float than the reverse, I have just done a brute force conversion and we have discussed with Tim and  Felipe that we will fix what we can in the review of the PR then by testing on platform (console don't support half anyway).

This is the decision we made after a meeting in Paris with Tim, Felipe, Evgenii, Me, Julien. But if anyone have a better idea we are open :)

And yes, for sure position, texture coordinate will be float, it is just a brute force conversion as a first step

sebastienlagarde:
>I am a bit concerned with using REAL or real since they are actual valid data types in some programming >languages, my first reaction was: 'What?? 'real' is now valid hlsl type?' can we have 'unity_real' instead to >avoid confusion?

open question, what are people thinking ? 
There is no "real" usage in GPU language, so don't think it is a problem, unity_real will just bloat the code from my point of view but can be safer. What are people thinking ?

Note: I have made the PR but we are a group of people on it :)

ScriptableRenderPipeline/Core/ShaderLibrary/Random.hlsl
```diff
@@ -30,42 +30,42 @@ uint JenkinsHash(uint4 v)
     return JenkinsHash(v.x ^ JenkinsHash(v.y) ^ JenkinsHash(v.z) ^ JenkinsHash(v.w));
 }
 
-// Construct a float with half-open range [0, 1) using low 23 bits.
+// Construct a real with real-open range [0, 1) using low 23 bits.
 // All zeros yields 0, all ones yields the next smallest representable value below 1.
```
EvgeniiG: This will not work with a half

aras-p:
@sebastienlagarde 
> Regarding lighting, half vs float is always a problem. With PBR, physically based unit + HDR, using float is way better which is the usage of HD. For LW, it make more sense to use half because the range will be lower and not use physical light unit anyway. For the lighting case we need to support both cases.
> The other domain is color. All function related color can or not have impact (particularly with HDR color grading).

Hmm ok good point. But for example once/if you start optimizing for PS4 Pro (or something else in PC/consoles that "actually" has FP16), what's your take on which parts you'll try to move to FP16?

I'm thinking that another possible approach could be doing something like a "color" typedef, that would be either float3 (for things that need super-bright HDR) or half3 (for mobile/low-end, where FP16 is enough).

But anyway, ok, fair enough, continue work :)

ScriptableRenderPipeline/Core/ShaderLibrary/Macros.hlsl
EvgeniiG: Line 31 above: INFINITY should be FLT_INF, HALF_INF has a different bitwise representation.
Could use (1.0 / 0.0), but this produces a FP div by 0 warning.

uygar3d:
Another approach would be to introduce new types for the various usages, e.g. pos4_t,, uvs3_t and use those everywhere. Then you could have a global type include where you can selectively shift from high to low precision for each specific usage. And code that actually uses float or half will never have its precision changed. It would be a pain to update all the shaders at this point, though.

sebastienlagarde:
>Hmm ok good point. But for example once/if you start optimizing for PS4 Pro (or something else in >PC/consoles that "actually" has FP16), what's your take on which parts you'll try to move to FP16?

My point of view for this is that for high end GPU, there is no point of using half at random place but only for highly optimized loop. I see that as the usage of SIMD/intrinsic for C++, you will spend time to optimize a given high specialized piece of code, and in this case you know what you want and will do you fp16 code just for this anyway. 

>I'm thinking that another possible approach could be doing something like a "color" typedef, that >would be either float3 (for things that need super-bright HDR) or half3 (for mobile/low-end, where >FP16 is enough).

>Another approach would be to introduce new types for the various usages, e.g. pos4_t,, uvs3_t and use >those everywhere. Then you could have a global type include where you can selectively shift from high >to low precision for each specific usage. And code that actually uses float or half will never have its >precision changed. It would be a pain to update all the shaders at this point, though.

Similar proposition for Aras and Uygar, having specialized typedef. 
Looks appealing. What are people thinking ?
I like the idea but also think it could be a pain in the ass to deal with all these variants, use correct type etc... (and also when you start to mix type inside an operation).

EvgeniiG:
> Another approach would be to introduce new types for the various usages, e.g. pos4_t,, uvs3_t and use those everywhere. Then you could have a global type include where you can selectively shift from high to low precision for each specific usage. And code that actually uses float or half will never have its precision changed. It would be a pain to update all the shaders at this point, though.

Using a half is an optimization is going to happen on a case-by-case basis which has little to do with semantics and more with "does it look good enough?" and "is it actually faster?".

Simple example for UVs. If you have a 4K texture, you cannot use a half. But it totally does with a 64x64 texture (a LUT, a tileable map or a face of a reflection probe).

Having more types also places more burden on the programmer (and not just us at Unity, but also hobbyists).

I've already converted obvious cases from real to float, which is probably good enough for now.

sebastienlagarde:
@ all, Evegnii have upgraded the PR with his various comment fixed, 
so currently we are with real and no specific naming for position, texcoord.

The PR as it is now is fine to be merge. Just open to discussion or maybe we can merge and discuss about specific renaming for a later work ?

Note for @aras-p after discussion with Thomas, color are fine in fp16 even for hdr color grading.

cc @phi-lira @stramit your opinion on various discussion ?

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/BC6H.hlsl
```diff
@@ -2,59 +2,59 @@
...
     err = err * err;
     return err.x + err.y + err.z;
 }
 
 // Quantification Helpers
-float3 Quantize7(float3 x)
+real3 Quantize7(real3 x)
 {
-    return (f32tof16(x) * 128.0f) / (0x7bff + 1.0f);
+    return (f32tof16(x) * 128.0) / (0x7bff + 1.0);
```
phi-lira: Do we want these to be real? If so, seems like we should have a realtof16 function. Similarly an f16toreal one.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -194,7 +194,7 @@ float3 SampleDirectionalLightmap(TEXTURE2D_ARGS(lightmapTex, lightmapSampler), T
     float3 illuminance = float3(0.0, 0.0, 0.0);
     if (encodedLightmap)
     {
-        half4 encodedIlluminance = SAMPLE_TEXTURE2D(lightmapTex, lightmapSampler, uv).rgba;
+        float4 encodedIlluminance = SAMPLE_TEXTURE2D(lightmapTex, lightmapSampler, uv).rgba;
```
phi-lira: Not sure I understand the reason to use float4 here.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/GeometricTools.hlsl
```diff
@@ -93,27 +93,27 @@ float3 RayPlaneIntersect(in float3 rayOrigin, in float3 rayDirection, in float3
 //-----------------------------------------------------------------------------
 
 // Box is AABB
-float DistancePointBox(float3 position, float3 boxMin, float3 boxMax)
+real DistancePointBox(real3 position, real3 boxMin, real3 boxMax)
```
phi-lira: This should be float3

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/GeometricTools.hlsl
```diff
@@ -93,27 +93,27 @@ float3 RayPlaneIntersect(in float3 rayOrigin, in float3 rayDirection, in float3
...
 {
-    return length(max(max(position - boxMax, boxMin - position), float3(0.0, 0.0, 0.0)));
+    return length(max(max(position - boxMax, boxMin - position), real3(0.0, 0.0, 0.0)));
 }
 
-float3 ProjectPointOnPlane(float3 position, float3 planePosition, float3 planeNormal)
+real3 ProjectPointOnPlane(real3 position, real3 planePosition, real3 planeNormal)
```
phi-lira: All of these functions should operate on floats imo, even planeNormal. I'm not sure if we pay for casting from/to half/float. Do we?

EvgeniiG:
@phi-lira 
Conversion from `half` to `float` and back is not free (the cost depends on the platform, likely 1 cycle).

The idea is to use `real` almost everywhere. That way, most functions are available in both half and full precision, depending on your requirements.

By default, `real` is defined as `float`. If you need functions with half precision, you `#define INCLUDE_HALF_PRECISION_FUNCTIONS` (we are missing this macro ATM), `#include your_file.hlsl`, then `#undef INCLUDE_HALF_PRECISION_FUNCTIONS`, which gives you half precision versions.

I see no reason to limit any functions to full precision unless you are sure that their half precision versions will not come useful at some point in some context.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/EntityLighting.hlsl
sebastienlagarde: it's a bug, you can fix it :)

sebastienlagarde:
>I'm a little bit concerned that mostly everything got converted to real. I can see already many functions >with precision issues if real is defined to half, which is what will happen on mobile platforms with LW. >These precision issues are a pain to debug/track.

Please, grab the branch and do all the modification that you think is fine. If you think location must be float, use explicit float :)

>I had the idea when we discussed that the approach was the other way around. Changing from float to >real based on what we know is required. This is much more safer IMO.

Ok, so I have understand the reverse, for me, everything is float and then you fix issue case by case. If you go the other way you will miss a lot of half.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/BC6H.hlsl
sebastienlagarde: fixed

sebastienlagarde:
For various reviewer, several point have been discussed offline, for now we go with this version of the PR.

PR #732, opened 2017-12-24
Use RenderTextureDescriptor in HDRP (XR infrastructure)

robbiesri:
Instead of passing a list of arguments to the RenderTexture APIs, we will use RenderTextureDescriptor to hold persistent information across RenderTexture allocs (width, height, texture dimension), and pass the varying parameters (color format, depth buffer bits, etc) as arguments.  HDCamera will own the base RenderTextureDescriptor.

The XR perspective to this is that the RenderTextureDescriptor can be quite unique depending on the XR rendering mode, making it preferable to hold onto an XR-aware descriptor.
Also, I turned enableStereo to true in FrameSettings and in the HDRenderPipelineAsset. This doesn't actually enable stereo.  It just allows stereo to be enabled with the correct Player mode selection and camera setup.  In the future, I think we want to have the FrameSetting bit turn everything on, but for now, we have a few hoops to jump through.
Anyway, if you turn stereo on, all it does right now is correctly allocate XR-aware RenderTextures.

The method is used in CoreUtils, since this would be useful in other pipelines (but not ported yet).

I tested this against the GraphicsTests, and manual testing, verified via the debug modes and RenderDoc.  For XR, the rendering isn't 'XR' yet, but I can see the underlying RenderTextures are allocated with the correct format (via RenderDoc).


robbiesri:
I'm aware this isn't merge-ready, but I just wanted to push this out for visibility.  The merge shouldn't affect the core functionality.

robbiesri:
Phew, merged in after the move/rename.  Git really hates moves 😢

robbiesri:
This is also preliminary work for adding MSAA to HDRP

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipeline/FrameSettings.cs
```diff
@@ -62,7 +62,7 @@ public class FrameSettings
         public bool enableDistortion = true;
         public bool enablePostprocess = true;
 
-        public bool enableStereo = false;
+        public bool enableStereo = true;
```
sebastienlagarde: caution: if you setup true here, it mean by default enableStereo is true. It is not what we aim for. The goal is that it is false, and it is the camera that enable it. But I suppose you enable it here for debugging purpose ? My advice will be to create your own VR camera with a custom render path that enableStereo in our test scene and just commit it disable.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipelineAsset.asset
```diff
@@ -32,7 +32,7 @@ MonoBehaviour:
     enableTransparentPostPass: 1
     enableDistortion: 1
     enablePostprocess: 1
-    enableStereo: 0
+    enableStereo: 1
```
sebastienlagarde: See Remark in FrameSettings (copy/paste here)

caution: if you setup true here, it mean by default enableStereo is true. It is not what we aim for. The goal is that it is false, and it is the camera that enable it. But I suppose you enable it here for debugging purpose ? My advice will be to create your own VR camera with a custom render path that enableStereo in our test scene and just commit it disable.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
```diff
@@ -77,7 +80,7 @@ public HDCamera(Camera cam)
             Reset();
         }
 
-        public void Update(PostProcessLayer postProcessLayer)
+        public void Update(PostProcessLayer postProcessLayer, bool stereoEnabled)
```
sebastienlagarde: It is ok, but lets pass FrameSettings instead of just this bool, as in the future it may be use for other purpose. thanks

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1011,6 +1033,7 @@ void RenderDistortion(CommandBuffer cmd, RenderPipelineResources resources)
 
             using (new ProfilingSample(cmd, "ApplyDistortion", GetSampler(CustomSamplerId.ApplyDistortion)))
             {
+                // TODO: Should this be using HDCamera.screenSize?
                 var size = new Vector4(m_CurrentWidth, m_CurrentHeight, 1f / m_CurrentWidth, 1f / m_CurrentHeight);
```
sebastienlagarde: yes it should use HDCamera.screensize, we shouln't have use m_CurrentWidth/m_CurrentHeight here

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1394,90 +1415,105 @@ void RenderGaussianPyramidColor(Camera camera, CommandBuffer cmd, ScriptableRend
 
             using (new ProfilingSample(cmd, "Gaussian Pyramid Color", GetSampler(CustomSamplerId.GaussianPyramidColor)))
             {
-                int w = camera.pixelWidth;
-                int h = camera.pixelHeight;
-                int size = CalculatePyramidSize(w, h);
+                var colorPyramidDesc = m_GaussianPyramidColorBufferDesc;
+                var pyramidSideSize = colorPyramidDesc.height;
```
sebastienlagarde: hmm... I don't think this will work. It will work only in the case of VR but we need to consider various configuration. it should be min (width, height) I think. Otherwise, yes I don't think it need to be a power of two as the original code. But if you make this change, you need to answer it still work. (Else I will check with Fred after Christmas vacation)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1394,90 +1415,105 @@ void RenderGaussianPyramidColor(Camera camera, CommandBuffer cmd, ScriptableRend
...
 
         void RenderPyramidDepth(Camera camera, CommandBuffer cmd, ScriptableRenderContext renderContext, FullScreenDebugMode debugMode)
         {
             using (new ProfilingSample(cmd, "Pyramid Depth", GetSampler(CustomSamplerId.PyramidDepth)))
             {
-                int w = camera.pixelWidth;
-                int h = camera.pixelHeight;
-                int size = CalculatePyramidSize(w, h);
+                var depthPyramidDesc = m_DepthPyramidBufferDesc;
+                var pyramidSideSize = depthPyramidDesc.height;
```
sebastienlagarde: Same remark than for color

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1604,7 +1648,7 @@ void RenderDebug(HDCamera camera, CommandBuffer cmd)
             }
         }
 
-        void InitAndClearBuffer(HDCamera hdCamera, bool enableBakeShadowMask, CommandBuffer cmd)
+        void InitAndClearBuffer(HDCamera hdCamera, bool enableBakeShadowMask, CommandBuffer cmd, bool stereoEnabled)
```
sebastienlagarde: you don't need to pass stereo enbaled, used m_FrameSettings instead

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -714,7 +739,7 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                     }
                     ConfigureForShadowMask(enableBakeShadowMask, cmd);
 
-	                InitAndClearBuffer(hdCamera, enableBakeShadowMask, cmd);
+	                InitAndClearBuffer(hdCamera, enableBakeShadowMask, cmd, m_FrameSettings.enableStereo);
```
sebastienlagarde: you don't need to pass stereo enbaled, used m_FrameSettings in the function instead

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1692,5 +1733,31 @@ static int CalculatePyramidSize(int w, int h)
...
+            desc.autoGenerateMips = false;
+
+            var pyramidSize = CalculatePyramidSize((int)hdCamera.screenSize.x, (int)hdCamera.screenSize.y);
+
+            var widthModifier = 1;
+            if (stereoEnabled && (desc.dimension != TextureDimension.Tex2DArray))
+                widthModifier = 2; // double-wide
+
+            desc.width = pyramidSize * widthModifier;
+            desc.height = pyramidSize;
```
sebastienlagarde: I don't know if the configuration will work with the current gaussian algorimth, don't know if it need to be square or not, maybe not, in this case it is ok.

sebastienlagarde:
Poking @JulienIgnace-Unity and @Chman as the work you do will need to be take into account for a future refactor of render target allocation.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
robbiesri: Sure, good idea.  I wasn't totally sure myself what would be better, so happy to hear this feedback.  I'll fix it up.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
robbiesri: Cool, I'll amend this.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
robbiesri: Yep, will change

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
robbiesri: The algorithms do work because of the explicit size parameters passed into the compute jobs.  However, your comment made me manually inspect this a bit more, and I found a little bug in my Size generation for the Color pyramid (fixed it). 

Anyway, right now, there's no need for me to double the width of the pyramid for double-wide, because everything is still rendering monoscopically, so I just commented out this code for now.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
robbiesri: Actually, I was totally dumb. I should have been just using the member var, as you indicated.  Holiday brain :(

robbiesri:
Thanks for the feedback @sebastienlagarde, fixed everything up according to your feedback.  Let me know if you need anything else!

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1394,90 +1414,105 @@ void RenderGaussianPyramidColor(Camera camera, CommandBuffer cmd, ScriptableRend
 
             using (new ProfilingSample(cmd, "Gaussian Pyramid Color", GetSampler(CustomSamplerId.GaussianPyramidColor)))
             {
-                int w = camera.pixelWidth;
-                int h = camera.pixelHeight;
-                int size = CalculatePyramidSize(w, h);
+                var colorPyramidDesc = m_GaussianPyramidColorBufferDesc;
+                var pyramidSideSize = colorPyramidDesc.height;
```
sebastienlagarde: Be careful, for consistency you should use width as you have replace by width in depth pyramid. I will merge and correct this :)

robbiesri: Hey Seb,

Actually, I think I do want to use height.  I should have left a comment here because it's not obvious.

In the monoscopic (and stereo texture array case), the width and height will be the same, so it doesn't matter whether I use width or height to calculate the pyramid size, which I will use to generate lodCounts and the *MipSize constant.

However, in the stereo double-wide case, the texture _will be_ (but isn't yet) two Guassian pyramids placed side by side, horizontally.  In that case, though the texture width is doubled, the 'true' format of the texture that will be sampled is still square, and the height will still be an accurate representation of the pyramid side size.

For now, this doesn't really matter, because I disabled the double-wide pyramid generation in BuildPyramidDescriptor.  In the future, I may actually have some other member variable stored somewhere that indicates the true pyramid size, instead of relying on some non-obvious knowledge about the texture height.

sebastienlagarde: Alright, so let's make it explicit. Create a function name GetPyramidSize() that return the size of the pyramid (so no more assign to height), and comment inside the function the why it is the height :). Thanks!

sebastienlagarde: (note: see the other PR I just push, I have replace to width for consistency, so feel free to update the two location + the setParam on the compute

robbiesri: Sounds good, I'll whip up the PR soon.  Thanks for checking all of this.

PR #736, opened 2017-12-29
Remove redundant clear of SSS filtering target

robbiesri:
Noticed this when doing inspection of target flow.  Verified redundant clear in RenderDoc.

The initial clear is handled at the top of SubsurfaceScatteringPass, where it checks for (sssParameters.useDisneySSS && NeedTemporarySubsurfaceBuffer()).    But then the clear is executed again in the useDisneySSS branch.


sebastienlagarde:
Indeed, well spotted, thanks!

PR #738, opened 2017-12-29
Shadow Refactor v2

phi-lira:
* Reduced shader variants by ~30% by combining shadow and light classification keywords
 * Removed all branching from shadow shader code. 
 * Moved bias and normal bias to caster side.
 * Bias values match Unity legacy. zclip bias is scaled by frustum depth range and normal bias use same kernel as built-in pipeline.
 * Fixed OpenGL clip space depth range while computing shadow projection.
 * Small improvement of shader ALU and constant access


phi-lira:
I'll merge this as we want to generate a new version of LW today. I'll address any comments in a following PR.

PR #739, opened 2018-01-02
 Updated lightweight to upload correct color when in gamma colorspace.

phi-lira:
As discussed in Paris, LW will allow devs to author in gamma colorspace. This PR converts color to active color space set in player settings.


phi-lira:
I'll merge this as we want to generate a new version of LW today. I'll address any comments in a following PR.

PR #751, opened 2018-01-05
Feature/hd camera editor

fredpointzero:
- Added HDCameraEditor
- Updated UI Drawer Framework
- Upgraded HDReflectionProbeEditor with UI Drawer Framework
- Updated HDRenderPipelineInspector to use the common FrameSettings inspector (shared with HDCameraEditor)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDAdditionalCameraData.cs
```diff
@@ -86,16 +90,19 @@ void Update()
 #endif
         }
 
-        void OnValidate()
+        void OnDisable()
```
sebastienlagarde: void OnValidate() is mandatory, let's discuss about it

PR #759, opened 2018-01-07
Decouple SSS and Transmission option

sebastienlagarde:
Rename EnableSSSAndTransmission as EnableSubsurfaceScattering and
EnableTransmission
Note: Disabling Transmission doesn't save the ALU if it is enabled in
the profile

+ Add supportSubsurfaceScattering in UI
+ Small cosmetic rearrangement at the beginning of lit.hlsl

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
@@ -2,45 +2,19 @@
 #define ENVMAP_FEATURE_INFLUENCENORMAL
 #define ENVMAP_FEATURE_PERFACEFADE
 
-//-----------------------------------------------------------------------------
-// SurfaceData and BSDFData
-//-----------------------------------------------------------------------------
-
 // SurfaceData is define in Lit.cs which generate Lit.cs.hlsl
 #include "Lit.cs.hlsl"
 #include "../SubsurfaceScattering/SubsurfaceScattering.hlsl"
 
-// Define refraction keyword helpers
-#define HAS_REFRACTION (defined(_REFRACTION_PLANE) || defined(_REFRACTION_SPHERE))
-#if HAS_REFRACTION
-# include "CoreRP/ShaderLibrary/Refraction.hlsl"
-
-# if defined(_REFRACTION_PLANE)
-#  define REFRACTION_MODEL(V, posInputs, bsdfData) RefractionModelPlane(V, posInputs.positionWS, bsdfData.normalWS, bsdfData.ior, bsdfData.thickness)
-# elif defined(_REFRACTION_SPHERE)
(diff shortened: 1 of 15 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -210,29 +221,28 @@ float3 ComputeFresnel0(float3 baseColor, float metallic, float dielectricF0)
...
+    if (bsdfData.enableTransmission)
     {
         bsdfData.thickness = _ThicknessRemaps[diffusionProfile].x + _ThicknessRemaps[diffusionProfile].y * thickness;
         bsdfData.useThickObjectMode = transmissionMode != TRANSMISSION_MODE_THIN;
 
 #if SHADEROPTIONS_USE_DISNEY_SSS
-            bsdfData.transmittance = ComputeTransmittanceDisney(_ShapeParams[diffusionProfile].rgb,
-                                                                _TransmissionTintsAndFresnel0[diffusionProfile].rgb,
-                                                                bsdfData.thickness, bsdfData.subsurfaceMask);
+        bsdfData.transmittance = _TransmittanceMultiplier * ComputeTransmittanceDisney( _ShapeParams[diffusionProfile].rgb,
```
EvgeniiG: The subsurface radius is unused, but why not pass it?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -210,29 +221,28 @@ float3 ComputeFresnel0(float3 baseColor, float metallic, float dielectricF0)
...
+                                                                                        _TransmissionTintsAndFresnel0[diffusionProfile].rgb,
+                                                                                        bsdfData.thickness, 1.0);
```
EvgeniiG: The subsurface radius is unused, but why not pass it?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
EvgeniiG: Use the tint instead of the multiplier

sebastienlagarde: done

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
sebastienlagarde: Because I have change the "semantic", mask instead of radius to attenuate the effect "artistically" rather than physically. Still even if we consider it as a radius, it should take worlds scale into account no ?

PR #763, opened 2018-01-08
Prototype/decals profiling

Paulm-Unity:


Paulm-Unity:
Fixes normal maps for decals, doesn't conflict with anything outside decal code.

PR #766, opened 2018-01-09
Added a RT abstraction layer to handle scale-based sizing and auto-resize

Chman:
This isn't pretty but will do just fine until we can get a better abstraction layer on top of (or better yet, in place of) the current render texture stuff.


robbiesri:
Infrastructure makes sense.  I kinda wish there was some more explanation about the intended usage/functionality, but I also understand if we have to defer that to later.

PR #771, opened 2018-01-10
Small fixes, cleanup

Chman:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1690,6 +1692,8 @@ public void BuildGPULightListsCommon(Camera camera, CommandBuffer cmd, RenderTar
 
             var w = camera.pixelWidth;
             var h = camera.pixelHeight;
+            s_TempIntArray[0] = w;
+            s_TempIntArray[1] = h;
```
robbiesri: I know I'm too late, but why not give this a more descriptive name?

Chman: I think that's because I planned to recycle this array for other things... Somehow I didn't. It could probably be given a more descriptive name then.

PR #772, opened 2018-01-10
Fixed gaussian pyramid first blit (point -> bilinear filtering)

Chman:
Should look smoother and reduce aliasing in motion.


Chman:
cc @fredpointzero

PR #776, opened 2018-01-11
Move script files to Editor folders if needed

robbiesri:
Some namespaces (e.g. NUnit) can't be used unless the script is inside an Editor folder.
This moves the minimum set of files to Editor folders, relative to the original file location.
There may be more files that should be re-arranged.


robbiesri:
cc @phi-lira 
Can you give this a try to see if this amends your local compilation issues?

PR #777, opened 2018-01-11
Updated Heightmap parametrization and converted all HDRP materials.

JulienIgnace-Unity:
Updated Heightmaps parametrization in the various layered/lit shaders
List of parameters for all the different cases
For per Pixel Displacement only the amplitude parameter remains (because we always displace inward)
For per Vertex Displacement and Tessellation we now have two parametrizations:
- Min/Max
   - Minimum Height (cm)
   - Maximum Height (cm)
   - Offset (cm)
- Amplitude
   - Amplitude (cm)
   - Base (maps the "base" of the texture between 0 and 1 in the texture)
   - Offset (cm)

A conversion script is available in the menu Internal/HDRenderPipeline/Update. This should be called once before changing any material in the project.


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/LitTessellation.shader
```diff
@@ -26,10 +26,22 @@ Shader "HDRenderPipeline/LitTessellation"
 
         _HeightMap("HeightMap", 2D) = "black" {}
         // Caution: Default value of _HeightAmplitude must be (_HeightMax - _HeightMin) * 0.01
-        [HideInInspector] _HeightAmplitude("Height Amplitude", Float) = 0.02 // In world units
-        _HeightCenter("Height Center", Range(0.0, 1.0)) = 0.5 // In texture space
+        // Those two properties are computed from the ones exposed in the UI and depends on the displaement mode so they are separate because we don't want to lose information upon displacement mode change.
```
sebastienlagarde: displaement

PR #778, opened 2018-01-11
Shader library fixes on GLES 2.0.

phi-lira:
* Fixed compiler errors on GLES 2.0
 * Follow up on comments from previous PR to Shader api functions (GLES, GLES3, GLCORE)
 * Not using texcubearray fallback to pano based on UNITY_NO_CUBEMAP_ARRAY because that depends on DirectionToLatLongCoordinate non-intrisic function


phi-lira:
I'll go ahead an merge this because QA is depending on it to build GLES 2.0. If you have any comments let me know and I can address in a following PR/

PR #779, opened 2018-01-11
Reduced GC pressure in LWRP

Chman:
6.9KB/frame -> 0.8KB/frame.

`LightweightPipeline.SetShaderKeywords()` can still be improved. The rest is engine-garbage we have no control over in user-land.

I haven't been able to run tests as the test framework is currently broken so just to be safe let's wait for it to be fixed & run it before we merge this in.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using UnityEngine.Rendering;
 
 namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 {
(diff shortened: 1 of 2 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
```diff
@@ -70,6 +69,19 @@ public float SquaredDistanceToCamera(Vector3 lightPos)
         }
     }
 
+    public class LightEqualityComparer : IEqualityComparer<VisibleLight>
+    {
+        public bool Equals(VisibleLight x, VisibleLight y)
+        {
+            return x.light.GetInstanceID() == y.light.GetInstanceID();
+        }
```
phi-lira: Accessing the light property is an expensive operation. I can expose the instance ID of the light to VisibleLight but meanwhile, do you think we could cache the instance ID?

Chman: It would definitely be nice to have it pre-cached in `VisibleLight` for sure. Caching it in userland will end up with a spaghetti mess.

Chman:
Tests are green (except for the ones that failed before the PR).

PR #795, opened 2018-01-13
Bring over Instancing support to HDRP shaders.

Wadarass:
Basic support for instancing to HDRP shaders.


Wadarass:
Don't know what happened with this merge.  Says I changed a lot more files than I actually did.

Wadarass:
@sebastienlagarde Do we want to make similar changes for lightweight as well?  I imagine so.

sebastienlagarde:
>Don't know what happened with this merge. Says I changed a lot more files than I actually did.
No worry,  a previous commit wasn't formatting files correctly (i.e space vs/tab and \n \n\r)
however in your PR you modify postprocess, I think you should update your postprocess submodule (postprocess shouldn't be part of the PR)

phi-lira:
It would be nice if instancing could be added to LW as well.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -0,0 +1,342 @@
...
+        #endif
+    UNITY_INSTANCING_BUFFER_END(unity_Builtins2)
+
+    #define unity_ObjectToWorld     UNITY_ACCESS_INSTANCED_PROP(unity_Builtins0, unity_ObjectToWorldArray)
+
+    #define MERGE_UNITY_BUILTINS_INDEX(X) unity_Builtins##X
+
+    #define unity_WorldToObject     UNITY_ACCESS_INSTANCED_PROP(MERGE_UNITY_BUILTINS_INDEX(UNITY_WORLDTOOBJECTARRAY_CB), unity_WorldToObjectArray)
+
+    inline float4 UnityObjectToClipPosInstanced(in float3 pos)
```
sebastienlagarde: we will need to redo these function for our purpose (for HD)

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -0,0 +1,342 @@
...
+        #define UNITY_DEFINE_INSTANCED_PROP(type, var)      type var;
+    #endif
+
+    #define UNITY_ACCESS_INSTANCED_PROP(arr, var)           var
+
+#endif // UNITY_INSTANCING_ENABLED
+
+#if defined(UNITY_INSTANCING_ENABLED) || defined(UNITY_PROCEDURAL_INSTANCING_ENABLED) || defined(UNITY_STEREO_INSTANCING_ENABLED)
+    // The following matrix evaluations depend on the static var unity_InstanceID & unity_StereoEyeIndex. They need to be initialized after UnitySetupInstanceID.
+    static float4x4 unity_MatrixMVP_Instanced;
```
sebastienlagarde: All these matrix are not needed.
 all the View and Projection matrix are setup by the render pipeline and we never combine the object to world with view or projection.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/VaryingMesh.hlsl
```diff
@@ -51,6 +51,9 @@ struct VaryingsMeshToPS
 #ifdef VARYINGS_NEED_COLOR
     float4 color;
 #endif
+
+UNITY_VERTEX_INPUT_INSTANCE_ID
```
sebastienlagarde: This shouldn't be require other than in Attribute struct (only from attribute to vertex shader), also we must apply the instance transform (to convert to positionWS) in the vertex shader (i.e before calling hull/domain)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/VaryingMesh.hlsl
```diff
@@ -86,13 +89,17 @@ struct PackedVaryingsMeshToPS
 #if defined(VARYINGS_NEED_CULLFACE) && SHADER_STAGE_FRAGMENT
     FRONT_FACE_TYPE cullFace : FRONT_FACE_SEMANTIC;
 #endif
+
+    UNITY_VERTEX_INPUT_INSTANCE_ID
```
sebastienlagarde: This shouldn't be require other than in Attribute struct (only from attribute to vertex shader), also we must apply the instance transform (to convert to positionWS) in the vertex shader (i.e before calling hull/domain)

sebastienlagarde:
Note: Currently this PR do nothing, just so you know, as instancing need to be enabled in the UI and we don't use instance id for transform

Wadarass:
@sebastienlagarde the UI gets updated automatically as long as the #pragma is in the shader.  @phi-lira I will add the necessary stuff to LW to enable instancing.

PR #807, opened 2018-01-16
GPU instancing in Lightweight renderpipeline

Wadarass:


ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -221,10 +216,11 @@
     #endif
 
     #if defined(UNITY_INSTANCED_SH) && !defined(LIGHTMAP_ON)
-        #if UNITY_SHOULD_SAMPLE_SH
+        //In HDRenderpipe we only decide to look at probe data based on Lightmap flags.
+        #if !defined(DYNAMICLIGHTMAP_ON) 
             #define UNITY_USE_SHCOEFFS_ARRAYS
         #endif
-        #if defined(UNITY_PASS_DEFERRED) && defined(SHADOWS_SHADOWMASK) && (UNITY_ALLOWED_MRT_COUNT > 4)
+        #if (SHADERPASS == SHADERPASS_GBUFFER) && defined(SHADOWS_SHADOWMASK)
```
seblagarde: It is not allowed to used this define here: SHADERPASS == SHADERPASS_GBUFFER
This is specific to HDRP and shouldn't be present in shader library

seblagarde: UNITY_USE_PROBESOCCLUSION_ARRAY => does this mean that shadowmak  is supported by instancing as well ?

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -324,10 +322,10 @@
     static float4x4 unity_MatrixITMV_Instanced;
     void UnitySetupCompoundMatrices()
     {
-        unity_MatrixMVP_Instanced = mul(UNITY_MATRIX_VP, unity_ObjectToWorld);
-        unity_MatrixMV_Instanced = mul(UNITY_MATRIX_V, unity_ObjectToWorld);
+        unity_MatrixMVP_Instanced = mul(UNITY_MATRIX_VP, UNITY_MATRIX_M);
```
sebastienlagarde: @phi-lira Do you have any used for these matrix ? We don't use it in HD (and we must not use them as we provide our own matrix from C#). I think we should remove this call  (UnitySetupCompoundMatrices and its content

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
sebastienlagarde: (SHADERPASS == SHADERPASS_GBUFFER) is not allowed here as we are in shader library and this is HDRP scpecific.

SRP (HD or other) handle shadow mask differently than legacy. There is no distinction between deferred and forward shadow mask for SRP. so this code should only test defined(SHADOWS_SHADOWMASK)

Note: I am wondering is shadow mask instancing is supported ? Seems that there is some work done on it, will ask Yao

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
phi-lira: I don't use them. not even the UNITY_MATRIX_MVP as it's not performance wise.

Wadarass:
@sebastienlagarde Updated with the changes requested.  Just nuked all the compound matrices from UnityInstancing.hlsl and removed Shaderpass reference.

PR #808, opened 2018-01-17
Prototype/decals drawinstance

Paulm-Unity:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Decal/DecalProjectorComponent.cs
```diff
@@ -10,17 +10,11 @@ public class DecalProjectorComponent : MonoBehaviour
         private static readonly int m_WorldToDecal = Shader.PropertyToID("_WorldToDecal");
```
sebastienlagarde: These are not required anymore right  ?
private static readonly int m_WorldToDecal = Shader.PropertyToID("_WorldToDecal");
          private static readonly int m_DecalToWorldR = Shader.PropertyToID("_DecalToWorldR");		          private static readonly int m_DecalToWorldR = Shader.PropertyToID("_DecalToWorldR");

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Decal/DecalData.hlsl
```diff
@@ -4,7 +4,7 @@
 #include "CoreRP/ShaderLibrary/Packing.hlsl"
 #include "CoreRP/ShaderLibrary/Sampling/SampleUVMapping.hlsl"
 
-void GetSurfaceData(float2 texCoordDS, out DecalSurfaceData surfaceData)
+void GetSurfaceData(float2 texCoordDS, float3x3 decalToWorld, out DecalSurfaceData surfaceData)
```
sebastienlagarde: hmm... I still need to think about it, but I think we could apply decalToWorld on normal outside of this function, i.e in shader pass DBuffer. I think it make more sense. Let's think about it for next PR

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/ShaderPassDBuffer.hlsl
```diff
@@ -4,30 +4,68 @@
...
+
+    FragInputs input = UnpackVaryingsMeshToFragInputs(packedInput.vmesh);
+    PositionInputs posInput = GetPositionInput(input.positionSS.xy, _ScreenSize.zw);
 
 	float d = LOAD_TEXTURE2D(_MainDepthTexture, posInput.positionSS).x;
 	UpdatePositionInput(d, UNITY_MATRIX_I_VP, UNITY_MATRIX_VP, posInput);
 
-	float3 positionWS = posInput.positionWS;
-	float3 positionDS = mul(_WorldToDecal, float4(positionWS, 1.0f)).xyz;
+	float3 positionWS = GetAbsolutePositionWS(posInput.positionWS);
```
sebastienlagarde: add a comment that you go in absolute position only to go in decal space, so there is no loss of precision here

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/VertMesh.hlsl
```diff
@@ -107,8 +107,9 @@ VaryingsMeshType VertMesh(AttributesMesh input)
     float3 normalWS = float3(0.0, 0.0, 0.0); // We need this case to be able to compile ApplyVertexModification that doesn't use normal.
 #endif
 
+	float4 tangentWS = float4(0.0, 0.0, 0.0, 0.0);
```
sebastienlagarde: why do you need to do that ? if ATTRIBUTES_NEED_TANGENT is not define it is expected to not use tangentWS

Paulm-Unity: Because to get the 3 interpolators, that I need, I define 

#define VARYINGS_NEED_POSITION_WS
#define VARYINGS_NEED_TANGENT_TO_WORLD

So if normalWS is not declared it fails compiler here

  #ifdef VARYINGS_NEED_TANGENT_TO_WORLD
    output.normalWS = normalWS;

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Decal/DecalProjectorComponent.cs
Paulm-Unity: That is right, this can be removed

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/VertMesh.hlsl
seblagarde: do you mean tangentWS ? or  normalWS  ?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/ShaderPassDBuffer.hlsl
Paulm-Unity: ok

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Decal/DecalData.hlsl
Paulm-Unity: Sure we can do that. For Lit shader it all happens inside GetSurfaceAndBuiltinData(...)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/ShaderPass/VertMesh.hlsl
Paulm-Unity: Sorry, I meant tangetWS, you are right.

PR #842, opened 2018-01-21
Update material features performance

sebastienlagarde:
This PR bring an optimization to the material classification.
The goal is that in case of the worst case where we have all material at the same time, we should test feature at pixel level and not at tile. Because it is very rare to have "really" all material.s It is more a combination that we have not handled (like SSS and clear coat => no anisotropy, no iridescence). With tile feature we do everything (like we consider to always have anisotropy even if there is none).
So this PR should favor case where we are coherent but don't handle this variant and it finish in catch all.

Test: 1920x1080, ps4

Variant0 only standard material no light
Before PR: 0.8930ms . After PR: 0.8542ms => save 0.3ms
Variant0 complex overlapping of material to get variant 26, no light
Before PR: 1.0361ms . After PR: 1.0039ms => save 0.03ms

Variant26 only SSS + clear coat material directional light
Before PR: 2.2322ms . After PR: 2.2089ms => save 0.02ms
Variant26 complex overlapping of material material directional light
Before PR: 1.1190ms . After PR: 1.1284ms => lose 0.0094ms

Conclusion: The difference for Variant26 is not huge, but the one for variant 0 make sense. The lose is because we have complex case that would benefit from coherency of tile approach but is also a case that we never have. Variant26 only SSS + clear coat is what we could have in real case scenario

Classification of the complex scenario (mix of aniso, clear coat, sss, standard, sss + clear coat, aniso + clear coat
image

PErf report of deferred.compute
[Deferred.perf.xlsx](https://github.com/Unity-Technologies/ScriptableRenderPipeline/files/1693271/Deferred.perf.xlsx)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -649,6 +649,12 @@ uint DecodeFromGBuffer(uint2 positionSS, uint tileFeatureFlags, out BSDFData bsd
     pixelFeatureFlags |= tileFeatureFlags & (pixelHasIridescence  ? MATERIALFEATUREFLAGS_LIT_IRIDESCENCE           : 0);
     pixelFeatureFlags |= tileFeatureFlags & (pixelHasClearCoat    ? MATERIALFEATUREFLAGS_LIT_CLEAR_COAT            : 0);
 
+    // If tileFeatureFlags == UINT_MAX it mean we are call from deferred.shader (or a debug mode) that have no classification
```
EvgeniiG: Given that deferred lighting doesn't have usable performance without material classification, we should probably just retire Deferred.shader.

sebastienlagarde: it still useful for debug as it have allowed many time to find problem :)

PR #849, opened 2018-01-22
Don't check textures slots in decal material validation

Paulm-Unity:
Add gizmo for selecting decals in scene view
Add material fold out in the DecalProjectorComponent inspector


sebastienlagarde:
I added a default decal material + support on amterial to disable decal received

PR #851, opened 2018-01-23
Add transparent priority support

sebastienlagarde:
This PR update the HDRenderQueue class. Main goal is to support priority for Transparent surface (i.e being able to force order of transparent via the RenderQueue). Value can goes from -100 to 100 for the transparent priority and define the "neutral" value of 0 as current TransparentQueue (so compatible with current Unity). This transparent Queue priority is added to the current RenderQueue allowing to change betwene preRefractoin and Transparent without trouble.

- Move all RenderQueueRange of the code in the HDRenderQueue
- Add min/max for preRefraction and Transparent

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/Material/Unlit/BaseUnlitUI.cs
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/Material/Unlit/BaseUnlitUI.cs
@@ -30,6 +30,8 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
             public static GUIContent transparentDepthPrepassEnableText = new GUIContent("Enable transparent depth prepass", "It allow to to fill depth buffer to improve sorting");
             public static GUIContent transparentDepthPostpassEnableText = new GUIContent("Enable transparent depth postpass", "It allow to fill depth buffer for postprocess effect like DOF");
             public static GUIContent transparentBackfaceEnableText = new GUIContent("Enable back then front rendering", "It allow to better sort transparent mesh by first rendering back faces then front faces in two separate drawcall");
+
+            public static GUIContent transparentQueuePriorityText = new GUIContent("Transparent Queue Priority", "Allow to define priority (from -100 to +100) to solve sorting issue with transparent");
             public static GUIContent enableTransparentFogText = new GUIContent("Enable fog", "Enable fog on transparent material");
             public static GUIContent enableBlendModePreserveSpecularLightingText = new GUIContent("Blend preserve specular lighting", "Blend mode will only affect diffuse lighting, allowing correct specular lighting (reflection) on transparent object");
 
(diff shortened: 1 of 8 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/Material/Unlit/BaseUnlitUI.cs
```diff
@@ -30,6 +30,8 @@ protected static class StylesBaseUnlit
             public static GUIContent transparentDepthPrepassEnableText = new GUIContent("Enable transparent depth prepass", "It allow to to fill depth buffer to improve sorting");
             public static GUIContent transparentDepthPostpassEnableText = new GUIContent("Enable transparent depth postpass", "It allow to fill depth buffer for postprocess effect like DOF");
             public static GUIContent transparentBackfaceEnableText = new GUIContent("Enable back then front rendering", "It allow to better sort transparent mesh by first rendering back faces then front faces in two separate drawcall");
+
+            public static GUIContent transparentQueuePriorityText = new GUIContent("Transparent Queue Priority", "Allow to define priority (from -100 to +100) to solve sorting issue with transparent");
```
JulienIgnace-Unity: Maybe "Transparent Sorting Priority" instead of "Queue" sounds more understandable for artists.
Missing a few 's'
Allow(s)
issue(s)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/Material/Unlit/BaseUnlitUI.cs
```diff
@@ -83,6 +85,8 @@ public enum BlendMode
         protected const string kTransparentDepthPostpassEnable = "_TransparentDepthPostpassEnable";
         protected MaterialProperty transparentBackfaceEnable = null;
         protected const string kTransparentBackfaceEnable = "_TransparentBackfaceEnable";
+        protected MaterialProperty transparentQueuePriority = null;
+        protected const string kTransparentQueuePriority = "_TransparentQueuePriority";
```
JulienIgnace-Unity: Maybe rename the shader variable as well.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderQueue.cs
```diff
@@ -2,14 +2,40 @@
...
-        Background = UnityEngine.Rendering.RenderQueue.Background,
-        Geometry = UnityEngine.Rendering.RenderQueue.Geometry,
-        AlphaTest = UnityEngine.Rendering.RenderQueue.AlphaTest,
-        GeometryLast = UnityEngine.Rendering.RenderQueue.GeometryLast,
-        PreRefraction = 2750,
-        Transparent = UnityEngine.Rendering.RenderQueue.Transparent,
-        Overlay = UnityEngine.Rendering.RenderQueue.Overlay
+        public const int k_TransparentPriorityQueueRange = 100;
+
+        public enum Priority
```
JulienIgnace-Unity: Maybe we could make these Priority match the names we use in HDRP (opaque instead of geometry, etc.)
Sure it won't be like legacy but do we care about that?

PR #855, opened 2018-01-24
Split resources into editor and runtime resources

stramit:
This allows updating of runtime resources between versions of unity in a nice and easy to update way.


phi-lira:
I was wondering that we should add a version number to these resource files. This way, if users create their own resources file, (nothing is preventing them to do so) and we upgrade to a new version we could provide an upgrade path to keep user custom data while upgrading with new resources.

PR #868, opened 2018-01-26
In LW the color target should be created in SRGB

stramit:


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -675,6 +675,7 @@ private void SetupIntermediateResourcesSingle(CommandBuffer cmd, FrameRenderingC
             }
 
             RenderTextureDescriptor colorRTDesc = new RenderTextureDescriptor(rtWidth, rtHeight, m_ColorFormat, kDepthStencilBufferBits);
+            colorRTDesc.sRGB = true;
```
phi-lira: You should probably check for the active colorspace since LW supports gamma. Look for GetColorInActiveSpace  function.

PR #872, opened 2018-01-26
HDRP + MSAA: Initial infrastructure

robbiesri:
This PR adds the initial infrastructure needed to enable MSAA with HDRP.  Multiple passes are disabled currently, but this at least lands the base support.  MSAA can be enabled, with the MSAA sample count sourced from QualitySettings.antiAliasing (will probably change this interface in the future).

The main things that are added in this PR are:
* Properly enable FrameSettings.enableMSAA, and make it ForwardOnly
* Make sure RTs that will use MSAA don't enable UAV/enableRandomWrite.  This does mean some passes will need to add Draw support in order to use MSAA, because compute can't be used.
* Detect sample count changes as part of a resolution change
* Enable correct MSAA passes in clustered tiled light generation
* Turn off stuff not needed for first pass

**Missing Functionality**
Currently, I have disabled these passes, which I will start working to re-add:

- No decals
- No distortion or refraction (need to generate Gaussian color pyramids from MSAA color)
- No transparent (need to generate Gaussian depth from MSAA depth)
- No post-process (need to hand off resolved buffer)
- No SSAO (hand off resolved buffer or add MSAA support to SSAO paths?)
- No SSS (will use Jimenez because Disney needs UAV.  I also hit issues with R8 + UAV +MSAA allocation, which is also needed for Disney)
- Deferred Shadow Buffer is currently 1x

I also need to do some work to prevent spurious resolves.  I probably need to add bindMS to all RTs, and resolve manually as needed.

**Bonus meta-note**
There is the question of "Why push out this PR if not all passes are supported?"  Well, the HDRP repo moves so fast, there are hazards to falling too far behind, and encountering large merge pains.  Other PRs have used this practice of 'partial functionality PRs' in order to keep up with the repo.


robbiesri:
Extra question: currently, I have almost all the 'disabling' in one method: FrameSettings.ConfigureMSAADependentSettings.  However, I currently disable FrameSettings.enableFptlForForwardOpaque inside LightLoopSettings.InitializeLightLoopSettings.  

I don't like that I'm doing enableMSAA checks in two different places.  Seems like it should just be in one place, but the original code is doing the enableMSAA check inside InitializeLightLoopSettings.  If nobody minds, I'll just move the check to disable enableFptlForForwardOpaque to ConfigureMSAADependentSettings, to keep everything in one place.  However, if people want me to leave it, I won't put up a fight.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -612,7 +612,20 @@ public void NewFrame(FrameSettings frameSettings)
...
+                }
+                else
+                {
+                    kernelName += "_NoDepthRT";
+                }
+                if (m_FrameSettings.lightLoopSettings.enableBigTilePrepass)
+                    kernelName += "_SrcBigTile";
+
                 s_GenListPerVoxelKernel = buildPerVoxelLightListShader.FindKernel(kernelName);
             }
```
Chman: String concatenation here will add GC pressure as it's called on every frame. You'll need to pre-compute string values to avoid that (or do it manually with an unsafe block and stackalloc - not very practical).

robbiesri:
Oh, this is great advice, thanks Thomas. I don’t know much about this stuff
in C#, whew

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
robbiesri: (try again)
Thanks for the assist here Thomas, will fix!

robbiesri: Sigh, this will make the code look so weird...

robbiesri: Ok, I changed it with 7f34c4a.  Instead, I pre-generate the kernel names, and store them into a 2D array.  Then I index into the array with the active BigTile state and preferred depth source.

robbiesri:
Hey @Chman, I changed the cluster kernel name selection code.  I now stash the kernel names in a static 2D array.  Then each frame generates an index to look into the array.

If this was overdoing it, let me know, and I can just slap in a big if/else statement.  But I think this works ok.  It also mirrors the organization of the kernel names in lightlistbuild-clustered.compute.

sebastienlagarde:
> "Why push out this PR if not all passes are supported?"
Perfectly fine, we know that thing are moving a lot :). and I encourage people to make frequent PR instead of big one .

>Missing Functionality
>Make sure RTs that will use MSAA don't enable UAV/enableRandomWrite
Main question:  can you confirm that UAV can't work with MSAA RT  ? Seems that it is the case.

We will indeed need to deal what we do for each of the passes. As this can be a problem. I don't provide solutoin yet, but will help you for each of the passes. Just giving information for now.

>No decals
Decals will use UAV in pixels shader for optimization. Mean we should disable this optimization with MSAA...

>No distortion or refraction (need to generate Gaussian color pyramids from MSAA color)
ok
Note that we currently rewrite a gaussian pyramid compute for HD (that will not perform square screen but will be really proportional to screen.

>No transparent (need to generate Gaussian depth from MSAA depth)
>No post-process (need to hand off resolved buffer)
>No SSAO (hand off resolved buffer or add MSAA support to SSAO paths?)
Nothing to add yet on this, same problem with directional shadow (but we can disable directional shadow)

>No SSS (will use Jimenez because Disney needs UAV. I also hit issues with R8 + UAV +MSAA allocation, which is also needed for Disney)
We can't rely on Jimenez. We have disable the dynamic switch for performance reasons (and also parameters are not the same than Disney). We will try to find a solution for this.

Just a first though, but it seems that we should go for the MSAA for geometry pass only and do all the lighting for 1X (but mean we need to average lighting inside shader (like for a deferred path). Need to think more about that, because if we go this way it will bloat the coat a bit, which we what we were wanting to avoid.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
```diff
@@ -171,7 +171,16 @@ public void Update(PostProcessLayer postProcessLayer, FrameSettings frameSetting
                 tempDesc = new RenderTextureDescriptor(camera.pixelWidth, camera.pixelHeight);
             }
 
-            tempDesc.msaaSamples = 1; // will be updated later, deferred will always set to 1
+            if (frameSettings.enableMSAA)
+            {
+                // this is already pre-validated to be a valid sample count by InitializeFrameSettings
+                var sampleCount = QualitySettings.antiAliasing;
```
sebastienlagarde: >var sampleCount = QualitySettings.antiAliasing;
We shouldn't realy on QualtiySettings here (this settings is not accessible anyway in HD).
Instead we must have this settings on FrameSettings for each individual camera.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipeline/FrameSettings.cs
```diff
@@ -169,12 +169,41 @@ public static void InitializeFrameSettings(Camera camera, RenderPipelineSettings
             aggregate.enableTransparentObjects = srcFrameSettings.enableTransparentObjects;
 
             aggregate.enableMSAA = srcFrameSettings.enableMSAA && renderPipelineSettings.supportMSAA;
+            if (QualitySettings.antiAliasing < 1)
```
sebastienlagarde: See comment above, we will not use QualtiySettings

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
JulienIgnace-Unity: This might actually be a problem with the RTHandle framework. Having the possibility to change the number of samples per camera means that we need to store many different version of the MSAA texture for each number of samples.
Do we really need this? IMO we should have one MSAA quality setting and then cameras can just switch MSAA on and off.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipeline/FrameSettings.cs
```diff
@@ -169,12 +169,41 @@ public static void InitializeFrameSettings(Camera camera, RenderPipelineSettings
...
         }
 
+        public void ConfigureMSAADependentSettings()
+        {
+            if (enableMSAA)
+            {
+                // Initially, MSAA will only support forward
+                enableForwardRenderingOnly = true;
+
+                // TODO: Should we disable enableFptlForForwardOpaque in here, instead of in InitializeLightLoopSettings?
```
sebastienlagarde: Lightloop should be seen as a module that could be replace (by another lighting architecture), so it is perfectly fine to check MSAA inside InitializeLightLoopSettings. I get that it doesn't gather all MSAA settings at one place but it gather the lightloop configuration at one place (else it will be split,  so it is MSAA settings split or lightloop setting split and it make more sense to me to split MSAA).

sebastienlagarde:
>Extra question: currently, I have almost all the 'disabling' in one method: 
See my comment of the code but it is fine to have MSAA settings in LightLoop as Lightloop is a separate module that could be change.

sebastienlagarde:
Hey, so we discussed with Julien, and here is what we should adopt for the MSAA pass (for perf reasons and code maintainability).

We resolve the MSAA after the Forward Opaque Pass and before the SubsurfaceScatteringPass.
So before SubsurfaceScatteringPass we resolve, so SSS, rough refraction, distortion and all postprocess pass will work with MSAA (for postprocess the target is resolved anyway before MSAA as the stack doesn't handle MSAA...).

So we only need to handle: Decal, deferred directional shadow and SSAO (SSAO could stay at original resolution).

The downside is that we will perform the tonemap after the resolve instead of before which is bad (Still it will be the case anyway due to postprocess). And I am wondering if we could do something smarter than an average here. Like instead of doing average of value we convert with tonemap, calculate the weight, then inverse tone maps and apply the weight on HDR value (i.e what is done with TAA). Anyway, I think it is the way to go as MSAA on transparent is less critical (and uselss on particles).

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
robbiesri: Yes, absolutely.  I'm working on a PR now to add this field to the pipeline asset.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipeline/FrameSettings.cs
robbiesri: Ah ok, good point.  I'll leave it as is :)

robbiesri:
> Main question: can you confirm that UAV can't work with MSAA RT ? Seems that it is the case.
Yep.  I admit that I wasn't clever enough to discover this via inspection.  I was flipping on MSAA, but my targets were failing creation.  I dug into the native API calls, and saw that the resource configuration was 'invalid', at which point I realized that I had MSAA + UAV on the same target, and that was causing the failure.

Your comment made me investigate a bit further, and I think there might be certain configs on DX11.1-HW where you can create a UAV + MSAA enabled resource (https://docs.microsoft.com/en-us/windows-hardware/drivers/display/directx-feature-improvements-in-windows-8#span-idunorderedspanspan-idunorderedspanuavs-with-multi-sample-anti-alias-sample-access), but I need to read a bit more on what the particulars are.  It didn't work on my local AMD or NV test machines, but I wouldn't be surprised if some random setting wasn't properly configured.

robbiesri:
@sebastienlagarde I like the proposed plan for what passes to use with MSAA and not.  

re: tonemap adjustment idea
I think your idea for a smarter tonemap resolve makes sense, but how would we de-couple it from the general Post-process infrastructure?  Something like how SSAO is implemented now?  If MSAA is enabled, hoist out the tonemap stage of the Post-processing Stack, and process it separately from the other effects?

sebastienlagarde:
For tonemap we do like for TAA. We don't use the tonemap of postprocess but a cheaper one that is invertible. Base on log/exp. Goal is just to gather weight in a perceptual space.

PR #878, opened 2018-01-29
Refactoring of HDRP render texture usage using RTHandles

JulienIgnace-Unity:
Adding Rob/Paul/Fred because this PR changes pretty much everything wrt Render Texture handling so you will want to check :)

The goal of this PR is to replace all our usage of TemporaryRT and plain RenderTextures by RTHandle to allow more efficient memory allocation across frames and better debugging.

- The RTHandle framework has been modified to account for MSAA.
In order for render textures to be able to switch between MSAA On/Off depending on the FrameSettings, the framework can hold potentially two versions of each texture. Each set of textures (MSAA/Non-MSAA) has its own reference size so that we don't waste un-needed memory (for example, main rendering at 1080p MSAA / Reflection rendering at 256 resolution w/o MSAA, non-MSAA ref-size will stay at 256).
In order for a render texture to be MSAA ready, it needs to be declared as MSAA upon creation. The system will then switch as needed.
- A new set of "SetRenderTarget" methods has been added to HDUtils. These should be used for setting up auto-scaled render target (basically, all "camera space" targets in the render loop). They take a HDCamera parameter so that they can also set the correct viewport depending on the currently needed resolution.
- New blit methods "BlitCameraTexture" have also been added. In the same way as the new SetRenderTarget methods, they should be used when blitting any "camera space" target so that they take into account the viewport inside the render target.


robbiesri:
Thanks for all your work on this @JulienIgnace-Unity .  Overall, this looks awesome, and I look forward to trying out some of the functionality in HDRP.

I have one major question:
How does one handle manually resolved MSAA RTs with this setup?  The current plan is to resolve the color buffer before transparent rendering, but I can't quite tell how I'd be able to switch MSAA modes for just a singular RTH.  It doesn't feel like the current model really supports that (yet).

ScriptableRenderPipeline/Core/CoreRP/Resources/GPUCopy.compute
```diff
@@ -17,7 +17,7 @@ Texture2D<float4> _Source4;
 [numthreads(8, 8, 1)]
 void KSampleCopy4_1_x(uint2 dispatchThreadId : SV_DispatchThreadID)
 {
-    _Result1[dispatchThreadId] = SAMPLE_TEXTURE2D_LOD(_Source4, sampler_LinearClamp, float2(dispatchThreadId) * _Size.zw, 0.0).x;
+    _Result1[dispatchThreadId] = LOAD_TEXTURE2D(_Source4, dispatchThreadId).x;
```
sebastienlagarde: This shader is generated note: Update the generator

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/RenderLoopSettings/RenderPipelineSettingsUI.cs
```diff
@@ -73,8 +73,12 @@ static void Drawer_SectionPrimarySettings(RenderPipelineSettingsUI s, Serialized
             EditorGUILayout.PropertyField(d.supportSSR, _.GetContent("Support SSR"));
             EditorGUILayout.PropertyField(d.supportSSAO, _.GetContent("Support SSAO"));
             EditorGUILayout.PropertyField(d.supportDBuffer, _.GetContent("Support Decal Buffer"));
-            EditorGUILayout.PropertyField(d.msaaSampleCount, _.GetContent("MSAA Sample Count"));
+            EditorGUILayout.PropertyField(d.supportMSAAAntiAliasing, _.GetContent("Support MSAA Anti-Aliasing"));
+            EditorGUILayout.PropertyField(d.MSAASampleCount, _.GetContent("MSAA Sample Count"));
             EditorGUILayout.PropertyField(d.supportSubsurfaceScattering, _.GetContent("Support Subsurface Scattering"));
+            EditorGUILayout.PropertyField(d.supportsForwardOnly, _.GetContent("Support Forward Only"));
```
sebastienlagarde: remove the "s" in d.supports for coherency

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -405,15 +240,89 @@ public HDRenderPipeline(HDRenderPipelineAsset asset)
             FrameSettings.RegisterDebug("Default Camera", m_Asset.GetFrameSettings());
 #endif
 
-            m_DebugColorPickerRT = HDShaderIDs._DebugColorPickerTexture;
-            m_DebugFullScreenTempRT = HDShaderIDs._DebugFullScreenTexture;
+            InitializeRenderTextures();
```
sebastienlagarde: Request :), could you handle a support debug ? (so we can remove them on target)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -405,15 +240,89 @@ public HDRenderPipeline(HDRenderPipelineAsset asset)
...
+        void InitializeRenderTextures()
+        {
+            // Initial state of the RTHandle system.
+            // Tells the system that we will require MSAA or not so that we can avoid wasteful render texture allocation.
+            // TODO: Might want to initialize to at least the window resolution to avoid un-necessary re-alloc in the player
+            RTHandle.Initialize(1, 1, m_Asset.renderPipelineSettings.supportMSAAAntiAliasing, m_Asset.renderPipelineSettings.msaaSampleCount);
+
+            m_GbufferManager.CreateBuffers();
+
+            if(m_Asset.renderPipelineSettings.supportDBuffer)
```
sebastienlagarde: "if (" not if(

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -990,28 +842,39 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                         }
                         else
                         {
-                            using (new ProfilingSample(cmd, "Blit to final RT", CustomSamplerId.BlitToFinalRT.GetSampler()))
+                            using (new ProfilingSample(cmd, "Blit to final RT", CustomSamplerId.CopyDepthForSceneView.GetSampler()))
```
sebastienlagarde: why "CopyDepthForSceneView" ?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -990,28 +842,39 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
...
+                    // On subtlety here is that our buffer can be bigger than the camera one so we need to copy only the corresponding portion
+                    // (it's handled automatically by the copy shader because it uses a load in pxiel coordinates based on the target).
+                    // This copy will also have the effect of re-binding this depth buffer correctly for subsequent editor rendering.
+
+                    // NOTE: This needs to be done before the call to RenderDebug because debug overlays need to update the depth for the scene view as well.
+                    // Make sure RenderDebug does not change the current Render Target
                     if (camera.cameraType == CameraType.SceneView)
                     {
-                        cmd.SetRenderTarget(BuiltinRenderTextureType.CameraTarget, m_CameraDepthStencilBufferRT);
+                        using (new ProfilingSample(cmd, "Copy Depth For SceneView", CustomSamplerId.BlitToFinalRT.GetSampler()))
```
sebastienlagarde: insn't there an inversion of CopyDepthForSceneView and BlitToFinalRT for the profiling ?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1239,10 +1098,10 @@ void RenderGBuffer(CullResults cull, HDCamera hdCamera, ScriptableRenderContext
             using (new ProfilingSample(cmd, m_CurrentDebugDisplaySettings.IsDebugDisplayEnabled() ? "GBufferDebugDisplay" : "GBuffer", CustomSamplerId.GBuffer.GetSampler()))
             {
                 // setup GBuffer for rendering
-                CoreUtils.SetRenderTarget(cmd, m_GbufferManager.GetGBuffers(), m_CameraDepthStencilBufferRT);
+                HDUtils.SetRenderTarget(cmd, hdCamera, m_GbufferManager.GetBuffersRTI(enableShadowMask), m_CameraDepthStencilBuffer);
                 if (m_FrameSettings.enableDBuffer)
                 {
-                    m_DbufferManager.SetHTile(m_GbufferManager.gbufferCount, cmd);
+                    m_DbufferManager.SetHTile(m_GbufferManager.GetBufferCount(enableShadowMask), cmd);
                 }
```
sebastienlagarde: So, Robs point me that we must still bind something here for PS4 for the SetHTile else it crash.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Builtin/BuiltinData.cs
```diff
@@ -72,19 +72,19 @@ public static RenderTextureFormat GetShadowMaskBufferFormat()
             return RenderTextureFormat.ARGB32;
         }
 
-        public static RenderTextureReadWrite GetShadowMaskBufferReadWrite()
+        public static bool GetShadowMask_sRGBFlag()
```
sebastienlagarde: dont put _  => GetShadowMaskSRGBFlag

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/GBufferManager.cs
```diff
@@ -0,0 +1,71 @@
+using UnityEngine.Rendering;
```
sebastienlagarde: must not be in Lit but in material. GBuffer here is generic

ScriptableRenderPipeline/Core/CoreRP/Resources/GPUCopy.compute
sebastienlagarde: ok

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/RenderLoopSettings/RenderPipelineSettingsUI.cs
sebastienlagarde: done

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
sebastienlagarde: done

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
sebastienlagarde: done

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
sebastienlagarde: corrected

ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
sebastienlagarde: corrected

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Builtin/BuiltinData.cs
sebastienlagarde: done

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/GBufferManager.cs
sebastienlagarde: done

robbiesri:
What happens when this system handles multiple cameras?  SetReferenceSize will cause all of the previous cameras associated RTHs to get thrown away, and rebuilt to match the current camera, right?  I understand the base case is a single camera, but if the intention is to minimize RT allocations, the multi-cam case works around that.

Probably not needed for the first merge, but something to address in the future.  First idea that comes to mind is to have a RTH pool per HDCamera...

JulienIgnace-Unity:
@robbiesri The system solves exactly this multi-camera case actually :)
RTH will re-allocate only if the new reference size is bigger. It means that basically your main rendering drives the max size and then all other smaller rendering use only a viewport inside the main RTs.

robbiesri:
@JulienIgnace-Unity Ah, you're totally right, my bad.  I wasn't tracking the viewport management code nor the reference size correctly, boo to me.  Thanks!

PR #882, opened 2018-01-29
Updating UI for clear coat and transmission

sebastienlagarde:
This PR update the UI to support correctly clear coat and SSS/Transmission.

In various previous PR the lighting model and GBuffer layout have been change to material features.
The standard lighting model is GGX + disney diffuse and we can enable material features on top of it.

Clear coat can be enabled on any of the feature, but specular color, SSS/Transmission, Anisotropy, iridescence can't (due to GBuffer storage). This PR update the UI accordingly to match this behavior. In case we select Subsurface Scattering and Transmission we can select if we have both, transmission only or sss only.

Also in Diffusion profile we remove diffuse transmission none as it is not needed anymore.

There is two script in Internal -> Render Pipeline -> HDRenderPipeline provided for upgrading the material accordingly, as clear coat was setup to 1 before by default.


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/Material/Lit/BaseLitUI.cs
```diff
@@ -26,7 +26,8 @@ protected static class StylesBaseLit
             public static GUIContent enableMotionVectorForVertexAnimationText = new GUIContent("Enable MotionVector For Vertex Animation", "This will enable an object motion vector pass for this material. Useful if wind animation is enabled or if displacement map is animated");
 
             // Material ID
-            public static GUIContent materialIDText = new GUIContent("Material type", "Subsurface Scattering: enable for translucent materials such as skin, vegetation, fruit, marble, wax and milk.");
+            public static GUIContent materialIDText = new GUIContent("Material type", "Select a material feature to enable on top of regular material");
```
EvgeniiG: Maybe we should decide on the naming and start calling it "standard" or "base" material.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -1044,7 +1042,9 @@ LightTransportData GetLightTransportData(SurfaceData surfaceData, BuiltinData bu
 
 bool PixelHasSubsurfaceScattering(BSDFData bsdfData)
 {
-    return bsdfData.diffusionProfile != DIFFUSION_PROFILE_NEUTRAL_ID &&  bsdfData.subsurfaceMask != 0 &&  HasFeatureFlag(bsdfData.materialFeatures, MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING);
+    // bsdfData.subsurfaceMask != 0 allow if sss is enabled for this pixels (i.e like per pixels feature) as in deferred case MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING alone is not sufficient
```
EvgeniiG: While this works, it is very inefficient, as you force both the diffusion profile and the subsurface mask to stay in VGPR until the end of the shader. We'll need to make a boolean for this.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -1940,7 +1940,9 @@ void PostEvaluateBSDF(  LightLoopContext lightLoopContext,
 #endif
 
     float3 modifiedDiffuseColor;
-    if (HasFeatureFlag(bsdfData.materialFeatures, MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING))
+    // bsdfData.subsurfaceMask != 0 allow if sss is enabled for this pixels (i.e like per pixels feature) as in deferred case MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING alone is not sufficient
+    // but keep testing MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING for forward case
+    if (HasFeatureFlag(bsdfData.materialFeatures, MATERIALFEATUREFLAGS_LIT_SUBSURFACE_SCATTERING) && bsdfData.subsurfaceMask != 0)
```
EvgeniiG: Same here.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/Lit/Lit.hlsl
sebastienlagarde: I am wondering if the compiler is able to figure it out. But yes, I can make the change in a next PR

PR #883, opened 2018-01-30
Pack normal on 1212bit in 888

sebastienlagarde:
Change normal packing to 1212 octahedral encode in 888 for better quality.

### DIFF
--- a/ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Packing.hlsl
+++ b/ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Packing.hlsl
@@ -489,6 +489,29 @@ real UnpackFloatFromR8G8(real2 f)
     return PackShort(cb);
 }
 
+// Pack float2 (each of 12 bit) in 888
+real3 PackFloat2To888(real2 f)
+{
+    uint2 i = (uint2)(f * 4095.0);
+    uint2 hi = i >> 8;
+    uint2 lo = i & 255;
+    // 8 bit in lo, 4 bit in hi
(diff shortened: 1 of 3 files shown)


ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Packing.hlsl
```diff
@@ -489,6 +489,29 @@ real UnpackFloatFromR8G8(real2 f)
     return PackShort(cb);
 }
 
+// Pack float2 (each of 12 bit) in 888
+real3 PackFloat2To888(real2 f)
+{
+    uint2 i = (uint2)(f * 4095.0);
```
EvgeniiG: Shouldn't we round here?

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Packing.hlsl
```diff
@@ -489,6 +489,29 @@ real UnpackFloatFromR8G8(real2 f)
...
+    uint2 hi = i >> 8;
+    uint2 lo = i & 255;
+    // 8 bit in lo, 4 bit in hi
+    uint3 cb = uint3(lo, hi.x | (hi.y << 4));
+
+    return cb / 255.0;
+}
+
+// Unpack 2 float of 12bit packed into a 888
+real2 Unpack888ToFloat2(real3 x)
```
EvgeniiG: I think we should be using an UINT target for this G-Buffer, This packing code is quite expensive, and I think it misses rounding.

sebastienlagarde:
We have discussed about it. It is fined. 
Improvement in picture:

image

To

image

PR #886, opened 2018-01-30
VR keywords are injected on build if the platform is needed. No need …

stramit:
…to add a multi-compile.


robbiesri:
Thanks for removing these. I assume @aet had visibility into this?

PR #892, opened 2018-02-01
Use fullscreen depth/color pyramids

fredpointzero:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/HDRenderPipeline.cs
```diff
@@ -1844,14 +1743,13 @@ void InitAndClearBuffer(HDCamera hdCamera, bool enableBakeShadowMask, CommandBuf
...
                     // Color and depth pyramids
-                    m_GaussianPyramidColorBufferDesc = BuildPyramidDescriptor(hdCamera, PyramidType.Color, m_FrameSettings.enableStereo);
+                    m_BufferPyramid.Initialize(hdCamera, m_FrameSettings.enableStereo);
                     cmd.ReleaseTemporaryRT(m_GaussianPyramidColorBuffer);
-                    cmd.GetTemporaryRT(m_GaussianPyramidColorBuffer, m_GaussianPyramidColorBufferDesc, FilterMode.Trilinear);
-
-                    m_DepthPyramidBufferDesc = BuildPyramidDescriptor(hdCamera, PyramidType.Depth, m_FrameSettings.enableStereo);
+                    cmd.GetTemporaryRT(m_GaussianPyramidColorBuffer, m_BufferPyramid.colorRenderTextureDescriptor, FilterMode.Trilinear);
 
+                    m_BufferPyramid.Initialize(hdCamera, m_FrameSettings.enableStereo);
```
robbiesri: I'm very late to the party here, but isn't this a duplicate of the call on 1746?  It won't cause an error, but BufferPyramid.Initialize only needs to be called once, right?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipelineResources/BufferPyramid.cs
```diff
@@ -0,0 +1,209 @@
...
+            {
+                Debug.LogWarningFormat("Cannot compute all mipmaps of the depth pyramid, max texture size supported: {0}", (2 << m_DepthPyramidMips.Length).ToString());
+                lodCount = m_DepthPyramidMips.Length;
+            }
+
+            cmd.ReleaseTemporaryRT(m_DepthPyramidMips[0]);
+
+            depthPyramidDesc.sRGB = false;
+            depthPyramidDesc.enableRandomWrite = true;
+            depthPyramidDesc.useMipMap = false;
```
robbiesri: It would be kinda nice if the RenderTextureDescriptor modifying code here in RenderDepthPyramid and in RenderColorPyramid was turned into some helper function like MakePyramidLevelDesc(sourceDesc).  And if the names of these descs indicated that they were being used for single-level allocation, that would also be cool (e.g. depthPyramidSingleLevelDesc).

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipelineResources/BufferPyramid.cs
```diff
@@ -0,0 +1,209 @@
...
+                    Mathf.CeilToInt(colorPyramidDesc.width / 8f),
+                    Mathf.CeilToInt(colorPyramidDesc.height / 8f),
+                    1);
+                cmd.CopyTexture(m_ColorPyramidMips[i + 1], 0, 0, targetTexture, 0, i + 1);
+
+                last = m_ColorPyramidMips[i + 1];
+            }
+
+            for (int i = 0; i < lodCount; i++)
+                cmd.ReleaseTemporaryRT(m_ColorPyramidMips[i + 1]);
```
robbiesri: I forgot to ask this when this code was in HDRenderPipeline, but why don't we use m_ColorPyramidMips[0]?  I get that the CopyTexture code makes a lot more sense when we copy from mip array i+1 to target texture mip level i+1.  I guess it's not really a big deal, it's just burning one extra int in the array.

robbiesri:
Post-merge, but I'm super happy to see this change.  Thanks for doing this @fredpointzero

PR #901, opened 2018-02-05
Add support of the ambient probe to volumetric lighting

EvgeniiG:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
```diff
@@ -338,6 +342,146 @@ static Vector4 ComputeLogarithmicDepthDecodingParams(float nearPlane, float farP
         return depthParams;
     }
 
+    public unsafe struct ZonalHarmonicsL2
+    {
```
sebastienlagarde: move all zonal harmonic in their own mathSH file

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
```diff
@@ -338,6 +342,146 @@ static Vector4 ComputeLogarithmicDepthDecodingParams(float nearPlane, float farP
...
+        SphericalHarmonicsL2 finalSH = PremultiplySH(Convolve(probeSH, phaseZH));
+
+        // Reorder coefficients in the MAD form:
+        // HornerForm[c_0 + c_1 y + c_2 z + c_3 x + c_4 x y + c_5 y z + c_6 (3 z^2 - 1) + c_7 x z + c_8 (x^2 - y^2)]
+        // = z (3 c_6 z + c_7 x + c_2) + y (-c_8 y + c_5 z + c_4 x + c_1) + x (c_8 x + c_3) + (c_0 - c_6)
+        Vector4[] coeffs = new Vector4[9];
+
+        const int r = 0, g = 1, b = 2;
+
+        coeffs[0] = new Vector3(finalSH[r, 0], finalSH[g, 0], finalSH[b, 0])
```
sebastienlagarde: What not use the traditional pack 7 vector4 coefficint of Peter Pike Sloan ? (http://www.realtimerendering.com/resources/shaderx/Tips_and_Tricks_with_DirectX_9.pdf - there is source with hlsl somewhere but don't remind))

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/Resources/VolumetricLighting.compute
```diff
@@ -16,14 +16,15 @@
 #if (SHADEROPTIONS_VOLUMETRIC_LIGHTING_PRESET == 1)
     // E.g. for 1080p: (1920/8)x(1080/8)x(128) =  4,147,200 voxels
     #define VBUFFER_TILE_SIZE   8
-    #define VBUFFER_SLICE_COUNT 128
+    #define VBUFFER_SLICE_COUNT 64
```
sebastienlagarde: question: Why this isn't deduce automatically from current resolution. For example on xbox if we switch to 900p should be less

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
```diff
@@ -278,9 +282,9 @@ static int ComputeVBufferSliceCount(VolumetricLightingPreset preset)
         switch (preset)
         {
             case VolumetricLightingPreset.Normal:
-                return 128;
+                return 64;
```
sebastienlagarde: this should be constant that match the define in hlsl, generate hlsl constant from it (or from screen size)

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
```diff
@@ -209,11 +209,13 @@ public void Update(PostProcessLayer postProcessLayer, FrameSettings frameSetting
         }
 
         // Warning: different views can use the same camera!
-        public int GetViewID()
+        public long GetViewID()
```
sebastienlagarde: I don't understand the need for long. And there is no comment about why this is needed

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
EvgeniiG: Done.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
EvgeniiG: Done.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
EvgeniiG: Because Camera.GetInstanceID() returns an int, and the range of returned values is not specified.
We use 0 for non-game cameras, and -1 for invalid values.
Therefore, we store an int inside a long, which gives us the extra range we need.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/Resources/VolumetricLighting.compute
EvgeniiG: The slice count does not depend on the resolution.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
EvgeniiG: Indeed, presets should probably generate their own HLSL files.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/Resources/VolumetricLighting.compute
sebastienlagarde: my mistake

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
sebastienlagarde: To do in next PR

PR #952, opened 2018-02-17
Fixed TAA only partially working in edit mode

Chman:
4be911b broke TAA in the game view. Unless you entered play mode the projection matrix wouldn't get jittered but TAA would still apply. This fix reverts to the older behavior where you'd see the effect of TAA after a the game view refreshes for a few frames.

Any particular reason why this change was made in the first place? Am I overlooking something?

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs
@@ -165,8 +165,7 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                 isFirstFrame = false;
             }
 
-            const uint taaFrameCount = 8;
-            taaFrameIndex = taaEnabled ? (uint)Time.renderedFrameCount % taaFrameCount : 0;
+            taaFrameIndex = taaEnabled ? (uint)postProcessLayer.temporalAntialiasing.sampleIndex : 0;
             taaFrameRotation = new Vector2(Mathf.Sin(taaFrameIndex * (0.5f * Mathf.PI)),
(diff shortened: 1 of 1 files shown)


Chman:
Btw, [this line of code](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/e81b8c38df3e58f52f40f88e93cb982af6dce571/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Camera/HDCamera.cs#L169) will only work properly in play mode as `Time.renderedFrameCount` is fixed in edit mode. I should probably expose an API in TAA to grab the current frame index.

Chman:
That should do it. Don't forget to update the postfx submodule.

EvgeniiG:
The reason it was changed because frame counters are broken in the Game View, unless you enter the Play Mode. You get an arbitrary number of updates (up to 30), then the frame counter gets stuck. Why/how your TAA works after this point, I have no idea (I suspect poorly, or not at all). You cannot get reliable temporal integration with a frozen frame counter.

If you think it works well enough for your needs, feel free to push this change.

Chman:
Our TAA doesn't rely on the frame counter provided by Unity because it's (as you experienced) unreliable outside of play mode, so it uses its own - basically increment after each rendered frame, as it should be. It worked fine since its introduction, however I got complains from artists recently about it "not working properly unless you switch to play mode" which kills their workflow velocity. This is what this PR fixes and it also sends the correct `taaFrameIndex` to shaders even in edit mode, which will in turn also fix other temporal effects using it in edit mode.

The reason why TAA was "half working" was because it wasn't disabled correctly in HDRP, that `taaEnabled` variable in `HDCamera` only controls the generation of the jittered projection matrix. So it resolved using the same unjittered frame.

EvgeniiG:
Good to see we have a working frame counter now, even if it's TAA-specific. :-)

PR #968, opened 2018-02-21
Fix shader graph submodule

pbbastian:
I accidentally added the Shader Graph submodule using SSH URL. This seems to fail if you do not have a GitHub account.


pbbastian:
hold on, don't merge this in yet

pbbastian:
Should be good to merge in :)

PR #970, opened 2018-02-21
Debug menu refactor

Chman:


ScriptableRenderPipeline/Core/CoreRP/Camera/CameraSwitcher.cs
```diff
@@ -48,7 +50,18 @@ void OnEnable()
...
+            var panel = DebugManager.instance.GetPanel("Camera", true);
+            panel.children.Add(m_DebugEntry);
+        }
+
+        void OnDisable()
+        {
+            if (m_DebugEntry != null && m_DebugEntry.panel != null)
+            {
+                var panel = m_DebugEntry.panel;
+                panel.children.Remove(m_DebugEntry);
```
JulienIgnace-Unity: Is the panel destroy when there are no more children?

ScriptableRenderPipeline/HDRenderPipeline/HDRP/RenderPipeline/FrameSettings.cs
```diff
@@ -234,85 +206,64 @@ public void ConfigureStereoDependentSettings()
...
-            DebugMenuManager.instance.RemoveDebugItem(menuName, kEnableAsyncCompute);
-
-            DebugMenuManager.instance.RemoveDebugItem(menuName, kEnableOpaqueObjects);
-            DebugMenuManager.instance.RemoveDebugItem(menuName, kEnableTransparentObjects);
-
-            DebugMenuManager.instance.RemoveDebugItem(menuName, kEnableMSAA);
-            DebugMenuManager.instance.RemoveDebugItem(menuName, kEnableShadowMask);
+            var panel = DebugManager.instance.GetPanel(menuName);
+
+            if (panel != null)
```
JulienIgnace-Unity: Can't we just remove the panel altogether. Would be shorter to with the same effect. And no need to keep around the debug entries.

PR #977, opened 2018-02-22
Added a first set of physical settings to HDCamera

Chman:
Aperture / Shutter Speed / ISO for now.
For Depth of Field control we'll need Focal Length (which will control FOV or the other way around) and Film size as well as a Priority setting so it can be overridden in Postfx volumes for special cases if needed.


robbiesri:
Maybe not the most appropriate place for this comment, but sometime in the future, there should be some sort of warning about modifying physical camera properties while using stereo.  They shouldn't be taken into consideration for stereo rendering.

sebastienlagarde:
Good point that for stereo we need to be careful. 
The purpose of these added parameters here is different than adjusting focal etc... Here the purpose is to calculate exposure for the lighting (now that we have physical light unit) correctly, and to link the MB and DOF to these parameters.

For focal lenght etc... this will come later and indeed require some adjustement for VR.

PR #978, opened 2018-02-22
Stereo-ized HDRP Lighting Update (WIP)

robbiesri:
Checking in my progress toward stereo lighting in HDRP.  It's not done, but is starting to 'look' right (at least for the left eye).  Remaining work is mostly targeted to the light tile generation.

This PR checks in: 
* More proper implementation of camera-relative rendering setup for stereo (offset matrices from the center eye)
* Modifying light tile resolution to be stereo-aware
* Build additional right-eye light culling information, as it is view-dependent
* Support stereo in screen-space AABBs

Remaining:
* Add stereo support to remaining light tile generation compute jobs (big-tile, FPTL, cluster)
* Clean up management of combined stereo view/projection matrices (e.g. explicit generation in HDRP code instead of fetching from generated cull parameters)
* Proper combined support for decals + stereo in tiled cull data
* Probably manage all stereo-dependent matrices in custom constant buffer instead of using engine-generated constant buffer


robbiesri:
I was talking to Paul about some difficulties I was having with the decal cull data being added to the light cull data (because it was adding in empty elements, and this was causing me trouble with generating stereo offsets into the master list).  I have an intermediate solution that I'll pop in, while Paul works on compacting the CPU side decal cull data before pushing it to the buffer.

PR #983, opened 2018-02-22
Upgrade Lightweight shaders and UI to closer match HD and Shader Graph

Kink3d:
- Upgrades to Standard, Simple Standard and Unlit
- Split Rendering Mode into Surface Type and Blending Mode
- Add premultiply option to Simple Standard
- Make Alpha Clip Orthogonal
- Add two sided toggle
- Add Additive and Multiply blend modes
- Rename all blend modes to match HD and SG
- Rename Alpha Cutoff to Clip Threshold
- Remove unused "_ALPHABLEND" keyword
- Split premultiply and clip keywords
- Update material upgrader to reflect changes
- Various shader GUI fixes

Keywords: Adds 1 to Simple, Removes 1 from Unlit, PBR stays the same

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderGUI/LightweightShaderGUI.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderGUI/LightweightShaderGUI.cs
@@ -3,12 +3,18 @@ using UnityEngine;
 
 public abstract class LightweightShaderGUI : ShaderGUI
 {
-    public enum BlendMode
+    public enum SurfaceType
     {
         Opaque,
-        Cutout,
-        Fade,   // Old school alpha-blending mode, fresnel does not affect amount of transparency
-        Transparent // Physically plausible transparency mode, implemented as alpha pre-multiply
+        Transparent
+    }
+
+    public enum BlendMode
+    {
+        Alpha,   // Old school alpha-blending mode, fresnel does not affect amount of transparency
+        Premultiply, // Physically plausible transparency mode, implemented as alpha pre-multiply
+        Additive,
+        Multiply
     }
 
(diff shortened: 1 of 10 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderGUI/LightweightShaderGUI.cs
```diff
@@ -50,53 +56,65 @@ protected void DoPopup(string label, MaterialProperty property, string[] options
 
     public static void SetupMaterialBlendMode(Material material)
     {
-        BlendMode blendMode = (BlendMode)material.GetFloat("_Mode");
-        switch (blendMode)
+        bool alphaClip = material.GetFloat("_AlphaClip") == 1;
```
sebastienlagarde: Be careful with the naming: "_AlphaClip"! vs _AlphaCutoff. 
We did meet a lot of trouble (same problem with _Color) in GI as it assume an _AlphaCutoff name. We have added specific code in lit.shader of HD to rebind _Color on our _BaseColor, you will require to do the same (i.e have a shadowed _AlphaCutoff, so it doesn't break Unity functionality. 
Sadly Unity have rely for years on fixed name of standard shader, this is not specific to SRP, but now we see the limitation of such a system, particularly with shader graph.

Kink3d: Interesting! I didnt realise that. However, in this case "_Cutoff" remains as the property for the threshold. This is merely a "bool" for if alpha clip is enabled or not.

sebastienlagarde: Ah ok :), false alarm then, misread the code :P

PR #986, opened 2018-02-23
add exact number of decals to bound and light volume arrays

Paulm-Unity:
don't add more decals than the length of decal data buffer

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
@@ -1778,15 +1778,15 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                 Debug.Assert(m_lightList.bounds.Count == m_lightCount);
                 Debug.Assert(m_lightList.lightVolumes.Count == m_lightCount);
 
-                if (DecalSystem.m_DecalDatasCount > 0)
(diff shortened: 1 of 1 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2110,7 +2110,7 @@ void UpdateDataBuffers()
             s_LightDatas.SetData(m_lightList.lights);
             s_EnvLightDatas.SetData(m_lightList.envLights);
             s_shadowDatas.SetData(m_lightList.shadows);
-            s_DecalDatas.SetData(DecalSystem.m_DecalDatas);
+            s_DecalDatas.SetData(DecalSystem.m_DecalDatas, 0, 0, Math.Min(DecalSystem.m_DecalDatasCount, k_MaxDecalsOnScreen)); // don't add more than the size of the buffer
```
robbiesri: What if we floated decalDatasCount to be defined outside of the loop, so we don't have to run Math.Min again here?

Paulm-Unity: Good suggestion, I will do that in next PR

robbiesri: Sounds good, wasn't urgent 💃

PR #990, opened 2018-02-23
Move Lightweight Shader Graph sub-shaders from SG repo

pbbastian:
This PR moves Shader Graph sub-shaders (backends for the master nodes) for the Lightweight RP to this repository, so that it can be delivered by the Lightweight package. This also means that the Lightweight package now has a dependency on Shader Graph, forcing the user to use a compatible set of Shader Graph and Lightweight versions.

It should not be merged in before the Shader Graph repo has the equivalent PR merged in. I will update in here once that is done, but do review in the meantime :)


pbbastian:
This is now pointing to the `2018.1` branch of ShaderGraph, and is therefor good to merge.

sebastienlagarde:
Alright. So next time we generate a new package, users will get automatically the shader graph package.

pbbastian:
Yep, and since there is a hard dependency, they will actually get the correct version.

PR #1000, opened 2018-02-24
Add profiling/debugging markers for LWRP cameras

robbiesri:
Mostly to improve the debugging experience in RenderDoc, but also has the side-effect of helping with other debuggers and the Unity profiler


robbiesri:
Ohh, what a boring PR for #1000 on this repo.

robbiesri:
Wait, I think I'm managing the command buffer here incorrectly.  Un momento.

robbiesri:
Ok, this is better.  Forgot about CommandBuffer.Clear(), my mistake.

PR #1020, opened 2018-03-01
LWRP XR Fixes

robbiesri:
Ok, this restores full single-pass doublewide and single-pass instancing support to LWRP.  I haven't tested Android Single-pass (multi-view) yet, but it should work in concept, as most of the infrastructure is shared with single-pass instancing (famous last words).

This is a brief summary of changes:
* Properly stereo-ize LWRP material shaders -They were lacking instancing/multi-view support
* Stereo-ize depth pre-pass in LWRP C# - This was being rendered monoscopically.
* Unify intermediate resource allocation - Previously, mono and stereo were allocating RTs independently, and stereo was being left behind.  Unifying the paths allows supports for both in one place.
* Add proper screen-space texture support for stereo - For double-wide, this means sampling from the correct portion of the screen-space texture.  For instancing/multi-view, this means sampling from the correct texture slice.
* Fix up screen-space shadow map pass - I ran into a lot of issues here.  Bindings the screen-space shadow map RT correctly, sampling depth correctly, setting up unity_StereoEyeIndex correctly.  There's actually some weird issue going on right now with how unity_StereoEyeIndex gets populated in stereo instancing, but I was able to work around it.  I'd like to investigate it further, but for now, this works.

There is some more work to do:
* Validate against Android multi-view
* Organize stereo shader functions/macros in a better way, maybe even move them to Core shader lib.
* Figure out how to manage both LWRP.RenderScale and XR.RenderViewportScale.


robbiesri:
@sebastienlagarde You are omniscient, so I'll just CC you here for visibility, but you don't need to review this.

PR #1026, opened 2018-03-01
Various light fix for HD

sebastienlagarde:
- Light was incorrectly initialize with Physical light unit when adding HDAdditonalLightData
- Hide "toggle light editor"
- Hide warning that spam the console with reflection probe (no impact on result)

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightUtils.cs
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightUtils.cs
@@ -53,7 +53,7 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
         }
 
         // convert intensity (lumen) to nits
-        public static float calculateLineLightArea(float intensity, float lineWidth)
+        public static float CalculateLineLightArea(float intensity, float lineWidth)
         {
             // The area of a cylinder is this:
             // float lineRadius = 0.01f; // 1cm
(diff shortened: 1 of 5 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/LightUtils.cs
```diff
@@ -53,7 +53,7 @@ public static float ConvertRectLightIntensity(float intensity, float width, floa
         }
 
         // convert intensity (lumen) to nits
-        public static float calculateLineLightArea(float intensity, float lineWidth)
+        public static float CalculateLineLightArea(float intensity, float lineWidth)
```
EvgeniiG: The implementation is correct, but the name of the function and the reasoning are wrong.

Line lights in the shader expect intensity (W/sr).
In the UI, we specify luminous flux (power) in lumens.
First, it needs to be converted to radiometric units (radiant flux, W).

Then we must recall how to compute power from intensity for a line light:

power = Integral{length, Integral{sphere, intensity}}.

For an isotropic line light, intensity is constant, so 

power = length * (4 * Pi) * intensity,
intensity = power / (length * (4 * Pi)).

sebastienlagarde: Double thanks! Fixed :). And thanks for the correct comment

PR #1027, opened 2018-03-01
Meta Fix, using world2clip instead of pbject2clip

Verasl:
Fixed non-working meta pass


robbiesri:
I'm late to the party, but I'd strongly recommend changing the name of 'vertex' to something like 'vertexWS'.  There's no reason to believe that the incoming position is anything but object-space, which is why the original object-to-clip transform was used.

PR #1028, opened 2018-03-02
XR support for HDRP Big Tile Pass

robbiesri:
This PR adds XR support for the big tile light cull pass.  I verified via monoscopic and stereo image comparison, and validation of the big tile output.

* Adds light cull utility file, LightCullUtils.hlsl.  The light cull code often generates indices into the various light cull buffers, and each shader re-calculates the indices.  There's no reason for repeating this logic everywhere.  We have the technology (functions).

* In the stereo case, the big tile code outputs a second set of tiles for the right eye.  This is because the transformation from world to screen space is unique for the right eye.  We already generated the light cull data for the right eye, which is in view space.  Inside the compute shader, we use the eye index to grab the right view-to-screen matrix, and offset into the correct sub-portions of the AABB screen bounds, LightVolumeData, and SFiniteLightBound buffers.

* The light index encoded into the big tile buffers corresponds to the 'original' light index.  Using the utility functions in LightCullUtils.hlsl, I can encode that index into buffer-specific indices that take the current eye index and number of lights into account.

* The dispatch encodes the stereo eye index into the GroupID.z value.  For monoscopic rendering, all the various eyeIndex-dependent calculations nop out, because eyeIndex will be 0.

Note: We really need to add some sort of LightLoop, or at least compute, validation of the code here.  The compute code here can be a bit opaque for an outsider, and making changes is kinda scary w/o good validation.  Of course, when I say that, I should probably investigate building out that infrastructure, and I will.  But if someone beats me to it...


robbiesri:
@mmikk I'm poking your code again!  This is a bit more invasive than the AABB shader change, but I think this change is mostly additive in nature, as opposed to manipulative of core logic (though I certainly did some refactoring).

sebastienlagarde:
> We really need to add some sort of LightLoop, or at least compute, validation of the code here.
We have a non tile pass mode with a pixel shader and we have also a compute pass without tiling, isn't it sufficient to compare with, we use it all the time as reference to detect bug with compute tile/cluster path. Or you think about something else ?

robbiesri:
Oh, I totally forgot about the PS pass.  And it also didn't even occur to me that I could just turn off tiling/cluster. I don't remember seeing that option before, but it's there now!

On the other hand, I still think a unit-test setup that explicitly validated stage output could be handy, especially if debugging is needed.

PR #1032, opened 2018-03-02
Make sure we get the correct lightmap decode parameters when in gamma…

stramit:
… mode

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/MaterialUtilities.hlsl
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/MaterialUtilities.hlsl
@@ -35,27 +35,34 @@ float3 SampleBakedGI(float3 positionWS, float3 normalWS, float2 uvStaticLightmap
 
 #ifdef UNITY_LIGHTMAP_FULL_HDR
     bool useRGBMLightmap = false;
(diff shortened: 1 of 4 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/MaterialUtilities.hlsl
```diff
@@ -35,27 +35,34 @@ float3 SampleBakedGI(float3 positionWS, float3 normalWS, float2 uvStaticLightmap
 
 #ifdef UNITY_LIGHTMAP_FULL_HDR
     bool useRGBMLightmap = false;
+    float4 decodeInstructions = float4(0.0, 0.0, 0.0, 0.0); // Never used but needed for the interface since it supports gamma lightmaps
 #else
     bool useRGBMLightmap = true;
+    #if defined(UNITY_LIGHTMAP_RGBM_ENCODING)
+        float4 decodeInstructions = float4(34.493242, 2.2, 0.0, 0.0); // range^2.2 = 5^2.2, gamma = 2.2
+    #else
+        float4 decodeInstructions = float4(0.0, 0.0, 0.0, 0.0); // range = 2.0^2.2 = 4.59
```
sebastienlagarde: wrong comment here. Minor

sebastienlagarde: I think we should still setup 2.0 and 2.2 here, just in case for unknow reasons something change :)

PR #1043, opened 2018-03-06
XR HDRP Clustered Lighting support

robbiesri:
Adds XR support to both the clustered light list generation, and light loop lookups.

The changes to lightlistbuild-clustered are pretty extensive, but I documented them thoroughly before implementation.  Mainly, the idea is to stereo-ize access to the cull data (screen-space AABBs, cull data, big tiles), and to stereo-ize the writes to the buffers configured by HDRP.

Tested via visual inspection, comparing to tiled lighting on/off.  Additionally, I used RenderDoc to do verification of index generation.  Easy to see if right eye buffer portion was being generated incorrectly.


sebastienlagarde:
Thanks for this work! I will just chek perf compare to without before merging (we have tools to make comparison on PS4.

robbiesri:
Sounds good, thanks Seb.  My guess is that perf should be relatively un-affected.  Probably a few more instructions, maybe an extra SGPR for the extra constants and GroupID.z (hopefully it doesn't push it across the 4-aligned boundary).

robbiesri:
Oh, it occurred to me that the code in LightLoopDef.hlsl could easily be gated with the stereo defines (USING_STEREO_MATRICES).  So if you hit any perf issues, I'll just patch that up.  There's always the latent chance that introducing the "#if defined" path could lead to divergence (and stagnation), but as long as we limit these cases and keep them in one place (like LightLoopDef.hlsl), it should be ok™️

PR #1059, opened 2018-03-07
Some shuffling around reduces frame time on older devices.

aleks01010101:
This saved additional 6.31ms on Mali-400.


robbiesri:
🥇

PR #1063, opened 2018-03-09
Fixes for cases when real != float. We need to declare both real and float variants

strandborg:


ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Common.hlsl
```diff
@@ -83,6 +83,8 @@
 #define TEMPLATE_2_REAL TEMPLATE_2_HALF
 #define TEMPLATE_3_REAL TEMPLATE_3_HALF
 
+#define HAS_HALF 1
```
EvgeniiG: This name is ambiguous. Nothing prevents you from using the _half_ data type (defined by HLSL) even if you _#define real float_. I'd suggest something like REAL_IS_HALF instead.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Macros.hlsl
```diff
@@ -148,23 +160,47 @@
...
-    void FunctionName(inout uint2  a, inout uint2  b) { uint2  t = a; a = b; b = t; } \
-    void FunctionName(inout uint3  a, inout uint3  b) { uint3  t = a; a = b; b = t; } \
-    void FunctionName(inout uint4  a, inout uint4  b) { uint4  t = a; a = b; b = t; } \
-    void FunctionName(inout bool   a, inout bool   b) { bool   t = a; a = b; b = t; } \
-    void FunctionName(inout bool2  a, inout bool2  b) { bool2  t = a; a = b; b = t; } \
-    void FunctionName(inout bool3  a, inout bool3  b) { bool3  t = a; a = b; b = t; } \
-    void FunctionName(inout bool4  a, inout bool4  b) { bool4  t = a; a = b; b = t; }
+    #if HAS_HALF
+        #define TEMPLATE_SWAP(FunctionName) \
+        void FunctionName(inout real  a, inout real  b) { real  t = a; a = b; b = t; } \
```
sebastienlagarde: This should half, not real

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Macros.hlsl
```diff
@@ -52,7 +52,11 @@
     half  FunctionName(half  Parameter1) { FunctionBody; } \
     half2 FunctionName(half2 Parameter1) { FunctionBody; } \
     half3 FunctionName(half3 Parameter1) { FunctionBody; } \
-    half4 FunctionName(half4 Parameter1) { FunctionBody; }
+    half4 FunctionName(half4 Parameter1) { FunctionBody; } \
+    float  FunctionName(float  Parameter1) { FunctionBody; } \
```
sebastienlagarde: This template is explictely for half, why adding float here ?

strandborg: This is because many call sites use explicit floats as arguments to these functions, so we'll need to declare both half and float variants for them

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Common.hlsl
sebastienlagarde: HAS_HALF  shoudn't define be there, it should be define inthe respective API file no? (Like in pssl.hlsl, d3d11.hlsl, vulkan.hlsl etc... Or I miss something. For me HAS_HALF mean support HALF type, so generate two version of the function as you write it

strandborg: HLSL doesn't actually define 'half' anymore (not since DX9/Cg days). It does define min16float. I guess I could rename this to HAS_NATIVE_HALF or something similar; I don't think REAL_IS_HALF is accurate either, because we'll have to have a way to know whether we should introduce both half and float versions of a function or not, and REAL_IS_HALF doesn't really convey that idea

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/MaterialUtilities.hlsl
```diff
@@ -11,7 +11,7 @@ float3 SampleBakedGI(float3 positionWS, float3 normalWS, float2 uvStaticLightmap
     if (unity_ProbeVolumeParams.x == 0.0)
     {
         // TODO: pass a tab of coefficient instead!
-        float4 SHCoefficients[7];
+        real4 SHCoefficients[7];
```
sebastienlagarde: As you made this change, not sure I understand why we need to create 2 version of Sample light probe.
Before the function was taking a real, so this will be correctly handled no? Why do you have duplicate it (Sorry there is certainly a compile issue I miss here).

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/EntityLighting.hlsl
```diff
@@ -35,24 +35,45 @@ real3 SHEvalLinearL2(real3 N, real4 shBr, real4 shBg, real4 shBb, real4 shC)
     return x2 + x3;
 }
 
-real3 SampleSH9(real4 SHCoefficients[7], real3 N)
+#if HAS_HALF
```
EvgeniiG: I suppose there is a case where we need both versions at the same time? We were trying to avoid this.

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/MaterialUtilities.hlsl
strandborg: I see some compute shaders also using SampleSH9, and it was not clear to me whether those parameters there were reals, so that's why I made 2 versions of this. In this case here the builtin Unity_SH* uniforms are declared as halfs, so this change makes sense.

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Filtering.hlsl
```diff
@@ -22,11 +22,29 @@ real2 BSpline2IntRight(real2 x)
...
+    // 1: lerp between middle and right
+    weights[0] = l + 0.5 * m;
+    weights[1] = r + 0.5 * m;
+    offsets[0] = -0.5 + 0.5 * m * rcp(weights[0]);
+    offsets[1] =  0.5 + r * rcp(weights[1]);
+}
+
+// If half is natively supported, create another variant
+#if HAS_HALF
+void BiquadraticFilter(half2 fracCoord, out half2 weights[2], out half2 offsets[2])
```
EvgeniiG: Why can't we use real here? And what happens if you have real defined as float? This function is preprocessed away?

ScriptableRenderPipeline/Core/CoreRP/ShaderLibrary/Common.hlsl
EvgeniiG: Ah, I see where you were going with this. :-) Aside from templates, is it possible (for now) to avoid having multiple definitions of a single function with varying precision? E.g. just fix the call site?

For multiple includes, the following model was proposed a while back:

// If you want real = half...
#include "define_real_as_half.hlsl"
#include "ShaderLibrary/SomeOtherHeader.hlsl"
#include "define_real_as_float.hlsl"

PR #1069, opened 2018-03-12
Lw bugfix

phi-lira:
- Fixed a regression that was causing shadow to not render with configurations that didn't require an intermediate render target.
 - Fixed shadowmap on lowend devices. GLES2 only support hard and non-cascade shadows for now until we can get the keyword stripping functionality. 
 - Fixed error about R8 texture format not being supported on old mobile devices.


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Shadows.hlsl
```diff
@@ -153,7 +159,11 @@ float4 ComputeShadowCoord(float4 clipPos)
 
 half RealtimeShadowAttenuation(float4 shadowCoord)
 {
+#if SHADOWS_SCREEN
     return SampleScreenSpaceShadowMap(shadowCoord);
+#endif
```
aleks01010101: I would probably do an "else" here, would be less changes if we introduce more shadow modes

PR #1070, opened 2018-03-12
Disable a number of scene view draw modes for LWRP

pbbastian:
As the draw modes haven't been looked at yet in Lightweight pipeline, I'm disabling them so that users cannot access them.


pbbastian:
Backported to `2018.1` in 2c10c306fca42aeaae997bf0372314f24c24c089

PR #1084, opened 2018-03-13
Add Distortion Texture functionality

Kink3d:
Similar to legacy renderer's **Grab Pass** except performed once in the `After Opaque` Pass. Provides functionality for shaders (such as particle shaders) to access the color values of the screen for distortion effects etc using the global texture `_CameraDistortionTexture`.

**Features**
- Get screen color values from `Opaque` pass when rendering in `Transparent` pass.
- Distortion texture can be downsampled to 1/2 or 1/4 using bilinear sampling.

**Changes**
- Add `TextureScale` enum for setting Distortion Texture size.
- Add `Distortion Texture` and `Distortion Scale` fields to Lightweight Asset.
- Add `LightweightSampling.shader` for performing bilinear downsampling.
- Add `LightweightSampling.shader` to Lightweight Resources object.
- Add Distortion render target to pipeline.
- Add Material for `LightweightSampling.shader` to pipeline.
- Add Distortion Pass to copy screen contents to `_CameraDistortionTexture` at the end of `After Opaque` pass.

**Risk**
- None really, apart from that this is my first PR to the render pipeline :D

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
@@ -36,6 +36,13 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         _8x = 8
     }
 
+    public enum TextureScale
+    {
+        Full,
+        Half,
+        Quarter
+    }
+
     public enum DefaultMaterialType
     {
         Standard = 0,
@@ -57,6 +64,8 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         [SerializeField] private bool m_SupportsVertexLight = false;
         [SerializeField] private bool m_RequireDepthTexture = false;
         [SerializeField] private bool m_RequireSoftParticles = false;
+        [SerializeField] private bool m_RequiresDistortionTexture = false;
+        [SerializeField] private TextureScale m_DistortionTextureScale = TextureScale.Half;
         [SerializeField] private bool m_SupportsHDR = false;
         [SerializeField] private MSAAQuality m_MSAA = MSAAQuality._4x;
         [SerializeField] private float m_RenderScale = 1.0f;
@@ -213,6 +222,17 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
             get { return m_RequireSoftParticles; }
         }
 
+        public bool RequireDistortionTexture
+        {
+            get { return m_RequiresDistortionTexture; }
+        }
+
+        public TextureScale DistortionTextureScale
+        {
+            get { return m_DistortionTextureScale; }
+            set { m_DistortionTextureScale = value; }
+        }
+
         public bool SupportsHDR
         {
             get { return m_SupportsHDR; }
@@ -341,6 +361,11 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         public Shader ScreenSpaceShadowShader
         {
             get { return resources != null ? resources.ScreenSpaceShadowShader : null; }
+        }
+
+        public Shader SamplingShader
+        {
+            get { return resources != null ? resources.SamplingShader : null; }
         }
     }
 }
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/LightweightAssetEditor.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/LightweightAssetEditor.cs
@@ -25,6 +25,10 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 
             public static GUIContent requireSoftParticles = new GUIContent("Soft Particles", "If enabled the pipeline will enable SOFT_PARTICLES keyword.");
 
+            public static GUIContent requireDistortionTexture = new GUIContent("Distortion Texture", "If enabled the pipeline will copy the screen to texture after opaque objects are drawn. For transparent objects this can be bound in shaders as _CameraDistortionTexture.");
+
+            public static GUIContent distortionTextureScale = new GUIContent("Distortion Scale", "The scale of the original screen size that is used for the distortion texture");
+
             public static GUIContent shadowType = new GUIContent("Type",
                     "Global shadow settings. Options are NO_SHADOW, HARD_SHADOWS and SOFT_SHADOWS.");
 
@@ -50,6 +54,8 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         }
 
         AnimBool m_ShowSoftParticles = new AnimBool();
+        AnimBool m_ShowDistortionTextureScale = new AnimBool();
+
 
         private int kMaxSupportedPixelLights = 8;
         private float kMinRenderScale = 0.1f;
@@ -59,6 +65,8 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         private SerializedProperty m_SupportsVertexLightProp;
         private SerializedProperty m_RequireDepthTextureProp;
         private SerializedProperty m_RequireSoftParticlesProp;
+        private SerializedProperty m_RequireDistortionTextureProp;
+        private SerializedProperty m_DistortionTextureScaleProp;
         private SerializedProperty m_ShadowTypeProp;
         private SerializedProperty m_ShadowNearPlaneOffsetProp;
         private SerializedProperty m_ShadowDistanceProp;
@@ -76,6 +84,8 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
             m_SupportsVertexLightProp = serializedObject.FindProperty("m_SupportsVertexLight");
             m_RequireDepthTextureProp = serializedObject.FindProperty("m_RequireDepthTexture");
             m_RequireSoftParticlesProp = serializedObject.FindProperty("m_RequireSoftParticles");
+            m_RequireDistortionTextureProp = serializedObject.FindProperty("m_RequiresDistortionTexture");
+            m_DistortionTextureScaleProp = serializedObject.FindProperty("m_DistortionTextureScale");
             m_ShadowTypeProp = serializedObject.FindProperty("m_ShadowType");
             m_ShadowNearPlaneOffsetProp = serializedObject.FindProperty("m_ShadowNearPlaneOffset");
             m_ShadowDistanceProp = serializedObject.FindProperty("m_ShadowDistance");
@@ -88,16 +98,20 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 
             m_ShowSoftParticles.valueChanged.AddListener(Repaint);
             m_ShowSoftParticles.value = m_RequireSoftParticlesProp.boolValue;
+            m_ShowDistortionTextureScale.valueChanged.AddListener(Repaint);
+            m_ShowDistortionTextureScale.value = m_DistortionTextureScaleProp.boolValue;
         }
 
         void OnDisable()
         {
             m_ShowSoftParticles.valueChanged.RemoveListener(Repaint);
+            m_ShowDistortionTextureScale.valueChanged.RemoveListener(Repaint);
         }
 
         void UpdateAnimationValues()
         {
             m_ShowSoftParticles.target = m_RequireDepthTextureProp.boolValue;
+            m_ShowDistortionTextureScale.target = m_RequireDistortionTextureProp.boolValue;
         }
 
         void DrawAnimatedProperty(SerializedProperty prop, GUIContent content, AnimBool animation)
@@ -126,6 +140,8 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
             EditorGUILayout.PropertyField(m_SupportsVertexLightProp, Styles.enableVertexLightLabel);
             EditorGUILayout.PropertyField(m_RequireDepthTextureProp, Styles.requireDepthTexture);
             DrawAnimatedProperty(m_RequireSoftParticlesProp, Styles.requireSoftParticles, m_ShowSoftParticles);
+            EditorGUILayout.PropertyField(m_RequireDistortionTextureProp, Styles.requireDistortionTexture);
+            DrawAnimatedProperty(m_DistortionTextureScaleProp, Styles.distortionTextureScale, m_ShowDistortionTextureScale);
             EditorGUILayout.PropertyField(m_HDR, Styles.hdrContent);
             EditorGUILayout.PropertyField(m_MSAA, Styles.msaaContent);
 
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
@@ -90,6 +90,9 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 
         // If soft particles are enabled and no depth prepass is performed we need to copy depth.
         public static int depthCopy;
+
+        // Camera distortion target. Used for accessing screen contents during transparent pass.
+        public static int distortion;
     }
 
     public class LightweightPipeline : RenderPipeline
@@ -135,6 +138,7 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         private RenderTargetIdentifier m_DepthRT;
         private RenderTargetIdentifier m_CopyDepth;
         private RenderTargetIdentifier m_Color;
+        private RenderTargetIdentifier m_DistortionRT;
 
         private bool m_IntermediateTextureArray;
         private bool m_RequireDepthTexture;
@@ -185,6 +189,7 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         private Material m_CopyDepthMaterial;
         private Material m_ErrorMaterial;
         private Material m_ScreenSpaceShadowsMaterial;
+        private Material m_SamplingMaterial;
         private int m_BlitTexID = Shader.PropertyToID("_BlitTex");
 
         private CopyTextureSupport m_CopyTextureSupport;
@@ -231,6 +236,7 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
             CameraRenderTargetID.copyColor = Shader.PropertyToID("_CameraCopyColorRT");
             CameraRenderTargetID.depth = Shader.PropertyToID("_CameraDepthTexture");
             CameraRenderTargetID.depthCopy = Shader.PropertyToID("_CameraCopyDepthTexture");
+            CameraRenderTargetID.distortion = Shader.PropertyToID("_CameraDistortionTexture");
 
             m_ShadowMapRT = new RenderTargetIdentifier(m_ShadowMapRTID);
             m_ScreenSpaceShadowMapRT = new RenderTargetIdentifier(m_ScreenSpaceShadowMapRTID);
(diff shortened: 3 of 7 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -498,6 +507,29 @@ private void DepthPass(ref ScriptableRenderContext context, FrameRenderingConfig
             StopStereoRendering(ref context, frameRenderingConfiguration);
         }
 
+        private void DistortionPass(ref ScriptableRenderContext context)
+        {
+            CommandBuffer cmd = CommandBufferPool.Get("Distortion");
+            switch(m_Asset.DistortionTextureScale)
+            {
+                case TextureScale.Half:
+                    cmd.GetTemporaryRT(CameraRenderTargetID.distortion, m_CurrCamera.pixelWidth / 2, m_CurrCamera.pixelHeight / 2, 0, FilterMode.Bilinear, m_ColorFormat);
```
phi-lira: Pipeline renders scale should be taken into consideration here and in the cases below.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
+Shader "Hidden/LightweightPipeline/Sampling"
+{
+    Properties
+    {
+        _MainTex("Albedo", 2D) = "white" {}
+    }
+
+    HLSLINCLUDE
+    #include "../ShaderLibrary/Core.hlsl"
```
phi-lira: Use full include path. If someone wants to make a copy/variantion of this shader it will keep working. "LWRP/ShaderLibrary/Core.hlsl"

ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
```diff
@@ -57,6 +64,8 @@ public class LightweightPipelineAsset : RenderPipelineAsset
         [SerializeField] private bool m_SupportsVertexLight = false;
         [SerializeField] private bool m_RequireDepthTexture = false;
         [SerializeField] private bool m_RequireSoftParticles = false;
+        [SerializeField] private bool m_RequiresDistortionTexture = false;
```
phi-lira: There's an inconsistency between Require and Requires. Requires does seem more correct but current api is already require in singular. Should change it to make it consistent.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/LightweightAssetEditor.cs
```diff
@@ -76,6 +84,8 @@ void OnEnable()
             m_SupportsVertexLightProp = serializedObject.FindProperty("m_SupportsVertexLight");
             m_RequireDepthTextureProp = serializedObject.FindProperty("m_RequireDepthTexture");
             m_RequireSoftParticlesProp = serializedObject.FindProperty("m_RequireSoftParticles");
+            m_RequireDistortionTextureProp = serializedObject.FindProperty("m_RequiresDistortionTexture");
```
phi-lira: Same require vs requires issue here.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -498,6 +507,29 @@ private void DepthPass(ref ScriptableRenderContext context, FrameRenderingConfig
...
+                    break;
+                default:
+                    cmd.GetTemporaryRT(CameraRenderTargetID.distortion, m_CurrCamera.pixelWidth, m_CurrCamera.pixelHeight, 0, FilterMode.Bilinear, m_ColorFormat);
+                    cmd.Blit(m_CurrCameraColorRT, CameraRenderTargetID.distortion);
+                    break;
+            }
+            SetRenderTarget(cmd, m_CurrCameraColorRT, m_DepthRT);
+            context.ExecuteCommandBuffer(cmd);
+            CommandBufferPool.Release(cmd);
+        }
```
robbiesri: This code won't work with XR.  Is that...expected?  Personally, I don't really care, but I guess there might be some expectation of everything 'just working' with XR and LWRP.

robbiesri: Oh yeah, this is because of how the distortion RT is allocated.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
```diff
@@ -36,6 +36,13 @@ public enum MSAAQuality
         _8x = 8
     }
 
+    public enum TextureScale
```
aleks01010101: Perhaps rename this to TextureSize for better readability? TextureScale.Full sounds a bit not right :)

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -498,6 +507,43 @@ private void DepthPass(ref ScriptableRenderContext context, FrameRenderingConfig
...
+
+            float renderScale = (m_CurrCamera.cameraType == CameraType.Game) ? m_Asset.RenderScale : 1.0f;
+
+            CommandBuffer cmd = CommandBufferPool.Get("Opaque Copy");
+            switch(m_Asset.OpaqueTextureScale)
+            {
+                case TextureScale.Half:
+                    opaqueDesc.width = (int)((float)opaqueDesc.width * renderScale * 0.5f);
+                    opaqueDesc.height = (int)((float)opaqueDesc.height * renderScale * 0.5f);
+                    cmd.GetTemporaryRT(CameraRenderTargetID.opaque, opaqueDesc, FilterMode.Bilinear);
```
aleks01010101: I'd probably mode this outside of the `switch`, since it's the same in all cases.

aleks01010101: Or even do a small method to setup the temporary texture, pass the 0.5f/0.25f/1.0f as scale based on the switch, and pass the fourth parameter to cmd.Blit in there

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -498,6 +507,43 @@ private void DepthPass(ref ScriptableRenderContext context, FrameRenderingConfig
...
+                    cmd.Blit(m_CurrCameraColorRT, CameraRenderTargetID.opaque, m_SamplingMaterial, 0);
+                    break;
+                case TextureScale.Quarter:
+                    opaqueDesc.width = (int)((float)opaqueDesc.width * renderScale * 0.25f);
+                    opaqueDesc.height = (int)((float)opaqueDesc.height * renderScale * 0.25f);
+                    cmd.GetTemporaryRT(CameraRenderTargetID.opaque, opaqueDesc, FilterMode.Bilinear);
+                    cmd.Blit(m_CurrCameraColorRT, CameraRenderTargetID.opaque, m_SamplingMaterial, 1);
+                    break;
+                default:
```
aleks01010101: A default: in an enum-based switch is error-prone, better to use the actual third value.
If someone adds a value to it, the compiler will immediately tell "You forgot this thing here".

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
aleks01010101: Also, the temporary RT will leak...

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+        UNITY_SETUP_INSTANCE_ID(i);
+        UNITY_TRANSFER_INSTANCE_ID(i, o);
+
+        o.pos = TransformObjectToHClip(i.vertex.xyz);
+
+        float4 projPos = o.pos * 0.5;
+        projPos.xy = projPos.xy + projPos.w;
+
+        o.texcoord.xy = i.texcoord;
+        o.texcoord.zw = projPos.xy;
```
aleks01010101: .zw is unused in fragment shader
Interpolators.texcoord can be 2 components

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+    struct VertexInput
+    {
+        float4 vertex   : POSITION;
+        float2 texcoord : TEXCOORD0;
+    };
+
+    struct Interpolators
+    {
+        half4  pos      : SV_POSITION;
+        half4  texcoord : TEXCOORD0;
```
aleks01010101: I think these two should be full floats.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+    {
+        float4 d = texelSize.xyxy * float4(-amount, -amount, amount, amount);
+
+        half4 s;
+        s =  (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.xy));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.zy));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.xw));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.zw));
+
+        return s * (1.0 / 4.0);
```
aleks01010101: Why not multiply by 0.25? :)
Also, please convert to half (e.g. 0.25h or 1.0h/4.0h).

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+            #pragma vertex Vertex
+            #pragma fragment FragBoxDownsample
+
+            TEXTURE2D(_MainTex);
+            SAMPLER(sampler_MainTex);
+            float4 _MainTex_TexelSize;
+
+            half4 FragBoxDownsample(Interpolators i) : SV_Target
+            {
+                half4 col = DownsampleBox4Tap(TEXTURE2D_PARAM(_MainTex, sampler_MainTex), i.texcoord, _MainTex_TexelSize.xy, 1.0);
```
aleks01010101: AFAIK a 4-tap box downsample with a raduis of 1 texel should not be done in the fragment shader.
A bilinear filter gives this for free if you sample between texels.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+            #pragma vertex Vertex
+            #pragma fragment FragBoxDownsample
+
+            TEXTURE2D(_MainTex);
+            SAMPLER(sampler_MainTex);
+            float4 _MainTex_TexelSize;
+
+            half4 FragBoxDownsample(Interpolators i) : SV_Target
+            {
+                half4 col = DownsampleBox4Tap(TEXTURE2D_PARAM(_MainTex, sampler_MainTex), i.texcoord, _MainTex_TexelSize.xy, 2.0);
```
aleks01010101: And we can leverage bilinear filtering here as well, by sampling 4 times between texels.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,112 @@
...
+
+        o.texcoord.xy = i.texcoord;
+        o.texcoord.zw = projPos.xy;
+
+        return o;
+    }
+
+    half4 DownsampleBox4Tap(Texture2D tex, SamplerState samplerTex, float2 uv, float2 texelSize, float amount)
+    {
+        float4 d = texelSize.xyxy * float4(-amount, -amount, amount, amount);
```
aleks01010101: This can be calculated either in the vertex shader or (better) in C# and passed in as a uniform.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
robbiesri: I have even more comments.  Most important, thanks for adding XR support to this.

Ok, since this code shares a lot with the setup in `SetupIntermediateRenderTextures`, you should functionalize the code.  Something like 
`RenderTextureDescriptor CreateRTDesc(FrameRenderingConfiguration config, float opaqueScaler = 1.0f)`

In the method, you select the correct base desc, and apply the `opaqueScaler `when you apply `renderScale`.  `SetupIntermediateRenderTextures `will just use the default param value for `opaqueScaler `of 1.0f.  `OpaqueTexturePass `will pass in the `TextureScale `enum generated value.  Then you can use the returned `RenderTextureDescriptor` in the RT allocation.

robbiesri: Oh, bonus note: What if you created an array of floats that represented opaque scaler modifers, and indexed into it with TextureScale?
```
const float[] opaqueScalerValues = {1.0f, 0.5f, 0.25f};
var scalerVal = opaqueScalerValues [m_Asset.OpaqueTextureScale];
```
I'm guessing there's also a way to make that array _only_ index-able by the enums, but that's outside my weak C#-fu.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -574,6 +620,11 @@ private void AfterOpaque(ref ScriptableRenderContext context, FrameRenderingConf
                 SetRenderTarget(cmd, m_CurrCameraColorRT, depthRT);
             context.ExecuteCommandBuffer(cmd);
             CommandBufferPool.Release(cmd);
+            
+            if(m_Asset.RequireOpaqueTexture)
```
phi-lira: At the beginning of AfterOpaque there's a check if depth texture is required and if not it returns. In that case this code won't execute. That needs to be udpated.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
Kink3d: Fixed

ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
Kink3d: Me and Felipe decided that Scale is better because its relative to the screen size. But I've changed `Full` to `One` as we feel thats better.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
Kink3d: Felipe: I think we should profile this before do this change. Shader is quite simple on ALU already and precomputing *might* increase L/S

Kink3d: I partially did this in the end, but I just passed the amount in as a uniform. Makes the box filter more multi-use.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -478,6 +493,34 @@ private void DepthPass(ref ScriptableRenderContext context, FrameRenderingConfig
...
+            
+            CommandBuffer cmd = CommandBufferPool.Get("Opaque Copy");
+            cmd.GetTemporaryRT(CameraRenderTargetID.opaque, opaqueDesc, m_Asset.OpaqueTextureScale == TextureScale.OnePoint ? FilterMode.Point : FilterMode.Bilinear);
+            switch(m_Asset.OpaqueTextureScale)
+            {
+                case TextureScale.OnePoint:
+                    cmd.Blit(m_CurrCameraColorRT, CameraRenderTargetID.opaque);
+                    break;
+                case TextureScale.HalfBilinear:
+                    cmd.Blit(m_CurrCameraColorRT, CameraRenderTargetID.opaque);
```
aleks01010101: I'm not sure if that would do a bilinear filter on the source properly...

ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
phi-lira: Not important to this PR, but maybe we should briefly discuss the name again since this is not just a scale anymore. We can do it in out meeting or chat if necessary.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
phi-lira: As a note, before the bilinear change we had to do multiple uniforms setups to achieve this. While now all that needs to be done is to reporpose the _SampleOffset. So no extra uniforms need to be set. This can be done in a next PR though.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightSampling.shader
```diff
@@ -0,0 +1,82 @@
...
+
+        half4 s;
+        s =  (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.xy));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.zy));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.xw));
+        s += (SAMPLE_TEXTURE2D(tex, samplerTex, uv + d.zw));
+
+        return s * (1.0h * 0.25h);
```
phi-lira: Compiler will optimize this but for comestic reasons you could remove 1.0h scale there.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
Kink3d: I thought it wouldnt but I checked it on lots of different target sizes and it was exactly the same so I guess it does?

TyelorD:
I seem to have a strange issue with the Camera Opaque Texture.

I'm using Unity 2018.2.1f1 and have LWRP updated to 3.0.0. I've enabled Opaque Textures on all the LWRP Quality variants, set the Texture2D property in my shader to not exposed and it's reference to _CameraOpaqueTexture. The following shader should (and does for another Unity user) produce a fairly simple "warped glass" refraction:
unitybug4

However it produces the following results:
unitybug3

Now there is no shader being applied at all, which seems strange. So I check the scene view, however the shader seems to be working fine in the scene view:
unitybug

Then when I go back to the play view (without ever stopping the game or pausing it), the shader starts working somewhat (although it flickers heavily for no apparent reason), but doesn't produce results identical to those in the scene view as you can see here (after moving the camera around a bit):
unitybug2

I tried to find the cause of this issue but haven't been able to yet. Nor can I find a work around, or what is causing it... It's strange that the shader works in the scene view, but not play mode (and it's even stranger that it starts partially working in play mode after switching to scene mode). There still shouldn't be any black (or very little since it was a PBR Master), and was completely translucent for the user that made the shader.

I tested a simple shader with just the _CameraOpaqueTexture referenced property linked to a Sample Texture 2D, and into an Unlit Master (which should in theory make an invisible mesh). Instead it produced a grey mesh, until I went into the scene view and back to the play view. After that the mesh was mostly translucent, but flickered heavily.

The last odd thing I noticed was when I added a Reflection Probe to the scene. Then it seemed that all of the shaders using the _CameraOpaqueTexture reference stopped working and just became simple reflection probe shaders.

I can provide you with more information, and/or project files if needed too! Thanks for contributions, and I hope you all have a good day! :)

PR #1096, opened 2018-03-14
 Added IgnoreProjector tag to shaders 

phi-lira:
Projectors are not supported on  SRP.


phi-lira:
I'll go ahead and merge this.

PR #1110, opened 2018-03-19
Added Switch support for SRP sample project.

kaychang-unity:
Added switch in the list of platforms where HDRP shaders should be compiled with.
Added missing api header for Switch.
Modified C# scripts to build without XR support on Switch (advised from platforms-vr).


kaychang-unity:
We have made Switch platform as similar as possible as D3D11, so it should be fine.
Sorry, I believe this PR was made against a\the tagged revision Unity-2018.1.0b2 which is too old. I may make a new PR again against the latest revision of the master branch (too many code changes happened).

sebastienlagarde:
Sorry for the delay, there is conflict now on this branch, can you merge latest?

kaychang-unity:
Yes, I'm working on it.

PR #1111, opened 2018-03-20
Lw/shader optimization

phi-lira:
Speeds up of 150% on Dungeon Throne room Scene in iPhone5
    * Added SHADOWS_ENABLED keyword to avoid worldToLight transform and the generate small shadow buffer pass.
    * defined min16float* as it is required by HLSLcc to propertly generate half precision
    * Fixed precision on shaders CB and removed some unused constants.
    * Bump scale is only performed when SHADER_HINT_NICE_QUALITY is enabled
    * Split simple lighting and PBS vertex shaders to avoid shininess computation on PBS (next PR we need to split the CB itself)
    * pre-computed F * V and 1.0 / 4.0 normalization term out of the light loop.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
@@ -69,9 +69,10 @@ half3 UnpackNormalScale(half4 packedNormal, half bumpScale)
 
 void OutputTangentToWorld(half4 vertexTangent, half3 vertexNormal, out half3 tangentWS, out half3 binormalWS, out half3 normalWS)
 {
+    // mikkts space compliant. only normalize when extracting normal at frag.
     half sign = vertexTangent.w * GetOddNegativeScale();
     normalWS = TransformObjectToWorldNormal(vertexNormal);
-    tangentWS = normalize(mul((half3x3)UNITY_MATRIX_M, vertexTangent.xyz));
+    tangentWS = TransformObjectToWorldDir(vertexTangent.xyz);
     binormalWS = cross(normalWS, tangentWS) * sign;
 }
 
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
@@ -124,7 +124,7 @@ half ComputeFogFactor(float z)
     // factor = (end-z)/(end-start) = z * (-1/(end-start)) + (end/(end-start))
     float fogFactor = saturate(clipZ_01 * unity_FogParams.z + unity_FogParams.w);
     return half(fogFactor);
-#elif defined(FOG_EXP2)
+#elif defined(FOG_EXP) || defined(FOG_EXP2)
     // factor = exp(-(density*z)^2)
     // -density * z computed at vertex
     return half(unity_FogParams.x * clipZ_01);
@@ -135,8 +135,12 @@ half ComputeFogFactor(float z)
 
 void ApplyFogColor(inout half3 color, half3 fogColor, half fogFactor)
 {
-#if defined (FOG_LINEAR) || defined(FOG_EXP2)
-#if defined(FOG_EXP2)
+#if defined(FOG_LINEAR) || defined(FOG_EXP) || defined(FOG_EXP2)
+#if defined(FOG_EXP)
+    // factor = exp(-density*z)
+    // fogFactor = density*z compute at vertex
+    fogFactor = saturate(exp2(-fogFactor));
+#elif defined(FOG_EXP2)
     // factor = exp(-(density*z)^2)
     // fogFactor = density*z compute at vertex
     fogFactor = saturate(exp2(-fogFactor*fogFactor));
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
@@ -13,6 +13,10 @@
 #endif
 #endif
 
+#ifndef BUMP_SCALE_NOT_SUPPORTED
+#define BUMP_SCALE_NOT_SUPPORTED !SHADER_HINT_NICE_QUALITY
+#endif
+
 ///////////////////////////////////////////////////////////////////////////////
 #ifdef _NORMALMAP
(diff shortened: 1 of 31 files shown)


phi-lira:
@sebastienlagarde I needed to add min16float* definitions to Common.hlsl. Shaders compiled with HLSLcc and HLSLPROGRAM need that to properly define half precision on mobile.

@arnaudcarre I've removed dynamic lightmap constant from LW as it doesn't support dynamic GI. Just want to make sure if we need to update the batcher because of that.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurface.hlsl
```diff
@@ -77,7 +76,7 @@ half4 MainTexture(float2 uv)
 half3 Normal(float2 uv)
 {
 #if _NORMALMAP
-    #if BUMP_SCALE_NOT_SUPPORTED
+    #if !SHADER_HINT_NICE_QUALITY
```
aleks01010101: This is incorrect. I added this define for the simple lighting shader, which doesn't expose _BumpScale to the user (it's always set to 1) to avoid normalising the unscaled normal extracted directly from the normal map. This doesn't depend on whether we want nice quality or not.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Lighting.hlsl
```diff
@@ -245,6 +246,7 @@ inline void InitializeBRDFData(half3 albedo, half metallic, half3 specular, half
     outBRDFData.perceptualRoughness = PerceptualSmoothnessToPerceptualRoughness(smoothness);
     outBRDFData.roughness = PerceptualRoughnessToRoughness(outBRDFData.perceptualRoughness);
     outBRDFData.roughness2 = outBRDFData.roughness * outBRDFData.roughness;
+    outBRDFData.normalizationTerm = (outBRDFData.roughness + 0.5h) * 4.0h;
```
aleks01010101: Please check the code generated by the shader compiler, it might be worth it to change to (outBRDFData.roughness * 4.0h + 2.0h)

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurface.hlsl
```diff
@@ -12,26 +12,25 @@
 #endif
 
 CBUFFER_START(UnityPerMaterial)
-half4 _MainTex_ST;
+float4 _MainTex_ST;
```
aleks01010101: Is this a typo (real4 intended?) or we really can't fit the UV transform in half4?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Lighting.hlsl
```diff
@@ -280,7 +282,7 @@ half3 DirectBDRF(BRDFData brdfData, half3 normalWS, half3 lightDirectionWS, half
     half d = NoH * NoH * (brdfData.roughness2 - 1.h) + 1.00001h;
```
aleks01010101: 'brdfData.roughness2 - 1.h' looks as invariant as 'brdfData.normalizationTerm'. Perhaps it's worth computing it once as well?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Particles.hlsl
```diff
@@ -161,7 +161,7 @@ half4 readTexture(TEXTURE2D_ARGS(_Texture, sampler_Texture), VertexOutputLit IN)
 half3 NormalTS(VertexOutputLit IN)
 {
 #if defined(_NORMALMAP)
-    #if BUMP_SCALE_NOT_SUPPORTED
+    #if !SHADER_HINT_NICE_QUALITY
```
aleks01010101: Same as above, this is incorrect.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandardParticlesSimpleLighting.shader
```diff
@@ -79,7 +79,6 @@ Shader "LightweightPipeline/Particles/Standard (Simple Lighting)"
             #pragma shader_feature _FADING_ON
             #pragma shader_feature _REQUIRE_UV2
 
-            #define BUMP_SCALE_NOT_SUPPORTED 1
```
aleks01010101: ...and this is the define that controls whether the shader actually supports normal scaling for the normal map.

aleks01010101: you can also find it in the 'LightweightStandardSimpleLighting.shader'

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurface.hlsl
phi-lira: You are right. I reverted the changes and when BUMP_SCALE_NOT_SUPPORTED is not defined definne it now as !SHADER_HINT_NICE_QUALITY.

phi-lira: On a side note. It would have been better if we had defined it to BUMP_SCALE_SUPPORTED instead. Shame I didn't think of that while reviewing PR that got that in. Not sure if it's wise to change it now.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurface.hlsl
phi-lira: texcoord are always float already. I changed it to avoid casting.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Core.hlsl
```diff
@@ -13,6 +13,10 @@
 #endif
 #endif
 
+#ifndef BUMP_SCALE_NOT_SUPPORTED
+#define BUMP_SCALE_NOT_SUPPORTED !SHADER_HINT_NICE_QUALITY
```
aleks01010101: I think it's still OK to change it to BUMP_SCALE_SUPPORTED :)
So basically we now disable normal scaling for normal maps for everything that doesn't say "I want the best picture", right?

phi-lira: Yeah. Performance by Default is the new black 😎

PR #1112, opened 2018-03-20
Added debug exposure parameter to several debug modes.

JulienIgnace-Unity:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Sky/SkyRenderer.cs
```diff
@@ -8,5 +8,15 @@ public abstract class SkyRenderer
...
         public abstract bool IsValid();
+
+        protected float GetExposure(SkySettings skySettings, DebugDisplaySettings debugSettings)
+        {
+            float debugExposure = 0.0f;
+            if(debugSettings != null && debugSettings.DebugNeedsExposure())
+            {
+                debugExposure = debugSettings.lightingDebugSettings.debugExposure;
+            }
+            return skySettings.exposure + debugExposure;
```
sebastienlagarde: I don't think this is correct to apply exposure here, you apply it on specular and diffuse lighting already, so here it will add up no ? You should be able to check it with the lux meter more. Check the value with exposure zero of debug diffuse lighting mode of a sphere under sky and do the same with varying debug exposue.

JulienIgnace-Unity: The debug exposure should only be applied when rendering into the main view, not the cube map so it will not affect lighting... there was a bug though, pushed a new commit to fix the issue.

PR #1121, opened 2018-03-23
Add local box volume support to volumetrics

EvgeniiG:
Constant density for now, 3D texture sampling coming later.

Side note: we should probably rename _public enum LightCategory_ to something like _LightLoopEntityType_.


ScriptableRenderPipeline/Core/CoreRP/GeometryUtils.cs
```diff
@@ -125,7 +130,7 @@ public static class GeometryUtils
             planes[0].distance = obb.extentX;
             planes[1].normal   = obb.up;
             planes[1].distance = obb.extentY;
-            planes[2].normal   = forward;
+            planes[2].normal   = obb.forward;
```
JulienIgnace-Unity: Not sure if the compiler is good enough (or even have the right) to optimize out the non-trivial "forward" getter here.
Maybe someone like @Chman can tell us :)

PR #1126, opened 2018-03-26
Revert modified package.json during prepare

alexlian:
The publish step requires a pristine tree, so revert to aim for cleanliness


alexlian:
Probably for automation there, but I'll follow up on that later.

PR #1127, opened 2018-03-27
Lw/surface cb split

aleks01010101:


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
```diff
@@ -0,0 +1,63 @@
...
+    return SAMPLE_TEXTURE2D(_MainTex, sampler_MainTex, uv);
+}
+
+half3 Normal(float2 uv)
+{
+#if _NORMALMAP
+    #if BUMP_SCALE_NOT_SUPPORTED
+        return UnpackNormal(SAMPLE_TEXTURE2D(_BumpMap, sampler_BumpMap, uv));
+    #else
+        return UnpackNormalScale(SAMPLE_TEXTURE2D(_BumpMap, sampler_BumpMap, uv), _BumpScale);
```
phi-lira: Not something to look at this point, but it seems wrong that common depends on each Surface include file. We could remove the dependency if we pass as argument the constants. (_BumpScale, _EmissionColor, etc) and we rename the functions SampleNormal, SampleEmission, etc. _MainTex_ST and Alpha functions there would need more thought.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Particles.hlsl
```diff
@@ -214,7 +226,7 @@ half AlphaBlendAndTest(half alpha)
 #else
     half result = 1.0h;
 #endif
-    AlphaDiscard(result, _Cutoff, 0.0001h);
+    AlphaDiscard(result, (half)_Cutoff, 0.0001h);
```
phi-lira: isn't _Cutoff already half?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/ParticlesPBR.hlsl
```diff
@@ -0,0 +1,39 @@
+#ifndef LIGHTWEIGHT_PARTICLES_PBS_INCLUDED
+#define LIGHTWEIGHT_PARTICLES_PBS_INCLUDED
+
+#include "Particles.hlsl"
+#include "SurfaceData.hlsl"
+
+TEXTURE2D(_MetallicGlossMap);   SAMPLER(sampler_MetallicGlossMap);
+
+CBUFFER_START(UnityPerMaterial_ParticlePBS)
```
phi-lira: I wonder if we have any tradeoff between CB and global. a CB for 2 halfs seems an overshoot at first.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/SurfaceData.hlsl
```diff
@@ -0,0 +1,17 @@
+#ifndef LIGHTWEIGHT_SURFACE_DATA_INCLUDED
+#define LIGHTWEIGHT_SURFACE_DATA_INCLUDED
+
+// Must match Lightweigth ShaderGraph master node
```
phi-lira: Did shader graph templates got updated with this? Or maybe we could add this to InputCommon.hlsl?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Particles.hlsl
aleks01010101: Hmm... It is, but the compiler complained about a possible precision loss

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/SurfaceData.hlsl
aleks01010101: It's included in InputSurfaceCommon

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/ParticlesPBR.hlsl
aleks01010101: Perhaps I should just put these two in the common particle CB and be done with it then? :)

PR #1130, opened 2018-03-27
Refactored the LWRP common code to not depend on...

aleks01010101:
…specific global variable names.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
@@ -45,7 +45,7 @@ half4 SampleAlbedoAlpha(float2 uv, TEXTURE2D_ARGS(albedoAlphaMap, sampler_albedo
     return SAMPLE_TEXTURE2D(albedoAlphaMap, sampler_albedoAlphaMap, uv);
 }
 
-half3 SampleNormal(float2 uv, float scale, TEXTURE2D_ARGS(bumpMap, sampler_bumpMap))
+half3 SampleNormal(float2 uv, TEXTURE2D_ARGS(bumpMap, sampler_bumpMap), half scale = 1.0h)
 {
 #if _NORMALMAP
     half4 n = SAMPLE_TEXTURE2D(bumpMap, sampler_bumpMap, uv);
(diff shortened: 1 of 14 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
```diff
@@ -25,50 +25,46 @@ struct SurfaceData
...
 
-half4 MainTexture(float2 uv)
+half4 SampleAlbedoAlpha(float2 uv, TEXTURE2D_ARGS(albedoAlphaMap, sampler_albedoAlphaMap))
 {
-    return SAMPLE_TEXTURE2D(_MainTex, sampler_MainTex, uv);
+    return SAMPLE_TEXTURE2D(albedoAlphaMap, sampler_albedoAlphaMap, uv);
 }
 
-half3 Normal(float2 uv)
+half3 SampleNormal(float2 uv, float scale, TEXTURE2D_ARGS(bumpMap, sampler_bumpMap))
```
phi-lira: scale should be half. You could also declare it with a default parameters like this:
SampleNormal(..., ..., half scale = 1.0h)

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
```diff
@@ -25,50 +25,46 @@ struct SurfaceData
...
 }
 
-half3 Emission(float2 uv)
+half3 SampleEmission(float2 uv, half3 emissionColor, TEXTURE2D_ARGS(emissionMap, sampler_emissionMap))
 {
 #ifndef _EMISSION
     return 0;
 #else
-    return SAMPLE_TEXTURE2D(_EmissionMap, sampler_EmissionMap, uv).rgb * _EmissionColor.rgb;
+    return SAMPLE_TEXTURE2D(emissionMap, sampler_emissionMap, uv).rgb * emissionColor.rgb;
```
phi-lira: emissionColor.rgb not necessary anymore here.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/InputSurfaceCommon.hlsl
aleks01010101: good catch!
Usually I have a personal preference to not have parameters with default values, as the parameter becomes somewhat hidden.

PR #1132, opened 2018-03-27
fix decal projector not reacting to change in transform, add more des…

Paulm-Unity:
…criptive comments to DecalUtilities shader code.


laurenth-unity:
It works in my previously failing test case

PR #1142, opened 2018-04-03
Refactor CoreRP root files

fredpointzero:
Moved root files into folder for easier maintenance


sebastienlagarde:
@JulienIgnace-Unity  @EvgeniiG  what do you think about "Miscellaneous" name, any though? wondering if we should have better classification

sebastienlagarde:
Or maybe use "Misc" instead

EvgeniiG:
My take: classes/templates should go into "Common", free functions should go into "Utilities".

PR #1144, opened 2018-04-03
No depth prepass when possible on LW

aleks01010101:


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -691,24 +691,19 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
             {
-                // If msaa is enabled we don't use a depth renderbuffer as we might not have support to Texture2DMS to resolve depth.
-                // Instead we use a depth prepass and whenever depth is needed we use the 1 sample depth from prepass.
-                // Screen space shadows require depth before opaque shading.
-                if (!msaaEnabled && !shadows)
-                {
-                    bool supportsDepthCopy = m_CopyTextureSupport != CopyTextureSupport.None && m_Asset.CopyDepthShader.isSupported;
-                    m_DepthRenderBuffer = true;
-                    intermediateTexture = true;
+                bool supportsDepthCopy = m_CopyTextureSupport != CopyTextureSupport.None && m_Asset.CopyDepthShader.isSupported;
```
phi-lira: Not related to your changes, but checking the code seems like this is a bug. It should be || instead of &&. We either issue a CopyTexture command or blit with the CopyDepthShader. Can you fix it?

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -691,24 +691,19 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
-                        configuration |= (supportsDepthCopy) ? FrameRenderingConfiguration.DepthCopy : FrameRenderingConfiguration.DepthPrePass;
-                }
-                else
-                {
-                    configuration |= FrameRenderingConfiguration.DepthPrePass;
-                }
+                // If we can't do a blit from an MSAA depth to resolve depth,
+                // we use a depth prepass and whenever depth is needed we use the 1 sample depth from prepass.
+                // Screen space shadows require depth before opaque shading.
+                bool requiresDepthPrePass = (shadows && m_ShadowSettings.screenSpace) || !supportsDepthCopy || m_Asset.PreferDepthPrepassOverCopy();
```
phi-lira: What happens on platforms that can't resolve multisampled depth without Texture2DMS? Don't we need an extra check for that here?

ScriptableRenderPipeline/LightweightPipeline/LWRP/Data/LightweightPipelineAsset.cs
```diff
@@ -67,6 +67,7 @@ public class LightweightPipelineAsset : RenderPipelineAsset
         [SerializeField] private ShadowCascades m_ShadowCascades = ShadowCascades.FOUR_CASCADES;
         [SerializeField] private float m_Cascade2Split = 0.25f;
         [SerializeField] private Vector3 m_Cascade4Split = new Vector3(0.067f, 0.2f, 0.467f);
+        [SerializeField] private bool m_PreferDepthPrepassOverCopy = false;
```
phi-lira: If I understood correctly the usage of having a toogle for requesting depth prepass would be to optimize overdraw. If that's the case IMO we should:
 1) generate a depth pre pass regardless of m_RequireDepthTexture below
 2) call this m_RequireDepthPrepass to be consistent with the other toogles (m_RequireDepthTexture, m_RequireSoftParticles). This is also more clear to what is the toogle. You get a depth prepass no matter what if you request it.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
phi-lira: Also, if we consider purposing the toogle to require depth prepass we should move the requiresDepthPrePass to outside the check of m_RequireDepthTexture.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -691,24 +691,19 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
+                m_DepthRenderBuffer = !shadows;
```
phi-lira: should be updated to !requiresDepthPrepass

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -691,24 +691,19 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
+                intermediateTexture = intermediateTexture || !shadows;
```
phi-lira: Same here. !requiresDepthPrepass instead of !shadows or to be more clear use m_DepthRenderBuffer

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -691,24 +691,19 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
+
+                // If requiring a camera depth texture we need separate depth as it reads/write to depth at same time
+                // Post process doesn't need the copy
+                if (shadows || (!m_Asset.RequireDepthTexture && postProcessEnabled))
+                    configuration |= (requiresDepthPrePass) ? FrameRenderingConfiguration.DepthPrePass : FrameRenderingConfiguration.DepthCopy;
```
phi-lira: This piece of code seems to be quite legacy and I think some things have changed since then + with theses changes it doesn't make sense anymore. DepthPrepass and DepthCopy are orthogonal to each other. Now if you have requiresDepthPrePass as true the Prepass flag should be set. DepthCopy is used to know if we should copy our depth texture to a secondary work depth texture. This is because sometimes you need to do read & write to depth at same time (e.g, soft particles). While with postprocess you don't need to copy.

For now I think it's safer to just always add DepthCopy flag to configuration here. (because we are already inside the m_RequireDepthTexture). Depth copy would **not** be needed only if the only reason you need a depth texture is because of postprocessing. It will require some reworking, but we can do later.

Issue #1145, opened 2018-04-03
BufferPyramid.cs uses Compute Dispatch UAV writes to the color pyramid, but doesn't have enableRandomWrite set for them


strandborg:
See topic.

PR #1146, opened 2018-04-03
Fix for issue 1145 and also fix standalone builds

strandborg:


robbiesri:
I'm not seeing the actual problem in BufferPyramid.cs. m_ColorPyramidBuffer doesn't get used as the _destination_ for a dispatch, so it doesn't need enableRandomWrite.

* We (scaled) blit the original color texture into m_ColorPyramidBuffer, so no need for UAV.
* We set m_ColorPyramidBuffer as our first 'src', but we only use it once, and just as a source texture in the m_ColorPyramidKernel dispatch.  For the rest of the loop iterations, the source is the previous level in 
m_ColorPyramidMips. No need for UAV.
* m_ColorPyramidBuffer is the eventual destination after our compute job, where we call CopyTexture from m_ColorPyramidMips into a mip level of m_ColorPyramidBuffer. Again, no need for UAV.

Ping @fredpointzero because now that I'm out of the org, I can't assign reviewers 😭 

Oh, GH doesn't seem to detect issue numbers in the title, so we can use this #1145.

strandborg:
@robbiesri at least as far as I see, TexturePadding does write to the color pyramid so it needs to have randomWrite enabled. Confusingly enough, the parameter for TexturePadding is still called source, but the compute shader does write to it.

robbiesri:
OMG, it's true, sorry about that @strandborg, my mistake.

Fred, this is really "not obvious" code.  I wish I had commented on this earlier, totally missed this :(

sebastienlagarde:
Rob.  These change has been added last week (texture padding). So you were not able to know about it before :).

fredpointzero:
Sorry for the naming, I'll make a PR to change this.
The Color Pyramid needs to be padded so it is sampled correctly when using NPOT viewport in a POT render texture, and I forgot the UAV declaration.

PR #1150, opened 2018-04-05
Build Automation fixes

alexlian:
Fix the scripts to point correctly at packages.

Though, 2018.1 is not up to date with the rest of the package consistency changes and this skips a couple of those files. Until that's done, validation will not pass on 2018.1

### DIFF
--- a/build.py
+++ b/build.py
@@ -131,8 +131,8 @@ def prepare(logger):
     import unity_package_build
     for name in publish_order:
         package_path = os.path.join(sub_package_folders[name])
-        unity_package_build.copy_file_to_project("LICENSE.md", ".", package_path, logger)
-        unity_package_build.copy_file_to_project("CHANGELOG.md", ".", package_path, logger)
+#        unity_package_build.copy_file_to_project("LICENSE.md", ".", package_path, logger)
+#        unity_package_build.copy_file_to_project("CHANGELOG.md", ".", package_path, logger)
 
 def cleanup(logger):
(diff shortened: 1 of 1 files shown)


sebastienlagarde:
>Though, 2018.1 is not up to date with the rest of the package consistency changes and this skips a couple of those files. Until that's done, validation will not pass on 2018.1

@alexlian What do you mean? We must update the build system with change that have been done in master?  (cc @stramit  - @phi-lira )

alexlian:
Various non-code structural issues of the package layout were addressed in master, but after 2018.1 branched. Minor things like converting files to .md, etc. Something that probably needs to be caught up on, but I leave it to the team to decide to do or not.

phi-lira:
Out of curiosity. I can see we copy SRPMARKER files. Do we really need that now?

pbbastian:
The test framework also uses the SRPMARKER

alexlian:
In the new model that Peter is working on, we can probably skip SRPMARKER....but until then, yes, still necessary in the automation.

sebastienlagarde:
Note to provide some update. We are preparing a new package with up to date code for 2018.1 in a separate branch: 2018.1-merge-master. We are currently testing it with QA to ensure stability. If it is ok, then I will update the change of this PR to this branch and we will made the package correctly. So in the mean time we don't merge this PR.

sebastienlagarde:
Note: Change are already present in main branch.

PR #1152, opened 2018-04-05
Lw/multipass

phi-lira:
First PR to scale how rendering is performed between PC/Console and Mobile.
  - Prepares work to start supporting multiple passes
  - On PC/Console
  --  Shades 1 directional light + up to 8 local lights (could be increased) in a single pass. 
  --  Uses StructuredBuffer to index lights
  --  No light popping
  - On Mobile and WebGL
  -- Shades 1 directional light + up to 4 local lights
  -- Fast light indexing with a couple of conditional moves
  -- Might have light popping (will be solved with multiple pass strategy)
  - This PR adds support to multiple shadow casting lights on LWRP. (1 directional and up to 4 local spots)
  -- Directional shadow uses a shadowmap atlas with screenspace shadows.
  -- Local lights shadows are baked into a separate shadow atlas. This is done so single directional light still has a fast path when shadows are enabled (shadowCoord computed on vert, no atlas bias/scale, bounds check)
  -- For now Local shadows are disabled on mobile (until we get keyword stripping to do it fast)
  - Moved shadow pass to a separate file. This is a first step to modularize LWRP into passes instead of a single file.
  - Fixed a few issues. 

Performance Improvements (Tested on S5 - Adreno 330):
 - 1 Directional Light
 -- 5ms (avg) improvement on nightmares scene
 -- 16ms (med) improvement on nightmares scene
       

What's planned for next PRs:
 - Additional Directional lights will be shaded each in an additional pass
 - On Mobile Additional Local Lights (exceeding 4) will be shaded in groups of 4 in additional passes. In order to make this fast we need to have exposed a way to mask out visible renderers in C#.
 - UI for Local Light shadows (Light Inspector and Pipeline Inspector)
 - Add keyword for local shadow support
 - Add support to point light shadows

DISCLAIMER:
 If you have Android Emulation set in the editor you might get light issues due to ES 3.0 not supporting ComputeBuffers. Be sure to turn off Android emulation.


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Lighting.hlsl
```diff
@@ -139,32 +136,35 @@ half4 GetMainLightDirectionAndAttenuation(LightInput lightInput, float3 position
...
-    lightInput.spotAttenuation = _MainLightSpotAttenuation;
-
-    half4 directionAndRealtimeAttenuation = GetMainLightDirectionAndAttenuation(lightInput, positionWS);
-
-    Light light;
-    light.direction = directionAndRealtimeAttenuation.xyz;
-    light.attenuation = directionAndRealtimeAttenuation.w;
-    light.subtractiveModeAttenuation = lightInput.distanceAttenuation.w;
-    light.color = lightInput.color;
+    Light light = (Light)0;
```
aleks01010101: I'm not sure if this would fill everything with zeroes first and then overwrite with the valies in the next lines. I guess setting the .lightIndex to 0 would do the same.

PR #1153, opened 2018-04-05
Initial push

Verasl:
-Fixed Shader Graph PBR template shadows and defines to match LW
-Fixed Shader Graph Unlit template defines to match LW
-Culling in depth pre pass fixed in LW PBR

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
@@ -136,7 +136,6 @@ Shader "LightweightPipeline/Standard (Physically Based)"
             // -------------------------------------
             // Material Keywords
             #pragma shader_feature _ALPHATEST_ON
-            #pragma shader_feature _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A
 
             //--------------------------------------
             // GPU Instancing
@@ -156,6 +155,7 @@ Shader "LightweightPipeline/Standard (Physically Based)"
 
             ZWrite On
             ColorMask 0
+            Cull[_Cull]
 
             HLSLPROGRAM
             // Required to compile gles 2.0 with standard srp library
@@ -169,7 +169,6 @@ Shader "LightweightPipeline/Standard (Physically Based)"
             // -------------------------------------
             // Material Keywords
             #pragma shader_feature _ALPHATEST_ON
-            #pragma shader_feature _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A
 
             //--------------------------------------
             // GPU Instancing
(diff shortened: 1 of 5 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
```diff
@@ -136,7 +136,6 @@ Shader "LightweightPipeline/Standard (Physically Based)"
             // -------------------------------------
             // Material Keywords
             #pragma shader_feature _ALPHATEST_ON
-            #pragma shader_feature _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A
```
phi-lira: This is required as we use it to decide if MainTexture/Albedo alpha is transparency/cutout or not.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
```diff
@@ -169,7 +169,6 @@ Shader "LightweightPipeline/Standard (Physically Based)"
...
```
phi-lira: Same as above. This can't be removed.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
```diff
@@ -156,6 +155,7 @@ Shader "LightweightPipeline/Standard (Physically Based)"
 
             ZWrite On
             ColorMask 0
+            Cull[_Cull]
```
phi-lira: Not really important for this PR but it would be nice if you can also add this to Standard (Simple Lighting) and Terrain.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
Verasl: Doesn't the alpha of albedo always get used, isn't this purely for if we use metallic alpha or albedo alpha for the smoothness?

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
Verasl: same reply as above :D

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
Verasl: Ah yea forgot about simple lighting, but terrain shouldn't need it as you shouldn't be not culling terrain, I don't even think we can?

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
phi-lira: No. If smoothness is packed into Albedo alpha we only use _Color.a, otherwise we modulate color with Albedo.a;

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightStandard.shader
phi-lira: We definitely can. Face culling is something fixed in the pipeline and works at triangle level. Not sure if Terrain engine has something forced in terms of face culling that disregard this setting though. Anyway, we also don't have UI for it so it's safe to disregard this change.

PR #1165, opened 2018-04-10
Lw/fix shadow tile initialization

phi-lira:
- Added Clear to Screenspace shadow occlusion texture to avoid Tile LOAD
 - Moved BlitQuad to a static mesh in LightweightUtils

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
@@ -65,7 +65,7 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
         public static void DrawFullScreen(CommandBuffer commandBuffer, Material material,
             MaterialPropertyBlock properties = null, int shaderPassId = 0)
         {
-            commandBuffer.DrawMesh(fullscreenMesh, Matrix4x4.identity, material, 0, 0, properties);
+            commandBuffer.DrawMesh(fullscreenMesh, Matrix4x4.identity, material, 0, shaderPassId, properties);
         }
 
         public static void StartStereoRendering(Camera camera, ref ScriptableRenderContext context, FrameRenderingConfiguration renderingConfiguration)
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using UnityEngine.Rendering;
 
 namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 {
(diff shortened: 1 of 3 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightShadowPass.cs
```diff
@@ -202,6 +202,8 @@ public void CollectShadows(Camera camera, FrameRenderingConfiguration frameRende
             // stereo functionality, we use the screen-space shadow map as the source (until we have
             // a better solution).
             // An alternative would be DrawProcedural, but that would require further changes in the shader.
+            cmd.SetRenderTarget(m_ScreenSpaceShadowmapTexture);
+            cmd.ClearRenderTarget(true, true, Color.white);
```
aleks01010101: Some GPUs prefer calling discard instead of clear, please check if this is exposed in the C# API.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
```diff
@@ -27,6 +28,46 @@ public enum FrameRenderingConfiguration
...
+
+                mesh.SetUVs(0, new List<Vector2>
+                {
+                    new Vector2(0.0f, bottomV),
+                    new Vector2(0.0f, topV),
+                    new Vector2(1.0f, bottomV),
+                    new Vector2(1.0f, topV)
+                });
+
+                mesh.SetIndices(new[] { 0, 1, 2, 2, 1, 3 }, MeshTopology.Triangles, 0, false);
```
aleks01010101: Nitpicking: we can save 1 vertex and 3 indices by drawing a triangle that covers the whole screen. Not sure if this would result in any performance changes though.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
```diff
@@ -27,6 +28,46 @@ public enum FrameRenderingConfiguration
...
+                mesh.UploadMeshData(true);
+                return mesh;
+            }
+        }
+
+        public static void DrawFullScreen(CommandBuffer commandBuffer, Material material,
+            MaterialPropertyBlock properties = null, int shaderPassId = 0)
```
aleks01010101: shaderPassId not used

phi-lira: Fixed

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightShadowPass.cs
phi-lira: There's nothing on CommandBuffer API. Only RenderTexture.DiscardContents. I could check that in a next PR.

ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipelineUtils.cs
phi-lira: I think the big improvements of the fullscreen triagle are from not setting the vertex/index buffer at all. That depends on VertexID being supported and requires some changes on the shader side. 

We could graphics caps and do that when supported. There's already C# and Shader api exposed to render a procedural fullscreen triangle.

PR #1168, opened 2018-04-10
HDRenderPipeline: Remove a warning

sebastienlagarde:
Remove this warning in Pyramid buffer: @fredpointzero any though about it ?


fredpointzero:
No this is good, dstMip is actually unused

PR #1187, opened 2018-04-12
Terrain Shader fixes

Verasl:
Some fixes from the branch 'lw/terrain-shader-fixes' and just the high P ones for 2018.1 release.
-Terrain no longer pink with more than 4 textures painted(terrain add pass)
-Terrain now casts shadows(fix since SSShadows were implemented)
-Terrain now has a base shader which takes over after 'Base Map Dist' distance
-Added WavingGrass & WavingGrassBillboard shaders converted for Lightweight
-Tweaked default grass billboard to be vertical billboarding rather than full billboards


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitGrass.hlsl
```diff
@@ -0,0 +1,158 @@
...
+
+    half3 emission = 0;
+    half4 specularGloss = 0.1;// SampleSpecularGloss(uv, diffuseAlpha.a, _SpecColor, TEXTURE2D_PARAM(_SpecGlossMap, sampler_SpecGlossMap));
+    half shininess = IN.posWSShininess.w;
+
+    InputData inputData;
+    InitializeInputData(IN, inputData);
+
+    return LightweightFragmentBlinnPhong(inputData, diffuse, specularGloss, shininess, emission, alpha);
+};
```
phi-lira: Doesn't grass support fog?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitGrass.hlsl
```diff
@@ -0,0 +1,158 @@
...
+    alpha *= IN.color.a;
+
+    half3 emission = 0;
+    half4 specularGloss = 0.1;// SampleSpecularGloss(uv, diffuseAlpha.a, _SpecColor, TEXTURE2D_PARAM(_SpecGlossMap, sampler_SpecGlossMap));
+    half shininess = IN.posWSShininess.w;
+
+    InputData inputData;
+    InitializeInputData(IN, inputData);
+
+    return LightweightFragmentBlinnPhong(inputData, diffuse, specularGloss, shininess, emission, alpha);
```
phi-lira: Is it intentional that grass is shader with BlinnPhong while Terrain is shaded with Physically Based?

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitTerrain.hlsl
```diff
@@ -0,0 +1,165 @@
...
+    splat_control = SAMPLE_TEXTURE2D(_Control, sampler_Control, IN.uvControlAndLM.xy);
+    weight = dot(splat_control, 1);
+
+#if !defined(SHADER_API_MOBILE) && defined(TERRAIN_SPLAT_ADDPASS)
+    clip(weight == 0.0f ? -1 : 1);
+#endif
+
+    // Normalize weights before lighting and restore weights in final modifier functions so that the overal
+    // lighting result can be correctly weighted.
+    splat_control /= (weight + 1e-3f);
```
phi-lira: Not related to your change but now we have HALF_MIN/FLOAT_MIN in SRP library. Fix in next PR.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitTerrain.hlsl
```diff
@@ -0,0 +1,165 @@
...
+
+    o.uvSplat01.xy = TRANSFORM_TEX(v.texcoord, _Splat0);
+    o.uvSplat01.zw = TRANSFORM_TEX(v.texcoord, _Splat1);
+    o.uvSplat23.xy = TRANSFORM_TEX(v.texcoord, _Splat2);
+    o.uvSplat23.zw = TRANSFORM_TEX(v.texcoord, _Splat3);
+    o.uvControlAndLM.xy = TRANSFORM_TEX(v.texcoord, _Control);
+    o.uvControlAndLM.zw = v.texcoord1 * unity_LightmapST.xy + unity_LightmapST.zw;
+
+#ifdef _TERRAIN_NORMAL_MAP
+    float4 vertexTangent = float4(cross(v.normal, float3(0, 0, 1)), -1.0);
```
phi-lira: IIRC this was copied from Standard terrain shader. Taking a look now it seems weird that we compute tangent this way, despite the tangent being available in VertexInput. Maybe we should check if normalmap is working.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitGrass.hlsl
Verasl: The main reason here is the grass has absolutely no settings you can choose other than the albedo texture, this means we would have to hardcode a set metallic and smoothness value which would work best in a lot of cases, terrain has metallic slider and a smoothness map/value, the original grass was a lambert shader and being that the user cannot choose otherwise what they use I thought for grass it would be best to take the cheaper approach.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitTerrain.hlsl
Verasl: Yea there was some weird stuff in here, but I can say that normal maps do work and seem to blend correctly.

ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Terrain/LightweightPassLitGrass.hlsl
phi-lira: Ok. As long as _SPECULAR_COLOR and _SPECGLOSSMAP (which I think it isn't) this will be lambert only as well.

PR #1190, opened 2018-04-12
 Fix shader include case sensitiviness

phi-lira:
Backport of #1171


phi-lira:
Added as reviewer for awareness. This is the same thing of #1171 but I actually rerunned the regexpression as files might have changed.

PR #1200, opened 2018-04-13
 Fixed negative pow warning.

phi-lira:
Fix negative pow warning. 
  For awareness. I understand the restrictions on warning in template, however IMO this is a risky change due to EntityLighting being used by all Phyically Based Shaders.


sebastienlagarde:
I don't understand your comment, this is safe. implementation of positivePow is:

TEMPLATE_2_REAL(PositivePow, base, power, return pow(abs(base), power))

abs() is free on many platform (but yes, maybe not on mobile), and .a is positive so all fine.

PR #1211, opened 2018-04-16
fix texture/sampler naming scheme for WIND_SETTINGS_Tex*

aet:
fixes pixelized trees issue with Fontainebleau on metal, probably on some other platforms too

-Set 2D Texture "WIND_SETTINGS_TexNoise" to slot 0 sampler slot -1
-Set 2D Texture "WIND_SETTINGS_TexGust" to slot 1 sampler slot -1
+Set 2D Texture "WIND_SETTINGS_TexNoise" to slot 0
+Set 2D Texture "WIND_SETTINGS_TexGust" to slot 1


sebastienlagarde:
Thanks for this correction. However I am wondering what is the meaning of this. On Windows, there is no problem of having a different name for sampler and texture, this is currently the goal. And we have keep Unity legacy convention just for convenience but this isn't required. So I deduce from this that the metal or HLSCC backend assume hardcoded name for sampler (or texture that match sampler) which feel to me wrong and would case mistake from users in the future. Or I missed the point ?

aet:
It looks like Unity code outside of Metal was looking up both "sampler" and "sampler_" prefix, fixed Metal codegen as well

PR #1223, opened 2018-04-18
Lw/get rid of depth prepass when possible

aleks01010101:


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -563,24 +568,17 @@ private void SetupFrameRenderingConfiguration(out FrameRenderingConfiguration co
...
-
-                    // If requiring a camera depth texture we need separate depth as it reads/write to depth at same time
-                    // Post process doesn't need the copy
-                    if (!m_Asset.RequireDepthTexture && postProcessEnabled)
-                        configuration |= (supportsDepthCopy) ? FrameRenderingConfiguration.DepthCopy : FrameRenderingConfiguration.DepthPrePass;
-                }
-                else
-                {
+                bool supportsDepthCopy = m_CopyTextureSupport != CopyTextureSupport.None || m_Asset.CopyDepthShader.isSupported;
+                bool requiresDepthPrepassToResolveMSAA = msaaEnabled && (SystemInfo.supportsMultisampledTextures == 0);
```
phi-lira: Shouldn't it be != 0?

aleks01010101: If MSAA is enabled and we can't use Tex2DMS, we need a prepass. I think `== 0` is correct.

phi-lira: Ah ok. The caps refer to Texture2DMS support.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightCopyDepthMSAA.shader
```diff
@@ -0,0 +1,71 @@
...
+
+        Pass
+        {
+            ZTest Always ZWrite On ColorMask 0
+
+            HLSLPROGRAM
+            // Required to compile gles 2.0 with standard srp library
+            #pragma prefer_hlslcc gles
+            #pragma exclude_renderers d3d11_9x
+            #pragma vertex vert
```
phi-lira: Should we also exclude gles2 and remove the prefer_hlslcc here?

aleks01010101: ` #pragma require msaatex` excludes es2 automatically, so I can just remove the line

PR #1230, opened 2018-04-19
Buffered RT History per camera

fredpointzero:
Added a Buffered RT system per camera.

- Moved the RTHandle class to an instantiable class.
- Added a BufferedRTHandleSystem
- Added a buffer history in HDCamera


ScriptableRenderPipeline/Core/CoreRP/Textures/BufferedRTHandleSystem.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace UnityEngine.Experimental.Rendering
+{
+    public class BufferedRTHandleSystem : IDisposable
+    {
+        Dictionary<int, RTHandleSystem.RTHandle[]> m_RTHandles = new Dictionary<int, RTHandleSystem.RTHandle[]>();
+
+        RTHandleSystem m_RTHandleSystem = new RTHandleSystem();
+        bool m_DisposedValue = false;
+
+        public RTHandleSystem.RTHandle GetFrameRT(int id, int index)
```
EvgeniiG: This function needs comments/renaming. 'id' and 'index' are almost interchangeable in my book, and I wouldn't be sure what to pass. I suppose 'id' is the RenderTargetIdentifier and 'index' is FrameIndex or FrameCount? Also, shouldn't you use modulo for indexing inside the function? I'm not sure it's a good idea for API users to have to take care of this.

ScriptableRenderPipeline/Core/CoreRP/Textures/BufferedRTHandleSystem.cs
```diff
@@ -0,0 +1,96 @@
...
+            foreach (var item in m_RTHandles)
+            {
+                var nextFirst = item.Value[item.Value.Length - 1];
+                for (int i = 0, c = item.Value.Length - 1; i < c; ++i)
+                    item.Value[i + 1] = item.Value[i];
+                item.Value[0] = nextFirst;
+
+                // First is autoresize, other are on demand
+                m_RTHandleSystem.SwitchResizeMode(item.Value[0], RTHandleSystem.ResizeMode.Auto);
+                m_RTHandleSystem.SwitchResizeMode(item.Value[1], RTHandleSystem.ResizeMode.OnDemand);
```
EvgeniiG: Can't you just resize the current buffer each frame? Literally:

var currentFrameBuffer = buffers[frameIndex & (bufferCount - 1)];
currentFrameBuffer.Resize();

ScriptableRenderPipeline/Core/CoreRP/Textures/DepthBits.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
```
EvgeniiG: Do we need a new file per enum?

ScriptableRenderPipeline/Core/CoreRP/Textures/BufferedRTHandleSystem.cs
fredpointzero: I'll update `index`to `frameIndex` and `id`to `frameId`

For the modulo, The user define the number of frame to store in history, so he should be aware of the size of the frame buffer.

To have safer code, I can clamp the frameIndex, but still I think that issuing a warning that the index is out of range will avoid confusion (as I do net return the requested frame)

ScriptableRenderPipeline/Core/CoreRP/Textures/BufferedRTHandleSystem.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            Assert.IsTrue(index >= 0 && index < m_RTHandles[id].Length);
+
+            return m_RTHandles[id][index];
+        }
+
+        public void AllocBuffer(
+            int id, 
+            Func<RTHandleSystem, int, RTHandleSystem.RTHandle> allocator,
+            int bufferSize
```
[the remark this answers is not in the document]
fredpointzero: Yes, will change that

ScriptableRenderPipeline/Core/CoreRP/Textures/DepthBits.cs
fredpointzero: I usually do this, one file per class / enum / struct. I find it easier to handle

ScriptableRenderPipeline/Core/CoreRP/Textures/BufferedRTHandleSystem.cs
EvgeniiG: SwitchResizeMode() doing anything other than switching the resize mode is misleading.

PR #1244, opened 2018-04-20
Lw/depth copy vr support

aleks01010101:


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/DepthCopy.hlsl
```diff
@@ -0,0 +1,81 @@
...
+    UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(o);
+    o.uv = i.uv;
+    o.position = TransformObjectToHClip(i.vertex.xyz);
+    return o;
+}
+
+#if defined(UNITY_STEREO_INSTANCING_ENABLED) || defined(UNITY_STEREO_MULTIVIEW_ENABLED)
+#define DEPTH_TEXTURE_MS Texture2DMSArray
+#define DEPTH_TEXTURE(name) TEXTURE2D_ARRAY(name)
+#define LOAD(uv, sampleIndex) LOAD_TEXTURE2D_ARRAY_MSAA(_CameraDepthTexture, uv, unity_StereoEyeIndex, sampleIndex)
```
phi-lira: For awareness. There's currently a SCREENSPACE_TEXTURE define that is defined to either TEXTURE2D_ARRAY or TEXTURE2D. But that define is causing some weird behavior in the compiler. We need to investigate the compiler issue.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightCopyDepthMSAA.shader
```diff
@@ -20,49 +20,10 @@ Shader "Hidden/LightweightPipeline/CopyDepthMSAA"
 
             #pragma require msaatex
 
-            #include "LWRP/ShaderLibrary/Core.hlsl"
+            #define MSAA_DEPTH 1
```
phi-lira: It seems that in the end we can have a single shader if we do a multi_compile and Enable/Disable the keyword based on the pipeline MSAA setting / and target Texture2DMS support

aleks01010101: But we won't be able to set the requirements properly.
And we need those to check if the shader is supported.

PR #1252, opened 2018-04-23
 Stripped invalid variants.

phi-lira:
Stripped 57.25% variants in Standard Shader due to being invalid variants.
Standard shader with minimum pipeline capabilities (single directional light - no shadows) has now only   2.08% surviving variants.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
@@ -74,6 +74,26 @@ namespace UnityEditor.Experimental.Rendering.LightweightPipeline
             return false;
         }
 
+        bool StripInvalidVariants(ShaderCompilerData compilerData)
+        {
+            bool isShadowVariant = compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.DirectionalShadows) ||
+                compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.LocalShadows);
+
+            if (compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.SoftShadows) && !isShadowVariant)
+                return true;
+
+            if (compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.VertexLights) &&
+                !compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.AdditionalLights))
+                return true;
+
+            // Note: LWRP doesn't support Dynamic Lightmap.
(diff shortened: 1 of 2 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
```diff
@@ -74,6 +74,26 @@ bool StripUnusedVariant(PipelineCapabilities capabilities, ShaderCompilerData co
...
+
+            if (compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.SoftShadows) && !isShadowVariant)
+                return true;
+
+            if (compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.VertexLights) &&
+                !compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.AdditionalLights))
+                return true;
+
+            // Note: LWRP doesn't support Dynamic Lightmap.
+            if (compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.DirectionalLightmap) &&
```
aleks01010101: What about SH? Aren't those incompatible with lightmaps?

ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
```diff
@@ -74,6 +74,26 @@ bool StripUnusedVariant(PipelineCapabilities capabilities, ShaderCompilerData co
             return false;
         }
 
+        bool StripInvalidVariants(ShaderCompilerData compilerData)
+        {
+            bool isShadowVariant = compilerData.shaderKeywordSet.IsEnabled(LightweightKeywords.DirectionalShadows) ||
```
aleks01010101: I'd make a local var out of `compilerData.shaderKeywordSet` and have less typing 😄

ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
phi-lira: We don't have any SH multi_compile variant. As it stands now, if not LM then SH is computed. We should look into that at some point, but SH is set perobject so we have to figure out which and if a keyword is set.

ScriptableRenderPipeline/LightweightPipeline/LWRP/Editor/ShaderPreprocessor.cs
phi-lira: Wise :D

PR #1259, opened 2018-04-24
 Fixed multiple local shadows.

phi-lira:
Fixes per-object local shadows. We use shadow strength value to decide if shadow is enabled for a particular light. Due to perobject indices we have to setup shadow buffer for all visible lights. 

As a next step we need to convert all light and shadow buffers to ComputeBuffer for platforms that support it and remove the MAX_VISIBLE_LIGHTS cap.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Shadows.hlsl
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Shadows.hlsl
@@ -38,8 +38,8 @@ float4      _ShadowmapSize; // (xy: 1/width and 1/height, zw: width and height)
 CBUFFER_END
 
 CBUFFER_START(_LocalShadowBuffer)
-float4x4    _LocalWorldToShadowAtlas[4];
-half        _LocalShadowStrength[4];
+float4x4    _LocalWorldToShadowAtlas[MAX_VISIBLE_LIGHTS];
+half        _LocalShadowStrength[MAX_VISIBLE_LIGHTS];
 half4       _LocalShadowOffset0;
 half4       _LocalShadowOffset1;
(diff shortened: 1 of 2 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/ShaderLibrary/Shadows.hlsl
```diff
@@ -227,7 +227,8 @@ half LocalLightRealtimeShadowAttenuation(int lightIndex, float3 positionWS)
     float4 shadowCoord = mul(_LocalWorldToShadowAtlas[lightIndex], float4(positionWS, 1.0));
     ShadowSamplingData shadowSamplingData = GetLocalLightShadowSamplingData();
     half shadowStrength = GetLocalLightShadowStrenth(lightIndex);
-    return SampleShadowmap(shadowCoord, TEXTURE2D_PARAM(_LocalShadowMapAtlas, sampler_LocalShadowMapAtlas), shadowSamplingData, shadowStrength);
+    half attenuation = SampleShadowmap(shadowCoord, TEXTURE2D_PARAM(_LocalShadowMapAtlas, sampler_LocalShadowMapAtlas), shadowSamplingData, shadowStrength);
+    return (shadowStrength > 0.0h) ? attenuation : 1.0h;
```
aleks01010101: `SampleShadowmap` already accounts for the shadowStrength by doing `attenuation = LerpWhiteTo(attenuation, shadowStrength)`, which expands to `attenuation = 1.0 - shadowStrength + attenuation * shadowStrength`. When shadowStrength is 0, attenuation becomes 1.

phi-lira: fixed.

PR #1263, opened 2018-04-24
Lw/fix copy depth to scene view camera

aleks01010101:


ScriptableRenderPipeline/LightweightPipeline/LWRP/LightweightPipeline.cs
```diff
@@ -318,24 +317,30 @@ public override void Render(ScriptableRenderContext context, Camera[] cameras)
...
                 if (sceneViewCamera)
+                {
+                    cmd = CommandBufferPool.Get("Copy Depth to Camera");
+                    cmd.DisableShaderKeyword(kMSAADepthKeyword);
                     CopyTexture(cmd, CameraRenderTargetID.depth, BuiltinRenderTextureType.CameraTarget, m_CopyDepthMaterial, true);
+                    context.ExecuteCommandBuffer(cmd);
+                    CommandBufferPool.Release(cmd);
+                }
 #endif
-                context.Submit();
```
phi-lira: I think the reason there was an extra context submit before releasing resources was because shadowmap texture needs to be immediate (RenderTexture.Get/Release). I'm not sure if that code is necessary. Not written by me. Please just check shadows are working and shadowmap is not leaking.

aleks01010101: The shadows work on a device and it doesn't crash, so it doesn't look like it leaks.

PR #1266, opened 2018-04-24
Fix decal atlas debug display so that it works correctly when both de…

Paulm-Unity:
…cal and shadow atlas debug views are enabled


Paulm-Unity:
Merged this myself since it was a fix to debug view only

PR #1268, opened 2018-04-24
Manager for the density volumes

Wadarass:
Use a manager instead of FindObjectsofType


EvgeniiG:
Looks good! Clean and simple to use.

Some thoughts for the future...

Perhaps we could store a DensityVolumeList for all volumes in the manager? This would accelerate the PrepareVisibleDensityVolumeList() step, as we don't have to create re-bounding boxes and convert DensityVolumeParameters (user-facing data) to DensityVolumeData (engine data) each frame.

(Side note: perhaps renaming them to DensityVolumeUserParameters and DensityVolumeEngineData would improve clarity?)

We also have 2x ComputeBuffers for visible volumes, one with OBBs and one with DensityVolumeData.
Perhaps it's more efficient to make these two buffers as persistent (not updated every frame) and instead upload an index buffer for visible volumes each frame? We'll just have to be careful with camera-relative rendering, as currently we transform the boxes into the camera-relative coordinate system on the CPU.

PR #1280, opened 2018-04-26
Volume improvements

Chman:
Added support for instantiated volume profiles in the inspector.


ScriptableRenderPipeline/Core/CoreRP/Editor/Volume/VolumeComponentListEditor.cs
```diff
@@ -233,16 +235,21 @@ internal void AddComponent(Type type)
             Undo.RegisterCreatedObjectUndo(component, "Add Volume Component");
 
             // Store this new effect as a subasset so we can reference it safely afterwards
-            AssetDatabase.AddObjectToAsset(component, m_Asset);
+            // Only when we're not dealing with an intantiated asset
```
JulienIgnace-Unity: instantiated instead of intantiated

Chman: Fixed

PR #1309, opened 2018-05-04
Move ShaderGraph into repository

pbbastian:
This PR moves the ShaderGraph repository inside this repository, including all history. Please test this out and see whether it also works on your machines.


phi-lira:
LGTM

PR #1315, opened 2018-05-04
Fixed warnings as requested by Spotlight customer

roberto-renzetti-unity:

PR #1324, opened 2018-05-07
Fixed warnings as requested by Spotlight customer

sebastienlagarde:
This PR replace the following close PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1315. It provide proper way to fix the shader warning
+ don't do a zero initialize on PreLightData to better track shader warning issue.


roberto-renzetti-unity:
LGTM - thanks for fixing the  warnings the proper way

PR #1329, opened 2018-05-08
 Renderer.flags has dynamic batching disabled by default. Enabled it …

phi-lira:
Dynamic batching was disabled by default in the renderer flags. Enabled it and made instancing flag also set explicitly.


phi-lira:
Created a new PR for this.

PR #1334, opened 2018-05-08
Add enableVolumetric in the FrameSettings

sebastienlagarde:
This PR introduce enableVolumetric flag
- It check the Framesettings instead of VolumetricPreset of 0 for volumetric to enable/disable the usage.

Note: Currently enableVolumetric have no visible effect but save performance (and display the last result), a second PR will introduce multicompile to disable the effect.

### DIFF
--- a/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/RenderLoopSettings/FrameSettingsUI.cs
+++ b/ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/RenderLoopSettings/FrameSettingsUI.cs
@@ -133,6 +133,7 @@ namespace UnityEditor.Experimental.Rendering
             EditorGUILayout.PropertyField(p.enableSSAO, _.GetContent("Enable SSAO"));
             EditorGUILayout.PropertyField(p.enableSubsurfaceScattering, _.GetContent("Enable Subsurface Scattering"));
             EditorGUILayout.PropertyField(p.enableTransmission, _.GetContent("Enable Transmission"));
+            EditorGUILayout.PropertyField(p.enableVolumetric, _.GetContent("Enable Volumetric"));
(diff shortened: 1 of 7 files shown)


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Editor/RenderLoopSettings/FrameSettingsUI.cs
```diff
@@ -133,6 +132,8 @@ static void Drawer_SectionLightingSettings(FrameSettingsUI s, SerializedFrameSet
             EditorGUILayout.PropertyField(p.enableSSAO, _.GetContent("Enable SSAO"));
             EditorGUILayout.PropertyField(p.enableSubsurfaceScattering, _.GetContent("Enable Subsurface Scattering"));
             EditorGUILayout.PropertyField(p.enableTransmission, _.GetContent("Enable Transmission"));
+            EditorGUILayout.PropertyField(p.enableAtmosphericScattering, _.GetContent("Enable Atmospheric Scattering"));
+            EditorGUILayout.PropertyField(p.enableVolumetric, _.GetContent("    Enable Volumetric"));
```
EvgeniiG: I suggest using "Volumetric**s**" rather than "Volumetric".

ScriptableRenderPipeline/HDRenderPipeline/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
```diff
@@ -278,8 +288,10 @@ public void UpdatePerCameraData(HDCamera hdCamera)
 
     void DestroyBuffers()
     {
-        RTHandles.Release(m_DensityBufferHandle);
-        RTHandles.Release(m_LightingBufferHandle);
+        if (m_DensityBufferHandle != null)
```
EvgeniiG: Probably better to do the null check inside the Release() call.

PR #1339, opened 2018-05-08
Stacklit

jeanfrancoisf:


ScriptableRenderPipeline/HDRenderPipeline/HDRP/Material/StackLit/StackLit.hlsl
```diff
@@ -1859,6 +2112,24 @@ float3 EvaluateTransmission(BSDFData bsdfData, float3 transmittance, float NdotL
     return intensity * transmittance;
 }
 
+void EvaluateBSDF_GetNormalUnclampedNdotV(BSDFData bsdfData, PreLightData preLightData, float3 V, out float3 N, out float unclampedNdotV)
+{
+#ifdef _MATERIAL_FEATURE_COAT_NORMALMAP
```
sebastienlagarde: Try to not use  _MATERIAL_FEATURE_COAT_NORMALMAP directly but follow same convention than for toher material features with hasFlags. The goal is to be able to share the code with shader graph later that will use the .materialFeatures properties convention.

PR #1350, opened 2018-05-11
LWRP XR Single-pass Instancing and Multi-view Support

brandon-unity3d:
Added support for single-pass instancing and multi-view support for XR with LWRP.

Changes Made:
1. Shaders that were not stereo aware now are.
2. Merged in most recent gles single pass instancing support changes that I made for the builtin pipeline.
3. Our shader compiler specifically looks for the global constant buffer to determine if it should add the gles multi-view extension.  So I converted the stereo CBUFFERS to GLOBAL_CBUFFERS which now matches the builtin shaders.
4. Fixed a blitting issue in which we were blitting from the wrong source texture.

### DIFF
--- a/ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightScreenSpaceShadows.shader
+++ b/ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightScreenSpaceShadows.shader
@@ -59,12 +59,7 @@ Shader "Hidden/LightweightPipeline/ScreenSpaceShadows"
         half4 Fragment(Interpolators i) : SV_Target
         {
             UNITY_SETUP_INSTANCE_ID(i);
-#if !defined(UNITY_STEREO_INSTANCING_ENABLED)
-            // Completely unclear why i.stereoTargetEyeIndex doesn't work here, considering
-            // this has to be correct in order for the texture array slices to be rasterized to
-            // We can limit this workaround to stereo instancing for now.
             UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(i);
-#endif
 
 #if defined(UNITY_STEREO_INSTANCING_ENABLED) || defined(UNITY_STEREO_MULTIVIEW_ENABLED)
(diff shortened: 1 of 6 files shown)


ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightScreenSpaceShadows.shader
```diff
@@ -59,12 +59,7 @@ Shader "Hidden/LightweightPipeline/ScreenSpaceShadows"
         half4 Fragment(Interpolators i) : SV_Target
         {
             UNITY_SETUP_INSTANCE_ID(i);
-#if !defined(UNITY_STEREO_INSTANCING_ENABLED)
-            // Completely unclear why i.stereoTargetEyeIndex doesn't work here, considering
-            // this has to be correct in order for the texture array slices to be rasterized to
-            // We can limit this workaround to stereo instancing for now.
             UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(i);
```
robbiesri: Did you not run into this issue on your machine?

robbiesri:
Thanks for doing this Brandon.  I think you should also add @Wadarass because he worked on adding the instancing to SRP

ScriptableRenderPipeline/LightweightPipeline/LWRP/Shaders/LightweightScreenSpaceShadows.shader
brandon-unity3d: I didn't run into this issue.  I tested on a variety of configurations including dx11, gles, desktop, and multiple mobile devices supporting the GL_AMD_vertex_shader_layer extension.  The reason this didn't work originally was that it didn't contain the support we added to the builtin pipeline to transfer the selected layer value from the vertex to the fragment shader.

phi-lira:
I'll go ahead and merge this as we need to prepare a new package. @Wadarass if you have comments we can address them in a next PR.

PR #1351, opened 2018-05-12
HDRP master node support

cdxntchou:


sebastienlagarde:
@pbbastian @cdxntchou  Hey, can you look at the conflict of this PR, seems like it assume that a submodule is still there, but there is no more shader graph submodule (or maybe another error). Thanks

PR #1352, opened 2018-05-13
Ensure the blackboard remains in view after deserialization

pbbastian:
_Note: This is a migration of [PR 375](https://github.com/Unity-Technologies/ShaderGraph/pull/375) towards the old repository._

Should resolve issues with the blackboard missing as reported in issues [#353](https://github.com/Unity-Technologies/ShaderGraph/issues/353) and [#365](https://github.com/Unity-Technologies/ShaderGraph/issues/365). However, I have not managed to reproduce the exact issue with the blackboard missing. The fix should ensure that the blackboard is at least not missing from the graph due to old/incompatible serialization.


pbbastian:
@KReguieg Thanks for the feedback :) Does that happen even with this fix? :/

PR #1353, opened 2018-05-13
Fix drag and drop

pbbastian:
_Note: This is a migration of [PR 384](https://github.com/Unity-Technologies/ShaderGraph/pull/384) towards the old repository._

Fixed drag and drop for texture2d, cubemap, subgraph and updated the blackboardfield drag and drop


pbbastian:
Can't add myself as a reviewer since I created the PR, but @martint-unity should be considered the creator and I should be considered a reviewer. I'll take care of merging.

PR #1365, opened 2018-05-18
Define REVERSE_ZBUF for Switch platform.

kaychang-unity:
There is currently a "bug" in the Unity for Switch plugin that incorrectly defines SHADER_API_GLCORE when building Switch shaders. However, Switch uses reversed z implementation, just as the D3D11 and PS4 implementations do on Unity.


sebastienlagarde:
I propose here a better way to solve this issue: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1369  (use the REVERSED Z flag correctly define :) )

sebastienlagarde:
Confirm it is good for you and close this PR if OK. (rely on UNITY_REVERSED_Z)

PR #1388, opened 2018-05-23
HDRP: add geometric filtering to lit shader

sebastienlagarde:


realJakubSmaga:
BaseLitUI.cs, Line: 62 says:

> "Enable geometric specular AA", "This reduce specular aliasing on highly dense mesh (Particularly useful when they don't use using normal map)"

Shouldn't that be:

> "Enable geometric specular AA", "This reduce specular aliasing on highly dense mesh (Particularly useful when they don't use normal map)"

?

sebastienlagarde:
Fixed
Thanks

PR #1404, opened 2018-05-25
 Fix LWRP camera nested code

phi-lira:
- Fixes a null reference due to accesing visibleLight.light for particles lights.
 - Fixed camera nested code. If a nullreference happens m_IsCameraRendering would be true and break all scenes.


phi-lira:
Seb, you have to do this as well for HD. I've looked into it but I don't know exactly where to hook the try/catch/finally block and if you want to do that. Let me know and I can help.

PR #1405, opened 2018-05-25
Add a node to get the texture2d texel size from texture2d inputs.

JapaMala:


alindmanUnity:
Thanks for the PR! This is a great idea, and we want to get this functionality to you faster so we recreated it in a separate PR with all of our specific changelog requirements so it can go through and get to you as soon as possible. See #1527 to track the progress! :)

PR #1423, opened 2018-05-29
Fixes for Subgraph

martint-unity:
Added a cache for subgraph. This was a fix for the de-serializing of the subgraph on every frame.
Removed legacy code. Old code that was not used anymore was removed.
Added (sub) to the node name if it is a subgraph. So it is easier to see which node is a subgraph node.


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -38,3 +38,5 @@ This adds gradient functionality via two new nodes. Sample Gradient node samples
 - Fixed an issue with editable sub graph paths, causing the search window to sometimes yield a null reference exception.
 - Ensure that the blackboard is within view when deserialized.
 - Fixed float value to string error and truncation warning for normal create node.
+- Deserialization of subgraphs is no longer causing an issue.
```
Katstodian: Please replace "is no longer causing an issue." with "now works correctly."

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -38,3 +38,5 @@ This adds gradient functionality via two new nodes. Sample Gradient node samples
...
+- Added suffix (sub) to subgraphs to make it easier to distinguish them from other nodes.
```
Added suffix (sub) to subgraphs to make it easier to distinguish them from other nodes. 

With this:
Sub graphs are now suffixed with (sub), so you can tell them apart form other nodes.

PR #1435, opened 2018-05-30
Import dependencies

pbbastian:
# Problem
If a sub-shader is changed the shader graphs using it does not get re-imported. Typically this will result in shader compilation errors, as the sub-shader was most likely changed for a reason. The same goes for sub-graphs.

# Solution
Adds the ability for a node to specify a list of source asset paths that should trigger re-import when changed. Similarly, a sub-shader can provide an equivalent list during shader generation.

For now, each sub-shader depends on the templates used (these are manually added) as well as _some_ `.cs` files.

Lightweight PBR depends on:
* lightweightPBRForwardPass.template
* lightweightPBRExtraPasses.template
* LightWeightPBRSubShader.cs

Lightweight Unlit depends on:
* lightweightUnlitForwardPass.template
* lightweightUnlitExtraPasses.template
* LightWeightUnlitSubShader.cs

HD PBR depends on:
* HDPBRSubShader.cs
* HDSubShaderUtilities.cs
* Any pass used via `GenerateShaderPass`. I modified it so that it adds the template if it exists.

HD PBR depends on:
* HDUnlitSubShader.cs
* HDSubShaderUtilities.cs
* Any pass used via `GenerateShaderPass`. I modified it so that it adds the template if it exists.

Note that the `*.cs` dependencies are only on that specific file. This means that if something outside of `HDPBRSubShader.cs` changes, but `HDPBRSubShader.cs` itself doesn't, then a re-import is not triggered. I still think this level of tracking is worth it though. We could also do an explicit `VERSION` file and track that.

# Notes
Change log coming soon.

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRSubShader.cs
+++ b/com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRSubShader.cs
@@ -537,7 +537,7 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
             return defines.GetShaderString(2);
         }
 
-        private static bool GenerateShaderPass(PBRMasterNode masterNode, Pass pass, GenerationMode mode, SurfaceMaterialOptions materialOptions, ShaderGenerator result)
+        private static bool GenerateShaderPass(PBRMasterNode masterNode, Pass pass, GenerationMode mode, SurfaceMaterialOptions materialOptions, ShaderGenerator result, List<string> sourceAssetDependencyPaths)
         {
             var templateLocation = Path.Combine(Path.Combine(Path.Combine(HDEditorUtils.GetHDRenderPipelinePath(), "Editor"), "ShaderGraph"), pass.TemplateName);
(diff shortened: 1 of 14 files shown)


pbbastian:
I think this is at a stage where it's okay for now, so I'm gonna merge it in. Later on, we can look into more advanced tracking of changes.

cdxntchou:
At Bungie we had a specific file (basically a global shader change log)
that was used for global versioning of shaders based on system code changes.

Anyone making a change that wanted to invalidate the cache / reimport all
shaders would touch that file (adding a description of their change / date
/ name).

This gave you manual control of when the global version was invalidated
(as  a lot of changes don't require every shader to be rebuilt from
scratch...)
And was better than a simple version number, as merges will automatically
force a reimport


On Wed, May 30, 2018 at 7:48 AM, Peter Bay Bastian <notifications@github.com

pbbastian:
Ah yes, that makes sense, since Git would even be able to merge that nicely. E.g. two PRs incrementing a version number at the same time would not result in the version number changing two times.

pbbastian:
Note to self: Do a change log change for this one

PR #1453, opened 2018-06-01
Fix for "Show Generated Code"

martint-unity:
Fix for open generated shader when external editor not set. Also tested on Mac.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -75,3 +75,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - SpaceMaterialSlot now reads correct slot.
 - Slider node control now functions correctly.
 - Shader Graphs no longer display an error message intended for Sub Graphs when you delete properties.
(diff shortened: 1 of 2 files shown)


martint-unity:
Not working yet on MAC

zadi0648:
On Windows, works well. Works with different editors, simple graphs, complex graphs, graphs with shader groups, ect.
On OsX, there is no error like in trunk but it opens up the external editor without opening up the generated shader in the external editor. Tried this with sublime text, Visual Studio, and TextEdit.

zadi0648:
Approving. Treating the OsX case as a separate bug, here is the bug case: https://fogbugz.unity3d.com/f/cases/1079798/

pbbastian:
Needs change log change

martint-unity:
Added change log
Merged with latest master
Running Katana right now.

martint-unity:
@pbbastian 
Test are successful
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Ffix-open-generated-shader-when-external-editor-not-set&automation-tools_branch=add-platform-filter&unity_branch=trunk

PR #1458, opened 2018-06-01
Fix texture and boolean property types in subgraphs

Kink3d:
## Fixes

- Multiple texture arguments in subgraph calls
- Texture2D array support in subgraphs
- Boolean support in subgraphs

## Risks

None


pbbastian:
LGTM, but please update the change log :)

PR #1476, opened 2018-06-06
Stack lit2

slunity:


com.unity.render-pipelines.core/CoreRP/ShaderLibrary/BSDF.hlsl
```diff
@@ -416,7 +416,14 @@ real3 EvalIridescence(real eta_1, real cosTheta1, real iridescenceThickness, rea
     // Force eta_2 -> eta_1 when Dinc -> 0.0
     // real eta_2 = lerp(eta_1, eta_2, smoothstep(0.0, 0.03, Dinc));
     // Evaluate the cosTheta on the base layer (Snell law)
-    real cosTheta2 = sqrt(1.0 - Sq(eta_1 / eta_2) * (1.0 - Sq(cosTheta1)));
+    real sinTheta2 = Sq(eta_1 / eta_2) * (1.0 - Sq(cosTheta1));
+#ifdef UNITY_MATERIAL_STACKLIT
```
sebastienlagarde: it is forbidden to use UNITY_MATERIAL_STACKLIT here, you are in shader library that have no knowledge of UNITY_MATERIAL_STACKLIT

sebastienlagarde: I will say, in the context of stacak lit, maybe you can do a separate EvalIridescence function

sebastienlagarde: I will fix it after the merge

PR #1485, opened 2018-06-06
Fix HD sub-shaders declaration order

pbbastian:
The input and output structs were being declared _after_ the functions that the graph provides. This meant that if any function tried to use the struct it would fail. Sub-graphs uses the struct.


cdxntchou:
This was backported to 2018.2 in commit 9e438c42224e3a93d8e45aa28e9043002c38536a

PR #1492, opened 2018-06-07
Decals/ss3 channel mask

Paulm-Unity:
- Add support for mesh decals
- Add flag to disable projector decals on transparent geometry to save performance and decal texture atlas space
- Add ability to use decal diffuse map as mask only


com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalUtilities.hlsl
```diff
@@ -121,8 +121,13 @@ void AddDecalContribution(PositionInputs posInput, inout SurfaceData surfaceData
 
                 if((decalData.diffuseScaleBias.x > 0) && (decalData.diffuseScaleBias.y > 0))
                 {
-                    ApplyBlendDiffuse(DBuffer0, mask, sampleDiffuse, DBUFFERHTILEBIT_DIFFUSE, decalBlend, lodDiffuse);
-                    alpha = alpha < decalBlend ? decalBlend : alpha;    // use decal alpha if it is higher than transparent alpha
+					ApplyBlendDiffuse(DBuffer0, mask, sampleDiffuse, DBUFFERHTILEBIT_DIFFUSE, decalBlend, lodDiffuse);
```
sebastienlagarde: Caution about indent

PR #1497, opened 2018-06-07
Added false color debug mode

alelievr:
False color debug mode in the rendering tab of the debug window for HDRP


com.unity.render-pipelines.high-definition/HDRP/Debug/DebugColorPicker.shader
```diff
@@ -51,6 +53,23 @@ Shader "Hidden/HDRenderPipeline/DebugColorPicker"
 
                 return output;
             }
+            
+            float InverseLerp(float min, float max, float t)
```
sebastienlagarde: InverseLerp => RangeRemap, move ni common.hlsl

com.unity.render-pipelines.high-definition/HDRP/Debug/DebugColorPicker.shader
```diff
@@ -158,7 +176,14 @@ Shader "Hidden/HDRenderPipeline/DebugColorPicker"
                 // _DebugExposure will be set to zero if the debug view does not need it so we don't need to make a special case here. It's handled in only one place in C#
                 mouseResult = mouseResult / exp2(_DebugExposure);
 
-                float4 finalResult = DisplayPixelInformationAtMousePosition(input, result, mouseResult, mousePixelCoord);
+                if (_FalseColor)
```
sebastienlagarde: reverse the operation, we want to have color picker before false color

PR #1505, opened 2018-06-08
Add export of normal+Roughness during depth prepass for forward

sebastienlagarde:
- This allow to do various effect that require normal in forward: DeferredDirectional light, SSR, SSAO

Note that this solve and issue when GBuffer is not clear and we have both lit and stacklit object in the scene.


pbbastian:
Hi guys, this broke support for Shader Graph :( Can you make a fix as per Oliver's instructions below? @sebastienlagarde @JulienIgnace-Unity 

> Oliver Schnabel [11:00]
> Hi Peter, ShaderGraph in current master is broken for HD, because of that:
> > - Change prototype of BSDFData ConvertSurfaceDataToBSDFData(SurfaceData surfaceData) to BSDFData ConvertSurfaceDataToBSDFData(uint2 positionSS, SurfaceData surfaceData)
> 
> You need to change https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
Line 393 to
> >BSDFData bsdfData = ConvertSurfaceDataToBSDFData(posInput.positionSS.xy, surfaceData);
> 
> Then it works again.

sebastienlagarde:
Arf my bad... sorry. Doing ASAP
@pbbastian can you move the .template to .template.hlsl as you propose last time? Will help to catch this more easily :P .

PR #1507, opened 2018-06-08
Naive hookup for cameraData.renderScale <-> XRSettings.eyeTextureResolutionScale in XR mode

Nerites:
In XR mode, cameraData.renderScale = XRSettings.eyeTextureResolutionScale. The functionality of cameraData.renderScale is analogous to XR's eyeTextureResolutionScale: it reallocates a scale*originalTextureSize texture as the render target. 

This simple hookup does not update the SRP asset to reflect the value of eyeTextureResolutionScale. As before, adjusting renderScale in the SRP asset with XR enabled will have no effect on renderScale.

### DIFF
--- a/com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
+++ b/com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
@@ -196,9 +196,24 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
 
             // Discard variations lesser than kRenderScaleThreshold.
             // Scale is only enabled for gameview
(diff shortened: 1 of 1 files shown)


Nerites:
I totally agree with you! This was intended to be a stopgap fix to 1) indicate that there should be dependency between renderScale and eyeTextureResolutionScale, 2) implicitly highlight the need for an overall decision on how XR settings will interact with SRP settings, and 3) be a proof of concept for me in terms of getting SRP up and running for the first time. 

As this change is intended to be overwritten at some point, it's totally up to you whether to accept it in the code base pending more permanent changes.

PR #1509, opened 2018-06-09
Fix for shader include paths

keijiro:
This pull request is to fix the relative paths written in the shader pass definition files. These files emit the "file not found" error when included from external shaders (e.g. custom shader defined in a user project).

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Material/Decal/ShaderPass/DecalSharePass.hlsl
+++ b/com.unity.render-pipelines.high-definition/HDRP/Material/Decal/ShaderPass/DecalSharePass.hlsl
@@ -13,4 +13,4 @@
 #endif
 
 // This include will define the various Attributes/Varyings structure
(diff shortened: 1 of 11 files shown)


keijiro:
They previously worked because these relative paths are based at the directory that contains the `.shader` files (e.g. `../../ShaderPass/VaryingMesh.hlsl` worked as `HDRP/Material/Lit/../../ShaderPass/VaryingMesh.hlsl`), but they broke when included from an external shader file.

The new relative paths are based at the directory that contains these `.hlsl` files themselves (e.g. `../../../ShaderPass/VaryingMesh.hlsl` work as `HDRP/Material/Lit/ShaderPass/../../../ShaderPass/VaryingMesh.hlsl`). They work no matter where they're included from.

Putting that aside, I agree with you. It's better to use simple absolute paths.

sebastienlagarde:
New PR here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1511

PR #1512, opened 2018-06-11
Update window title

martint-unity:
### Purpose

When changing the file name on disk, the title bars didn't update. With this PR it will update to the correct name.

### How to Test
- Open a shader  graph.
- Rename that shader graph in your project browser.
- Name of the titles should be updated.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -77,3 +77,4 @@ You can now see the generated code for any specific node. To do so, right-click 
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
+- Renaming a shader graph file on disk, will update the title of the shader graph window.
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -77,4 +77,4 @@ You can now see the generated code for any specific node. To do so, right-click 
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
-- Renaming a shader graph file on disk, will update the title of the shader graph window.
+- When you rename a shader graph or sub shader graph locally on your disk, the title of the Shader Graph window, black board and preview also updates.
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -77,4 +77,4 @@ You can now see the generated code for any specific node. To do so, right-click 
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
-- When you rename a shader graph or sub shader graph locally on your disk, the title of the Shader Graph window, black board and preview also updates.
+- When you rename a shader graph or sub shader graph locally on your disk, the title of the Shader Graph window, black board, and preview also updates.
--- a/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
@@ -148,8 +148,6 @@ namespace UnityEditor.ShaderGraph
             set { m_PreviewData = value; }
         }
 
-        public string name { get; set; }
-
         [SerializeField]
         string m_Path;
 
--- a/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
@@ -148,6 +148,8 @@ namespace UnityEditor.ShaderGraph
             set { m_PreviewData = value; }
         }
 
+        public string name { get; set; }
+        
         [SerializeField]
         string m_Path;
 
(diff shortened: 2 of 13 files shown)


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -32,6 +32,8 @@ public class MaterialGraphEditWindow : EditorWindow
 
         ColorSpace m_ColorSpace;
 
+        static MaterialGraphEditWindow s_Instance;
```
pbbastian: This won't work if the user has multiple Shader Graph windows open.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -73,6 +75,35 @@ public string selectedGuid
             private set { m_Selected = value; }
         }
 
+        // This is so that we can check if the file on disk has been renamed.
+        // If it has, we can update the Title of the shader graph tab.
+        class ShaderPostProcessor : AssetPostprocessor
```
pbbastian: This shouldn't be nested within `MaterialGraphEditWindow`. Actually, we already have an AssetPostprocessor in ShaderGraphImporter.cs. You should use that one instead :) You're very welcome to move that one into its own file now that you're touching it anyway :p

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -73,6 +75,35 @@ public string selectedGuid
...
+        {
+            static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
+            {
+                if (s_Instance != null)
+                {
+                    string ending = ".ShaderGraph";
```
pbbastian: This should be using `ShaderGraphImporter.ShaderGraphExtension` instead so that we keep that in one place

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -73,6 +75,35 @@ public string selectedGuid
...
+                    bool anyShaders = importedAssets.Any(val => val.EndsWith(ending));
+                    anyShaders |= movedAssets.Any(val => val.EndsWith(ending));
+                    if (anyShaders)
+                        s_Instance.UpdateAfterAssetChange();
```
pbbastian: You need to check whether the window actually has that asset open. Otherwise renaming _any_ Shader Graph asset will cause the window to change its title.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -77,3 +77,4 @@ You can now see the generated code for any specific node. To do so, right-click
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
+- Renaming a shader graph file on disk, will update the title of the shader graph window.
```
Renaming a shader graph file on disk, will update the title of the shader graph window.

With this:
When you rename a shader graph locally on your disk, the title of the Shader Graph window also updates.

zadi0648:
- Due to what Peter pointed out, title update frequently fails when multiple shader graphs are open at once. Furthermore once this happens the shader graph title will fail to ever update even if it becomes the current focus and the title is changed again.
- The name of the shader at the top of the blackboard does not update. Since both should always match, could we do that in this PR?

zadi0648:
Blackboard & master node updating looking good, as well as working well with multiple graphs open at once. However...
- Renaming & saving the "shader group" (the thing right under the Blackboard Title that defines a shader group, e.g. the Lit in Lit/LitLava) no longer updates (it did in the previous round of testing just fyi)
- Changing the name of the shader completely removes the shader group from title (e.g. if LitLava becomes MoltenLava, title goes from Lit/LitLava to MoltenLava instead of Lit/MotlenLava)

com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
```diff
@@ -149,7 +149,7 @@ public InspectorPreviewData previewData
         }
 
         public string name { get; set; }
-
+
```
pbbastian: remove this change

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -77,3 +77,4 @@ You can now see the generated code for any specific node. To do so, right-click
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
+- When you rename a shader graph or sub shader graph locally on your disk, the title of the Shader Graph window, black board and preview also updates.
```
pbbastian: @Katstodian What do you think about this?
> When you rename a shader graph locally on your disk, the title used in the Shader Graph window also updates.

com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
```diff
@@ -57,7 +57,17 @@ PreviewManager previewManager
             set { m_PreviewManager = value; }
         }
 
-        public GraphEditorView(EditorWindow editorWindow, AbstractMaterialGraph graph, string assetName)
+        public string assetName
+        {
+            get { return m_BlackboardProvider.blackboard.title; }
+            set
+            {
+                m_BlackboardProvider.blackboard.title = value.Split('/').Last();
```
pbbastian: hmm, shouldn't this be using the `assetName` property you just created on `BlackboardProvider`? also, it's not so good that the setter changes the value, because then you won't be getting back the same value you passed in

com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
pbbastian: @martint-unity you forgot about this one :p

PR #1527, opened 2018-06-12
Added Texel Size Node

alindmanUnity:
texel_size_node
### Purpose
This adds a new node to get the special texture properties of a Texture 2D Asset, from the `{texturename}_TexelSize` variable. This nodes takes in input of a Texture2D asset and outputs the width and height of the texel size in Vector1 format. It is important to note that using the default input to reference your texture asset will make your graph less performant, and it is recommended to always connect this node a separate Texture 2D Asset node (per image example).
### Changes

- Adds `Texel Size` node to node library.

- Changes the `TextureShaderProperty` file to also declare a variable for `{texture}_TexelSize`.
### Risks

- Changes to `TextureShaderProperty` declarations.


JapaMala:
Since the original variable exposes four outputs (width, height, 1/width, and 1/height) is there a reason why the new node only exposes width and height?

alindmanUnity:
@JapaMala We've established a convention that we're following with other nodes (such as `Screen`, which gives you the screen parameters) to not include the default 1/ outputs from Unity's built-in variables, for readability and user experience in the graph. We have the `Divide` and `Reciprocal` nodes which will get you those outputs and more clearly show on the graph view what calculations are happening. :)

JapaMala:
Okay, that makes sense.

I assume there's similar reason why there's no combined output given (a float 2)

PR #1532, opened 2018-06-14
New template parser

cdxntchou:
Created new TemplatePreprocessor class, moved all of the parser code into it, broke it up a bit so it's easier to read, added $include and $splice commands.

Name of new includes, and exactly what they contain is up for debate...   there's a bunch more shared code in the templates that we can move to shared files if we want.

Next up going to see if I can do $buildType(...) inline

EDIT: Added fix for FB case 1064118 as it would conflict with this PR otherwise


pbbastian:
Before we start reviewing this, can you rebase it on top of master so that we can avoid these conflicts? :) Just force push into this branch

cdxntchou:
Merged master in, and added $buildType

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -382,168 +378,395 @@ private static int SkipWhitespace(string str, int start, int end)
...
+            Dictionary<string, string> namedFragments;
+            string templatePath;
+            bool debugOutput;
+            string buildTypeAssemblyNameFormat;
+
+            // intermediates
+            HashSet<string> includedFiles;
+
+            // outputs
+            System.Text.StringBuilder result;
```
pbbastian: Can we use `ShaderStringBuilder` here? I'd like to expand the node tracking to master nodes at some point, so that we can do more precise error tracking. I'm open to any kinds of changes to `ShaderStringBuilder` that you feel like :)

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -382,168 +378,395 @@ private static int SkipWhitespace(string str, int start, int end)
...
+
+            public TemplatePreprocessor(HashSet<string> activeFields, Dictionary<string, string> namedFragments, bool debugOutput, string templatePath, List<string> sourceAssetDependencyPaths, string buildTypeAssemblyNameFormat, System.Text.StringBuilder outShaderCodeResult = null)
+            {
+                this.activeFields = activeFields;
+                this.namedFragments = namedFragments;
+                this.debugOutput = debugOutput;
+                this.templatePath = templatePath;
+                this.sourceAssetDependencyPaths = sourceAssetDependencyPaths;
+                this.buildTypeAssemblyNameFormat = buildTypeAssemblyNameFormat;
+                this.result = outShaderCodeResult;
```
pbbastian: replace this and the following `if` with:
```csharp
this.result = outShaderCodeResult ?? new System.Text.StringBuilder();
```

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -382,168 +378,395 @@ private static int SkipWhitespace(string str, int start, int end)
...
+            private bool Expect(string line, int location, char expected)
+            {
+                if ((location < line.Length) && (line[location] == expected))
+                {
+                    return true;
+                }
+                Error("Expected '" + expected + "'", line, location);
+                return false;
+            }
+            private void Error(string error, string line, int location)
```
pbbastian: Hmm, does this mean that if there are errors we just let the rest of the compilation process carry on?

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
cdxntchou: yay C#

pbbastian: I think it's nice syntax for a fairly common pattern :)

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
cdxntchou: Yes; it will skip the rest of the line, and continue trying to splice the rest of the template.  Almost always results in a compilation error of some kind.   We could force it to fail if we had a good way to message the failure & point back at the generated code?

pbbastian: It would be great if we could collect all the errors, and then force a failure :) Until we get a better error handling infrastructure, it might be good enough to just return a pink shader, and then log an error to the console.

cdxntchou:
Peter says we don't need the version bump anymore.
Merged latest master in.   Ready to go when approved!

pbbastian:
LGTM :) Make sure to "Squash and merge"

pbbastian:
oh no, looks like the Post-Processing submodule snuck in again :(

cdxntchou:
yeah, sorry.  git is confusing.

PR #1535, opened 2018-06-14
Warn about lack of compatible render pipeline asset

pbbastian:
This PR adds the ability for a sub-shader to express whether it is compatible with a given RenderPipelineAsset, and makes use of that for two things:
1. Master nodes now display an error badge if the current render pipeline is not compatible. We've been getting quite a few bugs about SG just displaying pink if no compatible RP is enabled. This change makes it a lot more clear to the user what's going on.
2. The preview shader compiles faster now by only considering the current render pipeline. I noticed that after we added the HD sub-shaders, compilation time for the preview shader got a lot slower. So the MasterNode abstract class now only asks a sub-shader to generate code if it's compatible with the current render pipeline. This also required me to add in a check to the window such that the UI is reloaded when the render pipeline changes.

Things I tested:
* Make a project and add Shader Graph as a package dependency, but not LWRP nor HDRP. Create a Shader Graph, open it and observe that an error badge is there. Hover the badge and observe the error message.
* Try it out with LWRP and HDRP, and verify that the previews are still working, and that the imported shader is also working.


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -77,3 +77,5 @@ You can now see the generated code for any specific node. To do so, right-click
 - Deserialization of subgraphs now works correctly.
 - Sub graphs are now suffixed with (sub), so you can tell them apart from other nodes.
 - The preview of a node does not obstruct the selection outliner anymore.
+- Master nodes now display an error badge if the current render pipeline is not compatible.
```
Master nodes now display an error badge if the current render pipeline is not compatible.

With this:
If the current render pipeline is not compatible, master nodes display an error badge. 

Reason:
Reasoning/intention/condition should be front-loaded. If readers' eyes glaze over, after "master nodes now do the thing", they might not clock onto the condition and think it's a permanent thing. :)

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -77,3 +77,5 @@ You can now see the generated code for any specific node. To do so, right-click
...
+- The preview shader compiles faster now by only considering the current render pipeline.
```
The preview shader compiles faster now by only considering the current render pipeline.

With this:
The preview shader only considers the current render pipeline. This makes the preview shader compile faster.

Reason:
Our styleguide says to avoid "ing-endings".

com.unity.shadergraph/CHANGELOG.md
pbbastian: Good point. How about:
> If the current render pipeline is not compatible, master nodes *now* display an error badge.
Maybe I'm wrong here, but otherwise it reads more like a fact than a change.

Katstodian: I agree.

pbbastian:
Thanks :D Merging it in then :) (assuming that @Katstodian's changes means that she approves)

Katstodian:
Yes to me having approved. :)

PR #1545, opened 2018-06-15
Lw/shadow optimization v2

phi-lira:
- Optimized shadow rendering for Mali Utgard arch by removing indexing and division in orthographic projections.
 - Removed 7x7 tent filtering. This was only happening when cascades were enabled and we resolve cascades in screenspace that later is sampled with bilinear filter.

### DIFF
--- a/com.unity.render-pipelines.lightweight/CHANGELOG.md
+++ b/com.unity.render-pipelines.lightweight/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - pipeline now uses XRSEttings.eyeTextureResolutionScale as renderScale when in XR 
 
 ### Changed
-- Reduced by 25% frame time in Mali Utgard arch in Overdraw benchmark scene by removing indexing in shadow code and avoiding division for orthographic projection.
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.lightweight/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - pipeline now uses XRSEttings.eyeTextureResolutionScale as renderScale when in XR 
 
 ### Changed
+- Optimized shadow rendering in Mali Utgard arch by removing indexing and avoid division on orthographic projections. Frame time reduce 25% on Overdraw benchmark.
```
pbbastian: Might wanna add kat in for more precise corrections here, but anyway here's my suggestion:
Shadow rendering has been optimized for the Mali Utgard architecture by removing indexing and avoiding divisions for orthographic projections. This reduces the frame time by 25% on the Overdraw benchmark.

phi-lira: I'll add Kat on the PRs starting monday.

phi-lira: For now I took your suggestion.

PR #1548, opened 2018-06-15
Fixed compilation errors for Switch platform.

kaychang-unity:
Minor compilation fix for Switch.


kaychang-unity:
XRSetting class has undefined members of Switch. The issue is known by the XR team and they are working on a solution to allow us to not constantly add guards to XRSetting related code.
Master merged, comments added in CHANGELOG.md.

PR #1550, opened 2018-06-15
Added Normal Derive Nodes

alindmanUnity:
normal_derive_nodes
### Purpose
There are two Normal Derive Nodes: `Normal From Height` and `Normal Reconstruct Z`.
`Normal From Height` uses Vector1 input to derive a normal map. `Normal Reconstruct Z` uses the X and Y components in Vector2 input to derive the proper Z value for a normal map.  
This also renames the `Normal Create` node to `Normal From Texture` for clarity.

### Changes
- Adds `Normal From Height` node.
- Adds `Normal Reconstruct Z` node.
- Renames `Normal Create` node to `Normal From Texture`.

### Risks
- Changing name of `Normal Create` node.


sebastienlagarde:
Hey, small suggestion:
 "Normal from value"  => "Heightmap to Normal"  or "Normal from HeightMap"
"Normal DeriveZ" => "Normal Reconstruct Z"

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -46,6 +46,12 @@ Note: This functionality requires that you have enabled **two sided** on the Mas
 
 This adds gradient functionality via two new nodes. The Sample Gradient node samples a gradient given a Time parameter. You can define this gradient on the Gradient slot control view. The Gradient Asset node defines a gradient that can be sampled by multiple Sample Gradient nodes using different Time parameters.
 
+## Normal Derive Nodes
+
+![](.data/normal_derive_nodes.png)
+
+This adds two new nodes for deriving normals. `Normal From Value` will take in Vector1 input and derive a normal map from that input. This normal map can be output in tangent space or world space based on the drop down menu.  `Normal Derive Z` will derive the proper Z value for a normal map given the X and Y components in a Vector2 input.
```
This adds two new nodes for deriving normals. `Normal From Value` will take in Vector1 input and derive a normal map from that input. This normal map can be output in tangent space or world space based on the drop down menu.  `Normal Derive Z` will derive the proper Z value for a normal map given the X and Y components in a Vector2 input.

With this:
There are two Normal Derive Nodes: Normal From Value and Normal Derive Z. 
`Normal From Value` uses Vector1 input to derive a normal map. `Normal Derive Z` uses the X and Y components in Vector2 input to derive the proper Z value for a normal map.

Reason:
- No-one likes Will. He's an arse! (The modality of **will** is hard to translate to other languages, and doesn't always make sense to non-native English speakers. Some Asian speakers will interpret it as "isn't sure to happen", and a lot of Scandinavians might interpret it as a declaration of intention instead of "this is going to happen". So it's best to just re-write to active verbs: It does the thing.
- I front-loaded the information about X and Y components to adhere to the Will-thing, but also to make the last sentence more "It does the thing". It's always good to front-load any referential information (like which components are used), because users can trail off before they hit the full stop.
- We have to try to limit time-specific language as much as possible. So I omitted "new". This isn't set in stone, so can be changed if you (or Peter, I see you reading my comments!) thinks so ;)

alindmanUnity: changes made! All those make sense - I also renamed the nodes so make sure everything still looks okay given the new names?

PR #1558, opened 2018-06-18
Fix apply Range Attenuation

sebastienlagarde:
The issue still remain for area light


EvgeniiG:
This PR breaks area lights. Please add me as a reviewer next time.
Let's discuss what you are trying to accomplish here and whether that makes sense.
FogBugz: https://fogbugz.unity3d.com/f/cases/1075168/

sebastienlagarde:
This was on purpose. The graphic test have been updated with this in mind and QA was notify.
We will discuss about it.

EvgeniiG:
Remy filed a FogBugz entry: https://fogbugz.unity3d.com/f/cases/1075168/

PR #1561, opened 2018-06-19
Shadowcascadegui class made internal

martint-unity:
Made shadowcascadegui internal, since it is internal in Trunk, so that we do not have to support it when leaving experimental.

### DIFF
--- a/com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightEditorUtils.cs
+++ b/com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightEditorUtils.cs
@@ -1,0 +1,48 @@
+using System;
+using UnityEngine;
+
+namespace UnityEditor.Experimental.Rendering
+{
+    internal static class LightweightEditorUtils
+    {
+        internal static void DrawCascadeSplitGUI<T>(ref SerializedProperty shadowCascadeSplit)
+        {
+            float[] cascadePartitionSizes = null;
+
+            System.Type type = typeof(T);
+            if (type == typeof(float))
+            {
+                cascadePartitionSizes = new float[] { shadowCascadeSplit.floatValue };
+            }
+            else if (type == typeof(Vector3))
+            {
+                Vector3 splits = shadowCascadeSplit.vector3Value;
+                cascadePartitionSizes = new float[]
+                {
+                    Mathf.Clamp(splits[0], 0.0f, 1.0f),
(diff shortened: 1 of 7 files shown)


com.unity.render-pipelines.lightweight/LWRP/Editor/Shadow/ShadowCascadeSplitGUI.cs
```diff
@@ -1,15 +1,10 @@
...
 using System.Linq;
-using System.Linq.Expressions;
-using System.Reflection;
-using System.Text;
 using UnityEngine;
 
 namespace UnityEditor.Experimental.Rendering
 {
-    static class ShadowCascadeSplitGUI
+    internal static class ShadowCascadeSplitGUI
```
pbbastian: Redundant since `internal` is the default modifier for classes. Style guide says redundant modifiers should be avoided.

com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightEditorUtils.cs
```diff
@@ -0,0 +1,48 @@
+using System;
+using UnityEngine;
+
+namespace UnityEditor.Experimental.Rendering
+{
+    internal static class LightweightEditorUtils
```
pbbastian: Redundant since internal is the default modifier for classes. Style guide says redundant modifiers should be avoided.

com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightEditorUtils.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        internal static void DrawCascadeSplitGUI<T>(ref SerializedProperty shadowCascadeSplit)
```
pbbastian: I think it's good practice to keep the members of an internal class as public instead of internal (unless there's a really good reason), as the class modifier will overwrite anyway, and then it's specified only in 1 place.

pbbastian:
@martint-unity @phi-lira Is this PR still relevant?

phi-lira:
Is shadow Cascade GUI exposed in Unity now?

phi-lira:
If it is then we just need to remove GUI from core.

martint-unity:
I don't think this widget is public. So we should probably make this internal.

phi-lira:
The following PR makes this one obsolete: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2332

PR #1562, opened 2018-06-19
Port PCSS shadows

alelievr:


com.unity.render-pipelines.core/CoreRP/ShaderLibrary/Shadow/PCSS.hlsl
```diff
@@ -0,0 +1,176 @@
...
+        real U = coord.x + offset.x;
+        real V = coord.y + offset.y;
+
+        //NOTE: We must clamp the sampling within the bounds of the shadow atlas.
+        //        Overfiltering will leak results from other shadow lights.
+        //TODO: Investigate moving this to blocker search.
+        if (U <= UMin || U >= UMax || V <= VMin || V >= VMax)
+            Sum += SAMPLE_TEXTURE2D_ARRAY_SHADOW(shadowMap, compSampler, real3(coord.xy, coord.z), slice);
+        else
+            Sum += SAMPLE_TEXTURE2D_ARRAY_SHADOW(shadowMap, compSampler, real3(U, V, coord.z), slice);
```
uygar3d: Needs to support the engine provided sampleBias here on the coord.z via 
`coord.z + dot( sampleBias, offset );`

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/Shadow/PCSS.hlsl
```diff
@@ -0,0 +1,176 @@
...
+        real U = coord.x + offset.x;
+        real V = coord.y + offset.y;
+
+        //NOTE: We must clamp the sampling within the bounds of the shadow atlas.
+        //        Overfiltering will leak results from other shadow lights.
+        //TODO: Investigate moving this to blocker search.
+        if (U <= UMin || U >= UMax || V <= VMin || V >= VMax)
+            Sum += SampleCompShadow_T2DA(shadowContext, texIdx, sampIdx, real3(coord.xy, coord.z), slice);
+        else
+            Sum += SampleCompShadow_T2DA(shadowContext, texIdx, sampIdx, real3(U, V, coord.z), slice);
```
uygar3d: Same here with the sample bias

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/Shadow/ShadowSampling.hlsl
```diff
@@ -526,6 +526,53 @@ real SampleShadow_MSM_1tap( ShadowContext shadowContext, inout uint payloadOffse
...
+
+    //2) Penumbra Estimation
+    real filterSize = shadowSoftnesss * PenumbraSize(tcs.z, averageBlockerDepth);
+    filterSize = max(filterSize, 0.000001);
+
+    //3) Filter
+    return PCSS(tcs, filterSize, scaleOffset, slice, sampleBias, shadowContext, texIdx, sampIdx, sampleCount);
+}
+
+real SampleShadow_PCSS( ShadowContext shadowContext, inout uint payloadOffset, real3 tcs, real4 scaleOffset, float slice, Texture2DArray tex, SamplerComparisonState compSamp, SamplerState samp )
```
uygar3d: Should receive the provided sampleBias as an input parameter

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/Shadow/ShadowSampling.hlsl
```diff
@@ -526,6 +526,53 @@ real SampleShadow_MSM_1tap( ShadowContext shadowContext, inout uint payloadOffse
...
+{
+    real2 params           = asfloat(shadowContext.payloads[payloadOffset].xy);
+    real shadowSoftnesss   = params.x;
+    int sampleCount        = params.y;
+    payloadOffset++;
+
+    real2 sampleBias = real2(sin(GenerateHashedRandomFloat(asuint(tcs.x))),
+                             cos(GenerateHashedRandomFloat(asuint(tcs.y))));
```
uygar3d: sampleBias already has a specific meaning in the system. Please choose a different name.

com.unity.render-pipelines.core/CoreRP/Shadow/Shadow.cs
```diff
@@ -168,6 +171,7 @@ override protected void Register(GPUShadowType type, ShadowRegistry registry)
             m_SupportedAlgorithms.AddUniqueUnchecked((int)ShadowUtils.Pack(ShadowAlgorithm.PCF, ShadowVariant.V2, precision));
             m_SupportedAlgorithms.AddUniqueUnchecked((int)ShadowUtils.Pack(ShadowAlgorithm.PCF, ShadowVariant.V3, precision));
             m_SupportedAlgorithms.AddUniqueUnchecked((int)ShadowUtils.Pack(ShadowAlgorithm.PCF, ShadowVariant.V4, precision));
+            m_SupportedAlgorithms.AddUniqueUnchecked((int)ShadowUtils.Pack(ShadowAlgorithm.PCSS, ShadowVariant.V0, precision)); //TODO: Consider the highest precision for blocker search.
```
uygar3d: Precision in this context refers more to the bit depth of the underlying render target, e.g. 16 bits vs. 32 bits and is being determined by whoever is creating the shadowmap.

com.unity.render-pipelines.core/CoreRP/Shadow/Shadow.cs
```diff
@@ -512,19 +559,28 @@ virtual protected void WritePerLightPayload(List<VisibleLight> lights, ShadowReq
...
+                        {
+                            sp.Set(shadowData[0] | (SystemInfo.usesReversedZBuffer ? 1 : 0), shadowData[1], 0, 0);
+                            payload[payloadOffset] = sp;
+                            payloadOffset++;
+                        }
+                        break;
                     }
-                    break;
+                }
+                else // PCSS
```
uygar3d: could have added it via else if( PCSS ) to make it more explicit and put an assert in the else branch

com.unity.render-pipelines.core/CoreRP/Shadow/Shadow.cs
```diff
@@ -165,7 +165,7 @@ virtual protected void CreateShadowmap(RenderTexture shadowmap)
         override protected void Register(GPUShadowType type, ShadowRegistry registry)
         {
             ShadowPrecision precision = m_ShadowmapBits == 32 ? ShadowPrecision.High : ShadowPrecision.Low;
-            m_SupportedAlgorithms.Reserve(5);
+            m_SupportedAlgorithms.Reserve(6);
```
sebastienlagarde: Can you add a comment here to say where this number come from? or can we use something like introspection on an enum to know how many alogrithm we have?

sebastienlagarde: i.e Enum.GetNames(typeof(ShadowAlgorithm)).Length;

uygar3d: It's not based on the enum size, though, as the class only supports an arbitrary subset. The reserve is only here so the array doesn't need to reallocate when adding the members below, that's all.

sebastienlagarde: ah ok

PR #1566, opened 2018-06-19
Fix missing matrices for Shader Graph in HDRP

Kink3d:
Adds the missing matrices required by Shader Graph for `Transformation Matrix` node. Zero risk as these are defined as just transpose or multiplication of previously defined matrices.

Fixes HUP bug case (Fogbugz case 1046087)


Kink3d:
@sebastienlagarde This is fine but we need to support these in Shader Graph. If you don't want to support MV in HD I can redefine it as identity matrix if you'd prefer, but it must be defined as something.

sebastienlagarde:
As explain, we can define them as a function:

float3 positionWS = TransformObjectToWorld(input.positionOS);
positionWS = GetCameraRelativePositionWS(positionWS);
output.positionVS = TransformWorldToView(positionWS);

Will work. I can handle this part if you want, so just put identity for now, and I will update it after hackweek.

Note: Current code for LW will not work with GPU instancing. in UnityInstancing.hlsl we don't have the macro for it (i.e this part)

"
        #undef UNITY_MATRIX_MVP
        #undef UNITY_MATRIX_MV
        #undef UNITY_MATRIX_T_MV
        #undef UNITY_MATRIX_IT_MV
        #define UNITY_MATRIX_MVP    unity_MatrixMVP_Instanced
        #define UNITY_MATRIX_MV     unity_MatrixMV_Instanced
        #define UNITY_MATRIX_T_MV   unity_MatrixTMV_Instanced
        #define UNITY_MATRIX_IT_MV  unity_MatrixITMV_Instanced
"

We have remove them because the View matrix must be setup by the render pipeline. It is the case for HD, and LW is suppose to do it at some point. As View matrix is control by the render pipeline, there is no way to provide unity_MatrixMV_Instanced from C++ side.

So to sum up current state:
- instancing don't work with shadergraph for UNITY_MATRIX_MVP /  UNITY_MATRIX_MV  macro
- HD don't support them

What could be done:
- define the macro with camera relative support for HD, it will be compatible with instancing
- redefine the instancing macro for LW to make it compatible with instancing

What should be done:
- define the macro with camera relative support for HD, it will be compatible with instancing
- Have LW setup camera matrix from C#
- define the macro for LW in two part: first objToWorld, then WorldToView to support instancing correctly

So short term action: 
- Add shader graph test with instancing :) with node that use M and V separated and MV together
- Define HD macro to identity
- Define UNITY_MATRIX_MV instancing macro on LW side

(Poking @phi-lira for awareness)

sebastienlagarde:
Now that this PR is in https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1587 
it is simpler to add MV macro if you want them.

Kink3d:
Replaced by #1652

PR #1568, opened 2018-06-19
Fix dielectric specular control

Kink3d:
Dielectric Specular node was resetting on node controls causing values to appear to not be serialized. This PR fixes all instances of this.

Fix for Fogbugz case 1043591


sebastienlagarde:
Sorry to hijack the PR with a comment but I noticed this:

m_IORSlider = new Slider(1, 5, changedIORSlider);

An IOR for dieletric should be in the range 1 - 2.5, not 1 - 5.

Look at this page: 
https://pixelandpoly.com/ior.html

Most material are 1.3 - 1.6
Gemstone are 1.8-2.5
Diamond is 2.5 and this is totally fine for game (there is one material that nobody know that is 2.6 but we don't care).

Let's avoid user to provide unrealistic value for such a node. If users wan to do crazy thing, they can provide spec color directly, if they use IOR, it is to get plausible value.

Kink3d:
Thanks Seb, Ill fix that. Previous chart I was looking at had values up to 5. Cant find it now though so cant say how. I'll assume you are correct though :D

PR #1569, opened 2018-06-19
Fix TextureSlotControl not accepting Render Textures

Kink3d:
TextureSlotControlView was casting input textures to type `Texture2D` which caused RTs to fail. This PR fixes this to make this input field behave the same as the ones on properties and `Texture2DAsset` node.

Fixed fogbugz case 1025884


com.unity.shadergraph/Editor/Drawing/Views/Slots/TextureSlotControlView.cs
```diff
@@ -22,7 +22,7 @@ public TextureSlotControlView(Texture2DInputMaterialSlot slot)
 
         void OnValueChanged(ChangeEvent<Object> evt)
         {
-            var texture = evt.newValue as Texture2D;
+            var texture = evt.newValue as Texture;
```
pbbastian: IMO it shouldn't be possible to select a RenderTexture if it's not valid in the given context, so the change should instead be:
```diff
-var objectField = new ObjectField { objectType = typeof(Texture2D), value = m_Slot.texture };
+var objectField = new ObjectField { objectType = typeof(Texture), value = m_Slot.texture };
```
Such that the object picker will only show Texture2D objects. Based on @zadi0648's comment, I'm guessing that a default slot can never use a RenderTexture?

Kink3d:
@zadi0648 This works fine for me when embedding a RT. Can you check this again?

zadi0648:
Weird, it does appear to work for me now so both texture 2d node and inline. I'm testing with my Mac atm (in the air onw to hack week) whereas before I was on windows, so that might be making a difference?

spaztronaut:
Tested on Windows

PR #1570, opened 2018-06-20
XRGConfig: New interface to XRSettings for SRP

nerites:
Most notable change in this PR is a rename of renderScale to TargetScale, to differentiate it from ViewportScale (which will be added in XRSettings integration). -> Reverted in most recent commit

I also fixed the naive renderScale <-> eyeTextureResolutionScale hookup I did in [this PR](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1507), which caused renderScale^2 to be applied when XR was enabled. 

Finally this PR has a little bit of plumbing for adding an XR section to the pipeline asset inspector. 

If people would rather I hold off on PRs related to XRSettings until I get to ramp up on SRP in 2 weeks, that's fine with me, but I thought this PR would be harmless enough and might prevent some conflicts that may arise in the intervening time.


com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightPipelineAssetEditor.cs
```diff
@@ -229,5 +247,15 @@ void DrawCapabilitiesSettings()
...
 
         }
+
+        void DrawXRSettings()
+        {
+            EditorGUILayout.LabelField(Styles.xrSettingsLabel, EditorStyles.boldLabel);
+            EditorGUILayout.PropertyField(m_XREnabledProp, Styles.xrEnabled);
+            EditorGUI.BeginDisabledGroup(m_ShowXREnabled.value);
+            EditorGUI.indentLevel++;
+            EditorGUI.indentLevel--;
```
daves-unity: The above two lines look like the cancel each other out.

com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
```diff
@@ -348,8 +348,8 @@ void SetupPerFrameShaderConstants()
 
         void SetupPerCameraShaderConstants(CameraData cameraData)
         {
-            float cameraWidth = (float)cameraData.camera.pixelWidth * cameraData.renderScale;
-            float cameraHeight = (float)cameraData.camera.pixelWidth * cameraData.renderScale;
+            float cameraWidth = (float)cameraData.camera.pixelWidth * cameraData.targetScale;
+            float cameraHeight = (float)cameraData.camera.pixelWidth * cameraData.targetScale;
```
daves-unity: Shouldn't the `cameraHeight` be based on the `cameraData.camera.pixelHeight`, and not on its width?

nerites: Oh yeah! Funnily enough I noticed this typo before but didn't get the fix in this changeset. Thanks for catching it!

com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightPipelineAssetEditor.cs
nerites: Yeah this section is a WIP; I'll be adding the stuff inside the indented region as I add more XRSettings. The call for this is commented out so it won't appear anyways.

If we're of a policy that we only merge in code that's being used, then I can comment this out as well.

daves-unity: Personally, I'm always in for the cleanest code possible, so I'd advocate removing lines 256 and 257, but they won't hold up approving the changes.

nerites:
Added XRGConfig struct which should be the SRP's interface to XRSettings from here on out. 
Updated LWRP to use XRGConfig in place of XRSettings and UNITY_SWITCH.

nerites:
I think this branch is good to go. The only thing outstanding is that segments of the LWRP's Inspector GUI that depend on XRGConfig.Enabled won't update with a new Enabled value unless in play mode.

com.unity.render-pipelines.core/CoreRP/XR/XRGConfig.cs
```diff
@@ -0,0 +1,144 @@
...
+        public void SetConfig()
+        { // If XR is enabled, sets XRSettings from our saved config
+            if (!Enabled)
+                return;
+            XRSettings.eyeTextureResolutionScale = renderScale;
+            XRSettings.renderViewportScale = viewportScale;
+            XRSettings.useOcclusionMesh = useOcclusionMesh;
+            XRSettings.occlusionMaskScale = occlusionMaskScale;
+            XRSettings.showDeviceView = showDeviceView;
+            XRSettings.gameViewRenderMode = gameViewRenderMode;
```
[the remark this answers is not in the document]
nerites: it returns before setting XRSettings if s_Enabled isn't true (ENABLE_VR is in s_Enabled)

PR #1571, opened 2018-06-20
Fix View Direction in Shader Graph with LWRP

Kink3d:
Simple removes the normalization in VS as we currently assume `SHADER_HINT_NICE_QUALITY`.

This fixes fogbugz case 1041684


zadi0648:
LW Normals are still a bit funky atm

Kink3d:
Fixed this PR

phi-lira:
If I got this right it should be just ok to define `#define SHADER_HINT_NICE_QUALITY 1` in the template for now. You should not normalize the tangent/bittangent. only the normal after the tangent space transform.

phi-lira:
I'd be interested in being the PRs related to LW and ShaderGraph. :D

PR #1573, opened 2018-06-20
Fix a couple of issues with density volume texture mask.

Wadarass:
Fixed bug when using more than 1 volume mask texture.


sebastienlagarde:
Can you do a PR for Master too? Thanks

sebastienlagarde:
+ update the changelog (caution: changelog of master an 2018.2 are different now, so need a different commit for each)

Wadarass:
Changelog udpated @sebastienlagarde

PR #1579, opened 2018-06-20
Fix initialization of vector properties when duplicating subgraphs

Kink3d:
Array of labels was undefined for Vector 1, 2 and 3 type input slots when using the constructor with no arguments. This caused nullrrefs when copy-pasting or duplicating sub-graph nodes with inputs of these types. 

Fixes fogbugz case: 1051391


com.unity.shadergraph/Editor/Data/Graphs/Vector1MaterialSlot.cs
```diff
@@ -19,7 +19,9 @@ public class Vector1MaterialSlot : MaterialSlot, IMaterialSlotHasValue<float>
         string[] m_Labels;
 
         public Vector1MaterialSlot()
-        {}
+        {
+            m_Labels = new[] { "X" };
```
pbbastian: This change will make the slot lose its labels. Instead, the `m_Labels` field should be marked `[SerializeField]`.

Kink3d:
@pbbastian Good call, fixed

PR #1585, opened 2018-06-21
Emissive mesh for area lights

alelievr:


sebastienlagarde:
Note: The color don't match when enabling the color temperature. This may be fix with your other PR that will handle color temperature with SetIntensity

sebastienlagarde:
@alelievr : I fixed light intensity mistmatch with emissive area light mesh + do some renaming. I will merge that. Please test that player work correctly with this and fix the issue with color temperature (+ light intensity unit change) in your PR on light unit.

PR #1587, opened 2018-06-21
HDRP: Improve camera relative rendering precision on object

sebastienlagarde:
Adding bunch of people like Juf, Raymond, Paul and Thomas for awareness of this change!
Juf, VFX editor template will require an update

This PR improve camera relative rendering precision on object.

We now apply camera translation on the model matrix, so before the TransformObjectToWorld(). This allow to save precision (as the object don't go back and forth to a far away position).

The PR add two functions to deal with it:
ApplyCameraTranslationToMatrix and ApplyCameraTranslationToInverseMatrix
in shaderVariable.hlsl. Also the PR rename positionWS to positionRWS for easier read.

Guideline are as follow:

- unity_WorldToObject and unity_ObjectToWorld must never be used directly!
- Use TransformObjectToWorld in vertex shader, it return the camera relative position (if enabled)
- The PR override the macro UNITY_MATRIX_M  / UNITY_MATRIX_I_M
- When a user want absolute world position it must use  GetAbsolutePositionWS(positionRWS)
- GetAbsolutePositionWS() need to be use for world space triplanar, decal and other case like this.
- The PR update the .templare and interpolator structure for shader graph

This is compatible with instancing.
For instancing it is require to define a macro MODIFY_MATRIX_FOR_CAMERA_RELATIVE_RENDERING so the instancing use our functions ApplyCameraTranslationToMatrix and ApplyCameraTranslationToInverseMatrix

Also PR Add a function GetObjectAbsolutePositionWS() in absolute position.

PR have been tested on Fontainbleau + on player PC. The shader graph have been tested manually with various programmer authored asset.

BTW: This should help with the other PR that try to add MV matrix macro, now it should be easier to implement: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1566


cdxntchou:
Yeah, seems like that line should call GetAbsolutePositionWS()
instead, if this behavior is on a toggle.



as an aside -- can we make a new name for the 'camera-relative world
space position'?

I like to have static definitions for things like 'world space'
whenever possible.  :)

cdxntchou:
To be clear -- both instances should use GetAbsolutePositionWS().

The  output.WorldSpacePosition values are meant to have absolute world
space (for things like triplanar, etc.)

sebastienlagarde:
>Yeah, seems like that line should call GetAbsolutePositionWS()
Which line do you refer to ?

>as an aside -- can we make a new name for the 'camera-relative world
>space position'?
Can you be more precise? you mean instead of having positionWS, you want something like positionCRWS (Camera Relative WS) or relativePositionWS everywhere ? (I am a bit afraid to do a big renaming of all the place where we use positionWS. As for example we need to rename all interpolator, all name in FragInputs, in Tessellation interpolator, in displacement function etc...). But yeah I understand that it may be obscur. Do you have a suggestion ?

>I like to have static definitions for things like 'world space'
What do you mean? That when we use "world space" it really mean world space ? :P . It is the case if you disable camera relative rendering, this is what is ambiguous. 

I am agree to try to rename stuff, just it appear I will need to touch ALL our shaders file + .template and I am afraid by such operation.

sebastienlagarde:
>The  output.WorldSpacePosition values are meant to have absolute world
>space (for things like triplanar, etc.)

If it is use only for Triplanar or mapping stuff it is fine. If it is use for displacement then the displacement should operate in camera relative space for better precision, if WorldSpacePosition is then sent to pixel shader for lighting, then it must be camera relative (or you use another interpolator for that ?)

sebastienlagarde:
Currently in lit.shader, when we need true world space position like for Triplanar, then we call GetAbsolutePosition(). So we always transmit the camera relative position and convert to absolute position when the operation require it.

sebastienlagarde:
BTW, just a notice: We keep non camera relative mode to be able to debug :), as when you implement a new effect you can miss stuff. But otherwise there is no reason to disable it.

cdxntchou:
We called it 'Local Space' / LS at bungie, but that's not a great name either, as it is kind of ambiguous.
Camera Space is already used for camera origin AND camera orientation.
Maybe 'Relative Space' or 'Relative World Space' / RWS?

cdxntchou:
WorldSpacePosition is used only for the 'World Space Position' graph node -- (anything the user wants to use it for) -- it doesn't directly feed into any of the pipeline specific lighting code, or vertex positions.

sebastienlagarde:
Ok, the PR is ready for final review and include the renaming from positionWS to positionRWS

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -293,13 +293,22 @@
         #endif
     UNITY_INSTANCING_BUFFER_END(unity_Builtins2)
 
+
     #undef UNITY_MATRIX_M
-    #define UNITY_MATRIX_M     UNITY_ACCESS_INSTANCED_PROP(unity_Builtins0, unity_ObjectToWorldArray)
+    #ifdef OVERRIDE_PRE_CAMERA_TRANSLATION_TO_MATRIX
```
EvgeniiG: PRE_CAMERA_TRANSLATION ?

com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
```diff
@@ -374,7 +372,7 @@ ${Graph}
         float3 bentNormalWS =                   surfaceData.normalWS;           // TODO : make bent normals work
 
         builtinData.opacity =                   surfaceDescription.Alpha;
-        builtinData.bakeDiffuseLighting =       SampleBakedGI(fragInputs.positionWS, bentNormalWS, fragInputs.texCoord1, fragInputs.texCoord2);    // see GetBuiltinData()
+        builtinData.bakeDiffuseLighting =       SampleBakedGI(fragInputs.positionRWS, bentNormalWS, fragInputs.texCoord1, fragInputs.texCoord2);    // see GetBuiltinData()
```
EvgeniiG: We should pass the absolute position here, and not compute the absolute position inside, IMHO.

com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
```diff
@@ -385,13 +383,13 @@ ${Graph}
             // however it will not optimize the lightprobe case due to the proxy volume relying on dynamic if (we rely must get right of this dynamic if), not a problem for SH9, but a problem for proxy volume.
             // TODO: optimize more this code.
             // Add GI transmission contribution by resampling the GI for inverted vertex normal
-            builtinData.bakeDiffuseLighting += SampleBakedGI(fragInputs.positionWS, -fragInputs.worldToTangent[2], fragInputs.texCoord1, fragInputs.texCoord2) * bsdfData.transmittance;
+            builtinData.bakeDiffuseLighting += SampleBakedGI(fragInputs.positionRWS, -fragInputs.worldToTangent[2], fragInputs.texCoord1, fragInputs.texCoord2) * bsdfData.transmittance;
```
EvgeniiG: Same.

com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
```diff
@@ -385,13 +383,13 @@ ${Graph}
...
         }
 
 $SurfaceDescription.Emission:     builtinData.emissiveColor =             surfaceDescription.Emission;
         builtinData.velocity =                  float2(0.0, 0.0);
 #ifdef SHADOWS_SHADOWMASK
-        float4 shadowMask = SampleShadowMask(fragInputs.positionWS, fragInputs.texCoord1);
+        float4 shadowMask = SampleShadowMask(fragInputs.positionRWS, fragInputs.texCoord1);
```
EvgeniiG: Same here, maybe? Let's discuss.

com.unity.render-pipelines.high-definition/HDRP/Material/MaterialUtilities.hlsl
```diff
@@ -24,8 +24,8 @@ float3 SampleBakedGI(float3 positionWS, float3 normalWS, float2 uvStaticLightmap
     }
     else
     {
-        // TODO: We use GetAbsolutePositionWS(positionWS) to handle the camera relative case here but this should be part of the unity_ProbeVolumeWorldToObject matrix on C++ side (sadly we can't modify it for HDRenderPipeline...)
-        return SampleProbeVolumeSH4(TEXTURE3D_PARAM(unity_ProbeVolumeSH, samplerunity_ProbeVolumeSH), GetAbsolutePositionWS(positionWS), normalWS, unity_ProbeVolumeWorldToObject,
+        // TODO: We use GetAbsolutePositionWS(positionRWS) to handle the camera relative case here but this should be part of the unity_ProbeVolumeWorldToObject matrix on C++ side (sadly we can't modify it for HDRenderPipeline...)
```
EvgeniiG: Are you sure we cannot modify it?

com.unity.render-pipelines.high-definition/HDRP/ShaderVariablesFunctions.hlsl
```diff
@@ -1,11 +1,37 @@
...
+// This function always return the absolute position in WS either the CameraRelative mode is enabled or not
+float3 GetAbsolutePositionWS(float3 positionRWS)
+{
+#if (SHADEROPTIONS_CAMERA_RELATIVE_RENDERING != 0)
+    positionRWS += _WorldSpaceCameraPos;
+#endif
+    return positionRWS;
+}
+
+// This function always return the camera relative position in WS either the CameraRelative mode is enabled or not
```
EvgeniiG: This comment is misleading (look at the code)

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/UnityInstancing.hlsl
sebastienlagarde: fixed

com.unity.render-pipelines.high-definition/HDRP/ShaderVariablesFunctions.hlsl
sebastienlagarde: Fixed

EvgeniiG:
There are certain cases where you want to switch between camera-relative and absolute coordinates.
You can check our codebase for the references to GetAbsolutePositionWS() for examples.
Most of them deal with texture mapping in world space (e.g. world space planar mapping).
I think it may be a good idea to expose the nodes with these transformations even if we manage to abstract most use cases.

sebastienlagarde:
I practice we don't need to expose camera relative node in shader graph. I haven't see such nodes at Frosbite , neither in Unreal . Chris have you one at bungie? Only thing that matter is a getwolrdposition node that only return absolute position.

sebastienlagarde:
Hmm... I have doubt. I think at frostbite every transform in shader graph was in absolute world space. Wondering if unreal provide anything or use the same technique

cdxntchou:
Though, admittedly, we didn't have the RWS node at Bungie.. it is kind of difficult to use effectively unless you are passing other data down in the same space.  We had a 'tiled time' node to maintain floating point time precision; we could do a similar 'tiled RWS' space that is basically quantized RWS -- useful for tiled effects as long as it is a multiple of the tiling rate.   Gets complicated...  :)

sebastienlagarde:
Can't find anything: https://docs.unrealengine.com/en-us/Engine/Rendering/Materials/ExpressionReference/Coordinates. 

Yes maybe do as you suggest Chris . A RWS space.

PR #1590, opened 2018-06-21
Generate correct depth and shadow passes for LW with alpha clip

Kink3d:
Previously the alpha and clip theshold port contents were not evaluated when creating these passes. Now there is a minimal pass definition for these in C# including only these pixel slots. This means we now only need one C# method to generate a pass in LWRP.

Fixes fogbugz cases 1051822 and 1051834


zadi0648:
This is going to take a while to test thoroughly :/

zadi0648:
Removing myself; Wyatt reviewed it instead.

pbbastian:
@Kink3d Are Katana tests green for this PR?

Kink3d:
All tests green

PR #1605, opened 2018-06-29
Stabilize the dithered LOD transition mask with respect to the camera…

EvgeniiG:
… rotation

Katana link; https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lod_transition_update&automation-tools_branch=add-platform-filter&unity_branch=trunk


EvgeniiG:
Could probably improve the quality with a LatLong parametrization, but the extra ALU cost for inverse trigonometry is prohibitive.

EvgeniiG:
Hm, that's unexpected, I will have another look.

PR #1606, opened 2018-06-29
Add an option to the HDRP asset to increase the resolution of volumetric lighting

EvgeniiG:
Will probably add an option to dynamically control the number of slices at some point in the future.


com.unity.render-pipelines.high-definition/HDRP/Editor/RenderPipeline/Settings/SerializedRenderPipelineSettings.cs
```diff
@@ -16,8 +16,9 @@ class SerializedRenderPipelineSettings
         public SerializedProperty supportOnlyForward;
         public SerializedProperty supportMotionVectors;
         public SerializedProperty supportStereo;
-        public SerializedProperty enableUltraQualitySSS;
-        public SerializedProperty supportVolumetric;
+        public SerializedProperty increaseSssSampleCount;
```
sebastienlagarde: "use formely know as"
 enableUltraQualitySSS
 C# attribute to not break compatiblity

sebastienlagarde: same for supportVolumetrics

EvgeniiG: Done.

PR #1610, opened 2018-07-02
Remove PostProcessing dependency from Core

pbbastian:
The only things that needed the dependency were 2 utility functions used only by HD, so I've moved those to `HDUtils` in the HD package. Marked as api-change because it removes API points from the Core package.


sebastienlagarde:
API change should be pretty safe, I don't think anybody use this outside of HD anyway. Thanks for the change.

PR #1614, opened 2018-07-02
Decals/v2/per channel mask

Paulm-Unity:


sebastienlagarde:
Close: replace by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1674

Issue #1595, opened 2018-06-22
LWRP Post Processing Issue


m3rt32:
I am trying to create image effects with shader graph. But because of OnRenderImage not supported, I have no clue how to make one. What is the way of making post processing effects via shader graph?

demonixis:
You've to use the PostProcessing StackV2. There is an API to [create your custom effects](https://github.com/Unity-Technologies/PostProcessing/wiki/Writing-Custom-Effects)

m3rt32:
Thanks for answer, I've tried that but I am getting "Invalid shader" error. Here is the example project that I've tried:
https://github.com/m3rt32/ShaderGraphCustomPostProcessingTest

sebastienlagarde:
This is currently not supported.

m3rt32:
So, we cant use shader graph for post processing/image effects right now ?

sebastienlagarde:
No.

PR #1618, opened 2018-07-03
Fixes issue #1595. [DON'T MERGE YET. This PR depends on another PostFX PR]

phi-lira:
This PR depends on another PR in 
 - Removed code that was forcing post being disabled on LWRP.
 - Fixed wrong texture flip


nerites:
Same comments as given by chman on the other half of this changeset! I think changing the changelog to "fixed wrong texture fix" would be more informative and less misleading but otherwise code changes look good.

PR #1620, opened 2018-07-03
Use EVSM for volumetrics

EvgeniiG:
We downsample the entire shadow atlas by 4x and apply 2-component EVSM to it.
We use box filtering for now.
MIP level selection code is not yet finished.

Pro: much faster than PCF. Will reduce shadow aliasing in the distance when the MIP code is finished.
Con: quality loss when the shadow is viewed up close (poor prefiltering).


com.unity.render-pipelines.high-definition/HDRP/Lighting/Volumetrics/VolumetricLighting.cs
```diff
@@ -626,39 +626,42 @@ public void VolumetricLightingPass(HDCamera hdCamera, CommandBuffer cmd, uint fr
...
+                bool highQuality        = m_Preset == VolumetricLightingPreset.High;
                 bool enableClustered    = hdCamera.frameSettings.lightLoopSettings.enableTileAndCluster;
                 bool enableReprojection = Application.isPlaying && hdCamera.camera.cameraType == CameraType.Game;
 
-                int kernel;
+                string kernelName = "VolumetricLighting";
+
+                if (m_UsePcf)
+                {
+                    kernelName += "Pcf";
```
sebastienlagarde: Building string like this cause memory GC call. Do a prebuild tab.

EvgeniiG:
Obsolete.

PR #1633, opened 2018-07-04
Stack lit2

slunity:
This PR add specular occlusion support to Lit shader and StackLit shader based on Spherical Pivot Distribution from Jonathan Dupuy


laurenth-unity:
I got an issue when using the object space normal / bent normal map mode.

slunity:
This is a problem in the Lit material. 
Seems the _BentNormalMapOS property is missing in LitProperties.hlsl, 
(ie no  "PROP_DECL_TEX2D(_BentNormalMapOS);"  ) 
but this problem is there in master and staging too.

sebastienlagarde:
yes, it still plan

sebastienlagarde:
reopen, close by mistake

PR #1652, opened 2018-07-10
Remove all legacy matrices from Shader Graph

Kink3d:
This PR replaces #1566. That PR reintroduced legacy matrices into HDRP for Shader Graph to use. This PR, however, removes those matrices from Shader Graph.

Fixes HUP bug case (Fogbugz case 1046087)

- Removes all legacy matrices from Matrix Transfomation node
   - Added enum remapper to preserve user selections that are still valid
- Removes all references to `MV`, `T_MV`, `IT_MV` and `MVP` matrices from core graph code
- Replaces `IT_MV` reference in Camera node with specific `transpose(mul(x,y))`


com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
```diff
@@ -6,8 +6,10 @@
 
 namespace UnityEditor.ShaderGraph
 {
+    // TODO - Remove this when leaving preview
```
pbbastian: Please don't merge in stuff that has comments like this :( What's the plan for this thing when we do leave preview? And can it be made internal?

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
```diff
@@ -18,39 +20,68 @@ public enum TransformationMatrixType
         WorldToObject
     };
 
+    public enum UnityMatrixType
```
pbbastian: Does this have to be public?

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
```diff
@@ -18,39 +20,68 @@ public enum TransformationMatrixType
...
+            {TransformationMatrixType.ObjectToWorld, UnityMatrixType.Model},
+            {TransformationMatrixType.WorldToObject, UnityMatrixType.InverseModel},
         };
 
+        // TODO - Remove this when leaving preview
         [SerializeField]
         private TransformationMatrixType m_matrix = TransformationMatrixType.ModelView;
 
+        [SerializeField]
+        private UnityMatrixType m_Matrix = UnityMatrixType.Model;
```
pbbastian: This only differs from `m_matrix` by the case of a single letter. `m_matrix` should've had a capital `m`, but due to Unity serialization we'll have to leave it. But please do rename this one to have a more specific name. Maybe `m_MatrixType`?

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
```diff
@@ -68,13 +99,21 @@ public override string documentationURL
 
         public sealed override void UpdateNodeAfterDeserialization()
         {
+            // TODO - Remove this when leaving preview
```
pbbastian: Same as above.

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
Kink3d: I assumed it would be ok because theres stuff like this in both HDRP and LWRP. My intention was we would just remove the remap and the old serialised property when we leave preview, chances of anyone needing that remap by this point is very slim...

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
Kink3d: Yup, thats fine

com.unity.shadergraph/Editor/Data/Nodes/Input/Matrix/TransformationMatrixNode.cs
Kink3d: All the other ones are, maybe well need to default matrices on another node. Can make it non-public though of course.

PR #1655, opened 2018-07-11
Fix DecalProjector material editor release (64 preview issue)

RSlysz:
Fix FB#1057395 for DecalProjector.


RSlysz:
When working on the scene, if you select a DecalProjector and you already
have done that 63 times before you have an issue with the preview creation.
It is due to no clean method call on the preview created. Here it is done
in the material editor. And it happens due to no release of this editor
when not needed anymore.

STR: make a scene with two DecalProjector. Alternate selecting between them
64 time or more. You get error with preview creation exceding 64.

Le mer. 11 juil. 2018 à 18:42, Paulm-Unity <notifications@github.com> a
écrit :

sebastienlagarde:
Replace by this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1698

PR #1658, opened 2018-07-11
Fixed distortion 1-frame delay

Chman:
For some reason distortion was using the color pyramid from the previous frame instead of the current one. This PR fixes that, distortion isn't delayed anymore. This should also potentially fix a bunch of weird tests behavior (cc @RemyUnity).


sebastienlagarde:
@RemyUnity we will need to regenerate the test screenshots. Maybe this explain our dark issue with katana?

PR #1668, opened 2018-07-12
Only initialize indirect buffer to size

jessebarker:
On iOS, we get a GPU hang/reset based upon the execution of the ClearDispatchIndirect kernel.  The buffer is created using the 'NUM_FEATURE_VARIANTS' constant (currently 27), but we're dispatching a threadgroup of 64 threads, so we are presumably going out of bounds (Metal GPU errors are not terribly descriptive, but this is a common cause).


runestubbe:
Looks good to me. On PC that test shouldn't be necessary but this is code is by no means performance critical, so I think just always doing the test is the most sensible thing.

runestubbe:
Argh, how do I actually set my status to green on this thing?

jessebarker:
I don't think the existing code is technically "legal" on any platform.  Some implementations (backed by hardware support, no doubt) may handle the out-of-bounds access gracefully (e.g., Vulkan or GL with robust buffer access enabled), but I think you are still guaranteed to generate the out-of-bounds access in the first place.  Does DX allow this?

I'm happy to add comments, etc., if you feel strongly about it (or if I'm wrong about other APIs allowing this more generally).

runestubbe:
In DirectX out-of-bounds reads are always zero and out-of-bounds writes are are a no op. But if the bounds check is required on some platforms, I think we should just add it. No need to add ifdefs as this shader only runs a single wave probably no more than once per frame, so adding a few instructions is going to do absolutely no difference in terms of performance.

sebastienlagarde:
Recreate the PR for HDRP/Staging, close this one, and merged into HDRP/Staging
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1694

PR #1673, opened 2018-07-13
Pipeline specific graph functions and nodes

Kink3d:
### Purpose 

This PR adds the concept of pipeline speific nodes via macro based "overrides". Default functions are implemented per node in `Functions.hlsl` (inside SG package) which can then be overriden by define and macro by each pipeline.

Each pipeline now has its own `GraphFunctions.hlsl` inside the respective package. This defines the "overrides" then references the defaults.

### Changes

- Define `graphRequirements` in all subshaders (union of all user defined requirements)
- Add interface `IMayRequireDepthTexture`
- Add define `REQUIRE_DEPTH_TEXTURE` based on `IMayRequireDepthTexture` in all RPs
- Add interface `IMayRequireCameraOpaqueTexture`
- Add define `REQUIRE_CAMERA_OPAQUE_TEXTURE` based on `IMayRequireDepthTexture` in LWRP
- Add new interfaces to requirement logic in `ShaderGraphRequirements`
- Change `Functions.hlsl` references in templates to RP specific `GraphFunctions.hlsl`
- Add Scene Depth node
- Define override function for Scene Depth node for default and all RPs
- Add Scene Color node
- Define override function for Scene Color node for default and LWRP
- Reposition `GraphFunctions.hlsl` references in HD templates so `_CameraDepthTexture` is defined

### Risks

- I dont think this needs an importer version bump because the base include file is still valid from previous version?
- `_CameraDepthTexture` may be defined elsewhere and end up defined twice, although I think I checked it all
- Adding new requirements always has an associated risk...


sebastienlagarde:
Random question but what not call the file GraphFunctions.hlsl, ShaderGraphFunctions.hlsl? At least this is explicit :)

com.unity.render-pipelines.high-definition/HDRP/GraphFunctions.hlsl
```diff
@@ -0,0 +1,19 @@
+#ifndef UNITY_GRAPHFUNCTIONS_HD_INCLUDED
+#define UNITY_GRAPHFUNCTIONS_HD_INCLUDED
+
+#define SHADERGRAPH_SAMPLE_SCENE_DEPTH(uv) shadergraph_HDSampleSceneDepth(uv);
+
+float shadergraph_HDSampleSceneDepth(float2 uv)
+{
+#if defined(REQUIRE_DEPTH_TEXTURE)
```
sebastienlagarde: I am curious but in which case is it possible to have a call to shadergraph_HDSampleSceneDepth without having REQUIRE_DEPTH_TEXTURE define ? 
If a shader graph use a node SampleDeth, then it call shadergraph_HDSampleSceneDepth . If it haven't, then there is no call to shadergraph_HDSampleSceneDepth and thus we don't need to care about REQUIRE_DEPTH_TEXTURE .

com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/GraphFunctions.hlsl
```diff
@@ -0,0 +1,46 @@
...
+    float rawDepth = SAMPLE_DEPTH_TEXTURE(_CameraDepthTexture, sampler_CameraDepthTexture, uv);
+#endif
+	return Linear01Depth(rawDepth, _ZBufferParams);
+#endif // REQUIRE_DEPTH_TEXTURE
+    return 0;
+}
+
+float3 shadergraph_LWSampleSceneColor(float2 uv)
+{
+#if defined(REQUIRE_OPAQUE_TEXTURE)
```
sebastienlagarde: Same remark here for REQUIRE_OPAQUE_TEXTURE and REQUIRE_DEPTH_TEXTURE than for HD

sebastienlagarde:
Question about the design.

I am worried that we add a GraphFunction.hlsl file in each pipeline that will most of the time duplicate already existing function in these renderpipeline (or just be a wrapper on them).  I think it will be simpler to just generate the correct call from the generator instead of relying on the macro system. For example

        static string Unity_SceneDepth(
+            [Slot(0, Binding.ScreenPosition)] Vector3 UV,
+            [Slot(1, Binding.None, ShaderStageCapability.Fragment)] out Vector3 Out)
+        {
+            Out = Vector3.one;
+            return
+                @"
+{
+    Out = SHADERGRAPH_SAMPLE_SCENE_DEPTH(UV);
+}
+";
+        }

HD or LW could derive the node to overload the function Unity_SceneDepth and simply generate the right call. This is simple to edit (as the macro system is) but don't add again another wrapper on top of existing function. 

So for HD it would be

        static string Unity_SceneDepth(
+            [Slot(0, Binding.ScreenPosition)] Vector3 UV,
+            [Slot(1, Binding.None, ShaderStageCapability.Fragment)] out Vector3 Out)
+        {
+            Out = Vector3.one;
+            return
+                @"
+{
float depth = LOAD_TEXTURE2D(_CameraDepthTexture, UV).x;
+    Out = LineazeZ01(depth );
+}
+";
+        }


Also side note: For texture fetch like depth for platform that support it, always use a LOAD instead of a sample nearest, it is faster.

Kink3d:
@sebastienlagarde Whilst I agree with you that doing the override on the node level would be a better solution it would still be a wrapper for the same code, just in C# string literal instead of an HLSL include...

The problem here is we dont have the structure on the node side to define overrrides to nodes per-pipeline. We could introduce something like this when we redesign the node API (as we couldnt have the override at this level anyway, it would have to be at the base node class which is much less simple) but this likely wont happen for a while...

In the meantime at least this solution doesnt stop us from doing that later, and wouldnt make already exisiting graphs non-functional when the switch is made.

As for duplicating code, I agree we need to minimize this. If each pipeline defines functions like "GetDepthTexture" which we would just call from these graph functions that would be better, I can refactor this PR to do that if you want?

sebastienlagarde:
Alright, let's go with this, we can still define the macro to point directly on function if needed.

So I fine with this PR, just would like to rename GraphFunctions.hlsl to ShaderGraphFunction.hlsl

Kink3d:
Alright, renamed :)

Kink3d:
All tests green!

PR #1681, opened 2018-07-13
Fix/coupling reflection probe reset request C++ PR#70483

RSlysz:
Fix contextual menu reset / remove for ReflectionProbe in inspector


sebastienlagarde:
Replace by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1796

PR #1684, opened 2018-07-13
Generate correct depth and shadow passes for LW with alpha clip

Kink3d:
Backport pull request #1590 into 2018.2


Kink3d:
All tests green

PR #1691, opened 2018-07-16
Changed version in HDAdditionalLightData from float to int

alelievr:


RSlysz:
Il manque le pendant pour les shadows non ?

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -47,11 +47,16 @@ struct TimelineWorkaround
     [ExecuteInEditMode]
     public class HDAdditionalLightData : MonoBehaviour, ISerializationCallbackReceiver
     {
-        public const float currentVersion = 1.1f;
+        public const int currentVersion = 2;
```
RSlysz: No need to be public

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -47,11 +47,16 @@ struct TimelineWorkaround
...
 
-        [HideInInspector]
+        [HideInInspector, SerializeField]
         [FormerlySerializedAs("m_Version")]
-        public float version = currentVersion;
+        [System.Obsolete("version is deprecated, use m_Version instead")]
+        private float version = currentVersion;
+// Currently m_Version is not used and produce a warning, remove these pragmas at the next version incrementation
+#pragma warning disable 414
+        private int m_Version = currentVersion;
```
RSlysz: Affectation must not be done here but inside the OnAfterDeserialization to be able to know from which version we upgrade and thus the task needed to be done in case of we migrate from before any versioning is done.

something like:
public void OnAfterDeserialize()
        {
            if (m_Version != currentVersion)
            {
                // Add here data migration code
                m_Version = currentVersion;
            }
        }

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -47,11 +47,16 @@ struct TimelineWorkaround
...
+#pragma warning restore 414
```
RSlysz: And thus, this pragma warning should not be usefull anymore

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -547,10 +556,13 @@ public void UpgradeLight()
                         intensity = areaIntensity;
                         break;
                 }
-#pragma warning restore 0618
+                needsIntensityUpdate_1_0 = false;
             }
 
+            m_Version = currentVersion;
```
RSlysz: And in fact you effectively done it there.

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
alelievr: I dont think assigning the version in OnAfterDeserialize is a good idea because in case we create a new light, we would be upgrading it as if it was old (before version system), furthermore setting the version here is a bad idea because we can't put the megration code in OnAfterDeserialize (because of monobehavior initialization).

com.unity.render-pipelines.high-definition/HDRP/Lighting/Light/HDAdditionalLightData.cs
alelievr: Fixed

RSlysz:
It is Ok to me now. But we do not support data from before v1.0 (but it seams Ok with Sebastien)

PR #1704, opened 2018-07-17
Refactoring of HDRP constant buffers

JulienIgnace-Unity:
In order to avoid having too many constant buffers bound at the same time (sometimes hitting the maximum limit), most constant buffers used during draw calls have been merged into one global constant buffer.
- Merged PerFrame/PerView into UnityGlobal
- Merged most sub systems constant buffers into UnityGlobal (decal, subsurface, lighting, ...)
- Moved AtmosphericScattering into the Lighting folder.

This should allow most draw calls to only bind 3 constant buffers.


com.unity.render-pipelines.high-definition/HDRP/HDRenderPipelineAsset.asset
```diff
@@ -24,7 +24,7 @@ MonoBehaviour:
     enableVolumetrics: 1
     diffuseGlobalDimmer: 1
```
sebastienlagarde: Don't commit HDRP Asset

com.unity.render-pipelines.high-definition/HDRP/Lighting/AtmosphericScattering/ShaderVariablesAtmosphericScattering.hlsl
```diff
@@ -0,0 +1,33 @@
+#ifdef SHADER_VARIABLES_INCLUDE_CB
```
sebastienlagarde: Add comment about this, why we have both, explain that the goal is to be included inside a single bufffer an maintain side by side in the file both CB and textures

sebastienlagarde: Add this at the top of each header

JulienIgnace-Unity: There is a comment in ShaderVariables.hlsl where everything is included

PR #1708, opened 2018-07-18
Lw/terrain optimization

aleks01010101:
Terrain shader optimizations.


com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/Terrain/LightweightPassLitTerrain.hlsl
```diff
@@ -54,45 +61,47 @@ void InitializeInputData(VertexOutput IN, half3 normalTS, out InputData input)
...
 
 #if !defined(SHADER_API_MOBILE) && defined(TERRAIN_SPLAT_ADDPASS)
-    clip(weight == 0.0f ? -1 : 1);
+    clip(weight == 0.0h ? -1.0h : 1.0h);
 #endif
 
     // Normalize weights before lighting and restore weights in final modifier functions so that the overal
     // lighting result can be correctly weighted.
-    splat_control /= (weight + 1e-3f);
+    splatControl /= (weight + 1e-3h);
```
phi-lira: maybe we should use HALF_MIN here

com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/Terrain/LightweightPassLitTerrain.hlsl
```diff
@@ -35,12 +40,14 @@ void InitializeInputData(VertexOutput IN, half3 normalTS, out InputData input)
     input.positionWS = IN.positionWS;
 
 #ifdef _TERRAIN_NORMAL_MAP
-    input.normalWS = TangentToWorldNormal(normalTS, IN.tangent, IN.binormal, IN.normal);
+    half3 viewDir = half3(IN.normal.w, IN.tangent.w, IN.binormal.w);
+    input.normalWS = TangentToWorldNormal(normalTS, IN.tangent.xyz, IN.binormal.xyz, IN.normal.xyz);
```
phi-lira: I'm almost sure we have to always normalize after transforming here to get correct results. I think this is a bug in lwrp shaders as well.

aleks01010101: Perhaps. Then this should be part of a fix for the whole :)

PR #1712, opened 2018-07-18
Update UI for decal material based on artists feedback + add upgrader

sebastienlagarde:
This PR perform various caption renaming + update the UI layout + fix the behavior of checkbox to not permit to uncheck the box when it is the latest active component

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Editor/Upgraders/UpgradeMenuItem.cs
+++ b/com.unity.render-pipelines.high-definition/HDRP/Editor/Upgraders/UpgradeMenuItem.cs
@@ -208,6 +208,23 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
             return;
         }
 
+        // Update decal material after we added AO and metal selection. It was default to 0 and need to default to 4 now.
+        static bool UpdateMaterial_DecalBlendMode(string path, Material mat)
+        {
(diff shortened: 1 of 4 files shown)


com.unity.render-pipelines.high-definition/HDRP/Editor/Upgraders/UpgradeMenuItem.cs
```diff
@@ -208,6 +208,23 @@ static void UpdateMaterialFile_EmissiveColor(string path)
...
+        // Update decal material after we added AO and metal selection. It was default to 0 and need to default to 4 now.
+        static bool UpdateMaterial_DecalBlendMode(string path, Material mat)
+        {
+            if (mat.shader.name == "HDRenderPipeline/Decal")
+            {
+                float maskBlendMode = mat.GetFloat("_MaskBlendMode");
+
+                if (maskBlendMode == 0.0f)
+                {
+                    mat.SetFloat("_MaskBlendMode", 4.0f);
```
Paulm-Unity: Should use enum here instead of 4.0,  Decal.MaskBlendFlags.Smoothness

sebastienlagarde: good point, added a fix

PR #1731, opened 2018-07-19
Removed XRSettings usage from HDRP. Protected XR usage in XRGraphicsConfig with ifdefs. 

nerites:
Also added a stereoRenderingPath method to XRGraphicsConfig. 

Pending Xbox build check from Sam Bickley. 
Tim- what other tests need to be run on this branch before PR approval?


nerites:
Solved the issue where XRGraphicsConfig was blocking builds (StereoRenderingPath is only defined in UnityEditor).

This should be ready for merge into master at this point.

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -147,11 +146,7 @@ public static void InitializeFrameSettings(Camera camera, RenderPipelineSettings
             // Planar and real time cubemap doesn't need post process and render in FP16
             aggregate.enablePostprocess = camera.cameraType != CameraType.Reflection && srcFrameSettings.enablePostprocess;
 
-#if UNITY_SWITCH
-            aggregate.enableStereo = false;
-#else
-            aggregate.enableStereo = camera.cameraType != CameraType.Reflection && srcFrameSettings.enableStereo && XRSettings.isDeviceActive && (camera.stereoTargetEye == StereoTargetEyeMask.Both) && renderPipelineSettings.supportStereo;
-#endif
+            aggregate.enableStereo = camera.cameraType != CameraType.Reflection && srcFrameSettings.enableStereo && XRGraphicsConfig.enabled && (camera.stereoTargetEye == StereoTargetEyeMask.Both);
```
sebastienlagarde: Please, continue to test && renderPipelineSettings.supportStereo.
Support may allow us to remove shader variant and save memory, so it need to be tested here.

com.unity.render-pipelines.core/CoreRP/Common/XRGraphicsConfig.cs
```diff
@@ -1,7 +1,13 @@
 using System;
 using UnityEditor;
-using UnityEngine.XR;
```
sebastienlagarde: I think we should move the file into a CoreRP/XR folder and not in common. A bit too generic. What do you think?

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/Settings/FrameSettings.cs
nerites: Yes, it's read-only access to XRSettings.enabled, which can only be set outside of playmode.

sebastienlagarde:
note: need to run katana before merging

sebastienlagarde:
katana is green

nerites:
Thanks for running that for me! Just to confirm: the mac failure is known?

sebastienlagarde:
yes, this mac failure is not due to mac, it happen sometime.

PR #1734, opened 2018-07-20
Hdrp/decals/v2 1/snap to grid

Paulm-Unity:


Paulm-Unity:
Improves decal extent editing in editor.

PR #1760, opened 2018-07-23
Refactor reflectionprobe/use separate influence volume class for hd reflection probe too

RSlysz:
- Move InfluenceVolume usage to HDProbe
- Adapte HDAdditionalData to use InfluenceVolume too
- Use InfluenceVolume drawer inside HDReflectionProbe drawer

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Lighting/Reflection/Volumes/InfluenceVolume.cs
+++ b/com.unity.render-pipelines.high-definition/HDRP/Lighting/Reflection/Volumes/InfluenceVolume.cs
@@ -1,5 +1,6 @@
 using System;
 using UnityEngine.Serialization;
+using UnityEngine.Experimental.Rendering;
 
 namespace UnityEngine.Experimental.Rendering.HDPipeline
 {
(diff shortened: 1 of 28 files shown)


com.unity.render-pipelines.high-definition/HDRP/Lighting/Reflection/Volumes/InfluenceVolume.cs
```diff
@@ -1,117 +1,185 @@
...
-        public Vector3 boxInfluenceNormalOffset { get { return (boxInfluenceNormalNegativeFade - boxInfluenceNormalPositiveFade) * 0.5f; } }
-        public Vector3 boxInfluenceNormalSizeOffset { get { return -(boxInfluenceNormalPositiveFade + boxInfluenceNormalNegativeFade); } }
+        /// <summary>Size of the InfluenceVolume in Box Mode.</summary>
+        public Vector3 boxSize
+        {
+            get { return m_BoxSize; }
+            set
+            {
+                m_BoxSize = value;
+                if (OnSizeChanged != null)
```
fredpointzero: Prefer to enclose the event caller in a method with arguments like:
```
void OnSizeChanged(Vector3 boxSize, Vector3 offset)
{
    if (SizeChanged != null)
        SizeChanged(boxSize, offset);
}
```

PR #1771, opened 2018-07-24
Hdrp/refactor reflection probe/clean proxy volume

RSlysz:
Clean ProxyVolume classes and associated editors


com.unity.render-pipelines.high-definition/HDRP/Lighting/Reflection/ProbeWrapper.cs
```diff
@@ -27,6 +27,19 @@ protected static EnvShapeType ConvertShape(Shape shape)
             }
         }
 
+        protected static EnvShapeType ConvertShape(ShapeOrInfinite shape)
+        {
+            switch (shape)
+            {
+                default:
+                case ShapeOrInfinite.Infinite:
```
fredpointzero: Prefer to use the sphere shape for infinite. (Sphere is cheaper in shader code)

com.unity.render-pipelines.high-definition/HDRP/Lighting/Reflection/Volumes/ShapeType.cs
```diff
@@ -5,4 +5,11 @@ public enum Shape
         Box,
         Sphere,
     }
+
+    public enum ShapeOrInfinite
```
fredpointzero: `ShapeOrInfinite` is a confusing name, I prefer `Shape`.

This is more confusing when reading core, it sounds more like a variable than a type name.

RSlysz:
Shape is already used for InfluenceVolume's shape which only contrains Box
and Sphere value. Are InfluenceShape and ProxyShape more suitable?

>

sebastienlagarde:
Intergate change from FRed, Verbal approve from FRed :)

PR #1780, opened 2018-07-25
Fix LWRP Shadow Pass incorrect assert and params setting

polar1225:


stramit:
Thanks for this PR, running it through our automated testing suite now.

stramit:
Hi, we have grafted these changes to a different PR ([here](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1808)), to allow it to run on our test framework (currently we can only run tests from our main repository, and you are PRing from an external repository). 

It should be merged in within 2 days and in the next release. Many thanks for this PR 👍

PR #1790, opened 2018-07-26
Indirect lighting control volume

alelievr:


com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -2051,6 +2051,10 @@ void PostEvaluateBSDF(  LightLoopContext lightLoopContext,
     // Subsurface scattering mode
     float3 modifiedDiffuseColor = GetModifiedDiffuseColorForSSS(bsdfData);
 
+    // Apply volume indirect multipler
+    builtinData.bakeDiffuseLighting *= _IndirectLightingMultiplier.x;
```
sebastienlagarde: After thinking more about it, we should do this in MaterialUtilities (and with new code I just merge in master :) )
For indirect diffuse

        // Apply control from the indirect lighting volume settings
        builtinData.bakeDiffuseLighting *= _IndirectLightingMultiplier.x;
        builtinData.backbakeDiffuseLighting *= _IndirectLightingMultiplier.x;

sebastienlagarde: For the other one indirect specular we will apply it here in C# in Lightloop.cs
GetEnvLightData() envLightData.multiplier = probe.multiplier * _IndirectLightingMultiplier.y;

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1364,10 +1364,11 @@ public bool GetEnvLightData(CommandBuffer cmd, Camera camera, ProbeWrapper probe
             // Build light data
             var envLightData = new EnvLightData();
 
+            float indirectSpecularMultiplier = VolumeManager.instance.stack.GetComponent<IndirectLightingController>().indirectSpecularIntensity;
```
sebastienlagarde: is it ok to do it for each light? Don't know if these value is cached or not, maybe ask Thomas what is better, calling it one time and chaching or if it is fine like this

PR #1802, opened 2018-07-26
Set correct default values for texture slot properties

Kink3d:
### Purpose

This PR removes the hardcoding of the default value for texture properties. This is now set via the `Texture2DInputMaterialSlot`.

Implemented for nodes:
- SampleTexture2D
- SampleTexture2DLOD
- Triplanar

### Changes

- Add `DefaultType` enum to `TextureShaderProperty`
- Serialise `DefaultType` member on `TextureShaderProperty`
- Append correct default value in `GetPropertyBlockString`
- Serialise `DefaultType` member on `Texture2DInputMaterialSlot`
- Set `defaultType` on shader property when created in `Texture2DInputMaterialSlot`
- Set correct default type on nodes listed above

### Risks

- Currently, when a node tries to change its property's default value based on an on-node control, the correct code is generated but the preview is not updated until it is redrawn again. This is due to the order in which this code is executed. This repros when the node is not connected to the master node, otherwise it updates correctly. Fixing this requires a big refactor and therefore, probably, should not be included in a bugfix PR.


com.unity.shadergraph/Editor/Data/Graphs/Texture2DInputMaterialSlot.cs
```diff
@@ -7,18 +7,29 @@
 
 namespace UnityEditor.ShaderGraph
 {
+    using DefaultType = TextureShaderProperty.DefaultType;
```
pbbastian: Unity style reference disallows putting `using` inside a namespace. Make sure to also get the other instances of this :)

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
```diff
@@ -76,6 +77,9 @@ public sealed override void UpdateNodeAfterDeserialization()
         // Node generations
         public virtual void GenerateNodeCode(ShaderGenerator visitor, GenerationMode generationMode)
         {
+            var textureSlot = FindInputSlot<Texture2DInputMaterialSlot>(TextureInputId);
+            textureSlot.defaultType = (textureType == TextureType.Normal ? DefaultType.Bump : DefaultType.White);
```
pbbastian: Not a fan of modifying global state during code generation. Could this be done immediately when selecting the type in the dropdown instead?

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
```diff
@@ -4,8 +4,7 @@
 using UnityEditor.ShaderGraph.Drawing.Controls;
 
 namespace UnityEditor.ShaderGraph
-{
-    
+{
```
pbbastian: ❤️

PR #1811, opened 2018-07-30
 Fixed _WorldToShadow warning.

phi-lira:
Fixed Property (_WorldToShadow) exceeds previous array size (5 vs 1) warning due caused because we cache the shader property ID. Solution was to remove the ifdef around the property. In the shader code the matrix will be turned into a vec4 array and we don't benefit from not indexing.

### DIFF
--- a/com.unity.render-pipelines.lightweight/CHANGELOG.md
+++ b/com.unity.render-pipelines.lightweight/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Fixed Property (_WorldToShadow) exceeds previous array size (5 vs 1) warning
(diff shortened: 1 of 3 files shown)


phi-lira:
Tests Green: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-All%20Tests/builds/324?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw/fix-shadowcascadewarning&unity_branch=trunk

com.unity.render-pipelines.lightweight/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+- Fixed Property (_WorldToShadow) exceeds previous array size (5 vs 1) warning
```
Katstodian: Please replace with:
When you change a Shadow Cascade option in the Pipeline Asset, this no longer warns you that you've exceeded an array size.

PR #1814, opened 2018-07-31
Lw/fix custom renderer lifetime 1061353

phi-lira:
Fixes case 1061353:
   - Materials are not cached anymore in the passes, instead the renderer is passed in execution and resources are accessed from it.


stramit:
Feel free to squash and merge when katana is green.

PR #1829, opened 2018-08-02
Convert LW / Core RP tests into pacakges

stramit:
Port LW and Core tests to live in packages. They will be built as part of the package build process and also published. This means the can be used on katana and similar for testing.

LW / core are 'first version' of this. We need 'real usage' before we port HD / Shader Graph.

Katana:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=packagify-tests&unity_branch=trunk


build.py
```diff
@@ -9,15 +9,13 @@ def packages_list():
...
+        ("com.unity.testing.srp.core", os.path.join("com.unity.testing.srp.core")),
+        ("com.unity.testing.srp.lightweight", os.path.join("com.unity.testing.srp.lightweight"))
     ]
 
 def test_packages_list():
     return [
-        "com.unity.render-pipelines.core",
-        "com.unity.render-pipelines.lightweight",
-        "com.unity.render-pipelines.high-definition",
-        "com.unity.shadergraph"
```
phi-lira: we are removing hd and shader graph from this list but not adding any test package to the one above. Is this expected?

com.unity.testing.srp.lightweight/Tests/Unity.Testing.SRP.Lightweight.Runtime.asmdef.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: c30d5ade4e19e114ba605fbd1faa5512
+guid: 7a8b1e213daa42642a24d29f85104bfc
```
phi-lira: Is this GUID change intentional?

build.py
stramit: yep this is correct for what we are doing.

com.unity.testing.srp.lightweight/Tests/Unity.Testing.SRP.Lightweight.Runtime.asmdef.meta
stramit: not intentional but does not matter for asm defs

PR #1830, opened 2018-08-02
Lw/receive shadows prop

phi-lira:
Adds supports to receive shadows property in LWRP. 
 - The property is exposed in the material instead of in the renderers.
 - Standard Physically Based and Simple Lighting shaders define shader_feature _RECEIVE_SHADOWS_OFF
 - Shaders that don't define and set this keyword keep the original behavior of receiving shadows.

Adding Matt Dean for awareness if we need something in ShaderGraph


com.unity.render-pipelines.lightweight/LWRP/Editor/ShaderGUI/LightweightShaderGUI.cs
```diff
@@ -26,6 +26,7 @@ private static class Styles
         public static GUIContent twoSidedText = new GUIContent("Two Sided", "Render front and back faces");
         public static GUIContent alphaClipText = new GUIContent("Alpha Clip", "Enable Alpha Clip");
         public static GUIContent alphaClipThresholdText = new GUIContent("Clip Threshold", "Threshold for alpha clip");
+        public static GUIContent receiveShadowText = new GUIContent("Receive Shadows", "Enables this material to receive shadows if there is at least a shadow casting light affecting it.");
```
aleks01010101: "...if there is at least one shadow casting light..."

com.unity.render-pipelines.lightweight/LWRP/Shaders/LightweightStandard.shader
```diff
@@ -51,6 +51,8 @@ Shader "LightweightPipeline/Standard (Physically Based)"
         [HideInInspector] _DstBlend("__dst", Float) = 0.0
         [HideInInspector] _ZWrite("__zw", Float) = 1.0
         [HideInInspector] _Cull("__cull", Float) = 2.0
+
+        _ReceiveShadows("Receive Shadows", Float) = 1.0
```
aleks01010101: Shouldn't this have [ToggleOn] or [ToggleOff] or somesuch?

phi-lira:
We don't have custom material inpector for terrains ATM and it seems the Terrains didn't originally had the option for receive shadows. At least for now I think it's safe to not have it. If people ask for it then we can discuss about adding it.

com.unity.render-pipelines.lightweight/LWRP/Shaders/LightweightStandard.shader
phi-lira: I couldn't find much docs about it but it seems this is used for EditorGUILayout.PropertyField. While drawing this property we use EditorGUILayout.Toggle which seems to make the [ToogleOn/Off] Not necessary.

com.unity.render-pipelines.lightweight/LWRP/Editor/ShaderGUI/LightweightShaderGUI.cs
phi-lira: fixed

PR #1832, opened 2018-08-02
Disable SSS when luxmeter is enabled

alelievr:


com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -785,7 +785,7 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
 
                 FrameSettings currentFrameSettings = new FrameSettings();
                 // Get the effective frame settings for this camera taking into account the global setting and camera type
-                FrameSettings.InitializeFrameSettings(camera, m_Asset.GetRenderPipelineSettings(), srcFrameSettings, ref currentFrameSettings);
+                FrameSettings.InitializeFrameSettings(camera, m_Asset.GetRenderPipelineSettings(), m_DebugDisplaySettings, srcFrameSettings, ref currentFrameSettings);
```
sebastienlagarde: Don't pass the m_DebugDisplaySettings to the function but do the modification after: 

for example just after this:

                   // Disable post process if we enable debug mode or if the post process layer is disabled
                    if (m_CurrentDebugDisplaySettings.IsDebugDisplayRemovePostprocess() || !HDUtils.IsPostProcessingActive(postProcessLayer))
                    {
                        currentFrameSettings.enablePostprocess = false;
                    }

PR #1854, opened 2018-08-07
Light attenuation matching that of baked GI

egomeh:
* Adds light request callback for lights when baking. This sets the falloff to inverse square, and the inner cone angle of the spotlights.
* Changes the light attenuation in `Lighting.hlsl` to match the baked GI.
* Adds a graphics test for light attenuation in spotlights.


com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
```diff
@@ -359,5 +365,127 @@ void SetupPerCameraShaderConstants(CameraData cameraData)
...
+            // In that case `light.bakingOutput.lightmapBakeType` would be baked, instead of mixed, whereas `light.lightmapBakeType` would still be mixed. But this difference is only relevant in editor builds
+#if UNITY_EDITOR
+            lightData.mode = LightmapperUtils.Extract(light.lightmapBakeType);
+#else
+            lightData.mode = LightmapperUtils.Extract(light.bakingOutput.lightmapBakeType);
+#endif
+
+            lightData.shadow = (byte)(light.shadows != LightShadows.None ? 1 : 0);
+
+            if (true)
```
phi-lira: I think this was left here by mistake.

com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
```diff
@@ -359,5 +365,127 @@ void SetupPerCameraShaderConstants(CameraData cameraData)
...
+                        lightData.coneAngle = light.spotAngle * Mathf.Deg2Rad; // coneAngle is the full angle
+                        lightData.innerConeAngle = LightmapperUtils.ExtractInnerCone(light) * 0.5f;
+#if UNITY_EDITOR
+                        lightData.shape0 = light.shadows != LightShadows.None ? light.shadowRadius : 0.0f;
+#else
+                        lightData.shape0 = 0.0f;
+#endif
+                        lightData.shape1 = 0.0f;
+                        lightData.type = UnityEngine.Experimental.GlobalIllumination.LightType.Spot;
+                        lightData.falloff = FalloffType.InverseSquared;
```
phi-lira: So, the only thing you want to override is the falloff right? Isn't there a default initializer for this lightData? If so, it would be nice to initialize it with the default value and just override what we need in case of Spot and Point lights.

com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
```diff
@@ -359,5 +365,127 @@ void SetupPerCameraShaderConstants(CameraData cameraData)
...
+                        lightData.innerConeAngle = 0.0f;
+
+#if UNITY_EDITOR
+                        lightData.shape0 = light.shadows != LightShadows.None ? light.shadowRadius : 0.0f;
+#else
+                        lightData.shape0 = 0.0f;
+#endif
+                        lightData.shape1 = 0.0f;
+                        lightData.type = UnityEngine.Experimental.GlobalIllumination.LightType.Point;
+                        lightData.falloff = true ? FalloffType.InverseSquared : FalloffType.InverseSquaredNoRangeAttenuation;
```
phi-lira: Also remove the if (true) here.

com.unity.render-pipelines.lightweight/LWRP/LightweightPipeline.cs
```diff
@@ -359,5 +365,127 @@ void SetupPerCameraShaderConstants(CameraData cameraData)
...
+            for (int i = 0; i < requests.Length; i++)
+            {
+                Light light = requests[i];
+#if UNITY_EDITOR
+                if (LightmapperUtils.Extract(light.lightmapBakeType) == LightMode.Realtime)
+                    lightData.InitNoBake(light.GetInstanceID());
+                else
+                    LightDataGIExtract(light, ref lightData);
+#else
+            ld.InitNoBake(l.GetInstanceID());
```
phi-lira: Isn't Lightmapping class Editor only? If so the delegates should be guarded by UNITY_EDITOR. Please test if you can build a player just to see there are no issues with Editor classes.

com.unity.render-pipelines.lightweight/LWRP/Passes/SetupLightweightConstanstPass.cs
```diff
@@ -203,7 +197,7 @@ void SetupMainLightConstants(CommandBuffer cmd, ref LightData lightData)
                 }
             }
 
-            cmd.SetGlobalVector(LightConstantBuffer._MainLightPosition, new Vector4(lightPos.x, lightPos.y, lightPos.z, lightDistanceAttenuation.w));
+            cmd.SetGlobalVector(LightConstantBuffer._MainLightPosition, new Vector4(lightPos.x, lightPos.y, lightPos.z, lightColor.w));
```
phi-lira: So I think we can use the lightPos.w here. Taking a look at the shader it's only used in the GetLightDirectionAndAttenuation to tell if it's a puntual or directional light but now since we have light classification that function is never called for a Directional Light.

com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/Lighting.hlsl
```diff
@@ -40,10 +40,9 @@
 struct LightInput
 {
     float4  position;
-    half3   color;
-    half4   distanceAttenuation;
+    half4   color;
```
phi-lira: if you use the lightPos.w remember to change this back to half3.

com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/Lighting.hlsl
```diff
@@ -128,7 +125,7 @@ half4 GetMainLightDirectionAndAttenuation(LightInput lightInput, float3 position
     half4 directionAndAttenuation = GetLightDirectionAndAttenuation(lightInput, positionWS);
```
phi-lira: We can remove this function.

PR #1857, opened 2018-08-07
Hdrp/decals/v2 1/normal buffer

Paulm-Unity:


com.unity.render-pipelines.high-definition/HDRP/Material/GBufferManager.cs
```diff
@@ -47,7 +47,7 @@ public override void CreateBuffers()
 
             for (int gbufferIndex = 0; gbufferIndex < m_BufferCount; ++gbufferIndex)
             {
-                m_RTs[gbufferIndex] = RTHandles.Alloc(Vector2.one, colorFormat: rtFormat[gbufferIndex], sRGB: sRGBFlags[gbufferIndex], filterMode: FilterMode.Point, name: string.Format("GBuffer{0}", gbufferIndex));
+                m_RTs[gbufferIndex] = RTHandles.Alloc(Vector2.one, colorFormat: rtFormat[gbufferIndex], sRGB: sRGBFlags[gbufferIndex], filterMode: FilterMode.Point, name: string.Format("GBuffer{0}", gbufferIndex), enableRandomWrite: gbufferIndex == 1); // normal buffer is used as RWTexture to composite decals in forward
```
sebastienlagarde: To avoid any break in case of change in lit.cs, introduce a tabl (like we have done for sRGBFlags[gbufferIndex]) for enableWrite. Thanks

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalNormalBuffer.shader
```diff
@@ -0,0 +1,77 @@
...
+        float4 FragNearest(Varyings input) : SV_Target
+        {
+            float4 DBufferNormal = LOAD_TEXTURE2D(_DBufferTexture1, input.texcoord * _ScreenSize.xy) * float4(2.0f, 2.0f, 2.0f, 1.0f) - float4(1.0f, 1.0f, 1.0f, 0.0f);
+            float4 GBufferNormal = _NormalBuffer[input.texcoord * _ScreenSize.xy];
+            NormalData normalData;
+            DecodeFromNormalBuffer(GBufferNormal, uint2(0, 0), normalData);
+            normalData.normalWS.xyz = normalize(normalData.normalWS.xyz * DBufferNormal.w + DBufferNormal.xyz);
+            EncodeIntoNormalBuffer(normalData, uint2(0, 0), GBufferNormal);
+            _NormalBuffer[input.texcoord * _ScreenSize.xy] = GBufferNormal;
+            return GBufferNormal;
```
sebastienlagarde: Why do we return GBufferNormal?

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.shader
```diff
@@ -440,6 +440,13 @@ Shader "HDRenderPipeline/Lit"
 
             ZWrite On
 
+            // decals rely on this to mask out the pixels that need to be updated in the normal buffer (forward)
+            Stencil
+            {
+                Comp Always
```
sebastienlagarde: Caution, we have 4 lit.shader (+ StackLit) + Cloth now...

sebastienlagarde: Big warning here, bit in stencil have a meaning, as it seems you don't clear the stencil after the DBuffer pass you should have your dedicated slot. See HDRednerPipeline.cs

        [Flags]
        public enum StencilBitMask
        {
            Clear           = 0,             // 0x0
            LightingMask    = 7,             // 0x7  - 3 bit
            ObjectVelocity  = 128,           // 0x80 - 1 bit
            All             = 255            // 0xFF - 8 bit
        }

sebastienlagarde: Note: Depth prepass in case of DBuffer is always perform, either in deferred or forward. So the code is a bit useless, you could only test decals. Ideally what we would like is to have a different pass (one for WRITE_BORMAL_BUFFER and one for the regular case so we can control the stencil. Sadly this mean adding an extra pass at compile time + extra management code. 

I have no good idea for now, but I don't like current one :), Note that we don't clear the normal buffer so we have no way to know if we are clean or not. Maybe we should clear just this one to 0 and thus in the applyDecal toNormal buffer we test for normal if is it 0. The question is, is it worth it ?

Anyway, let's cleanup this a bit if we don't find solution, just tag stencil on decal write and fill normal buffer for these bit, don't rely anything on depth buffer unless you find a way to detect that we are the depth only pass with WRITE_NORMAL_BUFFER enable (Will think about it tomrrow).

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalNormalBuffer.shader
Paulm-Unity: Since we do not need to write to actual render target in output merger, I was afraid the compiler would optimize code out if it is not returned, but to be honest I have not tested what happens.

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.shader
Paulm-Unity: Yes this is exactly what I was thinking too, i had a comment there in the shader about that. So we reserve a bit for decals 
[Flags]
    public enum StencilBitMask
    {
        Clear           = 0,             // 0x0
        LightingMask    = 7,             // 0x7  - 3 bit
        Decals = 8, // 0x8 - 1 bit
        ObjectVelocity  = 128,           // 0x80 - 1 bit
        All             = 255            // 0xFF - 8 bit
    }

And just set it during DBuffer then test for it during normal buffer compositing.

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalNormalBuffer.shader
sebastienlagarde: I don't think the compiler will optimize such a thing, it is like saying that when doing alpha test only for shadow and returning 0 the compiler remove all the testing code :)

Paulm-Unity:
I don't understand what you mean by extra pass in deferred, the normal buffer compositing is only done in forward:

 if ((hdCamera.frameSettings.enableForwardRenderingOnly) && (!hdCamera.frameSettings.enableMSAA)) // MSAA not supported
                    {
                        using (new ProfilingSample(cmd, "DBuffer Normal (forward)", CustomSamplerId.DBufferNormal.GetSampler()))


But in any case, you want me to profile that pass?

sebastienlagarde:
Ah I see.
So when you are in deferred you can still have forward material (Stacklit for example is forward while you are in deferred), so we need to support this normal buffer modification in deferred too.

Thus why I am talking about measuring the extra cost (but would be good to know even in forward : ))

Paulm-Unity:
Hmm so looking at NormalBufferManager.cs, normal buffer is basically GBuffer normal target. Which in deferred will get correctly updated by decals in the lit shader, because lit shader reads DBuffer (LitData.hlsl  AddDecalContribution(..)), so why do we need to do an extra pass on it?

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalNormalBuffer.shader
```diff
@@ -0,0 +1,83 @@
...
+    HLSLINCLUDE
+
+        #pragma target 4.5
+        #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
+        #include "CoreRP/ShaderLibrary/Common.hlsl"
+        #include "../../ShaderVariables.hlsl"
+        #include "Decal.hlsl"
+        #include "../NormalBuffer.hlsl"
+
+        TEXTURE2D(_DBufferTexture1);
```
sebastienlagarde: Yo Fix in next PR: you shouldn't use this but use DECODE_FROM_DBUFFER(), like we do for GBuffer. You don't have to know the layout of the DBuffer here.

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -136,6 +137,8 @@ public enum StencilBitMask
         {
             Clear           = 0,             // 0x0
             LightingMask    = 7,             // 0x7  - 3 bit
+            Decals          = 8,             // 0x8  - 1 bit
+            StackLit        = 16,            // 0x10  - 1 bit
```
sebastienlagarde: Fix in next PR, as explain call this DecalsForwardOutputNormalBuffer

sebastienlagarde:
Launched katana, 
will merge this PR (as I will go to siggraph next week then two weeks in vacation. Please contact @JulienIgnace-Unity if you have question)

please address the Two issue in next PR:
Stacklit name in the stencil mask
and 

Fix in next PR: you shouldn't use this but use DECODE_FROM_DBUFFER(), like we do for GBuffer. You don't have to know the layout of the DBuffer here.

Thanks!

sebastienlagarde:
The PR have failed katana test editor, you will need to look at it. Sorry I will not be able to help for the next 3 weeks, please contact @JulienIgnace-Unity or ask @stramit after siggraph is your are unable to solve your problem.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fdecals%2Fv2_1%2Fnormal_buffer&automation-tools_branch=add-platform-filter&unity_branch=trunk

Paulm-Unity:
Passed Katana

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fdecals%2Fv2_1%2Fnormal_buffer&unity_branch=trunk&automation-tools_branch=master

It was the enable multiple edit attribute that was failing player build since it is editor only.

PR #1861, opened 2018-08-08
Debug Light Volumes

anisunity:
Implement an option to display the light volumes of Point, Spot, Line Rectangular Lights and Reflection probes
Add a singleton that generates basic Meshes to avoid regenerating them at every frame in debug mode

### DIFF
--- a/com.unity.render-pipelines.core/CoreRP/Debugging/DebugShapes.cs
+++ b/com.unity.render-pipelines.core/CoreRP/Debugging/DebugShapes.cs
@@ -3,7 +3,7 @@ namespace UnityEngine.Experimental.Rendering
     public partial class DebugShapes
     {
         // Singleton
-        private static DebugShapes s_Instance = null;
+        static DebugShapes s_Instance = null;
 
         static public DebugShapes instance
         {
@@ -18,13 +18,13 @@ namespace UnityEngine.Experimental.Rendering
             }
         }
         
-        private Mesh m_sphereMesh = null;
-        private Mesh m_boxMesh = null;
-        private Mesh m_coneMesh = null;
-        private Mesh m_pyramidMesh = null;
+        Mesh m_sphereMesh = null;
+        Mesh m_boxMesh = null;
+        Mesh m_coneMesh = null;
+        Mesh m_pyramidMesh = null;
 
         // This code has been grabbed from http://wiki.unity3d.com/index.php/ProceduralPrimitives
-        private void BuildSphere(ref Mesh outputMesh, float radius, uint longSubdiv, uint latSubdiv)
+        void BuildSphere(ref Mesh outputMesh, float radius, uint longSubdiv, uint latSubdiv)
         {
             // Make sure it is empty before pushing anything to it
             outputMesh.Clear();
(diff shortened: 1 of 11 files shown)


com.unity.render-pipelines.high-definition/HDRP/Debug/DebugLightVolume.shader
```diff
@@ -0,0 +1,59 @@
...
+
+        ZWrite Off
+        Blend SrcAlpha OneMinusSrcAlpha
+
+        Pass
+        {
+            HLSLPROGRAM
+            #pragma vertex vert
+            #pragma fragment frag
+            #pragma enable_d3d11_debug_symbols
```
JulienIgnace-Unity: We should probably remove the debug pragma (even though that's a debug shader)

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2732,6 +2734,107 @@ public void RenderDebugOverlay(HDCamera hdCamera, CommandBuffer cmd, DebugDispla
                     HDUtils.NextOverlayCoord(ref x, ref y, overlaySize, overlaySize, hdCamera.actualWidth);
                 }
             }
+
+#if UNITY_EDITOR
```
JulienIgnace-Unity: Why only Editor? That's probably something that can be used at runtime as well.

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2732,6 +2734,107 @@ public void RenderDebugOverlay(HDCamera hdCamera, CommandBuffer cmd, DebugDispla
...
+            if (lightingDebug.displayLightVolumes)
+            {
+                // First of all let's do the regions for the light sources (we only support Poncutal and Area)
+                int numLights = cullResults.visibleLights.Count;
+                for (int lightIdx = 0; lightIdx < numLights; ++lightIdx)
+                {
+                    // Let's build the light's bounding sphere matrix
+                    Light currentLegacyLight = cullResults.visibleLights[lightIdx].light;
```
JulienIgnace-Unity: Needs to check visibleLight.light for null and early out. This can happen for particle lights.

com.unity.render-pipelines.core/CoreRP/Debugging/DebugShapes.cs
```diff
@@ -0,0 +1,506 @@
+namespace UnityEngine.Experimental.Rendering
+{
+    public partial class DebugShapes
+    {
+        // Singleton
+        private static DebugShapes s_Instance = null;
```
sebastienlagarde: Note, in unity c# convention we nevert write the private keyword

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.cs
sebastienlagarde: yes, this debug mode is for runtime too

anisunity:
@JulienIgnace-Unity @sebastienlagarde corrections done

PR #1865, opened 2018-08-08
Uielements label field refactor

jfparadis18:
This change is required for a major refactor done in the UIElements side where labels are added to the fields and old fields are all renamed *Input.

This will only work with this specific PR that is intended to be in 19.1
See trunk PR : https://ono.unity3d.com/unity/unity/pull-request/71970/_/editor/uielements/controls/input-label

Note : the code mentioned in the above PR for trunk will work only after the branch for 19.1. Meanwhile, if you take this Unity editor branch to test this package, you need to slightly change the code from this ShaderGraph PR from #if UNITY_2019_1_OR_NEWER to #if !UNITY_2019_1_OR_NEWER, where ever it is used.
Notice the "!" in front of the UNITY_2019_1_OR_NEWER


com.unity.shadergraph/Editor/Drawing/Blackboard/BlackboardFieldPropertyView.cs
```diff
@@ -10,6 +10,26 @@
 #if UNITY_2018_3_OR_NEWER
 using ContextualMenu = UnityEngine.Experimental.UIElements.DropdownMenu;
 #endif
+#if UNITY_2019_1_OR_NEWER
+using TextInputField = UnityEngine.Experimental.UIElements.TextInput;
```
pbbastian: Why alias it to a different name? It would be great if we could just use the new name, and then have conditional qualified usings for older versions. Then we can eventually just remove them once we don't need to support older versions anymore.

jfparadis18: I will modify the code to do so... You are right and this will be better !
Thank you.

jfparadis18:
I added changes into this PR to resolve code issues noticed by earlier comments.
Note that the PopupInput / PopupField cannot be the same as the others because of the template on the type.

Yelmond:
@pbbastian When you have a chance can you look over the new commits addressing your comments as per: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1865#issuecomment-411824059

jfparadis18:
Just a side note about that PR, I would like to leave it open and make sure it is reviewed, but **do not merge it** ...

I will know next week if we will do the changes like this or with a slightly different approach which will not require these changes...

Thank you !

com.unity.shadergraph/Editor/Drawing/Controls/ChannelEnumControl.cs
```diff
@@ -6,6 +6,12 @@
 using UnityEngine.Experimental.UIElements;
 using UnityEditor.Experimental.UIElements;
 
+#if UNITY_2019_1_OR_NEWER
+using PopupInput = UnityEditor.Experimental.UIElements.PopupInput<string>;
```
pbbastian: Can we change this to `PopupInputString`? Seems like it might get a bit confusing otherwise

pbbastian:
Closing this PR as the changes will instead be arriving in a bigger batch later. @jfparadis18 will delete the branch later, as he still needs the changes available.

PR #1866, opened 2018-08-08
Fabric Shader modifications

anisunity:
Correcting couple bugs in the GetSurfaceData of the Fabric Shader
Removing a detail mask map and adding a fuzzdetail map
Adding a test scene for the fabric material

Katana is green


anisunity:
I will trigger a katana testbed asap

PR #1878, opened 2018-08-11
Improvements to decal projector editor gizmo.

pastasfuture:
In response to feedback in: https://fogbugz.unity3d.com/f/cases/1065972

- An arrow is rendered at the center of it's bounds pointing in the direction of the decal projection.

- When a user is interacting with one of the face handles, the corresponding plane is highlighted and outlined in it's axis color (R, G or B for X, Y, or Z) for greater visibility.

- New "Crop Decal with Gizmo" option - When set treats face handles as crop tool, pinning the projection irrespective of the bounds (via automatically computing necessary scale, bias values). This is to allow artists to quickly crop area of influence of decal, without disturbing the position of the projection itself. Note: In the original fogbugz feedback @JulienHeijmans suggested mapping this functionality to a hot key. I opted for the GUI toggle approach, as I struggled to find a good hotkey for this toggle - ctrl and alt are already taken for uniform / symmetrical scale constrains on the gizmo handles. If anyone has any hotkey suggestions I'm happy to implement.

- New "Auto Center Transform" option - When set automatically centers the transform of the decal projector on the fly during manipulation. I added this feature as I found previously that it was easy to end up with decal projectors far off from their transform origin, after a number of projector manipulations. I made this a toggle (which defaults to ON) as I believe there are cases where artists would want to maintain offset from transform origin.

- Adds Undo / Redo support to decal projector gizmo modifications (was previously not hooked into the undo system).

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentHandle.cs
+++ b/com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentHandle.cs
@@ -1,0 +1,173 @@
+using System;
+using UnityEngine;
+
+namespace UnityEditor.IMGUI.Controls
+{
+    public class DecalProjectorComponentHandle : PrimitiveBoundsHandle
+    {
+        public DecalProjectorComponentHandle() : base()
+        {
+            midpointHandleDrawFunction = DrawHandleMidpoint;
+        }
+
+        public UnityEngine.Vector3 size { get { return GetSize(); } set { SetSize(value); } }
+
+        protected override void DrawWireframe()
+        {
+            Handles.DrawWireCube(center, size);
+            DrawArrowDownProjectionDirection();
+        }
+
+        protected void DrawArrowDownProjectionDirection()
+        {
+            int controlID = GUIUtility.GetControlID(GetHashCode(), FocusType.Passive);
+            Quaternion arrowRotation = Quaternion.LookRotation(Vector3.down, Vector3.right);
+            float arrowSize = size.y * 0.25f;
+            Handles.ArrowHandleCap(controlID, center, arrowRotation, arrowSize, EventType.Repaint);
+        }
+
+        // Could use a static readonly LUT, but this would require syncing with order with enum HandleDirection.
+        protected static Color ColorFromHandleDirection(HandleDirection handleDirection)
+        {
+            switch (handleDirection)
+            {
+                case HandleDirection.PositiveX:
+                case HandleDirection.NegativeX:
+                    return Handles.xAxisColor;
+
+                case HandleDirection.PositiveY:
+                case HandleDirection.NegativeY:
+                    return Handles.yAxisColor;
+
+                case HandleDirection.PositiveZ:
+                case HandleDirection.NegativeZ:
+                    return Handles.zAxisColor;
+
+                default:
+                    throw new ArgumentOutOfRangeException("handleDirection", "Must be PositiveX, NegativeX, PositiveY, NegativeY, PositiveZ, or NegativeZ");
+            }
+        }
+
+        protected static void PlaneVerticesFromHandleDirection(ref Vector3[] outVertices, HandleDirection handleDirection, Vector3 boundsSize, Vector3 boundsCenter)
+        {
+            Vector3 boundsMin = boundsSize * -0.5f + boundsCenter;
+            Vector3 boundsMax = boundsSize * 0.5f + boundsCenter;
+
+            switch (handleDirection)
+            {
+                case HandleDirection.PositiveX:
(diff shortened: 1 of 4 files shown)


JulienHeijmans:
Really great, trhnak you for that !
I think that maybe the Auto Center could be active all the time, I have trouble seeing when the user would like to have its pivot offset. The decal crop function will probably only(95%) be used to isolate part of a decal atlas texture.

(I used a similar function before, the pivot was always auto-centered, not a single time I would have wanted it not to auto center)

pastasfuture:
Thanks @JulienHeijmans I agree, auto centering should probably always be on. One valid OFF case I considered was rotating decals around a cylindrical object (ex: a pillar) where you want a specific offset radius. Though this might just be enough of an edge case that the user should be expected to create an explicit parent transform.

If that makes sense, I'll go ahead and make auto center always on (and remove the corresponding GUI).

JulienHeijmans:
Sounds good to me :)

sebastienlagarde:
For this review you should add Paul as reviewer as well. Change looks good to me. Do you have change to do in this PR or do you make them in another PR?

pastasfuture:
Thanks @sebastienlagarde I'll just go ahead and make the changes in this PR as they're fairly minor. Should have them in by the end of the day.

com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentHandle.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            PlaneVerticesFromHandleDirection(ref s_PlaneVertices, handleDirection, size, center);
+            Handles.DrawSolidRectangleWithOutline(s_PlaneVertices, planeColorFill, planeColorOutline);
+
+            Handles.color = handlesColorPrevious;
+        }
+
+        // Utility function for determining the handle direction (which face) we are currently rendering within DrawHandleMidpoint().
+        // Ideally, the base class PrimitiveBoundsHandle would expose the reverse lookup: HandleDirectionFromControlID().
+        // In lieu of an explicit way to handle this look up, we derive the handle direction from the handle rotation.
```
Paulm-Unity: Can we override:
protected virtual Bounds OnHandleChanged(HandleDirection handle, Bounds boundsOnClick, Bounds newBounds);
So we can store HandleDirection and avoid having to do this math here?

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalProjectorComponent.cs
```diff
@@ -18,6 +18,8 @@ public class DecalProjectorComponent : MonoBehaviour
         public Vector3 m_Size = new Vector3(1, 1, 1);
         private Material m_OldMaterial = null;
         private DecalSystem.DecalHandle m_Handle = null;
+        public bool m_IsTransformAutoCentered = true;
```
Paulm-Unity: When auto center is on, seems like Snap Settings are not working and the handle 'runs away' from the mouse cursor.

sebastienlagarde:
when problem from Paul are adressed, please launch a katana and if green @JulienIgnace-Unity  will merge it. Note: We will merge directly in master for this one. So please rebase the branch on master (can be done in webgithub after having merge lastest master (and be sure there is nothing from hdrp/staging. Or launch katana on staging and if green merge in master then merge this (want to avoid to have too much change on staging before merging to master)

com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentHandle.cs
pastasfuture: That was actually my first attempt implementing this, unfortunately I think OnHandleChanged() gets called after the handle is rendered, so the highlight color is one update cycle behind. The bigger problem though is that OnHandleChanged() only gets called on change, so there is no way to capture the exit state when the handle is no longer interacted with, so the handle would remain highlighted even when deselected.

com.unity.render-pipelines.high-definition/HDRP/Material/Decal/DecalProjectorComponent.cs
pastasfuture: I'll check that out, thanks

pastasfuture: Thanks for catching this Paul. I just fixed this issue. Because snapping is enforced in the local space of the gizmo, auto-centering was shifting the world space coordinates of the snap point.

Because this logic of this handle snapping is enforced in the unity runtime, I opted for a simple solution of side-stepping the problem: only perform auto-centering when the user is no longer interacting with the gizmo.

pastasfuture:
Closing this PR and replacing with PR rebased off of master.

PR #1879, opened 2018-08-13
Supporting Depth Prepass Normal Writing and Decal Support

anisunity:
Please review this PR! Thanks !

The only comment I have is that the 3 defines _AXF_BRDF_TYPE_SVBRDF _AXF_BRDF_TYPE_CAR_PAINT _AXF_BRDF_TYPE_BTF are mutually exclusive (right?) so we only need 2 of them to be defined for preprocessor macro's slots.

However, in the AxF.hlsl file I find couple places where it only handle 2/3 of the macros. Maybe I missing something?


Patapom:
> the 3 defines _AXF_BRDF_TYPE_SVBRDF _AXF_BRDF_TYPE_CAR_PAINT _AXF_BRDF_TYPE_BTF are mutually exclusive (right?) so we only need 2 of them to be defined for preprocessor macro's slots.

Right. I guess at the time I didn't have a clear understanding of how the shader variants were working (not sure I understand even now, clearly I needed them to work as an enum, not as flags).

> However, in the AxF.hlsl file I find couple places where it only handle 2/3 of the macros. Maybe I missing something?

Also right. That's because the **_AXF_BRDF_TYPE_BTF** type has been deemed deprecated (or at least seldom used by anyone) so it ended up not being implemented.
This results into the fact that the **_AXF_BRDF_TYPE_XXX** enum should even be made narrower to only support SVBRDF or CarPaint...

Patapom:
I approved your PR although you added a line to write the "specularLobe" as the perceptual roughness but 1) be aware that specularLobe is a float2 in case of anisotropic surfaces and 2) it's not perceptual but actual roughness (unless I'm mistaken)

anisunity:
I guess there is a RoughnessToPerceptualRoughness missing then

Patapom:
RoughnessToPerceptualRoughness() you mean?

anisunity:
Yes, i corrected my message!

com.unity.render-pipelines.high-definition/HDRP/Material/AxF/AxF.hlsl
```diff
@@ -72,12 +72,9 @@ NormalData ConvertSurfaceDataToNormalData(SurfaceData surfaceData)
 
     normalData.normalWS = surfaceData.normalWS;
 #if defined(_AXF_BRDF_TYPE_SVBRDF)
-    normalData.perceptualRoughness = surfaceData.specularLobe;
-#elif defined(_AXF_BRDF_TYPE_CAR_PAINT)
-    normalData.perceptualRoughness = 0.0;
+    normalData.perceptualRoughness = RoughnessToPerceptualSmoothness(surfaceData.specularLobe);
```
Patapom: Should be RoughnessToPerceptualRoughness

com.unity.render-pipelines.high-definition/HDRP/Material/AxF/AxF.hlsl
```diff
@@ -1911,9 +1881,7 @@ void PostEvaluateBSDF(LightLoopContext lightLoopContext,
     diffuseLighting = bsdfData.diffuseColor * lighting.direct.diffuse + builtinData.bakeDiffuseLighting;
     specularLighting = lighting.direct.specular + lighting.indirect.specularReflected;
 
-#if !defined(_AXF_BRDF_TYPE_SVBRDF) && !defined(_AXF_BRDF_TYPE_CAR_PAINT)
-    diffuseLighting = 10 * float3(1, 0.3, 0.01);  // @TODO!
-#endif
+    // diffuseLighting = 10 * float3(1, 0.3, 0.01);  // @TODO!
```
Patapom: You can remove this line completely

com.unity.render-pipelines.high-definition/HDRP/Material/AxF/AxF.hlsl
anisunity: yes indeed.

com.unity.render-pipelines.high-definition/HDRP/Material/AxF/AxF.hlsl
```diff
@@ -64,7 +66,18 @@ void GetBSDFDataDebug(uint paramId, BSDFData bsdfData, inout float3 result, inou
     }
 }
 
+NormalData ConvertSurfaceDataToNormalData(SurfaceData surfaceData)
+{
+    NormalData normalData;
 
+    normalData.normalWS = surfaceData.normalWS;
+#if defined(_AXF_BRDF_TYPE_SVBRDF)
+    normalData.perceptualRoughness = RoughnessToPerceptualSmoothness(surfaceData.specularLobe);
```
Patapom: Should be RoughnessToPerceptualRoughness() here

com.unity.render-pipelines.high-definition/HDRP/Material/AxF/AxF.hlsl
anisunity: Ok, will do.Can I remove all the debug code?

anisunity:
@sebastienlagarde I id indeed rename it because of the decals, so do i add a  #define inside the decalfunction?

Apologies if I misunderstood something. From what we discussed with @Patapom the axf shader will only support CarPaint and SVBRDF variants. Those two types are mutually exclusive right? So that becomes a similar case to the Silk/Wool and thus we can save one processing macro. What did I miss?

PR #1884, opened 2018-08-17
Fixing a bug in the specular term of the silk variant of the fabric shader

anisunity:
### Purpose of this PR
Fixing a bug in the specular term of the Fabric Shader (reported by QA)

---
### Release Notes
N/A


anisunity:
Just pushed an other corrections to the shading model. Waiting for @laurenth-unity 's input to validate

PR #1891, opened 2018-08-23
Disable front-to-back sorting of opaque objects on hardware with HSR

aleks01010101:
### Purpose of this PR
Disable front-to-back sorting of opaque objects on hardware that doesn't require it (HSR is present on the GPU). The user still has control over it using the Camera. opaqueSortMode enum, per camera and dynamically, so if someone experiences any perf troubles they can still force enable it.

---
### Release Notes
- Sorting of opaque objects now respects camera opaqueSortMode setting.
- Sorting of opaque objects disables front-to-back sorting flag when camera settings allow that and the GPU has hidden surface removal.

---
### Testing status
**Katana Tests**:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fno_front_to_back_sorting_of_opaques_on_gpu_with_hsr&unity_branch=trunk

**Manual Tests**: Checked if the sorting is set correctly based on camera settings and the SystemInfo.hasHiddenSurfaceRemovalOnGPU flag.

**Automated Tests**: none

Nope

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers


com.unity.render-pipelines.lightweight/LWRP/Passes/DepthOnlyPass.cs
```diff
@@ -59,7 +59,8 @@ public DepthOnlyPass()
                 context.ExecuteCommandBuffer(cmd);
                 cmd.Clear();
 
-                var drawSettings = CreateDrawRendererSettings(renderingData.cameraData.camera, SortFlags.CommonOpaque, RendererConfiguration.None, renderingData.supportsDynamicBatching);
+                var sortFlags = renderingData.cameraData.defaultOpaqueSortFlags;
```
phi-lira: Should we remove OptimizeStateChanges here? It should not be needed for depth prepass.

aleks01010101: The comments say "combination of: static batching, lightmaps, material sort key, geometry ID". Are you sure we don't need it?

phi-lira: Hum. Ok. so it seems we need that mainly due to geometry sorting and a new srpBatcherCompatible key. The rest should not matter for depth only and HSR GPUsthey are sorting to improve material batching and overdraw. 

I'm not sure if it makes sense to split those into two categories at some point. What do you think? It seems it might not be worth the trouble as most likely there won't be any noticeable performance diff in splitting these.

aleks01010101: Let's put this in the todo-list for some future where we don't have more important stuff to do :)

PR #1896, opened 2018-08-24
Update Shader Graph changelog format to match convention

pbbastian:
### Purpose of this PR
Shader Graph is not following the correct convention for change logs. This PR converts the current change log into the correct format. Note that with this PR we are going away from doing GIFs for the change log, at least for a while. It doesn't fit well into the new format, and I don't want to force people to write the same thing in 2 documents. That will be too error prone with backports and such.

---
### Release Notes
* Updated Shader Graph change log format to match the required convention.

---
### Testing status
This doesn't touch any code, so no testing is needed.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers
Something went wrong when people touched the old change log in their PRs, and it all got mushed into 1 version :( To deal with this I decided to put all of the items onto 3.4.0-preview rather than try to track down when the changes happened.


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -1,112 +1,45 @@
...
-
-### Rotate About Axis node
-
-![](.data/rotate_about_axis_node.png)
-
-With the Rotate About Axis node, you can rotate a 3D vector space around an axis. For the rotation, you can specify an amount of degrees or a radian value.
-
-### Bug fixes and minor changes
+# Changelog
+All notable changes to this package will be documented in this file.
```
Katstodian: Please change to:
All notable changes to this package are documented in this file.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -1,112 +1,45 @@
...
+- You can now edit the Reference name for a property. To do so, select the property and type a new name next to Reference. If you want to reset to the default name, right-click Reference, and select Reset reference.
+- In the expanded property window, you can now toggle whether the property is exposed.
+- You can now change the path of Shader Graphs and Sub Graphs. When you change the path of a Shader Graph, this modifies the location it has in the shader selection list. When you change the path of Sub Graph, it will have a different location in the node creation menu.
+- With this node, you can change graph output depending on the face sign of a given fragment. If the current fragment is part of a front face, the node returns true. For a back face, the node returns false. Note: This functionality requires that you have enabled **two sided** on the Master node.
```
Katstodian: Which node is this? 
It used to have a title - but now it's empy.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -1,112 +1,45 @@
...
+- Gradient functionality is now available via two new nodes. The Sample Gradient node samples a gradient given a Time parameter. You can define this gradient on the Gradient slot control view. The Gradient Asset node defines a gradient that can be sampled by multiple Sample Gradient nodes using different Time parameters.
```
Katstodian: Please change to this:
Gradient functionality now has the nodes Sample Gradient and Gradient Asset. The Sample Gradient node samples a gradient given a Time parameter. You can define this gradient on the Gradient slot control view. The Gradient Asset node defines a gradient that can be sampled by multiple Sample Gradient nodes using different Time parameters.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -1,112 +1,45 @@
...
+
+### Changed
+- The settings for master nodes now live in a small window that you can toggle on and off. Here, you can change various rendering settings for your shader.
+- There are two Normal Derive Nodes: `Normal From Height` and `Normal Reconstruct Z`.
+  `Normal From Height` uses Vector1 input to derive a normal map.
+  `Normal Reconstruct Z` uses the X and Y components in Vector2 input to derive the proper Z value for a normal map.
+- The Texture type default input now accepts render textures.
+- HD PBR subshader no longer duplicates surface description code into vertex shader.
+- If the current render pipeline is not compatible, master nodes now display an error badge.
+- The preview shader now only considers the current render pipeline. Because of this there is less code to compile, and therefore the preview shader will compile faster.
```
Katstodian: Please replace with this:
The preview shader now only considers the current render pipeline. Because of this there is less code to compile, so the preview shader compiles faster.

com.unity.shadergraph/CHANGELOG.md
pbbastian: We didn't have any gradient functionality before this though. The gradient functionality is the high level feature that has been added.

Katstodian: Ah! In that case:
Gradient functionality is now available via two new nodes: Sample Gradient and Gradient Asset. The Sample Gradient node samples a gradient given a Time parameter. You can define this gradient on the Gradient slot control view. The Gradient Asset node defines a gradient that can be sampled by multiple Sample Gradient nodes using different Time parameters.

PR #1898, opened 2018-08-24
Add vertex shader compatibility to subgraphs

Kink3d:
### Purpose of this PR
This PR removes the restriction of subgraph usage in vertex shader stage. The referenced graph now has an effective shader stage. This is calculated based on any vertex/fragment specific nodes in that subgraph and restricts the vert/frag usage of the subgraph in a regular graph. Otherwise the new restriction is that each instance of a subgraph node can only be used in a single stage. When an edge is connected from the subgraph to the master node this defines the effective stage of every slot on that node.

---
### Release Notes
Shouldnt this be specifiied in the changelog?

---
### Testing status
**Katana Tests**: Passed

**Manual Tests**: @zadi0648 @spaztronaut 

**Automated Tests**: No automated tests yet. This is being worked on by @zadi0648 and @spaztronaut 

Nope

---
### Overall Product Risks

---
### Comments to reviewers


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## [3.4.0]
 
+### Vertex shader Sub-graph support
+
+Shader Graph now supports connecting Sub-graph nodes to vertex shader slots. If any node in the referenced Sub-graph asset specifies a shader stage a Sub-graph node referencing it will be locked to that stage. Once an instance of a Sub-graph node is connected to a slot specifying a shader stage all slots on that instance will be locked to that stage.
```
Katstodian: Please replace with:
You can now connect Sub Graphs to vertex shader slots. If a node in the Sub Graph specifies a shader stage, that specific Sub Graph node is locked to that stage. When an instance of a Sub Graph node is connected to a slot that specifies a shader stage, all slots on that instance are locked to the stage.

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -264,16 +264,50 @@ public virtual void UpdateSlots()
             var subGraphOutputNode = outputNode;
```
pbbastian: Sorry to be nitpicky, but could we inline this one? Looks like it used to do a cast, but that's not needed anymore maybe.

com.unity.shadergraph/Editor/Data/SubGraph/SubGraphOutputNode.cs
```diff
@@ -66,6 +66,26 @@ public override PreviewMode previewMode
...
+        {
+            get
+            {
+                List<MaterialSlot> slots = new List<MaterialSlot>();
+                GetInputSlots(slots);
+
+                foreach(MaterialSlot slot in slots)
+                {
+                    ShaderStageCapability stage = NodeUtils.GetEffectiveShaderStageCapability(slot, true)
+                        & NodeUtils.GetEffectiveShaderStageCapability(slot, false);
```
pbbastian: I don't think you need the last call here. The output node will never have any output slots, so the test should always be going backwards.

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -264,16 +264,50 @@ public virtual void UpdateSlots()
...
+                foreach(MaterialSlot slot in slots)
+                    slot.stageCapability = outputStage;
+            }
+
+            ShaderStageCapability effectiveStage = ShaderStageCapability.All;
+
+            foreach(MaterialSlot slot in slots)
+            {
+                ShaderStageCapability stage = NodeUtils.GetEffectiveShaderStageCapability(slot, true)
+                    & NodeUtils.GetEffectiveShaderStageCapability(slot, false);
```
pbbastian: You shouldn't call this twice, but instead set `goingBackwards` based on whether it's an input or output slot. Otherwise you might have quite a lot of extra graph traversals, due to the unfortunate way that `GetEffectiveShaderStageCapability` works.

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -264,16 +264,50 @@ public virtual void UpdateSlots()
...
 
+        private void ValidateShaderStage()
+        {
+            List<MaterialSlot> slots = new List<MaterialSlot>();
+            GetOutputSlots(slots);
+
+            var subGraphOutputNode = outputNode;
+            if (outputNode != null)
+            {
+                var outputStage = ((SubGraphOutputNode)subGraphOutputNode).effectiveShaderStage;
```
pbbastian: Do we need to do this on _every_ validation? I think you could do it once when the slots are created initially. Ah wait, the get overridden later :/ This is ok for now, but the whole stage system really needs to be more centralized.

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
Kink3d: Correct, used to be a cast :P

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
Kink3d: Yup, as I was writing this it became clear we need to look at this code a lot later :P But it needs to stay like this for now

com.unity.shadergraph/Editor/Data/SubGraph/SubGraphOutputNode.cs
Kink3d: Fixed

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
Kink3d: Fixed

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
pbbastian: agree

PR #1902, opened 2018-08-27
Improvements to decal projector editor gizmo.

pastasfuture:
In response to feedback in: https://fogbugz.unity3d.com/f/cases/1065972

- An arrow is rendered at the center of it's bounds pointing in the direction of the decal projection.

- When a user is interacting with one of the face handles, the corresponding plane is highlighted and outlined in it's axis color (R, G or B for X, Y, or Z) for greater visibility.

- New "Crop Decal with Gizmo" option - When set treats face handles as crop tool, pinning the projection irrespective of the bounds (via automatically computing necessary scale, bias values). This is to allow artists to quickly crop area of influence of decal, without disturbing the position of the projection itself. Note: In the original fogbugz feedback @JulienHeijmans suggested mapping this functionality to a hot key. I opted for the GUI toggle approach, as I struggled to find a good hotkey for this toggle - ctrl and alt are already taken for uniform / symmetrical scale constrains on the gizmo handles. If anyone has any hotkey suggestions I'm happy to implement.

- Automatically centers the transform of the decal projector on the fly during manipulation. I added this feature as I found previously that it was easy to end up with decal projectors far off from their transform origin, after a number of projector manipulations. I made this a toggle (which defaults to ON) as I believe there are cases where artists would want to maintain offset from transform origin.

- Adds Undo / Redo support to decal projector gizmo modifications (was previously not hooked into the undo system).

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentEditor.cs
+++ b/com.unity.render-pipelines.high-definition/HDRP/Editor/Material/Decal/DecalProjectorComponentEditor.cs
@@ -19,10 +19,10 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
         private SerializedProperty m_UVScaleProperty;
         private SerializedProperty m_UVBiasProperty;
         private SerializedProperty m_AffectsTransparencyProperty;
-        private SerializedProperty m_Center;
         private SerializedProperty m_Size;
-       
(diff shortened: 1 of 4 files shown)


pastasfuture:
@JulienIgnace-Unity 

This is this PR, but rebased directly off of master, as requested by @sebastienlagarde 
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1878

Already ran and passed tests in Katana.

PR #1903, opened 2018-08-28
Hdrp/atmospheric scatter

Paulm-Unity:
### Purpose of this PR
Quick hack to fix atmospheric scattering on Ps4 release build.


Paulm-Unity:
Now that I had more time to look at it, I think a better fix would be to add another kernel to the compute shader that doesn't do the temporal and run it when the previous frame is not valid.

EvgeniiG:
Until we have working 3D texture clearing functionality, I propose the following (more simple) workaround:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/4bc2aaa01dda421a281b5493bb7d73f9b8caaedb

Would this be OK for you?

EvgeniiG:
Pushed the fix built upon yours: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/a5b7eac7a4802ba21ea6c023cb0a305896556cf3
Thanks!

PR #1904, opened 2018-08-28
Pyramid rework

Chman:
### Purpose of this PR
Fixes and optimizes the depth & color pyramids. Note that the previous color pyramid was incorrect, so this PR also indirectly improves the look of refraction and distortion.

Future work: move the depth map to an atlas instead of a pyramid and do it in-place to save the initial depth copy (which is slower than the pyramid itself...). It would also be nice to do the color pyramid in place to shave off more millisecond (0.2 on PS4) by removing the need to copy mip0 but the pyramid still needs to be in a separate buffer due to some passes writing to the color buffer while reading the pyramid.

---
### Testing status
Katana is [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=pyramid-rework&automation-tools_branch=add-platform-filter&unity_branch=trunk). Because this PR fixed a few visual artifacts some of the test images have been updated.


com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -298,14 +293,15 @@ void InitializeRenderTextures()
             m_SSSBufferManager.InitSSSBuffers(m_GbufferManager, m_Asset.renderPipelineSettings);
             m_NormalBufferManager.InitNormalBuffers(m_GbufferManager, m_Asset.renderPipelineSettings);
 
-            m_CameraColorBuffer = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: RenderTextureFormat.ARGBHalf, sRGB: false, enableRandomWrite: true, enableMSAA: true, name: "CameraColor");
+            m_CameraColorBuffer = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: RenderTextureFormat.ARGBHalf, sRGB: false, enableRandomWrite: true, enableMSAA: true, useMipMap: false, name: "CameraColor");
+            m_CameraColorBufferMipChain = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: RenderTextureFormat.ARGBHalf, sRGB: false, enableRandomWrite: true, enableMSAA: true, useMipMap: true, autoGenerateMips: false, name: "CameraColorBufferMipChain");
```
JulienIgnace-Unity: msaa is not needed here as transparent objects are going to be rendered post msaa resolve.

Chman: Fixed in latest commit.

PR #1907, opened 2018-08-28
Prototype for stereo support in compute and non-compute shaders

nerites:
### Purpose of this PR
- Prototype fix for stereo screen-space shadows 
- Prototype fix for stereo camera motion vectors (contains changeset from this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1933)
The completed fix for screen-space shadows will be slightly more involved so this seems like a good "save point".

---
### Release Notes
Please also review this PR as a representative of the compute shader strategy for stereo rendering: in C#, I added code to dispatch the screenspaceshadow.compute shader once per-eye, with the appropriate matrices attached. 

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp_xr&unity_branch=trunk

Enabled VR and ran with OpenVR SDK + HTC Vive. 

Nothing special- just wanted to make sure existing tests didn't break.

I can provide my project upon request.

---
### Overall Product Risks
Low

Low- higher impact changes are coming soon, which is why I wanted to get these changes in first.

---
### Comments to reviewers
Will wait for MSAA PR to complete to ensure these changes merge nicely (https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1916)


com.unity.render-pipelines.core/CoreRP/Common/XRGraphicsConfig.cs
```diff
@@ -25,7 +25,8 @@ public class XRGraphicsConfig
         public void SetConfig()
         { // If XR is enabled, sets XRSettings from our saved config
             if (!enabled)
-                Assert.IsFalse(enabled);
+                return;
+            Assert.IsTrue(enabled);
```
sebastienlagarde: Isn't this redudant? 
if (!enabled) you never reach the Assert, so what is the purpose of the assert? I will suggest to simply remove all the assert from the code here

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2531,11 +2531,30 @@ public void RenderScreenSpaceShadows(HDCamera hdCamera, RTHandleSystem.RTHandle
...
+                int numTilesX;
+                int numTilesY;
 
-                // TODO: Update for stereo
-                cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                if (hdCamera.camera.stereoEnabled)
+                {
+                    for (uint eye = 0; eye < 2; eye++)
+                    {
+                        numTilesX = ((hdCamera.actualWidth/2) + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
```
sebastienlagarde: you can put this two line (numTileX and TileY outside of the eye loop)

com.unity.render-pipelines.high-definition/HDRP/Material/NormalBuffer.hlsl
```diff
@@ -48,6 +48,13 @@ void DecodeFromNormalBuffer(uint2 positionSS, out NormalData normalData)
     DecodeFromNormalBuffer(normalBuffer, positionSS, normalData);
 }
 
+void DecodeFromNormalBuffer_Stereo(uint2 pixelCoord, uint2 positionSS, out NormalData normalData)
+{ // When rendering in single-pass stereo, positionSS != pixelCoord. Use pixelCoord instead to load from normal buffer.
+    float4 normalBuffer = LOAD_TEXTURE2D(_NormalBufferTexture0, pixelCoord);
```
sebastienlagarde: This feel wrong to me, we shouldn't have a dedicated function for VR here, we need to have a single way to sample a texure and it must work automatically with VR. I will discuss that with @JulienIgnace-Unity tomorrow to know what should be the best way to have a single function here.

sebastienlagarde: To give more explanation, if you require to do such a thing for normal buffer, it mean you need to do it for any buffer that we may sample inside a compute shader (like our Subsurface scattering), thus why I want to check with Julien for a proper solution here before we duplicate all our sampling functions

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1130,11 +1130,17 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                             using (new ProfilingSample(cmd, "Blit to final RT", CustomSamplerId.BlitToFinalRT.GetSampler()))
                             {
                                 // This Blit will flip the screen on anything other than openGL
-                                HDUtils.BlitCameraTexture(cmd, hdCamera, m_CameraColorBuffer, BuiltinRenderTextureType.CameraTarget);
+                                if (srcFrameSettings.enableStereo && (XRGraphicsConfig.eyeTextureDesc.vrUsage == VRTextureUsage.TwoEyes))
+                                {
+                                    cmd.BlitFullscreenTriangle(m_CameraColorBuffer, BuiltinRenderTextureType.CameraTarget); // If double-wide, only blit once (not once per-eye)
+                                } else
```
sebastienlagarde: formatting, else must be on new line

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipelineAsset.cs
```diff
@@ -7,7 +7,7 @@ public class HDRenderPipelineAsset : RenderPipelineAsset, ISerializationCallback
     {
         [HideInInspector]
         const int currentVersion = 1;
-
+
```
sebastienlagarde: Remove these space :)

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
```diff
@@ -25,6 +25,9 @@ float4  _PunctualLightPosition;
 float4  _ContactShadowParamsParameters;
 float4  _ContactShadowParamsParameters2;
 int     _SampleCount;
+float4x4 _Matrix_I_VP;
+float4x4 _Matrix_V;
```
sebastienlagarde: I am not sure to understand why we must declare these parameters
float4x4 _Matrix_I_VP;
float4x4 _Matrix_V;

Our macro is:
#define UNITY_MATRIX_V     _ViewMatrixStereo[unity_StereoEyeIndex]
So if you call 
cmd.SetComputeIntParam(screenSpaceShadowComputeShader, "unity_StereoEyeIndex", (int)eye);
instead of using  cmd.SetComputeIntParam(screenSpaceShadowComputeShader, HDShaderIDs._Eye, (int)eye);
This must work

or I miss something?

Then in the processing you can use unity_StereoEyeIndex instead of _Eye.

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
```diff
@@ -133,13 +137,13 @@ void DEFERRED_DIRECTIONAL(uint2 groupThreadId : SV_GroupThreadID, uint2 groupId
     if (depth == UNITY_RAW_FAR_CLIP_VALUE)
         return;
 
-    PositionInputs posInput = GetPositionInput(pixelCoord.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V, tileCoord);
+    PositionInputs posInput = GetPositionInput(pixelCoord.xy - (_Eye * float2(_ScreenSize.x, 0)), _ScreenSize.zw, depth, _Matrix_I_VP, _Matrix_V, tileCoord);
```
sebastienlagarde: Referring to the comment about havind done a stereo version of the DecodeNormalBuffer.
I am wondering if we can't get the correct positionSS, directly here.
If you don't apply  "- (_Eye * float2(_ScreenSize.x, 0))", don't we get the correct posInputSS from the transformation ?

Do you know if there is an offset bake inside the projection matrix? (i.e an offset for the screen to be at the correct location in the RT? ).  I suppose yes, so it should be good? Just a guess

com.unity.render-pipelines.high-definition/HDRP/Material/NormalBuffer.hlsl
sebastienlagarde: See my comment above about not applying Eye offset for GetPositionInput

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
sebastienlagarde: So I discussed with Julien, the projection matrix don't contain any offset this is perform by Viewport.

Here is something that we need to think about. We have a PR that I will merge before you that add support for MSAA. In this PR we have chose to move the cascade shadow map evaluation in the lightloop. This mean that cascade are no longer evaluated in this compute. 

My understand is that with this in mind, we don't need to split the rendering in two pass for the remaining contact shadow. So we should do only one pass like we do for deferred.compute (from what Julien told me there is a single pass for deferred lighting and not 2 right?)

So shadow contact should be "fix" automatically, now I think this mean we simply have move the issue with cascade shadow in light loop. Or not? Not sure :)

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
nerites: For compute shaders, we don't have certain shader keywords (notably UNITY_SINGLE_PASS_STEREO) defined. 

I think because of this, the compute shader is dispatched once per DispatchCompute call, while normal shaders are automatically dispatched twice (once per eye/viewport, with UNITY_MATRIX_* updated accordingly) when rendering in stereo. 

So UNITY_MATRIX_* will only have values for unity_StereoEyeIndex=0 in a compute shader. Likewise unity_StereoEyeIndex will only ever = 0, even when rendering the right eye.

nerites: The longterm fix is probably to add keyword support to compute shaders. In this case, VR support in HDRP would be bound to a newer version of Unity, which could be fine. What do you think we should do?

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
nerites: So for many of the other (non-compute) shaders I adjusted for stereo, I used a different version of GetPositionInputs that subtracts unity_StereoEyeIndex from the NDC when UNITY_SINGLE_PASS_STEREO is defined. 

This is because our texture is double-wide, resulting in NDC = 0-2 when SS coordinates are divided by _ScreenSize. Each eye should have NDC=0-1. 

If we're adding keyword support to compute shaders, we can do the same thing as we do for non-compute shaders.

com.unity.render-pipelines.high-definition/HDRP/Material/NormalBuffer.hlsl
nerites: We can probably modify LOAD_TEXTURE2D to use positionSS - unity_StereoEyeIndex*_ScreenSize. Since unity_StereoEyeIndex = 0 for non-stereo rendering, that should work for both stereo and non-stereo rendering.

nerites: The way positionSS is used, it seems like it's expected to be in the range 0-ScreenSize per-eye. Thus we can't use it to sample stereo textures (which should be sampled as a single double-wide texture).  Can we just use pixelCoord everywhere instead? It makes sense to me that SS = flat world coordinates, and pixel = location in texture.

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
sebastienlagarde: Ok got it. I remember now. As we need to add #pragma multi_compile _ UNITY_SINGLE_PASS_STEREO STEREO_INSTANCING_ON STEREO_MULTIVIEW_ON to compute shader and that they don't support multicompile it is a pain in the ass. Sorry have forgetten :P.

So we need to tackle it as we have a lot of compute shader in HDRP. It will be really painful to make workaround for all of them because of this (SSR, ScreenSpaceShadow, postprocessing, Lighting (in deferred :) ), SSAO, Volumetric lighting etc...)

Can you provide me more information. It is ok if we solely handle this UNITY_SINGLE_PASS_STEREO keyword in compute shader? or we need to handle them all (i.e STEREO_INSTANCING_ON STEREO_MULTIVIEW_ON) ? 
If compute shader only need this keyword and if the code is as simple as the change you have done in shadow contact, then we can afford to always perform it. So either we are in VR or nor we used the same shader and just update the eye index.

So the code above "should" work as expected if we add a 
#define UNITY_SINGLE_PASS_STEREO 
at the beginning of the file. Right ? And it shouldn't have other impact that defining the correct matrix I guess.

Thought?

com.unity.render-pipelines.high-definition/HDRP/Lighting/ScreenSpaceShadow.compute
nerites: Sorry, I didn't read your second message.

I think Julien is talking about the solution we used for the BOTD demo, where we evaluated pixelCoord.x to decide whether we are using left- or right-eye matrices to calculate positionWS from positionSS. 

In this case, we're saying that compute shaders that are worldspace-aware must be coded in a way that accounts for stereo rendering in order to behave properly in VR. Is this how we want to do it?

sebastienlagarde: I started double conversation, sorry :), confusing. Let's focus on the first one. 
Do you think we can declare 
#define UNITY_SINGLE_PASS_STEREO
at the beginning of the compute shader ? Is it sufficient for our need or are we suppose to support the other path ?

nerites: Having pulled the changes from the MSAA PR, I think we still need to add additional stereo logic for computing contact shadows, since ComputeContactShadow uses posInput.positionWS, which must be calculated using stereo matrices. Defining UNITY_SINGLE_PASS_STEREO isn't enough if we're only dispatching once, since there's nothing to update unity_StereoEyeIndex between the two eyes. 

If we want to only dispatch once, we can calculate "eye" on the fly by comparing pixelCoord.x against _ScreenSize.x/2. Do you know if that is better for performance, or dispatching twice with manually-bound eye value?

nerites:
I pulled in the changes from the MSAA support PR that just made it in. I've updated Deferred.shader and Deferred.compute (since that's where LightLoop is being invoked now), but there are still some artifacts in the right eye- most likely due to how tile offset is populated and used in those shaders. 

In this branch, if camera relative rendering is turned off, forward-only rendering looks fine in VR.

image

You can see that Bloom probably needs to be fixed for VR, but as this PR keeps growing I'm probably going to switch to another branch for that.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceShadow.compute
```diff
@@ -144,12 +148,12 @@ void DEFERRED_CONTACT_SHADOW_GENERIC(uint2 groupThreadId : SV_GroupThreadID, uin
     if (depth == UNITY_RAW_FAR_CLIP_VALUE)
         return;
 
-    PositionInputs posInput = GetPositionInput(pixelCoord.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V, tileCoord);
+    PositionInputs posInput = GetPositionInput(pixelCoord.xy - (_Eye * float2(_ScreenSize.x, 0)), _ScreenSize.zw, depth, _InvViewProjMatrixStereo[_Eye], _ViewMatrixStereo[_Eye], tileCoord);
```
sebastienlagarde: I still have this questoin, why not use our macro UNITY_MATRIX_I_VP here
instead of _InvViewProjMatrixStereo[_Eye]

you coudl do 

#if define UNITY_SINGLE_PASS_STEREO
#define  UNITY_MATRIX_I_VP _InvViewProjMatrixStereo[_Eye]


in shadervariables. Or there is an issue with this approach ? should be the same thing in .shader. Right?

nerites: When I tried this, unity_StereoEyeIndex wasn't being automatically toggled per-dispatch for compute shaders. 
But lemme try adding an #if SHADER_STAGE COMPUTE #define unity_StereoEyeIndex _Eye and see if that does anything.

sebastienlagarde: Have you try to setup unity_StereoEyeIndex with cmd.SetComputeVectorParam() ?
between the dispatch?

sebastienlagarde: #if SHADER_STAGE COMPUTE #define unity_StereoEyeIndex _Eye and see if that does anything.
Could be some kind of workaround. but will be weird that setting unity_StereoEyeIndex doesn't work

nerites: Looks like it works. Thanks, I'll kick off another katana run just in case.

PR #1908, opened 2018-08-29
Tests/hdrp/staging

RemyUnity:
All tests scene were cleaned up and optimized to be more readable end have smaller comparison images.


sebastienlagarde:
Katana is green, merging

PR #1909, opened 2018-08-29
Relocate NonCore files from core to HDRP

stramit:
### Purpose of this PR
Relocate NonCore files from core to HDRP. 

To LW to come out of experimental the core module also needs to come out of experimental. This means that we need to document all public API points and ensure test coverage. This is a substantial amount of work and will also lock the API. To make this work easier / not lock the api we are relocating much of the core library (only used by HD) into the HD package. This will improve import times for LW as well as simplify the release process. 

The idea is that as HD matures and features become documented then we can move them back into core. Examples for this is the Debug system, and many utility functions. 

---
### Release Notes
Move many utility functions from core to HD

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-All%20Tests/builds/535?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=cleanup-core-package&unity_branch=trunk

Green katana

---
### Overall Product Risks
**Technical Risk**: 
Low

**Halo Effect**: 
Low
---


sebastienlagarde:
Merge master to solve conflict. Now merging the PR.

PR #1911, opened 2018-08-29
Lighting Nodes

Kink3d:
### Purpose of this PR
This PR builds on the previous pipeline specific node PR by:
- Adding various lighting nodes for both pipelines
- Fixing nodes that failed to compile when using HDRP by making them pipeline specific

Changes:
#### Baked GI node (edit):
- Previously Light Probe node
- Works in both LWRP and HDRP
- Returns probes or lightmap in LWRP
- Returns full baked GI in HDRP

#### Reflection Probe node (edit):
- Made LWRP specific (returns black in HDRP)

#### Ambient node (edit):
- Made LWRP specific (returns black in HDRP)

#### Shadow Mask node (new):
- Works in HDRP (returns white in LWRP)

#### Main Light node (new)
- Works in LWRP
- Returns a "helpful constant values" for preview (returns same values in HDRP)

#### Additional Light node (new)
- Works in LWRP
- Returns a "helpful constant values" for preview (returns same values in HDRP)

#### Integer Input Material Slot
Adds a new input slot type that only allows integers. This is actually just a Vector 1 type slot but wont allow decimal input. This is also supported in CodeFunctionNode. This was added to support the AdditionalLight node and requires testing also.

---
### Release Notes
Changelog...

---
### Testing status
**Katana Tests**: Running...

**Manual Tests**: @zadi0648 / @spaztronaut 

**Automated Tests**: Nothing yet...

@zadi0648 / @spaztronaut 

---
### Overall Product Risks
Low - Only adds new nodes and fixes a few bugs. No graph changes

Low - Only adds new nodes and fixes a few bugs. No graph changes

---
### Comments to reviewers
See changes outlined at top...


Kink3d:
@stramit For what? The new nodes? Agreed but @zadi0648 and @spaztronaut are working on rebuilding our automated tests, so now is a tough time for us to be adding tests :/

sebastienlagarde:
I am really concern about some of the lighting node that you are adding here as "General".
I am agree that static diffuse lighting node make sense as general node (i.e lightmaps/Lightprobe/shadowmask).  However any other lighting node are very to LW RP: i.e MainLight, AdditionLight, ReflectionProbe, AmbientXXX.

We must move these node as specific to pipeline and they shouldn't appear when you work with HDRP.
I think we really miss the concept of RP specific node that only appear when you work with a given pipeline. Otherwise users will think their shader graph can work whereas it is totally unrelated.

what I propose is that we add specific RP node API. Each RP can add any of this specific node and if a user add it to the shader graph this produce an error message if you don't use the right pipeline (but IIRC this is already something plan or supported). And all these extra lighting node enter in this category: LW specific. That's way you don't need to provide a "default" value for adding light for other RP. It don't make sense that for any specific node we need to implement a default interface for the other RP.

Also have in mind that when LW will support deferred rendering, these nodes will not work anymore, you don't have access to lights during GBuffer pass in deferred, so even for LW you will need to handle that in a way or another, this can become tricky, returning black when deferred can be disturbing for users as it mean that if you are opaque you have no lighting, then if you are transparent you have lighing...

My true suggestion here will be to not expose any lighting node (outside of static GI), if we don't want to be in trouble.

com.unity.render-pipelines.lightweight/LWRP/Editor/ShaderGraph/Nodes/MainLightNode.cs
```diff
@@ -0,0 +1,47 @@
...
+            get { return "https://github.com/Unity-Technologies/ShaderGraph/wiki/Main-Light-Node"; }
+        }
+
+        protected override MethodInfo GetFunctionToConvert()
+        {
+            return GetType().GetMethod("Unity_MainLight", BindingFlags.Static | BindingFlags.NonPublic);
+        }
+
+        static string Unity_MainLight(
+            [Slot(0, Binding.None)] out Vector1 Attenuation,
```
phi-lira: You don't need attenuation for a directional light.

com.unity.render-pipelines.lightweight/LWRP/ShaderLibrary/ShaderGraphFunctions.hlsl
```diff
@@ -39,6 +46,37 @@ float3 shadergraph_LWSampleSceneColor(float2 uv)
     return 0;
 }
 
+void shadergraph_LWMainLight(out float attenutation, out float3 direction, out float3 color)
+{
+    Light light = GetMainLight();
+    attenutation = light.attenuation;
```
phi-lira: Same here. GetMainLight will initilize attenuation properly.

RyanJEC:
@phi-lira  Attenuation is needed for cel/toon Shading

pbbastian:
> 
> 
> @phi-lira Attenuation is needed for cel/toon Shading

Even for directional lights? What do you define the attenuation as?

RyanJEC:
I'm sorry I got confused. No not for directional lights. Please forgive.

pbbastian:
hehe, no worries :)

sebastienlagarde:
is this PR suppose to be replace by this one? https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2143
in this case can you close it?

Kink3d:
Yes I can close the PR. No-one delete this branch though :P

PR #1922, opened 2018-08-31
Color Picker Issue

anisunity:
### Purpose of this PR
Fixing a bug in the colorpicker

### DIFF
--- a/com.unity.render-pipelines.core/CoreRP/Debugging/MousePositionDebug.cs
+++ b/com.unity.render-pipelines.core/CoreRP/Debugging/MousePositionDebug.cs
@@ -132,17 +132,37 @@ namespace UnityEngine.Experimental.Rendering
 #endif
         }
 
-        public Vector2 GetMousePosition(float ScreenHeight)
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.core/CoreRP/Debugging/MousePositionDebug.cs
```diff
@@ -132,17 +132,29 @@ public void Cleanup()
...
             }
+            else
+            {
+                if(EditorApplication.isPlayingOrWillChangePlaymode)
+                {
+                    return Input.mousePosition;
+                }
+                else
+                {
+                    return new Vector2(-1.0f, -1.0f);
```
sebastienlagarde: is it really expected that we multiply x coordinate by -1 ?

anisunity: It is not a multiplication, let's talk

sebastienlagarde:
Katana green and I approved, don't know why I can update my review. MErging

PR #1923, opened 2018-09-03
Lw/lw pipeline asset inspector

martint-unity:
### Purpose of this PR
Changing the way the XR Properties were showing in the LWRP Asset Inspector

---
### Release Notes
Updates to LWRP inspector were XR Settings is now a foldout widget like the rest of the settings.

---
### Testing status
Kata Tests

---
### Overall Product Risks
Low

Low

---


com.unity.render-pipelines.lightweight/LWRP/Editor/LightweightPipelineAssetEditor.cs
```diff
@@ -2,6 +2,7 @@
 using UnityEditor;
 using UnityEditor.AnimatedValues;
 using UnityEditor.Experimental.Rendering;
+using UnityEditor.Rendering.PostProcessing;
```
phi-lira: remove postprocessing reference.

phi-lira:
@martint-unity we had a meeting today and completely redefined the UI for the settings here. This will make this PR obsolete.

phi-lira:
@martint-unity Can we delete this PR?

phi-lira:
deleting as not necessary anymore

PR #1924, opened 2018-09-03
Feature/hdrp/area light

Patapom:
### Purpose of this PR

- New disk/sphere area lights + viewport gizmos
- Switches between area light types in the UI are less surprising, energy perceptually stays the same (by a quick switch to luminance before changing the light shape)
- Refactored BSDF_Area() lighting code in Lit.hlsl to attempt a new code path with less branching
- Added code to support new LTC tables (only used in Lit.hsl at the moment, need to extend its use to other material types)
- Added a LTC Tables generator
- Added a small "horizon fade" to attempt to fix the backlighting issue with LTC tables (it helps but it's not enough, the fade range should depend on the size of the sphere somehow)

---
### Release Notes
- The refactored Area Light code should be extended even further by making the linear light compliant to a "form factor" computation like the other light types so we can eventually end up with a single code path.
- The disk light is the most expensive one, its form factor computation is very complicated (when compared to the one for the rectangle). We should attempt to find a simpler way to do that.
- The LTC tables generator can be accessed via the **Menu > Window > Render Pipeline > LTC Tables Generator**

---
### Testing status
**Katana Tests**:  (not sure what Katana is?)

**Manual Tests**: 
- Created a simple test scene
- Compared old and new LTC tables
- Compared new area light types and new LTC tables to ground truth

**Automated Tests**:  Nothing.

---
### Overall Product Risks
**Technical Risk**: Medium. Rect/Line area lights might be broken (although it's been checked they're still working). New refactored code *should* be equivalent or faster than former one. Not measured.

**Halo Effect**: Not sure what the halo effect is, or where it applies here.

---
### Comments to reviewers
Good luck! :D


sebastienlagarde:
adding @EvgeniiG for visibility and as he write the code for rect and line

EvgeniiG:
@Patapom I will have a look. Stay tuned. :-)

EvgeniiG:
image

I've noticed an artifact which I believe is new. Notice the wavy pattern on the reflection of the light.

EvgeniiG:
Looking at the VGPR numbers of Deferred_Indirect_ShadowMask_Fptl_Variant2 (LIGHTFEATUREFLAGS_SKY | LIGHTFEATUREFLAGS_DIRECTIONAL | LIGHTFEATUREFLAGS_AREA | MATERIALFEATUREFLAGS_LIT_STANDARD).

Old VGPR count: 84.
New VGPR count: 124.

EvgeniiG:
OK, merged the fix for an upstream regression which broke feature classification. Stats look a little better.

EvgeniiG:
Please adhere to the coding convention.
Don't use spaces inside brackets like "( this )", put { } on separate lines, etc.
In shaders, it's roughly the same as for C#: https://ono.unity3d.com/unity-extra/unity-meta/files/@/ReferenceSource/CSharp/Assets/CSharpReference.cs

EvgeniiG:
The license file for the fitter is missing: https://github.com/selfshadow/ltc_code/blob/master/LICENSE

com.unity.render-pipelines.core/CoreRP/ShaderLibrary/AreaLighting.hlsl
```diff
@@ -144,12 +164,22 @@ real PolygonIrradiance(real4x3 L)
...
+
     // Clamp invalid values to avoid visual artifacts.
-    real f2         = saturate(dot(F, F));
-    real sinSqSigma = min(sqrt(f2), 0.999);
-    real cosOmega   = clamp(F.z * rsqrt(f2), -1, 1);
+    real    f = length(F);
+    real    cosOmega   = clamp(F.z / f, -1, 1);
+    real    sqSinSigma = min(f, 0.999);
 
-    return DiffuseSphereLightIrradiance(sinSqSigma, cosOmega);
```
EvgeniiG: Why did you rename this? It is now more ambiguous.
sqSinSigma = Sq(sinSigma). You can literally type this in our shader and it will compile (Sq() is defined in Common.hlsl).
'sinSqSigma' is ambiguous. Is it sin^2(sigma)? Or sin(Sq(sigma))? Hard to know without context.

Patapom:
Okay so I noticed the wavy pattern at low roughness too. It seems to disappear for perceptual smoothness above 0.985 (so roughness below 0.000225). I believe it's a precision issue since we allow the roughness to be super small (1e-5 IIRC) in the new LTC tables computer, whereas before it was limited to 1e-3 IIRC.
The light shapes stay super sharp and don't exhibit the wave pattern if we don't go below 0.000225 in roughness, the simplest way would be to clamp to this value in the precomputer I suppose?

> image
> 
> I've noticed an artifact which I believe is new. Notice the wavy pattern on the reflection of the light.
> 
> image
> 
> Same for the line light.

Okay so I noticed the wavy pattern at low roughness too.
It seems to disappear for perceptual smoothness above 0.985 (so roughness below 0.000225). I believe it's a precision issue since we allow the roughness to be super small (1e-5 IIRC) in the new LTC tables computer, whereas before it was limited to 1e-3 IIRC.
The light shapes stay super sharp and don't exhibit the wave pattern if we don't go below 0.000225 in roughness, the simplest way would be to clamp to this value in the precomputer I suppose?

Patapom:
> Please adhere to the coding convention.
> Don't use spaces inside brackets like "( this )", put { } on separate lines, etc.
> In shaders, it's roughly the same as for C#: https://ono.unity3d.com/unity-extra/unity-meta/files/@/ReferenceSource/CSharp/Assets/CSharpReference.cs

I'm trying to!

PR #1925, opened 2018-09-04
Fix for Y-flip issue with tile debug view

keijiro:
### Purpose of this PR

Fix an issue where the tile/cluster debug view flips upside-down.

before

after

---

### Release Notes

- Fix an issue where the tile/cluster debug view flips upside-down.

---

### Testing status

**Manual Tests**: Tested on DX11 and Metal.

---

### Comments to reviewers

It's not clear when this issue was introduced.

Not tested on platforms where `UNITY_UV_STARTS_AT_TOP` is defined as 0 (as far as I know there is no such platform under HDRP support at the moment).

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Debug/DebugViewTiles.shader
+++ b/com.unity.render-pipelines.high-definition/HDRP/Debug/DebugViewTiles.shader
@@ -139,6 +139,12 @@ Shader "Hidden/HDRenderPipeline/DebugViewTiles"
 
             float4 Frag(Varyings input) : SV_Target
             {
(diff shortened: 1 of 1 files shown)


sebastienlagarde:
Thanks for the Fix! We don't merge it yet, just to remember us to try to apply your method of testing projection matrix at other place, as currently we have used another way and your way is better :)

EvgeniiG:
Thanks for the PR, Keijiro!
I integrated your fix into a larger PR (https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1945), which basically applies your technique in a bunch of places. It's a nice solution. :-)

PR #1927, opened 2018-09-04
Alpha Version of SSR

EvgeniiG:
First experimental release: expect quality, performance and stability problems.

Katana is green: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=ssr_rewrite&automation-tools_branch=add-platform-filter&unity_branch=trunk


com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDUtils.cs
```diff
@@ -363,5 +363,72 @@ public static string GetCorePath()
...
+            public Vector2Int   textureSize;
+            public int          mipLevelCount;
+            public Vector2Int[] mipLevelSizes;
+            public Vector2Int[] mipLevelOffsets;
+        }
+
+        // We pack all MIP levels into the top MIP level to avoid the Pow2 MIP chain restriction.
+        // We compute the required size iteratively.
+        // This function is NOT fast, but it is illustrative, and can be optimized later.
+        public static PackedMipChainInfo ComputePackedMipChainInfo(Vector2Int textureSize)
```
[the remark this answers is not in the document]
EvgeniiG: Done.

EvgeniiG:
Katana is green. @Chman @sebastienlagarde any feedback before I merge?

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -192,6 +193,24 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
         //  1. Screen Space Refraction / Reflection
         //  2. Environment Reflection / Refraction
         //  3. Sky Reflection / Refraction
+
+    #if !defined(_SURFACE_TYPE_TRANSPARENT) && defined(UNITY_MATERIAL_LIT)
```
sebastienlagarde: We must not have a test on defined(UNITY_MATERIAL_LIT) here. 
the Lightloop is independent of the the material. 

Why you don't use 

        if (featureFlags & LIGHTFEATUREFLAGS_SSREFLECTION)
        {
            IndirectLighting lighting = EvaluateBSDF_SSLighting(    context, V, posInput, preLightData, bsdfData, envLightData,
                                                                    GPUIMAGEBASEDLIGHTINGTYPE_REFLECTION, reflectionHierarchyWeight);
            AccumulateIndirectLighting(lighting, aggregateLighting);
        }

it was exactly the purpose of this call

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.hlsl
```diff
@@ -49,6 +49,8 @@ TEXTURE2D(_LightLayersTexture);
 TEXTURE2D(_ShadowMaskTexture); // Alias for shadow mask, so we don't need to know which gbuffer is used for shadow mask
 #endif
 
+TEXTURE2D(_SsrLightingTexture);
```
sebastienlagarde: This must be move in ShaderVariablesScreenSpaceLighting.hlsl

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -634,11 +673,8 @@ bool IsConsolePlatform()
 
         bool NeedDepthBufferCopy()
         {
-            // For now we consider all console to be able to read from a bound depth buffer.
-            // We always need it for SSR since depth textures do not support MIP maps.
-            return !IsConsolePlatform()
-                || m_Asset.GetFrameSettings().enableRoughRefraction
-                || m_Asset.GetFrameSettings().enableSSR;
+            // We always need a copy for SSR, so make one in all cases.
+            return true;
```
sebastienlagarde: Why change this test? if we don't enable SSR and Refraction, no need for copy. The test was fine.

sebastienlagarde: i.e we generate depth pyramid only if needed. Maybe the test should rather be rename: use depth pyramid?

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1000,18 +1036,19 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
...
-                    RenderDepthPyramid(hdCamera, cmd, FullScreenDebugMode.DepthPyramid);
-
                     // TODO: In the future we will render object velocity at the same time as depth prepass (we need C++ modification for this)
                     // Once the C++ change is here we will first render all object without motion vector then motion vector object
                     // We can't currently render object velocity after depth prepass because if there is no depth prepass we can have motion vector write that should have been rejected
                     RenderObjectsVelocity(m_CullResults, hdCamera, renderContext, cmd);
                     RenderCameraVelocity(m_CullResults, hdCamera, renderContext, cmd);
 
+                    // Needs the depth pyramid and the motion vectors, as well as the render of the previous frame.
+                    RenderSSR(hdCamera, cmd);
```
sebastienlagarde: In the future we need to move this a bit later to be done async during shadow

Chman:
Looks good to me.

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: See my comment which is a part of the PR:

// TODO: reading the depth buffer with a compute shader will cause it to decompress in place.
// On console, to preserve the depth test performance, we must NOT decompress the 'm_CameraDepthStencilBuffer' in place.
// We should call decompressDepthSurfaceToCopy() and decompress it to 'm_CameraDepthBufferMipChain'.

In short, we always want to make a copy, both on PC and console. Additionally, this allows to make the code simpler.
We can probably avoid building the pyramid if it's not necessary, but this is an optimization which doesn't necessarily have to be a part of this PR.

com.unity.render-pipelines.high-definition/HDRP/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: Or move shadows up. This is the earliest point at which SSR can be computed.

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.hlsl
EvgeniiG: I can do this, but this texture will most likely go away in a future PR.

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: The purpose of light feature flags is feature classification.
I don't want SSR to be a part of feature classification because this will increase the performance overhead of classification at almost zero benefit.
Additionally, SSR only supports the Lit shader right now, and since I don't know how StackLit works, I will not be adding StackLit support myself.

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/Lit.hlsl
EvgeniiG: This does not work because it is declared as a texture here, and I also declare it as an RW_texture in the SSR compute shader under the same name.
Personally, I also do not like the design where we put resources into headers. Constant buffers, and by some stretch shared samplers are tolerable, but buffers are not.

EvgeniiG: Note that including ShaderVariables.hlsl pulls in not just constant buffers and samplers (which I need), but also all the textures (which I don't).

The shader itself also becomes harder to read. Just look at this:

#ifdef DEBUG
    RW_TEXTURE2D(float4, _SsrDebugTexture);
#endif

#ifdef SSR_TRACE
    // TEXTURE2D(        _NormalBufferTexture0); // Declared in NormalBuffer.hlsl.
    // TEXTURE2D(        _DepthPyramidTexture);  // Declared in ShaderVariablesScreenSpaceLighting.hlsl
    RW_TEXTURE2D(float2, _SsrHitPointTexture);
#else
    // TEXTURE2D(        _NormalBufferTexture0); // Declared in NormalBuffer.hlsl.
    // TEXTURE2D(        _ColorPyramidTexture);  // Declared in ShaderVariablesScreenSpaceLighting.hlsl
       TEXTURE2D(        _VelocityTexture);
       TEXTURE2D(        _SsrHitPointTexture);
    RW_TEXTURE2D(float4, _SsrLightingTexture);
#endif

It's overly complicated for no reason.

com.unity.render-pipelines.high-definition/HDRP/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: In the future, I will also add a uniform branch to avoid evaluation for opaques. The transparent check needs to stay.

com.unity.render-pipelines.high-definition/HDRP/RenderPipelineResources/ScreenSpaceReflections.compute
```diff
@@ -0,0 +1,311 @@
...
+
+#include "CoreRP/ShaderLibrary/Common.hlsl"
+#include "HDRP/ShaderVariables.hlsl"
+
+#ifdef DEBUG_DISPLAY
+    #include "CoreRP/ShaderLibrary/Debug.hlsl"
+    #include "HDRP/Debug/DebugDisplay.hlsl"
+#endif
+
+#include "HDRP\Material\NormalBuffer.hlsl"
```
sebastienlagarde: be consistent: #include "HDRP/Material/NormalBuffer.hlsl"

EvgeniiG:
Addressed the feedback.

PR #1928, opened 2018-09-06
HDRP UI changes for XR + Minimal functional changeset

nerites:
### Purpose of this PR
My hdrp_xr branch is bogged down with the question of how compute shaders ought to be programmed for stereo. This branch contains only the minimal changeset for XR. Notably, I left in the VR override, such that HDRP still officially does not support VR. 
---
### Release Notes
+ UI integration of HDRP with XRGraphicsConfig
+ Used DrawFullScreen for final blit when stereo is enabled
+ Removed gameViewRenderMode and mirrorToGameView from XRGraphicsConfig
+ Execute current command buffers before start/stop multieye calls

---
### Testing status
Tests are currently running: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=xr_uichanges&unity_branch=trunk&automation-tools_branch=add-platform-filter

Loaded modified SRP in a test HRDP scene.

---
### Overall Product Risks
Low.

Low.

---
### Comments to reviewers
These changes are a subset of the changes in the hdrp_xr branch.


com.unity.render-pipelines.core/CoreRP/Common/XRGraphicsConfig.cs
```diff
@@ -19,24 +19,22 @@ public class XRGraphicsConfig
...
-        public bool showDeviceView;
-        public GameViewRenderMode gameViewRenderMode;
-        
+
         public void SetConfig()
         { // If XR is enabled, sets XRSettings from our saved config
             if (!enabled)
-                Assert.IsFalse(enabled);
+                return;
+            Assert.IsTrue(enabled);
```
JulienIgnace-Unity: Shouldn't the assert be before the test?  This way user would be warned before the method silently returns.

sebastienlagarde:
I am good with the PR, except the thing with the Assert as they are useless. See Julien comment. You will also need to solve conflict from master

com.unity.render-pipelines.core/CoreRP/Common/XRGraphicsConfig.cs
nerites: I will remove the asserts. If I place them before the return, then when VR is disabled, the assertions will fire, causing Katana tests to fail even if stereo is not intended to be enabled for those tests.

nerites:
Removed assertions and extraneous commits. Kicked off katana again just to check for typos.

PR #1932, opened 2018-09-06
Update Decal contribution of shader per material API

sebastienlagarde:
### Purpose of this PR

This PR change the way the decal contribution is done. Now it is the material that handle the application of decal part.

The AddDecalContribution have been rename GetDecalSurfaceData. It now only fill a DecalSurfaceData which is apply on the material side.

Any shader willing to work with decal need to add the following when building its surface data:

    if (_EnableDecals)
    {
        DecalSurfaceData decalSurfaceData = GetDecalSurfaceData(posInput, alpha);
        ApplyDecalToSurfaceData(decalSurfaceData, surfaceData);
    }

The ApplyDecalToSurfaceData is control by the material.

Note for Russel for shader graph: You will need to upgrade the template to deal with this way of handling decal.

Note for anis: You will need to update the AxF PR with this.

---
### Release Notes

This doesn't affect any content, no release note.

---
### Testing status
**Katana Tests**: Green

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=update-shader-decal-API&automation-tools_branch=add-platform-filter&unity_branch=trunk#
---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Material/Lit/LitDecalData.hlsl
+++ b/com.unity.render-pipelines.high-definition/HDRP/Material/Lit/LitDecalData.hlsl
@@ -1,0 +1,28 @@
+void ApplyDecalToSurfaceData(SurfaceData surfaceData, DecalSurfaceData decalSurfaceData)
+{
+    // using alpha compositing https://developer.nvidia.com/gpugems/GPUGems3/gpugems3_ch23.html
+    if (decalSurfaceData.HTileMask & DBUFFERHTILEBIT_DIFFUSE)
+    {
+        surfaceData.baseColor.xyz = surfaceData.baseColor.xyz * decalSurfaceData.baseColor.w + decalSurfaceData.baseColor.xyz;
+    }
+
+    if (decalSurfaceData.HTileMask & DBUFFERHTILEBIT_NORMAL)
+    {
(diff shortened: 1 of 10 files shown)


russoh:
OK, I will update the template code once this is merged. Thanks for the heads up.

com.unity.render-pipelines.high-definition/HDRP/Material/Lit/LitDecalData.hlsl
```diff
@@ -0,0 +1,28 @@
+void ApplyDecalToSurfaceData(DecalSurfaceData decalSurfaceData, inout SurfaceData surfaceData)
```
Paulm-Unity: Can we avoid duplicating this code somehow?

sebastienlagarde: I don't think so.
I am not of fan of having users to test HTileMask and check DECALS_4RT but as this is fully dependents on the material implementation we don't really have the choice. That's said, I could indeed remove DECALS_4RT test and put neutral value for decalSurfaceData.MAOSBlend so users at least don't need to bother with it. Will do that in a next PR.

sebastienlagarde:
Added @tim to be sure the history of this PR is fine regarding recent master problem. Should be ok as the faulty master branch was not involve here, but as I merge lastest master, when sanity check :)

PR #1933, opened 2018-09-06
Added support for stereo camera motion vectors

nerites:
### Purpose of this PR
I'm using this PR to review my method for updating non-compute shaders. This PR contains changes to enable single-pass stereo rendering of motion vectors, including changes to GetPositionInputs. Please give feedback on these changes, to set the template for future stereo-enabling non-compute-shader changes.

---
### Release Notes
+ Removed jitter from stereo (maybe not necessary?)
+ Added GetPositionInputs_Stereo
+ GetPositionInputs now invokes GetPositionInputs_Stereo with eye = 0
+ Added stereo matrix history
In order to fully enable TAA, need changes to postprocessing:
https://github.com/Unity-Technologies/PostProcessing/tree/pp_vr

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp_xr_taa&unity_branch=trunk&automation-tools_branch=add-platform-filter

- Used RenderDoc to verify per-eye NDC/WS and to visually inspect motion vectors
- Visually confirmed no additional artifacts with temporal anti-aliasing enabled

---
### Overall Product Risks
Low

High- GetPositionInputs is used in many places, and these changes will set the template for all future shader changes to enable stereo.

---
### Comments to reviewers
Looking for HDRP-specific feedback from either Sebastien or Julien, and VR coding style feedback from Brandon. (:


com.unity.render-pipelines.high-definition/HDRP/RenderPipelineResources/CameraMotionVectors.shader
```diff
@@ -32,13 +32,17 @@ Shader "Hidden/HDRenderPipeline/CameraMotionVectors"
...
             float depth = LOAD_TEXTURE2D(_CameraDepthTexture, input.positionCS.xy).x;
 
-            PositionInputs posInput = GetPositionInput(input.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V);
+            PositionInputs posInput = GetPositionInput_Stereo(input.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V, unity_StereoEyeIndex);
 
             float4 worldPos = float4(posInput.positionWS, 1.0);
             float4 prevPos = worldPos;
-
+#if UNITY_SINGLE_PASS_STEREO
+            float4 prevClipPos = mul(_PrevViewProjMatrixStereo[unity_StereoEyeIndex], prevPos); // Don't need to jitter for VR
```
sebastienlagarde: I don't understand the comment.
You say in comment "don't need to jitter for VR". But motion vector must not used jitterred matrix. This is what happen in the non VR code, we used unjittered matrix (prevViewProj is non jitter also.
Whereas you use MATRIX_VP which is jittered. 

I am missing something?

sebastienlagarde: I think here to make it cleaner, we could create macro for prev matrix and non jittered, so we could handle the eye case like the other shader.

i.e define a UNITY_MATRIX_PREV_VP and UNITY_MATRIX_UNJITTERED_VP

com.unity.render-pipelines.high-definition/HDRP/RenderPipelineResources/CameraMotionVectors.shader
```diff
@@ -48,6 +52,11 @@ Shader "Hidden/HDRenderPipeline/CameraMotionVectors"
 #if UNITY_UV_STARTS_AT_TOP
             velocity.y = -velocity.y;
 #endif
+
+#if UNITY_SINGLE_PASS_STEREO
+            // Since texture is double-wide, we need to halve again the velocity in x
+            velocity.x = velocity.x * 0.5;
```
sebastienlagarde: I think we could have a uniform here to perform this * 0.5. So the C# side control it and the compute code stay the same. Just a guess.

nerites: I'm happy to do it however you prefer, but what is the motivation for doing it this way?

sebastienlagarde:
Hey, we have discuss with Julien and here is what we think:
- For compute shader we always define UNITY_SINGLE_PASS_STEREO, so compute shader are always aware of VR.
- We implement your API about GetPositionInputs_Stereo. Compute shader always use this version, regular shader use GetPositionInputs.
- on C# side we always provide eye index and scale/bias (maybe only scale?) uniform for compute shader (Currently, eye should be what we already provide).
- The C# need to do the two dispatch if VR
- compute shader is implement with eye index as uniform and an offset

So we always pay the cost for VR code in compute shader, but as it is just a uniform eye indexaction, shit and scale it is negligeable and we don't need to add more variant or whatever, as we are already heavy on shader compilation.

Thought ?

com.unity.render-pipelines.high-definition/HDRP/RenderPipelineResources/CameraMotionVectors.shader
nerites: I have updated the shader per your recommendations. (:

nerites:
This all sounds good, just a question about GetPositionInputs: Is the positionSS input closer to pixelCoord or positionCS? Specifically, do we expect to be able to convert positionSS to positionWS? If we do, then positionSS should be 0-eyeWidth for both eyes, otherwise it should be 0-eyeWidth for the left eye, and eyeWidth-2*eyeWidth for the right eye (which is what it is right now).

nerites:
One more question about the programming model for compute shaders: Is this intended to be the final direction for compute shaders? 

Brandon and I were thinking of adding keyword support and doing per-eye dispatches of compute shaders on the native side. This would bring compute shaders much closer to vert/frag shaders in how they behave under the hood.

But maybe the C# control of stereo dispatches for compute shaders is more in line with the design goal of giving more control to SRP developers?

nerites:
> - For compute shader we always define UNITY_SINGLE_PASS_STEREO, so compute shader are always aware of VR.
> - We implement your API about GetPositionInputs_Stereo. Compute shader always use this version, regular shader use GetPositionInputs.

Regular shaders still need to pass the eye index into GetPositionInputs, since Common.hlsl does not have access to unity_StereoEyeIndex (which is defined in ShaderVariables.hlsl). For stereo-independent shaders, we can call GetPositionInputs, which will still call GetPositionInputs_Stereo and assume eye = 0. 

I kicked off Katana again on my most recent changes; please let me know what else needs to be done to merge in this changeset.

nerites:
Merging this PR with the compute shaders PR, since there seems to be some conceptual linkage anyways: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1943

com.unity.render-pipelines.high-definition/HDRP/ShaderVariablesMatrixDefsHDCamera.hlsl
```diff
@@ -13,6 +13,8 @@
 #define UNITY_MATRIX_I_P   _InvProjMatrixStereo[unity_StereoEyeIndex]
 #define UNITY_MATRIX_VP    _ViewProjMatrixStereo[unity_StereoEyeIndex]
 #define UNITY_MATRIX_I_VP  _InvViewProjMatrixStereo[unity_StereoEyeIndex]
+#define UNITY_MATRIX_UNJITTERED_VP _ViewProjMatrixStereo[unity_StereoEyeIndex] // Since VR doesn't need to add jitter, just use normal VP matrix
```
sebastienlagarde: Do you expect to not use TAA in the future for VR?
// Since VR doesn't need to add jitter, just use normal VP matrix

Just a question, nothing related to the current Pr :)

com.unity.render-pipelines.high-definition/HDRP/Camera/HDCamera.cs
```diff
@@ -201,9 +208,55 @@ public void Update(FrameSettings currentFrameSettings, PostProcessLayer postProc
...
+                        {
+                            prevViewProjMatrixStereo[eyeIndex] = GetViewProjMatrixStereo(eyeIndex); // Grabbing this before ConfigureStereoMatrices updates view/proj
+                        }
+
+                        isFirstFrame = false;
+                    }
+                }
+                isFirstFrame = true; // So that mono vars can still update when stereo active
+
+                screenWidth = XRGraphicsConfig.eyeTextureWidth;
```
JulienIgnace-Unity: Seems like those two temp variables are not used?

sebastienlagarde:
Waiting for Katana to finish then will merge

sebastienlagarde:
@nerites Katana is failing.
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp_xr_taa&automation-tools_branch=add-platform-filter&unity_branch=trunk#

Can you load the project HDRP_Test, and see if there is no error? as this is not the graphic test that are failing apparently

nerites:
040_UpgradeScene is the known failure on master (I didn't pull in the disable of the test on this branch yet). Unless you're seeing another failure?

Anyways, I can pull in master and rerun.

sebastienlagarde:
Looks like the last run of Katana have work, so it should be fine. I will merge :)

sebastienlagarde:
Thinking about it, this PR miss to define UNITY_SINGLE_PASS_STEREO for compute shader. So this will not work, right? (@nerites )

nerites:
I didn't see a compute shader for camera motion vectors. This PR was a prototype for modification of vert/frag shaders to support stereo.

sebastienlagarde:
Ah right, sorry :P. But then the define is still missing, you need to add the multi-compile right?

PR #1968, opened 2018-09-12
fix reflection within volume

RSlysz:
### Purpose of this PR
Fix influance point used to determine if volume apply on reflection.
For Planar, the influence point used is now the viewing camera position to remove artefact when surface is not at center of volume.

---
### Release Notes
Fix reflection in PlanarReflectionProbe used in combination of Volumes.

---
### Testing status
**Katana Tests**: 

**Manual Tests**: tested locally

**Automated Tests**: 

---
### Overall Product Risks
**Technical Risk**: Low
(only change a few lines)

**Halo Effect**: Medium
(change capture camera configuration for Reflections (used frequently in scene))

---


RSlysz:
Well, I will try to modify graphic test 2202_ReflectionProbes_Volume to cover this case too.

PR #1971, opened 2018-09-13
xr/multi view/shadow fix - (DO NOT MERGE INTO TRUNK)

brandon-unity3d:
### Purpose of this PR
Currently shadows do not render correctly in the right eye when using multi-view.  This PR fixes that issue.

---
### Release Notes
Shadows will now render correctly in both eyes when using multi-view.

---
### Testing status
Running katana tests...

I tested on a samsung s9 with gearvr + multiview and I tested on a pixel 2 with daydream + multi-view.
I also verified that single-pass double wide works correctly.

---
### Overall Product Risks
Low

Low


sebastienlagarde:
Caution: Files have move, there is no more LWRP folder

PR #1972, opened 2018-09-13
Hdrp/graphic test/fix reflection with volume

RSlysz:
### Purpose of this PR
This PR follows PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1968 adding a regression graphic test on the volume anchor point used in reflection's camera.

---
### Release Notes

---
### Testing status
**Katana Tests**: Green
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fgraphic_test%2Ffix_reflection_with_volume&automation-tools_branch=add-platform-filter&unity_branch=trunk#

**Manual Tests**: work launching locally.

**Automated Tests**: Added graphic test

In SRP's HDRP_Test, it could be found in scene 2210.

---
### Overall Product Risks
**Technical Risk**: ~~High~~
~~(For this to work, it actually require a hack. Need some investigation on the Volume component.)~~
None-Low

**Halo Effect**: None
(it is just a test for PR#1968)

---
### Comments to reviewers
~~@Chman You could use this as test project to debug Volume not working at start of scene.
Note that the hack let it work when launching the test but it did not work when loading the scene.
See test result to see what we should see at start/loading of scene.~~

~~The hack consist in flickering once the volume to re-register it.~~

@RemyUnity The test require a lot of things as it is a cross of Reflection and Volume. Do not hesitate to amend it. Also, I have not used Viewport Rect as this features is not finished for HDRP. This explain the high number of asset required and the strange scene set up..


Chman:
The issue with volumes in this scene is simply a priority issue, as all volumes have the same priority it's LIFO. Drop the SceneSettings volume priority down and it works.

RSlysz:
Should be good now and without hack. Thx @Chman

PR #1975, opened 2018-09-13
Package Shader Include Paths

stramit:
### Purpose of this PR
Update include paths to be package relative instead of using shader include path export as that code is flakey

---
### Release Notes
-Modified the shader paths to be package relative to package for future proofing. 

---
### Testing status

---
### Overall Product Risks
Medium, will need to bump major semvar


sebastienlagarde:
Ping @nerites @Paulm-Unity @fredpointzero @RemyUnity @RSlysz @anisunity @EvgeniiG @alelievr @julienf-unity @PaulDemeulenaere @iTris666  @brandon-unity @slunity  for awareness. This PR remove the "HDRP", "CoreRP", "LWRP" from the respective pipeline.

PR #1982, opened 2018-09-13
4-channel UV support in HDRP ShaderGraphs

cdxntchou:
(Doesn't change non-shadergraph interpolators)

### Purpose of this PR

Speedtree needs four channel UV support for their tree animation node.
We need speedtree shaderes in SRP, so shadergraph is an easy way to get it.

---
### Release Notes
This modifies FragInputs texcoords to be float4, but the .shader interpolators are still two channel.
Shadergraph builds it's own interpolators, and makes them float4 (though a future work item is to make 4-channel UV's something that is requested rather than the default..)

---
### Testing status

Load up my test project with four channel UV1, rebuild template project.

---
### Overall Product Risks
Low: worst case is shader compilation failures we would have to fix.

Medium at the shader level: modifies HDRP FragInputs definition of texcoords
Low outside of shaders

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
+++ b/com.unity.render-pipelines.high-definition/HDRP/Editor/ShaderGraph/HDPBRPass.template
@@ -244,7 +244,7 @@ $include("SharedCode.template.hlsl")
 
         // Builtin Data
         // For back lighting we use the oposite vertex normal 
-        InitBuiltinData(surfaceDescription.Alpha, surfaceData.normalWS, -fragInputs.worldToTangent[2], fragInputs.positionRWS, fragInputs.texCoord1, fragInputs.texCoord2, builtinData);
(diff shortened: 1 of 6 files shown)


cdxntchou:
Seems to work... fairly simple change at the shadergraph level.  Only problem is we need to pipe the float4 values through FragInputs which is an HDRP defined structure ... so I upgraded it to float4 and fixed the code that converts from interpolators and to debug output.   If we have extensive HDRP shader tests we should run that to see if there are any regressions.

cdxntchou:
image
4 channel UV works!

cdxntchou:
Made a new branch instead of trying to solve the conflicts merging this one.
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2090
Also fixed all of the other references to FragInputs.texCoord* in that PR.

cdxntchou:
Closing this PR.   Use sg/vec4hdrp2 instead!

PR #1987, opened 2018-09-14
HD Lit Master Node - HDRP compatible ShaderGraph master node.

russoh:
### Purpose of this PR
Implementation of Lit master node providing functionality exposed by HDRenderPipeline/Lit shader. Includes changes to nodes to support additional functionality, such as derivative maps, instances etc.

---
### Release Notes
The intention is to support all features of HDRP Lit, with the exception of things that can be implemented using graph functionality.
I.e. things deliberately excluded include:
-Vertex and pixel displacement
-Wind
-Detail maps

---
### Testing status

Lots of tests in this branch, created by zacharyd, comparing ShaderGraph to HDRenderPipeline/Lit for a large number of permutations and features:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/sg/lit-master-node-tests
Have also tested all features for parity during development.

zacharyd is in the process of setting up automated tests based on the above test branch

https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/sg/lit-master-node-tests
Consult zacharyd. There are numerous tests scenes in this branch.

---
### Overall Product Risks
Medium

**Halo Effect**:
Medium

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+        public const string ThicknessSlotName = "Thickness";
+        public const string DiffusionProfileSlotName = "DiffusionProfile";
+        public const string IridescenceMaskSlotName = "IridescenceMask";
+        public const string IridescenceThicknessSlotName = "IridescenceThickness";
+        public const string SpecularSlotName = "Specular";
+        public const string CoatMaskSlotName = "CoatMask";
+        public const string EmissionSlotName = "Emission";
+        public const string MetallicSlotName = "Metallic";
+        public const string SmoothnessSlotName = "Smoothness";
+        public const string OcclusionSlotName = "Occlusion";
```
sebastienlagarde: Will suggest to call it "Ambient Occlusion", as we need to add 'Specular Occlusion slot". 
Could be ok for "occlusion" but it become too generic as one is for indirect diffuse, the other for indirect specular

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+        public const int AnisotropySlotId = 20;
+        public const int SpecularAAScreenSpaceVarianceSlotId = 21;
+        public const int SpecularAAThresholdSlotId = 22;
+        public const int RefractionIndexSlotId = 23;
+        public const int RefractionColorSlotId = 24;
+        public const int RefractionDistanceSlotId = 25;
+        public const int DistortionSlotId = 26;
+        public const int DistortionBlurSlotId = 27;
+
+        public enum MaterialType
```
sebastienlagarde: For a next PR: For a next PR we will need to share this various enum between our inspector UI and shader graph UI. Maybe we will move them in another shared class.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+
+        // Don't support Multiply
+        public enum AlphaModeLit
+        {
+            Alpha,
+            PremultipliedAlpha,
+            Additive,
+        }
+
+        public enum ProjectionModelLit
```
sebastienlagarde: Note: I will remove ProjectionModelLit enum in few days. Just awareness for a next PR

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/DiffusionProfileNode.cs
```diff
@@ -0,0 +1,105 @@
+using System.Collections.Generic;
+using System.Globalization;
+using UnityEngine;
+using UnityEditor.Graphing;
+using UnityEditor.ShaderGraph.Drawing.Controls;
+using UnityEngine.Experimental.Rendering.HDPipeline;
+
+namespace UnityEditor.ShaderGraph
+{
+    [Title("Input", "PBR", "Diffusion Profile")]
```
Kink3d: Not a fan of this menu placement. Should be in an "HD Render Pipeline" specific section.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/DiffusionProfileNode.cs
```diff
@@ -0,0 +1,105 @@
...
+    public class DiffusionProfileNode : AbstractMaterialNode, IGeneratesBodyCode
```
Kink3d: In future we need to have this defined in Shader Graph package and have its behaviour overriden by an HDRP specific implementation. This will be needed for the capability that handles HDRP SSSSS. Just a note for @pbbastian re: new node API. Ignore this comment wrt this PR.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Slots/DiffusionProfileInputMaterialSlot.cs
```diff
@@ -0,0 +1,82 @@
...
+using UnityEditor.ShaderGraph.Drawing.Slots;
+using UnityEngine;
+using UnityEngine.Experimental.UIElements;
+using UnityEditor.ShaderGraph.Drawing.Controls;
+using UnityEngine.Experimental.Rendering.HDPipeline;
+
+namespace UnityEditor.ShaderGraph
+{
+    [Serializable]
+    public class DiffusionProfileInputMaterialSlot : Vector1MaterialSlot
```
Kink3d: Formatting in this class

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Slots/DiffusionProfileInputMaterialSlot.cs
```diff
@@ -0,0 +1,82 @@
...
+                    for (int i = 0; i < profiles.Length; i++)
+                    {
+                        m_DiffusionProfile.popupEntries.Add(profiles[i].name);
+                    }
+                    m_DiffusionProfile.selectedEntry = Mathf.Min(m_DiffusionProfile.selectedEntry, profiles.Length + 1);
+                }
+                else
+                {
+                    m_DiffusionProfile.selectedEntry = 0;
+                    // Need something equivalent, perhaps via implementation of a warning interface for the node.
```
Kink3d: If were not going to do anything with this we should remove the comment.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+using UnityEngine.Experimental.Rendering.HDPipeline;
+
+
+namespace UnityEditor.ShaderGraph
+{
+    [Serializable]
+    [Title("Master", "HDLit")]
+    public class HDLitMasterNode : MasterNode<IHDLitSubShader>, IMayRequirePosition, IMayRequireNormal, IMayRequireTangent
+    {
+        public const string AlbedoSlotName = "Albedo";
```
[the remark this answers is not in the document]
Kink3d: This wont work moving forward. Slots cant/shouldnt change their labels/behaviour based on other capabilities (on modular master nodes).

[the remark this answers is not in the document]
russoh: I'm inclined to agree with Matt here, since Albedo could end up being the composition of multiple colors/texture (e.g. added detail) from the graph.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+        public const string NormalSlotName = "Normal";
+        public const string BentNormalSlotName = "BentNormal";
+        public const string TangentSlotName = "Tangent";
+        public const string SubsurfaceMaskSlotName = "SubsurfaceMask";
+        public const string ThicknessSlotName = "Thickness";
+        public const string DiffusionProfileSlotName = "DiffusionProfile";
+        public const string IridescenceMaskSlotName = "IridescenceMask";
+        public const string IridescenceThicknessSlotName = "IridescenceThickness";
+        public const string SpecularSlotName = "Specular";
```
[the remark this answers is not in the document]
russoh: Will do.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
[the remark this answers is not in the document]
russoh: Also, not entirely sure I understand the comment regarding Specular Occlusion.  Are you saying that ShaderGraph should not follow the behavior of Lit.shader, and should provide a different slot explicitly for specular occlusion?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
russoh: Will do, Also, this comment sounds like it confirms my question on the previous comment about having another slot.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
russoh: OK.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
russoh: OK.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+        public const string EmissionSlotName = "Emission";
+        public const string MetallicSlotName = "Metallic";
+        public const string SmoothnessSlotName = "Smoothness";
+        public const string OcclusionSlotName = "Occlusion";
+        public const string AlphaSlotName = "Alpha";
+        public const string AlphaClipThresholdSlotName = "AlphaClipThreshold";
+        public const string AlphaClipThresholdDepthPrepassSlotName = "AlphaClipThresholdDepthPrepass";
+        public const string AlphaClipThresholdDepthPostpassSlotName = "AlphaClipThresholdDepthPostpass";
+        public const string AnisotropySlotName = "Anisotropy";
+        public const string PositionName = "Position";
```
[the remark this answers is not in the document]
russoh: Will do.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+    //-------------------------------------------------------------------------------------
+
+    #pragma vertex Vert
+    #pragma fragment Frag
+
+    #define UNITY_MATERIAL_LIT      // Need to be define before including Material.hlsl
+
+    // Use surface gradient normal mapping as it handle correctly triplanar normal mapping and multiple UVSet
+    // this modifies the normal calculation
+    // #define SURFACE_GRADIENT
```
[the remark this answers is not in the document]
russoh: Yes, I'll purge some old comments.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSettingsView.cs
```diff
@@ -0,0 +1,510 @@
...
+                    {
+                        row.Add(new EnumField(HDLitMasterNode.AlphaModeLit.Additive), (field) =>
+                        {
+                            field.value = GetAlphaModeLit(m_Node.alphaMode);
+                            field.OnValueChanged(ChangeBlendMode);
+                        });
+                    });
+                }
+
+                ps.Add(new PropertyRow(new Label("    Blend Preserves Specular")), (row) =>
```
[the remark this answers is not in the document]
russoh: I may do something super simple and just have a prefix string where I manage the indentation with some IncreaseIndentation(), DecreaseIndentation(). Maybe :)

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitUI.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using UnityEngine;
+
+namespace UnityEditor.ShaderGraph
+{
+    public class HDLitGUI : ShaderGUI
+    {
+        public override void OnGUI(MaterialEditor materialEditor, MaterialProperty[] props)
+        {
+			materialEditor.PropertiesDefaultGUI(props);
```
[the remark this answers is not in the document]
russoh: Tabs in that line by the looks of it. Your tab size is probably different to mine? Does the Git code not get formatted like we do for Mercurial?

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/DiffusionProfileNode.cs
```diff
@@ -0,0 +1,105 @@
...
+        public DiffusionProfileNode()
+        {
+            name = "Diffusion Profile";
+            UpdateNodeAfterDeserialization();
+        }
+
+        public override string documentationURL
+        {
+            // This still needs to be added.
+            get { return "https://github.com/Unity-Technologies/ShaderGraph/wiki/Diffusion-Profile-Node"; }
```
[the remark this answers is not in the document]
russoh: That does have a wiki page, it just lives in ShaderGraph currently. When is this all getting rolled into documentation in the manual?

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/DiffusionProfileNode.cs
russoh: Yes, this needs to be renamed.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Slots/DiffusionProfileInputMaterialSlot.cs
russoh: More tabs

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Slots/DiffusionProfileInputMaterialSlot.cs
russoh: Do we have anything we can use currently?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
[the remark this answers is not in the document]
sebastienlagarde: ok, I can get that it can be problematic technically to change the label, still I suggest to use BaseColor rather than Albedo as "metallic" is the default parametrization. This is more artists friendly (and the correct name should be Diffuse Albedo). Thought? This is also to have uniform naming with our other shader.

Regarding specular occlusion, yes I suggest to not follow what Lit.shader is doing and have a dedicated slot for specular occlusion. We want artist to be able to provide what they want for this input but we we still providean helper node that is doing what Lit shader is doing (In practice this allow user to chose between different spec occlusion method).

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
sebastienlagarde: yup

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+
+        // This could also be a simple array. For now, catch any mismatched data.
+        SlotMask GetActiveSlotMask()
+        {
+            switch (materialType)
+            {
+                case MaterialType.Standard:
+                    return StandardSlotMask;
+
+                case MaterialType.SubsurfaceScattering:
```
sebastienlagarde: I don't see clear coat mask, is it expected?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+                m_DrawBeforeRefraction = value.isOn;
+                UpdateNodeAfterDeserialization();
+                Dirty(ModificationScope.Topological);
+            }
+        }
+
+        [SerializeField]
+        ScreenSpaceLighting.RefractionModel m_RefractionModel;
+
+        public ScreenSpaceLighting.RefractionModel refractionModel
```
sebastienlagarde: As explain above, this will be remove in a future PR.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -0,0 +1,830 @@
...
+
+                AddSlot(new ColorRGBMaterialSlot(RefractionColorSlotId, RefractionColorSlotName, RefractionColorSlotName, SlotType.Input, Color.white, ColorMode.Default, ShaderStageCapability.Fragment));
+                validSlots.Add(RefractionColorSlotId);
+
+                AddSlot(new Vector1MaterialSlot(RefractionDistanceSlotId, RefractionDistanceSlotName, RefractionDistanceSlotName, SlotType.Input, 0.0f, ShaderStageCapability.Fragment));
+                validSlots.Add(RefractionDistanceSlotId);
+            }
+            if (HasDistortion())
+            {
+                AddSlot(new Vector2MaterialSlot(DistortionSlotId, DistortionSlotName, DistortionSlotName, SlotType.Input, new Vector2(2.0f, -1.0f), ShaderStageCapability.Fragment));
```
sebastienlagarde: We have an option that is "distortion only", in this case we should only display distortion slot. Don't know if this is possible?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+    #else
+        // used for shaders that don't need lighting
+        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
+    #endif
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/BuiltinUtilities.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/MaterialUtilities.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalUtilities.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitDecalData.hlsl"
+
+    int _ObjectId;
```
sebastienlagarde: you can add a comment that it is use for selection in edtior

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+
+        // tangent-space normal
+        float3 normalTS = float3(0.0f, 0.0f, 1.0f);
+        $SurfaceDescription.Normal: normalTS = surfaceDescription.Normal;
+
+        // compute world space normal
+        GetNormalWS(fragInputs, normalTS, surfaceData.normalWS);
+
+#ifdef _BENT_NORMAL
+        GetNormalWS(fragInputs, surfaceDescription.BentNormal, bentNormalWS);
```
sebastienlagarde: Note for a future PR:
There is a PR coming with LightingNod: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1911
This Lighting node take as input a normal for baked GI.
In the shader graph we shouldn't expose any BentNormal input, BentNormal can be use by the artists or not inside the node graph itself. They are free to read a bentNormal texture, and feed it to the lighting node, which is exactly what we do under the hood here. Goal is to expose this part in the shader graph.

So for next PR after lighting node is merge, plan to remove BentNormal input (and so the code above). I can help of course.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+        bentNormalWS = surfaceData.normalWS;
+#endif
+
+#ifdef _HAS_REFRACTION
+        surfaceData.transmittanceMask = 1.0 - surfaceDescription.Alpha;
+        surfaceDescription.Alpha = 1.0;
+#endif
+
+#if _TANGENT_MAP
+        surfaceData.tangentWS = TransformTangentToWorld(surfaceDescription.Tangent, fragInputs.worldToTangent);
```
sebastienlagarde: Caution: Tangent must be an input on the master node exactly like normal input. This is the same thing, we shouldn't have any define for _TANGENT_MAP.

The inputs of master node are in Tangent Space (IIRC), so user must provide a tangent in this space.

So the code here should be something lile this:
float3 tangentWS = normalize(fragInputs.worldToTangent[0].xyz);
$SurfaceDescription.Normal: tangentTS = surfaceDescription.Tangent;
$SurfaceDescription.Normal:  TransformTangentToWorld(Tangent, fragInputs.worldToTangent);
 surfaceData.tangentWS =tangentWS ;

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+#else
+        surfaceData.tangentWS = normalize(fragInputs.worldToTangent[0].xyz);    // The tangent is not normalize in worldToTangent for mikkt. TODO: Check if it expected that we normalize with Morten. Tag: SURFACE_GRADIENT
+#endif        
+        surfaceData.tangentWS = Orthonormalize(surfaceData.tangentWS, surfaceData.normalWS);
+
+        // By default we use the ambient occlusion with Tri-ace trick (apply outside) for specular occlusion.
+        // If user provide bent normal then we process a better term
+        surfaceData.specularOcclusion = 1.0;
+#if defined(_BENT_NORMAL_SPECULAR_OCCLUSION)
```
sebastienlagarde: So as explain in a previous comment remove this whole section

#if defined(_BENT_NORMAL_SPECULAR_OCCLUSION)
        // If we have bent normal and ambient occlusion, process a specular occlusion
        surfaceData.specularOcclusion = GetSpecularOcclusionFromBentAO(V, bentNormalWS, surfaceData.normalWS, surfaceData.ambientOcclusion, PerceptualSmoothnessToPerceptualRoughness(surfaceData.perceptualSmoothness));
#elif defined(_OCCLUSION) && defined(SPECULAR_OCCLUSION)
        surfaceData.specularOcclusion = GetSpecularOcclusionFromAmbientOcclusion(ClampNdotV(dot(surfaceData.normalWS, V)), surfaceData.ambientOcclusion, PerceptualSmoothnessToRoughness(surfaceData.perceptualSmoothness));
#else
        surfaceData.specularOcclusion = 1.0;
#endif

We will provide node to do GetSpecularOcclusionFromBentAO and GetSpecularOcclusionFromAmbientOcclusion in shader graph

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+        SurfaceDescription surfaceDescription = SurfaceDescriptionFunction(surfaceDescriptionInputs);
+
+        // Perform alpha test very early to save performance (a killed pixel will not sample textures)
+        // TODO: split graph evaluation to grab just alpha dependencies first? tricky..
+        $AlphaTest:         DoAlphaTest(surfaceDescription.Alpha, surfaceDescription.AlphaClipThreshold);
+        $AlphaTestPrepass:  DoAlphaTest(surfaceDescription.Alpha, surfaceDescription.AlphaClipThresholdDepthPrepass);
+        $AlphaTestPostpass: DoAlphaTest(surfaceDescription.Alpha, surfaceDescription.AlphaClipThresholdDepthPostpass);
+
+        float3 bentNormalWS;
+        BuildSurfaceData(fragInputs, surfaceDescription, V, surfaceData, bentNormalWS);
```
sebastienlagarde: For a future PR, as explain previously we will use lightingNode, let's chat for this future to know what we do here, as we will not call BuildSurfaceData as it is in current state

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+
+#if HAVE_DECALS && _DECALS
+        DecalSurfaceData decalSurfaceData = GetDecalSurfaceData(posInput, surfaceDescription.Alpha);
+        ApplyDecalToSurfaceData(decalSurfaceData, surfaceData);
+#endif
+
+        // Builtin Data
+        // For back lighting we use the oposite vertex normal 
+        InitBuiltinData(surfaceDescription.Alpha, bentNormalWS, -fragInputs.worldToTangent[2], fragInputs.positionRWS, fragInputs.texCoord1, fragInputs.texCoord2, builtinData);
```
sebastienlagarde: For a future PR, as explain previously we will use lightingNode, let's chat for this future to know what we do here, as we will not call BuildSurfaceData as it is in current state

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+
+#ifdef _EMISSION
+        builtinData.emissiveColor = surfaceDescription.Emission;
```
sebastienlagarde: Why do you need to test for _EMISSION here? should be 
builtinData.emissiveColor = 0;
$Emissive:Color builtinData.emissiveColor = surfaceDescription.Emission;
$AlbedoAffectsEmissive: builtinData.emissiveColor *= surfaceDescription.Albedo;

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -0,0 +1,865 @@
...
+        {
+            Name = "GBuffer",
+            LightMode = "GBuffer",
+            TemplateName = "HDLitPass.template",
+            MaterialName = "Lit",
+            ShaderPassName = "SHADERPASS_GBUFFER",
+
+            ExtraDefines = new List<string>()
+            {
+                "#pragma multi_compile _ DEBUG_DISPLAY",
```
sebastienlagarde: We need to share this definition of multicompile. 

We have a set of common multicompile for lighting (shadowmask / GI) + DECALS
Forward pass is exactly the same thing than deferred + some other option, there is redudency here (and with pass for backface too...

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -0,0 +1,865 @@
...
+            },
+            Includes = new List<string>()
+            {
+                "#include \"Runtime/ShaderPass/ShaderPassGBuffer.hlsl\"",
+            },
+            RequiredFields = new List<string>()
+            {
+                "FragInputs.worldToTangent",
+                "FragInputs.positionRWS",
+                "FragInputs.texCoord1",
```
sebastienlagarde: For future PR. Once with have lighting node, we can provide our own UV, and texcoord1 and texcoord2 will not be mandatory anymore! \0/

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -0,0 +1,865 @@
...
+            VertexShaderSlots = new List<int>()
+            {
+                HDLitMasterNode.PositionSlotId
+            },
+            UseInPreview = false
+        };
+
+        Pass m_SceneSelectionPass = new Pass()
+        {
+            Name = "SceneSelectionPass",
```
sebastienlagarde: We could be clever here. The selection pass is only required when we have displacement, vertex animation or alpha test. Don't know if we want to optimize for this or not. This can save time on shader compilation in editor

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: That define (_TANGENT_MAP) is actually set based on whether or not we have a tangent bound to  a slot. I just did that for convenience, so it only does the transformation when the slot is connected.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
russoh: To be honest, nearly all the time is spent compiling the forward pass, so I don't think it would have a tremendous impact in shader graph compile time, but we could easily guard that pass generation if needed.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -0,0 +1,865 @@
...
+        Pass m_PassForward = new Pass()
+        {
+            Name = "Forward",
+            LightMode = "Forward",
+            TemplateName = "HDLitPass.template",
+            MaterialName = "Lit",
+            ShaderPassName = "SHADERPASS_FORWARD",
+            ExtraDefines = new List<string>()
+            {
+                "#pragma multi_compile _ DEBUG_DISPLAY",
```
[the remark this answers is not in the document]
russoh: Yes, those multi-compiles should probably be moved to a common include location. Before or after this PR?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -0,0 +1,865 @@
...
+                GenerateShaderPassLit(masterNode, m_SceneSelectionPass, mode, subShader, sourceAssetDependencyPaths);
+
+                if (opaque)
+                {
+                    GenerateShaderPassLit(masterNode, m_PassDepthOnly, mode, subShader, sourceAssetDependencyPaths);
+                }
+
+                if (velocity)
+                {
+                    GenerateShaderPassLit(masterNode, m_PassMotionVectors, mode, subShader, sourceAssetDependencyPaths);
```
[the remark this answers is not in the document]
russoh: Yep, going to do that.

com.unity.shadergraph/Editor/Data/MasterNodes/SpecularOcclusionMode.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace UnityEditor.ShaderGraph
+{
+    public enum SpecularOcclusionMode
```
sebastienlagarde: As discuss, this class must be removed

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: Right, sounds good.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: Will do.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
russoh: I actually noticed that today. Need to look at this, not sure yet.

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
```diff
@@ -60,6 +60,39 @@ public TextureType textureType
...
+                if (m_NormalMapSpace == value)
+                    return;
+
+                m_NormalMapSpace = value;
+                Dirty(ModificationScope.Graph);
+            }
+        }
+
+        [SerializeField]
+        private bool m_DerivativeMap = false;
```
sebastienlagarde: Hmm... I have check the PR but didn't see anything about composing normal map with surface gradient. Maybe I missed it. You see to just have node to create surface gradient, right? 
For it to be transparent we need to add a node, like composeNormal that will perform compositing (either a add for surface gradient or a ReorientednormalMapping for tangent space normal map (or RNM in world space for object space normal).

Note: Derivative map isn't visible or use at all. What we do is that we convert a normal map to a derivative map, then to surface gradient, but artist never see derivative map, nor need to produce them.

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
```diff
@@ -106,7 +139,32 @@ public virtual void GenerateNodeCode(ShaderGenerator visitor, GraphContext graph
             visitor.AddShaderChunk(result, true);
 
             if (textureType == TextureType.Normal)
-                visitor.AddShaderChunk(string.Format("{0}.rgb = UnpackNormalmapRGorAG({0});", GetVariableNameForSlot(OutputSlotRGBAId)), true);
+            {
+                if (normalMapSpace == NormalMapSpace.Tangent)
+                {
+                    if (derivativeMap.isOn)
```
sebastienlagarde: so correct name here is "surfaceGradient", not derivative map

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -0,0 +1,332 @@
...
+        // Reproduce the energy conservation done in legacy Unity. Not ideal but better for compatibility and users can unchek it
+        surfaceData.baseColor *= (1.0 - Max3(surfaceData.specularColor.r, surfaceData.specularColor.g, surfaceData.specularColor.b));
+#endif
+
+        // tangent-space normal
+        float3 normalTS = float3(0.0f, 0.0f, 1.0f);
+        $SurfaceDescription.Normal: normalTS = surfaceDescription.Normal;
+
+        // compute world space normal
+        GetNormalWS(fragInputs, normalTS, surfaceData.normalWS);
```
sebastienlagarde: So I am not sure to understand how this work with surface gradient (I have added comment at the end of the PR about it). GetNormalWS will convert from a surface gradient to world, or from normalTS to world. But you need to define SURFACE_GRADIENT for this, and I haven't see it (or I missed it). Also what is happening is that the whole graph need to be surface gradient, not only some node. You can't mix and match surface gradient with regular normal in tagent space.

My guess to deal with surface gradient is a global switch, where all the normal map node use surface gradient, or they all use regular normal, and artists need to use composeNormal node to add normal (this will also help for normal map filtering in the future). We should also add a node: GetNormal() or something like that, that will convert the current normal as surface gradient to real normal, so it can be use as input for other node (like lighting node). We can discuss about it, no need to fix all this in this PR, also I haven't tried surface gradient of your PR, so maybe I missed how this is implemented. LEt's talk!

russoh: I'm basing the derivative map option on my understanding of:
#ifdef SURFACE_GRADIENT
        real2 deriv = UNPACK_DERIVATIVE_FUNC(SAMPLE_TEXTURE_FUNC(textureName, samplerName, uvMapping.uv, param), scale);
        return SurfaceGradientFromTBN(deriv, uvMapping.tangentWS, uvMapping.bitangentWS);
#else

I thought that it was expected that the input was a derivative map, since the shader unpacks it like that, and then generates the surface gradient. Did I miss some intermediate step somewhere?

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
russoh: I'm basing the derivative map option on my understanding of:
#ifdef SURFACE_GRADIENT
        real2 deriv = UNPACK_DERIVATIVE_FUNC(SAMPLE_TEXTURE_FUNC(textureName, samplerName, uvMapping.uv, param), scale);
        return SurfaceGradientFromTBN(deriv, uvMapping.tangentWS, uvMapping.bitangentWS);
#else

I thought that it was expected that the input was a derivative map, since the shader unpacks it like that, and then generates the surface gradient. Did I miss some intermediate step somewhere?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
sebastienlagarde: isn't what this code is doing? 

float3 tangentWS = normalize(fragInputs.worldToTangent[0].xyz);
$SurfaceDescription.tangent: tangentWS = TransformTangentToWorld(surfaceDescription.Tangent, fragInputs.worldToTangent);
surfaceData.tangentWS = tangentWS ;

=> doesn't this mean that if a slot is connected, then you do a transform, otherwise we use normalize of TBN[0] ? 

Then no need for _TANGENT_MAP. 
Looks cleaner to me. Just wondering why tangent will be processed differently than normal whereas it is the same thing.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
sebastienlagarde: isn't it cleaner to not rely on any define? Why do we need to add define? 
EmissiveColor is no different from BaseColor for shader graph, and we don't use define for shader graph. And this remove an option (AlbedoAffectsEmissive ) from the UI which is good :)

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
sebastienlagarde: Sadly #pragma multicompile aren't supported in include with unity shader compiler. What I suggest is just to shader the definition in C#, so moving the string definition in a common accessible class.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
sebastienlagarde: ok, fair enough :)

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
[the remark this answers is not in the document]
sebastienlagarde: after PR is fine

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
sebastienlagarde: indeed, the naming in lit.shader is also wrong :P, when we do this: UNPACK_DERIVATIVE_FUNC we are really reading a normal map, convert it to a derivative map, then convert it to surface gradient. But there is also something missing. Where is the code that generate the tangent basis on the fly? (i.E uvMapping.tangentWS, uvMapping.bitangentWS but for the given texcoord?)
I think to be sure to avoid redundancy we should have somewhere in the shader graph something that define which texcoord require to use surface gradient, and generate tangent basis on the fly for them. Such why I suggest a global switch for the shader graph to say the whole graph rely on surface gradient. This become a bit complex and I suggest we discuss this in a call and don't care for this PR.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: Yes, agreed, AlbedoAffectsEmissive not needed in this case, since it's just a simple multiply on the input color,which  can be performed in the graph.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
russoh: Let's try to sort this out in a follow up PR. Is that OK?

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
russoh: OK

com.unity.shadergraph/Editor/Data/Nodes/Input/Texture/SampleTexture2DLODNode.cs
russoh: Let's figure out the correct way to set this up, for another PR.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: Sounds like we need guidance with appropriate tests for this, which Zack can setup. We did some simple testing using the tangent map, for which the derivative map unpack was added. It sounds like we need additional tests, and perhaps more functionality.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: It's the same thing. As you say, if we have $SurfaceDescription.tangent, the compiler is just going to use the transformation anyway. Was mostly just for clarity at a glance, I don't mind either way.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
russoh: Will do this in a follow up PR, since it requires new nodes and slots, (and test changes).

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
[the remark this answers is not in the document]
russoh: OK, using BaseColor on the HD node. Specular occlusion will be changed in a later PR.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Slots/DiffusionProfileInputMaterialSlot.cs
russoh: Fixed.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
russoh: Will try to add this in a later PR.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
russoh: OK, we'll have to sort this out in a follow up PR.

sebastienlagarde:
Adding one comment. We would also need to expose Depth offset as input.

PR #2003, opened 2018-09-17
Hdrp shadow system

alelievr:
### Purpose of this PR

There is a new plan for the shadow framework: https://docs.google.com/document/d/1zWX1Qj5HAAZk-luulC51oKn6iD3njrb8v2gspvN-bFE/edit?usp=sharing

This PR introduce the first update to the shadow framework to follow principles of above document.
As we move a large part of the system under HDRP control, there is various part that have be simplified, and less indirection are required. The memory is now entirely control by HDRP and we rely on the RT handle system for RT allocation. Resource management now goes through HDRP.

The PR allow to switch between old and new system just to validate that nothing have been broken, a follow up PR will disable the shadow system present in Core.

The atlas management have also been update. Shadowmap size scale with distance and when there is no room in the atlas anymore the whole atlas is resize. This also fix some issues reported in Fogbugz on previous atlas allocation.

In this PR only the PCF and PCSS filter have been ported. Another PR will add the moment shadow map and ESM. The filter can now be dynamically selected between low, medium and high which bind currently on PCF3x3, PCF5x5 and PCSS.

There is a new HDShadoData definition for hlsl, as on shader side we give the control to HDRP. For now the AdditionalShadowData are still separate from the HDAdditionalLightData but in a futrue PR they will be merge.

All the shadow bias algorithm are the same but they don't rely on ShadowContext anymore.

---
### Release Notes
In this PR the Core shadow system is just disabled, it's possible to switch (by changing some code) between the two systems

---
### Testing status
**Katana Tests**:  they're running, waiting for results

**Manual Tests**:  Simple local tests & HDRP graphic tests

**Automated Tests**:  No tests was added (as this system should look the same as before and we already have tests for shadows)

---
### Overall Product Risks
**Technical Risk**: Medium: the switch of shadow system should be pretty smooth as i didn't change where the shadow datas are stored, no upgrade should be necessary.

**Halo Effect**: Medium


com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
```diff
@@ -120,12 +124,12 @@ public override string ToString()
 
             public int elementCount
             {
-                get { return rows * cols; }
+                get { return rows * cols * (arraySize > 0 ? arraySize : 1); }
```
EvgeniiG: This could be just a max(arraySize, 1)

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
```diff
@@ -247,7 +252,7 @@ bool ExtractComplex(FieldInfo field, List<ShaderFieldInfo> shaderFields)
                     Error(mismatchErrorMsg);
                     return false;
                 }
-                EmitPrimitiveType(PrimitiveType.Float, floatFields.Count, field.Name, comment, shaderFields);
+                EmitPrimitiveType(PrimitiveType.Float, floatFields.Count, -1, field.Name, comment, shaderFields);
```
EvgeniiG: Why -1?

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
```diff
@@ -620,37 +658,37 @@ public bool Generate()
...
                 {
-                    if (field.FieldType == typeof(float))
-                        EmitPrimitiveType(PrimitiveType.Float, 1, field.Name, "", m_ShaderFields);
-                    else if (field.FieldType == typeof(int))
-                        EmitPrimitiveType(PrimitiveType.Int, 1, field.Name, "", m_ShaderFields);
-                    else if (field.FieldType == typeof(uint))
-                        EmitPrimitiveType(PrimitiveType.UInt, 1, field.Name, "", m_ShaderFields);
-                    else if (field.FieldType == typeof(bool))
-                        EmitPrimitiveType(PrimitiveType.Bool, 1, field.Name, "", m_ShaderFields);
+                    if (fieldType == typeof(float))
```
EvgeniiG: Probably better to use switch() here

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.Styles.cs
```diff
@@ -54,6 +54,7 @@ sealed class Styles
             public readonly GUIContent shadowFadeDistance = new GUIContent("Fade Distance", "The shadow will fade at distance ShadowFadeDistance before being culled to minimize popping.");
             public readonly GUIContent shadowDimmer = new GUIContent("Dimmer", "Aim to be use with script, timeline or animation. It allows dimming one or multiple shadows. This can also be used as an optimization to fit in shadow budget manually and minimize popping.");
             public readonly GUIContent contactShadows = new GUIContent("Enable Contact Shadows", "Enable support for contact shadows on this light. Better for lights with a lot of visible shadows.");
+            public readonly GUIContent useDynamicViewportResize = new GUIContent("Use Dynamic Viewport Resize", "Scale the shadow map size using on screen size of the light to gain space for more shadows in the atlas");
```
EvgeniiG: Nice

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
```diff
@@ -53,6 +82,22 @@ public BaseShaderPreprocessor()
         // -
         protected bool CommonShaderStripper(HDRenderPipelineAsset hdrpAsset, Shader shader, ShaderSnippetData snippet, ShaderCompilerData inputData)
         {
+            // Strip every useless shadow configs
+            // TODO: test if it actually works
```
EvgeniiG: Does it?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDShadowInitParametersUI.cs
```diff
@@ -0,0 +1,55 @@
+using UnityEngine;
+using UnityEngine.Experimental.Rendering;
+using UnityEditor.Experimental.Rendering;
+
+namespace UnityEditor.Experimental.Rendering.HDPipeline
+{
+    using _ = CoreEditorUtils;
```
EvgeniiG: Please don't do that. It doesn't really improve the clarity of the code.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -170,6 +171,167 @@ public uint GetLightLayers()
...
+            // If the shadow is too far away, we don't render it
+            if (m_Light.type != LightType.Directional && Vector3.Distance(cameraPos, transform.position) >= m_ShadowData.shadowFadeDistance)
+                return -1;
+
+            Vector2 viewportSize = new Vector2(m_ShadowData.shadowResolution, m_ShadowData.shadowResolution);
+
+            if (useDynamicViewportResize)
+            {
+                // resize viewport size by the normalized size of the light on screen
+                viewportSize *= visibleLight.screenRect.size.magnitude / Mathf.Sqrt(2);
```
EvgeniiG: IMHO we should not rescale but rather set the size directly.
The size should be max(screenRect.size.x, screenRect.size.y), there's no need to account for the length of the diagonal.
If we really want to rescale, the code should probably be different.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -170,6 +171,167 @@ public uint GetLightLayers()
...
+            shadowRequest.viewBias = new Vector4(m_ShadowData.viewBiasMin, m_ShadowData.viewBiasMax, m_ShadowData.viewBiasScale, 2.0f / shadowRequest.projection.m00 / m_ShadowData.shadowResolution * 1.4142135623730950488016887242097f);
+            shadowRequest.normalBias = new Vector3(m_ShadowData.normalBiasMin, m_ShadowData.normalBiasMax, m_ShadowData.normalBiasScale);
+            shadowRequest.flags = 0;
+            shadowRequest.flags |= m_ShadowData.sampleBiasScale     ? (int)HDShadowFlag.SampleBiasScale : 0;
+            shadowRequest.flags |= m_ShadowData.edgeLeakFixup       ? (int)HDShadowFlag.EdgeLeakFixup : 0;
+            shadowRequest.flags |= m_ShadowData.edgeToleranceNormal ? (int)HDShadowFlag.EdgeToleranceNormal : 0;
+            shadowRequest.edgeTolerance = m_ShadowData.edgeTolerance;
+
+            // Make light position camera relative:
+            // TODO: think about VR (use different camera position for each eye)
```
EvgeniiG: Pretty sure we already have some kind of head position rather than eye position.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1651,57 +1713,72 @@ public bool IsBakedShadowMaskLight(Light light)
                     rightEyeWorldToView = WorldToViewStereo(camera, Camera.StereoscopicEye.Right);
                 }
 
+                // TODO: Only for dev purpose remove once the new shadow system is stable
+                if (useNewShadowSystem)
+                    Shader.DisableKeyword("USE_CORE_SHADOW_SYSTEM");
```
EvgeniiG: But it's not a keyword? This does nothing?

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.shader
```diff
@@ -315,6 +315,14 @@ Shader "HDRenderPipeline/Fabric"
             #pragma multi_compile USE_FPTL_LIGHTLIST USE_CLUSTERED_LIGHTLIST
             #pragma multi_compile DECALS_OFF DECALS_3RT DECALS_4RT
 
+            // TODO: remove this once new shadow system works
+            // Only for dev/test purpose, allow to switch dynamically between HD and Core shadow system
+            // #pragma multi_compile _ USE_CORE_SHADOW_SYSTEM
+            
+            // Supported shadow modes per light type
+            #pragma multi_compile PUNCTUAL_SHADOW_LOW PUNCTUAL_SHADOW_MEDIUM PUNCTUAL_SHADOW_HIGH
```
EvgeniiG: I would argue that it is useless to have different shadow quality between local and dir lights.

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
alelievr: The HLSLArray dont support complex type arrays as it will make the struct containing this not blittable, should i add a comment for this ?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/Shadow/ShadowSamplingTent.hlsl
```diff
@@ -0,0 +1,244 @@
+// ------------------------------------------------------------------
```
EvgeniiG: This file is now twice?

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
EvgeniiG: Aren't you passing the size of -1 here? Just pass 0?

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
alelievr: Unfortunately, there is no switch case for types in our version of C#, we'll have to wait pattern matching in C# 7

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
alelievr: Oops, I forgot to remove this, yes it actually works :)

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/HDShadowInitParametersUI.cs
alelievr: I tried to keep consistency between all the render pipeline UI code so it's alnost a copy-paste of the ShadowInitParameterUI. If you want to change this it should probably better to do it in `HDRenderPipelineUI`, `FrameSettingsUI`, `GlobalDecalSettingsUI`, `GlobalLightLoopSettingsUI`, `LightLoopSettingsUI`, amd `RenderPipelineSettingsUI` as well

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -120,6 +125,48 @@ void GetPunctualLightVectors(float3 positionWS, LightData lightData, out float3
...
+    float2 positionNDC = positionCS * 0.5 + 0.5;
+
+    // Manually clamp to border (black).
+    cookie.rgb = SampleCookie2D(lightLoopContext, positionNDC, lightData.cookieIndex, false);
+    cookie.a   = isInBounds ? 1 : 0;
+
+    return cookie;
+}
+
+float4 EvaluateCookie_Point(LightLoopContext lightLoopContext, LightData lightData,
```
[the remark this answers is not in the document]
alelievr: At some point i had two different functions for spot and point shadow evaluation but since it's not the case anymore, there is no reason to be splitted.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/Shadow/ShadowSamplingTent.hlsl
alelievr: it reappeared during the big pr that moved everything, i'll delete the other one

com.unity.render-pipelines.high-definition/Editor/CoreRP/Editor/ShaderGenerator/ShaderTypeGeneration.cs
alelievr: ok

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
alelievr: This is intended to switch between core and hd sahdow systems (ofc you have to uncomment the multi compile in the shaders)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
alelievr: This approximation as far as i have tested gives good results, i understand that it can make far shadow a bit blurry but i think the compromise between gained space and quality is good. We'll have to ask to the artists for some feedback about this feature.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.Styles.cs
JulienIgnace-Unity: I think "Use Dynamic Viewport Resize" is not super clear (the tooltip is though). Maybe more something like "Dynamic Shadow Rescale" ?
Another thing, I'm not sure we want this setting per light. This adds potentially a lot of work for artists especially with a default value at false when in fact we'll want that for almost all lights.
My suggestion is that this should be a global setting set to true by default (for debug purpose) and then it can be disabled per light for cinematic purpose for example.
We should check with the artists for the workflow

com.unity.render-pipelines.high-definition/Runtime/CoreRP/CoreResources/GPUCopy.cs
```diff
@@ -76,13 +76,15 @@ public GPUCopy(ComputeShader shader)
                 if (dispatch8Rect.width > 0 && dispatch8Rect.height > 0)
                 {
                     var r = dispatch8Rect;
+                    // Caution: passing parameters to SetComputeIntParams() via params generate 48B several times at each frame here !
```
JulienIgnace-Unity: Afaik, this can be fixed by creating a local array on the stack and passing it exlicitely instead of relying on the variadic version of SetComputeIntParams.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -170,6 +171,167 @@ public uint GetLightLayers()
...
+                viewportSize *= visibleLight.screenRect.size.magnitude / Mathf.Sqrt(2);
+
+                // Prevent flickering caused by the floating size of the viewport
+                viewportSize.x = Mathf.Round(viewportSize.x);
+                viewportSize.y = Mathf.Round(viewportSize.y);
+
+                viewportSize = Vector2.Max(viewportSize, new Vector2(32, 32));
+            }
+
+            for (int faceIndex = 0; faceIndex < shadowRequests.Length; faceIndex++)
```
JulienIgnace-Unity: Minor complaint but I'd rather have requestIndex than faceIndex. Face index is very specific to point lights and we want this to be as generic as possible.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.shader
alelievr: I don't know, with the core shadow framework we could choose different algorithms per light type so for me it makes sense to keep this feature.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Shadow.hlsl
```diff
@@ -35,15 +35,15 @@
 // point
 #define SHADOW_DISPATCH_POINT_TEX 0
 #define SHADOW_DISPATCH_POINT_SMP 0
-#define SHADOW_DISPATCH_POINT_ALG GPUSHADOWALGORITHM_PCF_TENT_3X3
+#define SHADOW_DISPATCH_POINT_ALG GPUSHADOWALGORITHM_PCF_TENT_5X5
```
JulienIgnace-Unity: What's the reason of the bump in quality?

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.shader
JulienIgnace-Unity: I think it's interesting to split punctual and directional shadow quality.
Very often you have a much better texel to pixel density for punctual lights so you can get away with cheaper filtering.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Shadow.hlsl
EvgeniiG: 5x5 is pretty expensive for Low Quality IMHO.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightEditor.Styles.cs
alelievr: I've added an option in the HDRP asset to disable the shadow resolution rescale, and renamed the option to "Dynamic Shadow Rescale"

com.unity.render-pipelines.high-definition/Runtime/CoreRP/CoreResources/GPUCopy.cs
alelievr: Sure, I think we need to do a complete batch where we remove all the garbage generated at each frames.

alelievr:
Changes to the lighting have been reverted in c7d03eb19232a53ac1ae0c58af09c0a8f50e9583. The core shadow algorithm have also been reverted

PR #2012, opened 2018-09-19
Stereo ProceduralSky/Atmospherics + Stereo MSAA Color Resolve fix

nerites:
### Purpose of this PR
- Fixed skybox rendering for single-pass stereo
- Fixed atmospherics for single-pass stereo
- Fixed MSAA color resolve shader for single-pass stereo
---
### Release Notes
- Includes changesets from the pending hdrp_xr PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1907

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=stereo_msaa_skybox&unity_branch=trunk&automation-tools_branch=add-platform-filter

With forward rendering on, visually confirmed correct skybox + atmospherics with single-pass stereo enabled and MSAA on and off: 
MSAA off:
image

MSAA on:
image

No additional tests.

HDRP template scene should demonstrate working atmospherics

---
### Overall Product Risks
Medium: Modified view direction computation for proceduralsky shader if UNITY_SINGLE_PASS_STEREO defined. Texture coordinate modifications for color resolve, and opaque atmospheric scattering shaders. 

Low: Changes are in common shaders but are protected by #ifdef UNITY_SINGLE_PASS_STEREO.

---
### Comments to reviewers
Might be easier to see the changes specific to this commit after merging the PR mentioned above.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
@@ -101,6 +101,56 @@ namespace UnityEngine.Experimental.Rendering
                 name
                 );
         }
+        public static RTHandleSystem.RTHandle Alloc(
+            Vector2 scaleFactor,
+            bool stereoEnabled,
+            DepthBits depthBufferBits = DepthBits.None,
+            RenderTextureFormat colorFormat = RenderTextureFormat.Default,
+            FilterMode filterMode = FilterMode.Point,
+            TextureWrapMode wrapMode = TextureWrapMode.Repeat,
+            TextureDimension dimension = TextureDimension.Tex2D,
+            bool sRGB = true,
+            bool enableRandomWrite = false,
+            bool useMipMap = false,
+            bool autoGenerateMips = true,
+            int anisoLevel = 1,
+            float mipMapBias = 0,
+            bool enableMSAA = false,
+            bool bindTextureMS = false,
+            bool useDynamicScale = false,
+            RenderTextureMemoryless memoryless = RenderTextureMemoryless.None,
+            string name = ""
+            )
+        {
+            int slices = 1;
+            VRTextureUsage vrUsage = VRTextureUsage.None;
+            if (stereoEnabled)
+            {
+                slices = (XRGraphicsConfig.eyeTextureDesc.dimension == TextureDimension.Tex2DArray) ? 2 : 1; // TODO VR: double-check that this works as expected
+                vrUsage = XRGraphicsConfig.eyeTextureDesc.vrUsage;
+            }
+            return s_DefaultInstance.Alloc(
+                scaleFactor,
+                slices,
+                depthBufferBits,
+                colorFormat,
+                filterMode,
+                wrapMode,
+                dimension,
+                sRGB,
+                enableRandomWrite,
+                useMipMap,
+                autoGenerateMips,
+                anisoLevel,
+                mipMapBias,
+                enableMSAA,
+                bindTextureMS,
+                useDynamicScale,
+                vrUsage,
+                memoryless,
+                name
+                );
+        }
 
         public static RTHandleSystem.RTHandle Alloc(
             ScaleFunc scaleFunc,
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
@@ -2540,11 +2540,22 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                 cmd.SetComputeTextureParam(screenSpaceShadowComputeShader, kernel, HDShaderIDs._DeferredShadowTextureUAV, deferredShadowRT);
 
                 int deferredShadowTileSize = 16; // Must match DeferreDirectionalShadow.compute
-                int numTilesX = (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
+                int numTilesX = hdCamera.frameSettings.enableStereo ? ((hdCamera.actualWidth / 2) + (deferredShadowTileSize - 1)) / deferredShadowTileSize : (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
                 int numTilesY = (hdCamera.actualHeight + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
 
-                // TODO: Update for stereo
-                cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                if (hdCamera.frameSettings.enableStereo)
+                {
+                    for (uint eye = 0; eye < 2; eye++)
+                    {
+                        cmd.SetComputeIntParam(screenSpaceShadowComputeShader, HDShaderIDs._Eye, (int)eye);
+                        cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                    }
+                }
+                else
+                {
+                    cmd.SetComputeIntParam(screenSpaceShadowComputeShader, HDShaderIDs._Eye, 0);
+                    cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                }
 
                 cmd.SetGlobalTexture(HDShaderIDs._DeferredShadowTexture, deferredShadowRT);
             }
@@ -2580,13 +2591,13 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                 // Compute path
                 if (m_FrameSettings.lightLoopSettings.enableTileAndCluster && m_FrameSettings.lightLoopSettings.enableComputeLightEvaluation)
                 {
-                    int w = camera.pixelWidth;
-                    int h = camera.pixelHeight;
-                    int numTilesX = (w + 15) / 16;
+                    int w = hdCamera.actualWidth;
+                    int h = hdCamera.actualHeight;
+                    int numTilesX = m_FrameSettings.enableStereo ? ((w / 2) + 15) / 16 : (w + 15) / 16;
                     int numTilesY = (h + 15) / 16;
                     int numTiles = numTilesX * numTilesY;
 
-                    bool enableFeatureVariants = GetFeatureVariantsEnabled() && !debugDisplaySettings.IsDebugDisplayEnabled();
+                    bool enableFeatureVariants = GetFeatureVariantsEnabled() && !debugDisplaySettings.IsDebugDisplayEnabled() && !hdCamera.frameSettings.enableStereo; // TODO VR: Reenable later
 
                     int numVariants = 1;
                     if (enableFeatureVariants)
@@ -2627,15 +2638,28 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                         // always do deferred lighting in blocks of 16x16 (not same as tiled light size)
 
                         if (enableFeatureVariants)
-                        {
+                        { // TODO VR: variants support (solve how g_TileListOffset and surrounding math should work with stereo)
                             cmd.SetComputeBufferParam(deferredComputeShader, kernel, HDShaderIDs.g_TileFeatureFlags, s_TileFeatureFlags);
                             cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs.g_TileListOffset, variant * numTiles);
                             cmd.SetComputeBufferParam(deferredComputeShader, kernel, HDShaderIDs.g_TileList, s_TileList);
+                            cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._Eye, 0);
                             cmd.DispatchCompute(deferredComputeShader, kernel, s_DispatchIndirectBuffer, (uint)variant * 3 * sizeof(uint));
                         }
                         else
                         {
-                            cmd.DispatchCompute(deferredComputeShader, kernel, numTilesX, numTilesY, 1);
+                            if (m_FrameSettings.enableStereo)
+                            {
+                                for (int eye = 0; eye < 2; eye++)
+                                {
+                                    cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._Eye, eye);
+                                    cmd.DispatchCompute(deferredComputeShader, kernel, numTilesX, numTilesY, 1);
+                                }
+                            }
+                            else
+                            {
+                                cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._Eye, 0);
+                                cmd.DispatchCompute(deferredComputeShader, kernel, numTilesX, numTilesY, 1);
+                            }
                         }
                     }
                 }
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -1052,11 +1052,11 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                     // We can't currently render object velocity after depth prepass because if there is no depth prepass we can have motion vector write that should have been rejected
                     RenderCameraVelocity(m_CullResults, hdCamera, renderContext, cmd);
 
+                    StopStereoRendering(cmd, renderContext, hdCamera);
                     // Caution: We require sun light here as some skies use the sun light to render, it means that UpdateSkyEnvironment must be called after PrepareLightsForGPU.
                     // TODO: Try to arrange code so we can trigger this call earlier and use async compute here to run sky convolution during other passes (once we move convolution shader to compute).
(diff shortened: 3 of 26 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2540,11 +2540,22 @@ public void RenderScreenSpaceShadows(HDCamera hdCamera, RTHandleSystem.RTHandle
...
                 int deferredShadowTileSize = 16; // Must match DeferreDirectionalShadow.compute
-                int numTilesX = (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
+                int numTilesX = hdCamera.frameSettings.enableStereo ? ((hdCamera.actualWidth / 2) + (deferredShadowTileSize - 1)) / deferredShadowTileSize : (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
                 int numTilesY = (hdCamera.actualHeight + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
 
-                // TODO: Update for stereo
-                cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                if (hdCamera.frameSettings.enableStereo)
+                {
+                    for (uint eye = 0; eye < 2; eye++)
```
sebastienlagarde: I am wondering what happen when we are more than two eye (This is the case with StarVR headset where you have up to 4 view). I guess with should iterate over a given number of eye here and not hardcoded 2. (Just a guess). This can be fix in a future PR.

Also maybe we can simplify the code?

int eye = hdCamera.frameSettings.enableStereo ? 2 : 1
for (uint eye = 0; eye < 2; eye++)
{
                        cmd.SetComputeIntParam(screenSpaceShadowComputeShader, HDShaderIDs._Eye, (int)eye);
                        cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
}

to avoid to duplicate  dispatch  call.

sebastienlagarde: Same for others call below

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -450,7 +450,7 @@ bool SetRenderingFeatures()
                 return false;
             }
 
-            // VR is not supported currently in HD
+            // VR is not fully supported in HD
             //if (XRGraphicsConfig.enabled)
```
sebastienlagarde: Remove this comment and the commented code, we don't need it anymore

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -630,8 +630,11 @@ public void PushGlobalParams(HDCamera hdCamera, CommandBuffer cmd, DiffusionProf
 
                 // Set up UnityPerView CBuffer.
                 hdCamera.SetupGlobalParams(cmd, m_Time, m_LastTime, m_FrameCount);
-                if (hdCamera.frameSettings.enableStereo)
-                    hdCamera.SetupGlobalStereoParams(cmd);
+
+                // TODO VR: Current solution for compute shaders grabs matrices from
+                // stereo matrices even when not rendering stereo in order to reduce shader variants.
+                // After native fix for compute shader keywords is completed, qualify this with stereoEnabled.
+                hdCamera.SetupGlobalStereoParams(cmd);
```
sebastienlagarde: Shouldn't we move call to (SetupGlobalStereoParams) inside hdCamera.SetupGlobalParams ? Or there is exception?

sebastienlagarde: Avoid duplciation as there is also a call below

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -204,7 +204,7 @@ public void ConfigureStereoDependentSettings()
             if (enableStereo)
             {
                 // Force forward if we request stereo. TODO: We should not enforce that, users should be able to chose deferred
-                enableForwardRenderingOnly = true;
+                //enableForwardRenderingOnly = true;
```
sebastienlagarde: Nice :), please remove this code then :)

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
```diff
@@ -18,6 +18,7 @@ Shader "Hidden/HDRenderPipeline/Sky/ProceduralSky"
     #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
     #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl"
     #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/CommonLighting.hlsl"
+    #include "../../ShaderVariables.hlsl"
```
sebastienlagarde: Please use absolute path

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
```diff
@@ -119,10 +120,16 @@ Shader "Hidden/HDRenderPipeline/Sky/ProceduralSky"
 
     float4 Frag(Varyings input) : SV_Target
     {
+#if defined(UNITY_SINGLE_PASS_STEREO)
```
sebastienlagarde: I don't get this one. There is no multicompile, 
so UNITY_SINGLE_PASS_STEREO is never call, what is the trick?

Can you add a comment about what is this code doing?
Isn't it missing something like
float3 dir = normalize(posInput.positionWS - cameraPositionWS);

Have you assume you are camera relative here?

com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
```diff
@@ -101,6 +101,56 @@ public static class RTHandles
...
+            bool useDynamicScale = false,
+            RenderTextureMemoryless memoryless = RenderTextureMemoryless.None,
+            string name = ""
+            )
+        {
+            int slices = 1;
+            VRTextureUsage vrUsage = VRTextureUsage.None;
+            if (stereoEnabled)
+            {
+                slices = (XRGraphicsConfig.eyeTextureDesc.dimension == TextureDimension.Tex2DArray) ? 2 : 1; // TODO VR: double-check that this works as expected
```
sebastienlagarde: If it is only to add this test, you should modify directly the main function with an option argument rather than doing a new prototype function. Or I miss something?

sebastienlagarde: @JulienIgnace-Unity @anisunity ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStringConstants.cs
```diff
@@ -199,6 +199,9 @@ public static class HDShaderIDs
         public static readonly int _DirectionalContactShadowSampleCount = Shader.PropertyToID("_SampleCount");
         public static readonly int _DirectionalLightDirection = Shader.PropertyToID("_DirectionalLightDirection");
         public static readonly int _PunctualLightPosition = Shader.PropertyToID("_PunctualLightPosition");
+        public static readonly int _Matrix_I_VP = Shader.PropertyToID("_Matrix_I_VP");
```
sebastienlagarde: Both _Matrix_I_VP and _Matrix_V aren't required anymore.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStringConstants.cs
```diff
@@ -199,6 +199,9 @@ public static class HDShaderIDs
...
+        public static readonly int _Matrix_V = Shader.PropertyToID("_Matrix_V");
+        public static readonly int _Eye = Shader.PropertyToID("_Eye");
```
sebastienlagarde: Maybe it should be call _EyeIndex instead of just _Eye.

com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
JulienIgnace-Unity: As Seb said, we should probably just add a new named parameter with a false default value. This way we don't have a new overload.
Side note: Is this only for stereo instanced mode? Do we even support that in HDRP?

sebastienlagarde:
Genereal comment.
As we are reviewing this PR, maybe you can close the previous one (https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/1907). 

Katana is Green (you have a fail but it is related to a know LW issue, so we can consider all as green).

So just address / Answer question to the comment and we are fine to merge this PR after :)
(And sorry for the delay, we have a lot of PR to review :( )

com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
anisunity: agreed

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
nerites: I love this it's much cuter, thank you!!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
nerites: Tiled light list population is still broken for deferred stereo rendering so it's not 100% working... I'll make a note of it here

nerites:
Oh haha maybe I should have read your last comment first here (:

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
nerites: There is native code that defines the UNITY_SINGLE_PASS_STEREO keyword for non-compute shaders when context.StartMultiEye is called. 

When setting up _ViewMatrixStereo (which is bound per-eye to UNITY_MATRIX_I_V for non-compute shaders), HDRP code manually adjusts the view matrix if camera-relative rendering is set. 

I just disabled camera-relative rendering and confirmed that the skybox still renders as it should.

nerites: For context, my previous attempt at fixing this shader involved passing the stereo view matrix into HDUtils.ComputePixelCoordToWorldSPaceViewDirectionMatrix. I'm not sure if it was because the fov being passed in wasn't correct, or if the omission of the translation column from the 3x3 pixelCoordToViewDir matrix meant the eye offset wasn't correct, but Skybox was still rendering incorrectly even with stereo pixelCoordToViewDir matrices.

com.unity.render-pipelines.high-definition/Runtime/CoreRP/Textures/RTHandles.cs
nerites: I think it is only needed for stereo instanced mode... so I'm good to remove these changes for now.

I think the reason I had this as a required argument instead of an optional argument was because with the optional argument, it was ambiguous with some other override for RTHandles.Alloc. But that's probably something I can resolve if this every becomes necessary and isn't needed for this PR.

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
sebastienlagarde: ok, if @JulienIgnace-Unity is fine with the code, I am good :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2540,11 +2540,15 @@ public void RenderScreenSpaceShadows(HDCamera hdCamera, RTHandleSystem.RTHandle
                 cmd.SetComputeTextureParam(screenSpaceShadowComputeShader, kernel, HDShaderIDs._DeferredShadowTextureUAV, deferredShadowRT);
 
                 int deferredShadowTileSize = 16; // Must match DeferreDirectionalShadow.compute
-                int numTilesX = (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
+                int numTilesX = hdCamera.frameSettings.enableStereo ? ((hdCamera.actualWidth / 2) + (deferredShadowTileSize - 1)) / deferredShadowTileSize : (hdCamera.actualWidth + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
                 int numTilesY = (hdCamera.actualHeight + (deferredShadowTileSize - 1)) / deferredShadowTileSize;
 
-                // TODO: Update for stereo
-                cmd.DispatchCompute(screenSpaceShadowComputeShader, kernel, numTilesX, numTilesY, 1);
+                int numEyes = hdCamera.frameSettings.enableStereo ? 2 : 1; // FIXME generalize this once XRSDK support for >2 eyes is ready
```
sebastienlagarde: just a random thought, but if you want to have to avoid to update all the code later, maybe we should add a hdCamera.numEye, and init it with other parameter in HDCamera at a single place and then always use that. So we have  int numEyes = hdCamera.frameSettings.enableStereo ? 2 : 1; at a single place in the code :)

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
sebastienlagarde: ok, I am stupide, so the multicompile is define outside of the shader itself, it is provide at shader compilation right? So how can we strip it if our project is not using VR? or how can we avoid to generate this variant in the player? In HDRP we use variant stripper, is it still working?

sebastienlagarde:
ok, I am fine with the PR, just try to clean a bit the code for the  
int numEyes = hdCamera.frameSettings.enableStereo ? 2 : 1; mention ,
and would like tought of Julien about the comment above.

fix merge issue with changelog and launch katana. Thanks

com.unity.render-pipelines.high-definition/Runtime/Sky/ProceduralSky/ProceduralSky.shader
nerites: I don't know what #pragma multi_compile is used for, so please feel free to question/inform me on the subject!

In the native ShaderKeywords class, there is the following line:
> CreateBuiltin("UNITY_SINGLE_PASS_STEREO", kSinglePassStereo, kShaderKeywordBuiltinAutoStripped);

I have no idea how that interacts with variant generation in the player, but does that add any info for you?

sebastienlagarde: So  #pragma multi_compile will create several variant of a shader. When you have 
#pragma multi_compile _ UNITY_SINGLE_PASS_STEREO STEREO_INSTANCING_ON STEREO_MULTIVIEW_ON
it mean that we will generate a 4 skyShader variant, one with each type of define.

We are suppose to explicitly declare this multicompile. But it looks like for VR, this is implicit.
But then I don't know exactly where this is control, maybe when you build a player you know what you want to support?

>CreateBuiltin("UNITY_SINGLE_PASS_STEREO", kSinglePassStereo, kShaderKeywordBuiltinAutoStripped);
This is just the keyword declaraction on C++ side.

sebastienlagarde:
I merged master and fix conflict

Katana is green: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=stereo_msaa_skybox&automation-tools_branch=add-platform-filter&unity_branch=trunk#

Ready to merge

sebastienlagarde:
Note: @nerites could you add a section in HDRP documentation: https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki to explain how to enable VR (from users point of view), and maybe also add a note about what is currently supported and what is not (i.e a bit like the tab that you have done). thanks!

nerites:
> Note: @nerites could you add a section in HDRP documentation: https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki to explain how to enable VR (from users point of view), and maybe also add a note about what is currently supported and what is not (i.e a bit like the tab that you have done). thanks!

https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki/VR-in-HDRP

PR #2016, opened 2018-09-19
Merge HDRP/Staging

RSlysz:
- Fix issue when disabling contact shadow
- Hide environment ligthing panel
- Fix wrong tab allocation for reflection probe

Katana is green: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fstaging&automation-tools_branch=add-platform-filter&unity_branch=trunk#


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -212,6 +212,8 @@ public int GetDecalAtlasMipCount()
         ComputeBuffer m_DebugScreenSpaceTracingData = null;
         ScreenSpaceTracingDebug[] m_DebugScreenSpaceTracingDataArray = new ScreenSpaceTracingDebug[1];
 
+        static int livingHDRenderPIpelineCounter = 0;
```
JulienIgnace-Unity: There's a type (P'I'peline)
Why do we need  this counter though? How can we have more than one pipe active at a time?

PR #2027, opened 2018-09-21
Micro shadowing

FrancescoC-unity:
### Purpose of this PR
Added micro shadowing support that fakes some directional light shadowing based on AO and normal data. This is based on "The Technical Art of Uncharted 4" [Brinck and Maximov 2016] and adds an opacity slider to control the effect. 

comparison

Worth noting that this is heavily dependent on good content. Dark AO maps will overly darken the objects. 

---
### Release Notes
Implement micro shadowing support for directional light. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=micro-shadowing&automation-tools_branch=add-platform-filter&unity_branch=trunk  [Still running] 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
I used specular occlusion as we don't have AO around all the time, while this is not strictly correct (AO is used to approximate a visibility cone), it still leads pleasant enough results.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Shadow.hlsl
```diff
@@ -123,6 +122,14 @@ float GetPunctualShadowClosestDistance( ShadowContext shadowContext, SamplerStat
 
 #endif
 
+// From "The Technical Art of Uncharted 4" [Brinck and Maximov 2016]
+float GetMicroshadowing(float NdotL, float AO)
```
sebastienlagarde: Move this to CommonLighting.hlsl

FrancescoC-unity:
Katana is now all green

PR #2035, opened 2018-09-21
Fix tangent maps

EvgeniiG:


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitDataIndividualLayer.hlsl
```diff
@@ -294,7 +294,8 @@ float ADD_IDX(GetSurfaceData)(FragInputs input, LayerTexCoord layerTexCoord, out
 
 #ifdef _TANGENTMAP
     #ifdef _NORMALMAP_TANGENT_SPACE_IDX // Normal and tangent use same space
-    float3 tangentTS = SAMPLE_UVMAPPING_NORMALMAP(_TangentMap, sampler_TangentMap, layerTexCoord.base, 1.0);
+    // Tangent space vectors always use only 2 channels.
+    float3 tangentTS = UnpackNormalmapRGorAG(SAMPLE_UVMAPPING_TEXTURE2D(_TangentMap, sampler_TangentMap, layerTexCoord.base), 1.0);
```
EvgeniiG: Note: I bypass the surface gradient here.
What this means for triplanar is that we blend the vectors directly.
That's probably not a good idea, while blending in the slope space usually is.
I need to think about it. :-)

EvgeniiG:
Added a comment to the code:

Note: I bypass the surface gradient here.
What this means for triplanar is that we blend the vectors directly.
That's probably not a good idea, while blending in the slope space usually is.
I need to think about it. :-)

EvgeniiG:
Let's merge this, I will revisit better blending later.

PR #2036, opened 2018-09-21
Avoid recomputing FGD tables every frame

FrancescoC-unity:
### Purpose of this PR
Check if FGD tables for AxF have been already init, if so don't recompute them again. 
---
### Release Notes
Avoid preintegration of AxF's FGD tables every frame. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=AxF-fix-preintegrated-FGD-every-frame&automation-tools_branch=add-platform-filter&unity_branch=trunk [Still running]

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.cs
@@ -181,6 +181,7 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
             m_preIntegratedFGD_Ward = null;
             m_preIntegratedFGDMaterial_Ward = null;
             m_preIntegratedFGDMaterial_CookTorrance = null;
+            m_precomputedFGDTablesAreInit = false;
 
             // LTC data
             CoreUtils.Destroy(m_LtcData);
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.cs
```diff
@@ -196,6 +198,8 @@ public override void RenderInit(CommandBuffer cmd)
                 CoreUtils.DrawFullScreen(cmd, m_preIntegratedFGDMaterial_Ward, new RenderTargetIdentifier(m_preIntegratedFGD_Ward));
                 CoreUtils.DrawFullScreen(cmd, m_preIntegratedFGDMaterial_CookTorrance, new RenderTargetIdentifier(m_preIntegratedFGD_CookTorrance));
             }
+
+            m_precomputedFGDTablesAreInit = true;
```
sebastienlagarde: you miss to reset it to false 
m_precomputedFGDTablesAreInit = false 
in Cleanup() function

sebastienlagarde: Note: HDRP is frequently rebuild (i.e init/cleanup cycle happen often)

FrancescoC-unity: Ha, yes,  I ignored the existence of the Cleanup function, I just hastily did the change :-) Will fix immediately!

sebastienlagarde:
Note: katana is green except the test from LW that was failing on master so we are good. Merging: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=AxF-fix-preintegrated-FGD-every-frame&automation-tools_branch=add-platform-filter&unity_branch=trunk#

PR #2038, opened 2018-09-21
Hdrp/gc cleanup

Paulm-Unity:
### Purpose of this PR
Cleanup some memory leaks in HDRP

---
### Testing status
**Katana Tests**:  Green
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fgc-cleanup&unity_branch=trunk&automation-tools_branch=master

### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Low


sebastienlagarde:
Thanks for this cleanup Paul!
Poking @EvgeniiG @JulienIgnace-Unity @anisunity  for awareness

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -196,6 +196,17 @@ public int GetDecalAtlasMipCount()
...
+        RenderTargetIdentifier[] m_MRTWithSSS;
+        string m_ForwardPassProfileName;
+
+        int[] m_DepthPyramidMipLevelOffsetsX;
+        int[] m_DepthPyramidMipLevelOffsetsY;
+
+        Vector2Int m_PyramidSizeV2I = new Vector2Int();
+        Vector4 m_PyramidSizeV4F = new Vector4();
+        Vector4 m_PyramidScaleLod = new Vector4();
+        Vector4 m_PyramidScale = new Vector4();
```
Chman: I'm not sure I understand why these four fields need to be promoted to class members. In fact, they will make the code slower as they're now allocated on the heap whereas they were fine on the stack (and calling `Set()` on them is slower).

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.cs
```diff
@@ -304,6 +304,8 @@ public void BeginCull()
                 m_BoundingDistances[0] = DecalSystem.instance.DrawDistance;
                 m_NumResults = 0;
                 m_CullingGroup = new CullingGroup();
+                // GC.Alloc
+                // m_CullingGroup gets .Dispose() called on it later, not clear why this leaks
```
Chman: A GC alloc isn't a leak. It does alloc because `CullingGroup` is a class and a reference is new'd on every frame.

PR #2054, opened 2018-09-26
culling group manager, to get rid of more GC.Alloc's

Paulm-Unity:
### Purpose of this PR
Centralize culling group allocation. Remove GC.Alloc's related to culling group context allocation.

---
### Testing status
**Katana Tests**: Green
https://katana.hq.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fculling-group-manager&unity_branch=trunk&automation-tools_branch=master

### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/CullingGroupManager.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using UnityEngine.Experimental.Rendering;
+using UnityEngine.Rendering;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+
+    public class CullingGroupManager
+    {
+        static CullingGroupManager m_Instance;
```
Chman: I've seen this pattern in a few places in the SRP repository, this isn't a correct/safe singleton pattern (with regards to thread safety etc). Don't change it, we'll do a pass on all of them at some point, I'm just leaving this note so I don't forget about it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/CullingGroupManager.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            get
+            {
+                if (m_Instance == null)
+                    m_Instance = new CullingGroupManager();
+                return m_Instance;
+            }
+        }
+
+        private List<CullingGroup> m_FreeList = new List<CullingGroup>();
```
Chman: This list is used as a stack (push/pop behavior), so I'd use a [Stack&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1?view=netframework-4.7.2) for consistency and speed.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/CullingGroupManager.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using UnityEngine.Experimental.Rendering;
+using UnityEngine.Rendering;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+
+    public class CullingGroupManager
+    {
+        static CullingGroupManager m_Instance;
```
Paulm-Unity: Changed List to Stack, have no idea where Thomas's comment about it went so can't reply to it. Seems to have disappeared when I pushed my new commit...

PR #2057, opened 2018-09-26
Added a custom callback to bypass HDRP rendering for a specific camera

JulienIgnace-Unity:
### Purpose of this PR
Added a custom callback to bypass HDRP rendering for a specific camera.
This works by adding an event to the HDAdditionalCameraData component of a camera.

### Testing status
Manual Testing in HDRP
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FcustomRenderCallback&automation-tools_branch=add-platform-filter&unity_branch=trunk


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -854,15 +854,6 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                 // This is the main command buffer used for the frame.
                 var cmd = CommandBufferPool.Get("");
 
-                // Specific pass to simply display the content of the camera buffer if users have fill it themselves (like video player)
```
tlaedre: FullscreenPassthrough is still needed to be able to use UI Canvas cameras without having to add an empty callback *from code* to simulate pass through. The ExecuteCommandBuffer+Release+Submit doesn't really make sense to me, though., it should be moved above the CBPool.Get and just do nothing (but arguably still invoke beginCameraRendering)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -927,6 +918,14 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                         VolumeManager.instance.Update(hdCamera.volumeAnchor, hdCamera.volumeLayerMask);
                     }
 
+                    if (additionalCameraData != null && additionalCameraData.ExecuteCustomRender(renderContext, hdCamera))
+                    {
+                        renderContext.ExecuteCommandBuffer(cmd);
```
tlaedre: I'm not sure if ExecuteCommandBuffer makes sense here or not, maybe it does. But if it does, should it not invoke *before* the custom render callback in that case?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs
```diff
@@ -30,7 +31,6 @@ public enum RenderingPath
         {
             UseGraphicsSettings,
             Custom,  // Fine grained
-            FullscreenPassthrough  // Hard coded path
```
tlaedre: Still needed, see below.

PR #2059, opened 2018-09-26
Update shadow system

alelievr:
### Purpose of this PR
Clean shadow files
---

---
### Testing status
**Katana Tests**:  [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=UpdateShadowSystem&automation-tools_branch=add-platform-filter&unity_branch=trunk)

**Manual Tests**:  Local tests

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Low


sebastienlagarde:
Katana is green (except metal that still running as always), merging as better fro users if we remove the old code before release

PR #2060, opened 2018-09-26
Optimize SSR by using a structured buffer instead of a constant buffer for mip level offsets

FrancescoC-unity:
### Purpose of this PR

Optimize SSR by sending depth pyramid mip offsets via a structured buffer rather than cbuffer

This goes around a quirk in the consoles' HLSL compiler that caused over allocation of VGPRs.

With this change on PS4:
VGPRs:  44 -> 28
SGPRs: 54+2 -> 46+2
Timing (64spp): 7.819ms -> 6.193ms

On Xbox One X
VGPRs: 36 -> 32
SGPRs: 64+2 -> 48+2
Timing (64spp): 14.683 -> 7.835

This increases the fetch cost, but we are never bound by it on this pass.

Also, this PR includes a small bug fix: if the color pyramid is not available, then bind a white texture instead. This solves a problem during the first frame of rendering when the color pyramid is invalid. 

Note that SSR is still producing NaNs on Xbox One due to NaNs appearing at the second frame (at least in my example) in SH coefficients. I opened a relevant bug with repro and GI team has been made aware of it. 

---
### Release Notes

Optimising SSR by going around a console HLSL compiler quirk, passing depth pyramid mip  offsets via structured buffer instead of using a constant buffer. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=ssr-lut-to-structure-buffer&automation-tools_branch=add-platform-filter&unity_branch=trunk [Mac still running, windows's green]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


EvgeniiG:
"Also, this PR includes a small bug fix: if the color pyramid is not available, then bind a white texture instead." Why would it not be available? I think the proper fix is to make the color pyramid available if SSR is on.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -180,8 +176,8 @@ void ScreenSpaceReflectionsTracing(uint2 positionSS : SV_DispatchThreadID)
         rayPos.xy += raySign.xy * satEdgeDist;
 
         int2 mipCoord  = (int2)rayPos.xy >> mipLevel;
-        int2 mipOffset = int2(depthPyramidMipLevelOffsetsX[(mipLevel + 0) >> 1],
-                              depthPyramidMipLevelOffsetsY[(mipLevel + 1) >> 1]);
+        int2 mipOffset = int2(_SsrDepthPyramidMipLevelOffsets[(mipLevel + 0) >> 1].x,
+                              _SsrDepthPyramidMipLevelOffsets[(mipLevel + 1) >> 1].y);
```
EvgeniiG: Suggestion for improvement: use
int2 mipOffset = _SsrDepthPyramidMipLevelOffsets[mipLevel].xy;
and update the LUT accordingly (it will have some duplicate data).
It most likely makes sense to trade storage (a few byes) to halve the number of fetches.

FrancescoC-unity:
> "Also, this PR includes a small bug fix: if the color pyramid is not available, then bind a white texture instead."
> 
> Why would it not be available? I think the proper fix is to make the color pyramid available if SSR is on.

I wasn't clear on that, it is referring to the first frame, when obviously we do not have it kicking around (as it is generated after SSR).

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
FrancescoC-unity: That's a good point and I actually wanted to do that, but when I saw fetch was far from being a bottleneck I moved on and eventually forgot about it :-)

FrancescoC-unity:
Profiled the new change reducing the number of fetches on Xbox One X and it decreased the cost by roughly 5%, I am guessing mostly due to having 3 VALU ops less per iteration of the loop

PR #2061, opened 2018-09-26
Fix DynamicValueMaterialSlot decimal separator

pbbastian:
### Purpose of this PR
Found yet another instance where we incorrectly convert a float to a string, which causes shader compile errors on certain locales. Bumping importer version because it changes shader generation for some people.

---
### Release Notes
The dynamic value slot type now uses the correct decimal separator during HLSL generation.

---
### Testing status
**Katana Tests**: Running [here](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Ffix-dynamicvaluematerialslot-decimal-separator&automation-tools_branch=add-platform-filter&unity_branch=trunk)

**Manual Tests**: Open up Unity and check that stuff still generally works. This is the same fix that we have been applying to other instances of the issue for a long time.

**Automated Tests**: Nothing new.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---


pbbastian:
Change log updates approved by @Katstodian on Slack

PR #2062, opened 2018-09-26
Update shader processor and fix render resources asset and motion vectors

sebastienlagarde:
Goal of this PR is:
- To remove Autodesk shader from resource asset pipeline
- Apply the stripper rules to all shaders regardless of their name, so they can work with shader graph
- Fix issue with motion vectors (Terrain, AxF, Fabric, StackLit, connect objectMotionVector and MotionVector option)

Katana: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=Fix-shader-processor-to-work-correctly-with-all-shader&automation-tools_branch=add-platform-filter&unity_branch=trunk)


sebastienlagarde:
@anisunity Observe that there is additional fix for motion vector

PR #2065, opened 2018-09-26
Tests/framework/fixes

RemyUnity:
Fix the "replace template" function of the test result window.
Option to disable XML logging (prevent the freeze of test runner if an image comparison failed).

### DIFF
--- a/com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
+++ b/com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
@@ -83,7 +83,7 @@ namespace UnityEngine.TestTools.Graphics
             {
                 foreach (var testCase in provider.GetTestCases())
                 {
-                    TestCaseParameters parms = new TestCaseParameters( new object[]{ m_OutputAsString? testCase.ScenePath : (object) testCase } )
+                    TestCaseParameters parms = new TestCaseParameters( new object[]{ (m_TestParameters==TestParameters.ScenePath)? testCase.ScenePath : (object) testCase } )
                     {
                         ExpectedResult = new object(),
                         HasExpectedResult = true,
--- a/com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
+++ b/com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
 using UnityEngine.Rendering;
 using Attribute = System.Attribute;
 
@@ -9,18 +10,25 @@ namespace UnityEngine.TestTools.Graphics
 {
     /// <summary>
     /// Marks a test which takes <c>GraphicsTestCase</c> instances as wanting to have them generated automatically by
-    /// the scene/reference-image management feature in the framework. 
+    /// the scene/reference-image management feature in the framework.
     /// </summary>
-    public class UseGraphicsTestCasesAttribute : Attribute, ITestBuilder
+    public class UseGraphicsTestCasesAttribute : UnityEngine.TestTools.UnityTestAttribute, ITestBuilder
     {
         string m_ReferenceImagePath = string.Empty;
 
-        public UseGraphicsTestCasesAttribute()
-        {}
+        NUnitTestCaseBuilder _builder = new NUnitTestCaseBuilder();
 
-        public UseGraphicsTestCasesAttribute(string referenceImagePath)
+        bool m_OutputAsString = false;
+
+        public UseGraphicsTestCasesAttribute(bool outputAsString = false)
+        {
+            m_OutputAsString = outputAsString;
+        }
+
+        public UseGraphicsTestCasesAttribute(string referenceImagePath , bool outputAsString = false)
         {
             m_ReferenceImagePath = referenceImagePath;
+            m_OutputAsString = outputAsString;
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


RemyUnity:
Added a commit to be able tu run the tests individually. Will help a lot.

com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
```diff
@@ -2,25 +2,33 @@
...
         string m_ReferenceImagePath = string.Empty;
 
-        public UseGraphicsTestCasesAttribute()
-        {}
+        NUnitTestCaseBuilder _builder = new NUnitTestCaseBuilder();
 
-        public UseGraphicsTestCasesAttribute(string referenceImagePath)
+        bool m_OutputAsString = false;
+
+        public UseGraphicsTestCasesAttribute(bool outputAsString = false)
```
richard-fine: It'd be much clearer if you used an enum instead of a bool for this. `[UseGraphicsTestCases(false)]` doesn't have an obvious meaning.

richard-fine: I think you can do it as a freestanding enum, it doesn't have to be nested inside UseGraphicsTestCasesAttribute. Like `ImageOutputMode.TestProperty` vs `ImageOutputMode.Disk` or something.

richard-fine: This said, will outputting directly to disk actually work properly outside of the Editor?

com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
```diff
@@ -97,5 +110,19 @@ IEnumerable<TestMethod> ITestBuilder.BuildFrom(IMethodInfo method, Test suite)
             Console.WriteLine("Generated {0} graphics test cases.", results.Count);
             return results;
         }
+
+        public static IEnumerable<GraphicsTestCase> GraphicsTestCaseList()
+        {
+            return new UnityEditor.TestTools.Graphics.EditorGraphicsTestCaseProvider(null).GetTestCases();
```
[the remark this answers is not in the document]
RemyUnity: Will totally remove this method. It's something that I forgot to erase when testing. Not used.

com.unity.testframework.graphics/Runtime/UseGraphicsTestCasesAttribute.cs
RemyUnity: > This said, will outputting directly to disk actually work properly outside of the Editor?

Maybe, but all direct output to disk code is encapsulated in editor code.

RemyUnity: And I totally confused the subject of your comment with other parts of the pr, as this bool setting is used to fix the inability to run the tests individually.
But I will replace by an enum.

richard-fine: See, bools are confusing :)

If it's fixing the ability to run the tests individually, when are we ever _not_ going to want to do it? And how does the fix actually work?

RemyUnity:
@richard-fine Are those changes okay for you, or do you think that more things should be corrected ?

richard-fine:
@RemyUnity it's still not clear to me exactly what is going on in the change to `UseGraphicsTestCasesAttribute`. What is your solution actually doing, and why does it work?

RemyUnity:
@richard-fine 
I did some tests with how you can generate test cases with test runner, and apparently if you generate a test case list with a custom class or struct IEnumerable, you will encounter the issue of running all test cases for a method instead of running a single one.
If you generate the list with for example, string, float or int IEnumerable, you can run a test case individually.

This is apparently how NUnit is supposed to work.

So my workaround is to generate the test cases list from the scenes path (string) list instead of the original GraphicTestCase class list. I then use UseGraphicsTestCasesAttribute.GetCaseFromScenePath(scenePath) to re-create the GraphicTestCase object in the test method.

I guess it's not optimal, but no people can run a single graphic test locally in the editor to check for their fixes in case of single test failure.

Also, the other fix in this PR, disables logging the actual and diff images in the TestContent and directly writes them to the filesystem.
The previous behaviour is kept by default (for Katana), but when running tests locally in editor, you can use the menu item "Tests/XML Logging/Disable" that creates a "DisableXMLLogging" file in the Library folder that will act as a local setting to what was mentioned earlier.

When tests are run locally in editor, going through the TestContext method to encode PNG + encode base 64 and then wait for the end of the tests to decode and write to file (and TestRunner to handle the data also I guess) will make the editor hang a long time at the end of the tests run.
I tried to run 22 tests forcing them to fail (no reference image, and thus, generate 22 actual images without diffs), it took more than 3 minutes with the original behaviour, and a bit less than 30 seconds with this fix, so it is not a negligible gain.
Also, it is wrapped up in Unity_Editor conditions and because the checking file only exists in the local library, it should change the behaviour of our automated tests, but act only in developer machines.

richard-fine:
> This is apparently how NUnit is supposed to work.

How did you confirm that? From what you're describing this just sounds like a straight up NUnit bug that we should fix there, not by working around it here...

RemyUnity:
Fixed a compilation issue introduced in commit e45c5da

@richard-fine :
I found a way to run the tests individually without the enum :
When building the tests list with a GraphicsTestCase object as parameter, the TestMethod object must be create with a TestCaseData argument instead of TestCaseParameters. While TestCaseData inherits from TestCaseParameters, it has a method to set the readonly property "TestName", that is used I guess by testrunner in its TestFilter to run selected tests.

richard-fine:
Ah, awesome! Does the resulting test name (in the TestRunner) have the complete "MyGraphicsTest(SceneName)", or does it only say "SceneName"?

RemyUnity:
Only "SceneName", like it was originally.

richard-fine:
Alright, if that's the same as it was before then that's OK. It might be nice to try out a pattern that includes the full test name using https://github.com/nunit/docs/wiki/Generation-of-Test-Names-Spec at some point, but this is :+1: from me for now anyway.

sebastienlagarde:
Katana test are green: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=tests%2Fframework%2Ffixes&automation-tools_branch=add-platform-filter&unity_branch=trunk#

PR #2066, opened 2018-09-26
Fixed re-create of DepthPyramid RT when loading renderdoc

JulienIgnace-Unity:
### Purpose of this PR
Fixed re-create of DepthPyramid RT when loading renderdoc

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Ffix-depth-pyramid-create&automation-tools_branch=add-platform-filter&unity_branch=trunk#


sebastienlagarde:
Merging as katana is block but windows editor is green which is sufficient for this single line PR.

PR #2070, opened 2018-09-27
Disable object motion vector optimization

sebastienlagarde:
Disable motion vector optimization until we get a C++ PR that allow to enable it.


anisunity:
As discussed, we should also move back the RenderObjectVelocity next to the Camera Velocity!

sebastienlagarde:
@anisunity added the move of the object render

PR #2071, opened 2018-09-27
Send missing stereo informations to compute shaders

FrancescoC-unity:
### Purpose of this PR
Currently stereo matrices where not configured when stereo was disabled and some information (world space stereo camera pos and stereo projection matrices) where missing altogether. 

This caused Screen space shadows to be broken in master. 

---
### Release Notes
Fixed Contact shadows in master. Allow access to additional stereo information in compute shaders.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix-compute-shader-XR-issues-when-no-XR&automation-tools_branch=add-platform-filter&unity_branch=trunk [ Running ]

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Medium - it may impact other shaders using stereo informations.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
@@ -24,7 +24,7 @@
     #define unity_CameraInvProjection unity_StereoCameraInvProjection[unity_StereoEyeIndex]
     #define unity_WorldToCamera unity_StereoWorldToCamera[unity_StereoEyeIndex]
     #define unity_CameraToWorld unity_StereoCameraToWorld[unity_StereoEyeIndex]
-    #define _WorldSpaceCameraPos unity_StereoWorldSpaceCameraPos[unity_StereoEyeIndex]
+    #define _WorldSpaceCameraPos _WorldSpaceCameraPosStereo[unity_StereoEyeIndex].xyz
 #endif
 
 #define UNITY_LIGHTMODEL_AMBIENT (glstate_lightmodel_ambient * 2)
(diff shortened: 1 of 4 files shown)


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -24,7 +24,7 @@
     #define unity_CameraInvProjection unity_StereoCameraInvProjection[unity_StereoEyeIndex]
     #define unity_WorldToCamera unity_StereoWorldToCamera[unity_StereoEyeIndex]
     #define unity_CameraToWorld unity_StereoCameraToWorld[unity_StereoEyeIndex]
```
FrancescoC-unity: Should we rename the other fields so that they use variables from UnityPerPassStereo instead?

nerites:
Hi Francesco, thanks for catching this! When I first started enabling stereo for compute shaders, I routed the projection and view matrices through the stereo matrices on the C# side- see lines 313-320 in HDCamera.cs. Would that work for the changes you're making here? It'd be nice if we had a unified way of routing through stereo matrices for compute.

FrancescoC-unity:
> Hi Francesco, thanks for catching this! When I first started enabling stereo for compute shaders, I routed the projection and view matrices through the stereo matrices on the C# side- see [here](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/8cd1f5faf29d411a2230e3a8694b639fc21f92ea/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs#L312). Would that work for the changes you're making here? It'd be nice if we had a unified way of routing through stereo matrices for compute.

Hey! I think it would work, I am just slightly wary of duplicating  code. By putting the ConfigureStereoMatrices outside of the` if (m_frameSettings.enableStereo)`, we'd be consistent on how those matrices are computed and then sent over, regardless of stereo being enabled or not. We can move my changes to the part of code you linked, but wouldn't that be a sort-of duplication what's already happening in ConfigureStereoMatrices ?

nerites:
That's a great point and I totally agree with you. I made some changes to streamline the matrices as you suggested, could you check out https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2082 and see if that aligns better?

sebastienlagarde:
Hey, some I am ok with the change in #2082. I am merging #2082 in this current branch. 
It is fine to break camera relative and allow this PR as we know it doesn't work currently, but not sure to get why it is fully broken now. Still next step is to fix camera relative rendering :). We need this PR to fix the regular rendering

sebastienlagarde:
Katana is green and the PR include change from @nerites . Moving forward and merging.

PR #2075, opened 2018-09-28
Sg/hdrp automated tests

zadi0648:
### Purpose of this PR

Adds Automated Tests for HDRP Shader Graph.

Three of these scenes (81__) included an updated test settings which makes the scene do two image comparisons/assets; the second one makes sure the Shader Graph objects are consistent with HDRP/Lit objects.

In these same scenes you can play them and then press space to switch between the different objects.

The other three (82__) are standard image comparison.

---
### Release Notes

First SG HDRP automated tests!

---
### Testing status

I  confirmed that tests pass, that they take two images, that they fail when something is different, ect.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


zadi0648:
Oh, I forgot to add; you may notice that the transparent scene has any empty lower row. That is being distortion seems to be broken, and I'll readd them later. But we want to get automated test in soon.

PR #2080, opened 2018-09-28
2019.1/upgrade sceneview api

karl-:
### Purpose of this PR

2019.1.0a3 deprecates `SceneView.m_SceneLighting` and `SceneView.onSceneGUIDelegate`. This PR updates references to their new signatures (`sceneLighting` and `duringSceneGui`).

https://ono.unity3d.com/unity/unity/pull-request/73821/_/wb/expose-onPreSceneGUI

These properties were not marked `[Obsolete]` on the initial PR because SRP is included in the EditorTests package cocktail, and would therefore cause the tests to fail due to unexpected logs about deprecated API. 

Once this PR is merged to SRP and a new version is pushed to Katana, these properties will be marked obsolete on trunk.

---
### Testing status

Manual, no functionality is changed.

---
### Overall Product Risks

**Technical Risk**:  None

**Halo Effect**:  None

---


sebastienlagarde:
Note: We will need to create a 2018.3 first before merging this branch, or we will require to have #ifdef based on version around those change. Tought? @stramit

stramit:
Yeah, these should be guarded with a unity version tag and have both implementations.... or they should be supported via the script updater and happen automatically. We don't want to branch SRP just yet and landing this now makes sense.

karl-:
At the moment the script updater does not quietly update package API in tests (though it is planned), so it will need to use define guards until the API updater can run without triggering test failures. I'll amend the PR.

sebastienlagarde:
This is done:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2100

sebastienlagarde:
Ok, we can go forward with this PR for 2019.1. Need to solve conflict first

karl-:
@sebastienlagarde I've updated the branch and added `#define` guards, ready for review.

EvgeniiG:
The #define guards appear to be swapped, so this breaks on 2018.3.

sebastienlagarde:
Thanks, fixed the issue here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2208

PR #2083, opened 2018-09-29
Shader error badges on nodes

marctem:
### Purpose of this PR
We want to show shader errors on the nodes that caused them as much as possible. They still appear in the Console but now will also badge the node itself (for the preview shader only - the main shader compilation does not properly report errors out to C# yet).

---
### Release Notes
Shader Graph now shows errors on the nodes that caused them.

---
### Testing status
**Katana Tests**: Passed (https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=shadergraph%2Ferror-handling&unity_branch=graphics%2Fshadergraph%2Ferror-bindings)

**Manual Tests**: Created a node that generates broken shader code and verified the error badge appears on the node. QA also verified no behavior change with error handling.

**Automated Tests**: Added three tests in MaterialNodeTests to ensure error API behaves as expected. Fixed one test broken by API change.

---
### Overall Product Risks
**Technical Risk**: Low - new API in trunk, small change to existing node api
**Halo Effect**: Low - affects trunk code and bindings but limited to ShaderGraph

---
###Notes to Reviewers
Requires 19.1 trunk changes in https://ono.unity3d.com/unity/unity/pull-request/73478/_/graphics/shadergraph/error-bindings, so don't merge until after SRP 18.3 branchoff.


com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -388,6 +388,12 @@ void UpdateShaders()
                                     message.AppendLine("Shader compilation error in {3} at line {1} (on {2}):\n{0}", error.message, error.line, error.platform, node != null ? string.Format("node {0} ({1})", node.name, node.guid) : "graph");
                                     message.AppendLine(error.messageDetails);
                                     message.AppendNewLine();
+
+                                    var errNode = m_Graph.GetNodes<AbstractMaterialNode>().First(graphNode => graphNode == node);
+                                    if (errNode != null)
+                                    {
+                                        errNode.AddError(error);
```
pbbastian: I'm not sure that I like this architecture. `AbstractMaterialNode` should generally only be used for persistent data, not as a grabbag for temporary data. The issue here is that `PreviewManager` is part of the UI layer, and now you have an extra cycle in the update loop, where the graph changes, then `PreviewManager` uses that to generate preview shaders, and now `PreviewManager` feeds that back into the graph such that another UI system can read that.

I think `PreviewManager` should own this data, and maybe expose it via a method that takes a node ID. (Be sure to look at the temp ID, which is fast for lookups and won't change unless the window is closed.) Also see my comment below in `MaterialNodeView.cs` as that relates a lot to this.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -149,13 +149,17 @@ public void Initialize(AbstractMaterialNode inNode, PreviewManager previewManage
...
                 {
-                    IconBadge wrongPipeline = IconBadge.CreateError("The current render pipeline is not compatible with this master node.");
-                    Add(wrongPipeline);
-                    VisualElement title = this.Q("title");
-                    wrongPipeline.AttachTo(title, SpriteAlignment.LeftCenter);
+                    AttachError("The current render pipeline is not compatible with this master node.");
                 }
             }
 
+            if (node.hasErrors)
```
pbbastian: This function is only called when the node is added, so if the node gets into an error state during its lifetime the badge won't show. Also, how are badges removed again? There needs to be some form of communication from the `PreviewManager` that the error state has now changed, and that badges should be updated. Preferably in a fashion similar to how created/added nodes are tracked in `AbstractMaterialGraph`/`GraphEditorView`.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -196,6 +200,14 @@ public void Initialize(AbstractMaterialNode inNode, PreviewManager previewManage
             }
         }
 
+        public void AttachError(string errString)
+        {
+            var wrongPipeline = IconBadge.CreateError(errString);
```
pbbastian: Rename this one to `badge`

com.unity.shadergraph/Editor/Data/Nodes/AbstractMaterialNode.cs
```diff
@@ -437,9 +469,13 @@ public virtual void ValidateNode()
             GetOutputSlots(s_TempSlots);
             isInError |= s_TempSlots.Any(x => x.hasError);
             isInError |= CalculateNodeHasError();
-            hasError = isInError;
 
-            if (!hasError)
+            if (isInError)
+            {
+                // TODO: Specify actual validation errors after cleaning up this validation code
```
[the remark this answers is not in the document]
marctem: Fine if you don't want TODO's, but this is really part of whatever refactoring work we do about node processing - any errors anywhere need to be collected and bubbled up to where we can show them in the UI. Right now we just drop a lot of them on the floor.

Rather than a single card, it's a checkbox on refactors of shader compiler output processing, node validation, and preview generation.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
marctem: It's not a generic badge, though, it's specifically an error. IconBadge.CreateComment() uses different styling, and we could support that eventually when we need it. But for now we're just showing errors.

com.unity.shadergraph/Editor/Data/Nodes/AbstractMaterialNode.cs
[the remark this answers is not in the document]
pbbastian: Right, that makes a lot of sense :) But it needs to be tracked on Favro then, because otherwise it will be forgotten. It's alright to leave it in, but I'd like the comment to have more info on what's blocking these changes, and when it is expected to happen. Maybe even a reference to a Favro card.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
pbbastian: I was talking about the variable `wrongPipeline`, which should be renamed to `badge`, not the method.

marctem:
Closing to try again

PR #2089, opened 2018-10-01
Fix for contact shadows not rendering correctly when multiple views were active

FrancescoC-unity:
### Purpose of this PR
Fixes an issue with contact shadows when multiple cameras are present or when the editor is resized to a smaller dimension than previous. 
The issue was that to access the half res mip, I was jumping vertically by ScreenSize (i.e. viewport height), whereas it needs to be the size of the render target as it might be a mismatch for that. 

I have also updated the reference visual tests. Notice that they have been generated considering the fix present in PR [2071](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2071), so the following katana run should actually fail.

---
### Release Notes
Fixed contact shadows visual tests.
Fixed contact shadows with multiple cameras. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix-contact-shadows-multiple-view&automation-tools_branch=add-platform-filter&unity_branch=trunk

**Manual Tests**:  Run visual tests locally and checked them via result window. Tried contact shadows with game **and** scene view open at the same time

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low


FrancescoC-unity:
Katana is now all green for this :-)

PR #2093, opened 2018-10-02
Fix shadow near plane limit

alelievr:
### Purpose of this PR
Fix shadow near plane limit, 0.0001 instead of 0.1

---
### Release Notes

---
### Testing status
**Katana Tests**:  [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=FixShadowNearPlaneLimit&automation-tools_branch=add-platform-filter&unity_branch=2018.3%2Fstaging)

**Manual Tests**:  Manual tests

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -720,6 +717,10 @@ public void UpgradeLight()
                 }
                 needsIntensityUpdate_1_0 = false;
             }
+            if (m_Version != currentVersion)
```
sebastienlagarde: you need to test which version! otherwise any future version increase will do this code

something like

if (m_Version <= 2.0)
{
// ShadowPlane have been move to HDRP as default legacy unity clamp it to 0.1 and we need to be able to go below that. 
shadowNearPlane = m_Light.shadowNearPlane;
}

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -67,17 +67,14 @@ struct TimelineWorkaround
     [ExecuteInEditMode]
     public class HDAdditionalLightData : MonoBehaviour, ISerializationCallbackReceiver
     {
-        private const int currentVersion = 2;
+        private const int currentVersion = 3;
```
sebastienlagarde: // Add comment about what happen with version

// 3. Move shadowNearPlane to HDRP additional light data

alelievr: Done

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
alelievr: Yes, sorry I've fixed this

PR #2094, opened 2018-10-02
Allow materials to not receive SSR

FrancescoC-unity:
### Purpose of this PR
This adds a checkbox on a per-material basis that allow the user to exclude specific materials from receiving SSR (if unchecked). This could be useful in case of instabilities. 

onoff

This is achieved by modifying the stencil copy generation, the stencil copy has now also an UAV and, after a first clear+mark pixel that receive light pass, the content is OR'd to receive a SSR-specific bit flag.
The whole stencil handling would be faster if we could create an SRV to read directly the stencil or if we could guarantee access to SV_StencilRef in shader. In absence of those two, this was the fastest way I could think of. 

---
### Release Notes
Select on a per-material basis whether SSR should be received or not. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=allow-materials-to-not-receive-ssr&automation-tools_branch=add-platform-filter&unity_branch=trunk [Running]

**Manual Tests**: Tested both in forward only and deferred in editor and in game mode, both on PC and Xbox. 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Low

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
@@ -81,11 +81,10 @@ void ScreenSpaceReflectionsTracing(uint2 positionSS : SV_DispatchThreadID)
 {
     // If we are in forward only, the exclusion happens at a later stage. This is detected with _SsrStencilExclusionValue < 0
     // TODO: It would be good to find a way to skip some work here even in forward.
-    bool doesntReceiveSSR = false;
     UNITY_BRANCH if (_SsrStencilExclusionValue > 0)
     {
-        uint   stencilVal = UnpackByte(LOAD_TEXTURE2D(_StencilTexture, positionSS).r);
-        doesntReceiveSSR = (stencilVal & _SsrStencilExclusionValue) != 0;
+        uint stencilVal = UnpackByte(LOAD_TEXTURE2D(_StencilTexture, positionSS).r);
+        bool doesntReceiveSSR = (stencilVal & _SsrStencilExclusionValue) != 0;
 
         if (doesntReceiveSSR)
         {
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
@@ -39,6 +39,7 @@
 #ifdef SSR_TRACE
     // TEXTURE2D(        _NormalBufferTexture); // Declared in NormalBuffer.hlsl.
     // TEXTURE2D(        _DepthPyramidTexture);  // Declared in ShaderVariablesScreenSpaceLighting.hlsl
+    TEXTURE2D(           _StencilTexture);
     RW_TEXTURE2D(float2, _SsrHitPointTexture);
 #else
     // TEXTURE2D(        _NormalBufferTexture); // Declared in NormalBuffer.hlsl.
@@ -59,6 +60,7 @@ CBUFFER_START(UnityScreenSpaceReflections)
     float _SsrRoughnessFadeEndTimesRcpLength;
     float _SsrEdgeFadeRcpLength;
     int   _SsrDepthPyramidMaxMip;
+    int   _SsrStencilExclusionValue;
 CBUFFER_END
 
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -98,8 +99,6 @@ void ScreenSpaceReflectionsTracing(uint2 positionSS : SV_DispatchThreadID)
     float3 positionWS  = ComputeWorldSpacePosition(positionNDC, deviceDepth, UNITY_MATRIX_I_VP); // Jittered
     float3 V           = GetWorldSpaceNormalizeViewDir(positionWS);
 
-    bool killRay = deviceDepth == UNITY_RAW_FAR_CLIP_VALUE;
```
FrancescoC-unity: I removed this as I assumed it was to test for skybox, which should be marked as no-lighting anyway. Please let me know if this was not the only purpose and I'll re-instate the line

FrancescoC-unity: And now, thinking about it myself, it'd be good to keep around for Forward-sake, I'll reinstate it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -77,14 +79,27 @@ void WriteDebugInfo(uint2 positionSS, float4 value)
...
 {
+    // If we are in forward only, the exclusion happens at a later stage. This is detected with _SsrStencilExclusionValue < 0
+    // TODO: It would be good to find a way to skip some work here even in forward.
+    UNITY_BRANCH if (_SsrStencilExclusionValue > 0)
+    {
+        uint stencilVal = UnpackByte(LOAD_TEXTURE2D(_StencilTexture, positionSS).r);
+        bool doesntReceiveSSR = (stencilVal & _SsrStencilExclusionValue) != 0;
+        // Skip SSR for unlit materials as well. 
+        doesntReceiveSSR = doesntReceiveSSR || (stencilVal == 0);
+        if (doesntReceiveSSR)
```
EvgeniiG: Again, here you could just check the SSR bit. If it's not set (which is the case for 0 stencil), you set (killRay = true) or return. As is, the code is a bit awkward.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
EvgeniiG: This is indeed the skybox test.
We always have a depth prepass in forward. I think we should remove this test.
Doesn't it work?

EvgeniiG: Have you looked into using the stencil test during the resolve pass as well? Can't recall whether it makes sense or not.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -158,7 +158,8 @@ public enum StencilBitMask
             Clear                           = 0,    // 0x0
             LightingMask                    = 7,    // 0x7  - 3 bit
             Decals                          = 8,    // 0x8  - 1 bit
-            DecalsForwardOutputNormalBuffer = 16,   // 0x10  - 1 bit
+            DecalsForwardOutputNormalBuffer = 16,   // 0x10 - 1 bit
+            DoesntReceiveSSR                = 32,   // 0x20 - 1 bit
```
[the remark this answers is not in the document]
FrancescoC-unity: The main reason is performance. Hopefully receiving SSR is the standard and not-receiving is the exception. If we mark in stencil what should receive SSR, it will result in lots of stencil pass and the stencil copy update will cost more (a full screen is around 0.45ms on Scorpio)

com.unity.render-pipelines.high-definition/Runtime/Material/SharedRTManager.cs
```diff
@@ -61,7 +61,7 @@ public void InitSharedBuffers(GBufferManager gbufferManager, RenderPipelineSetti
             m_CameraDepthBufferMipChain = RTHandles.Alloc(ComputeDepthBufferMipChainSize, colorFormat: RenderTextureFormat.RFloat, filterMode: FilterMode.Point, sRGB: false, enableRandomWrite: true, name: "CameraDepthBufferMipChain");
 
             // Technically we won't need this buffer in some cases, but nothing that we can determine at init time.
-            m_CameraStencilBufferCopy = RTHandles.Alloc(Vector2.one, depthBufferBits: DepthBits.None, colorFormat: RenderTextureFormat.R8, sRGB: false, filterMode: FilterMode.Point, name: "CameraStencilCopy"); // DXGI_FORMAT_R8_UINT is not supported by Unity
+            m_CameraStencilBufferCopy = RTHandles.Alloc(Vector2.one, depthBufferBits: DepthBits.None, colorFormat: RenderTextureFormat.R8, sRGB: false, filterMode: FilterMode.Point, enableRandomWrite: true, name: "CameraStencilCopy"); // DXGI_FORMAT_R8_UINT is not supported by Unity
```
EvgeniiG: I don't think it needs the UAV flag, does it?

EvgeniiG: Nevermind, I see how it works. :-)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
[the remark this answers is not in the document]
EvgeniiG: Keep in mind that our 1-bit-at-a-time stencil hacks are going to go away. :-)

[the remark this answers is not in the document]
sebastienlagarde: >Keep in mind that our 1-bit-at-a-time stencil hacks are going to go away. :-)
Finger cross :), I will believe it when it is in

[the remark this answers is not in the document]
sebastienlagarde: I am fine with negation as the default is activation. + Update, we will not have stencil read before 2019.2 :(

FrancescoC-unity:
Katana has now passed against 2018.3

PR #2099, opened 2018-10-02
fix decal draw order on transparent materials

Paulm-Unity:
Fixed decal draw order on transparent material

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=HDRP%2Ffogbugz-1073011&unity_branch=2018.3%2Fstaging


sebastienlagarde:
Note: katana have fail. 
2004_animationcookie is instable, you can disregard it, however 9005_Decals-stress-test is broken by your PR.

PR #2100, opened 2018-10-02
Add guard for 2019.1 function to stay compatible with 2018.3

sebastienlagarde:
Add guard for 2019.1 function to stay compatible with 2018.3.

Just launched the auto upgrade script and guarded all the updated function (here GPUFence)

Tested manually with latest 2019.1 and 2018.3


sebastienlagarde:
@stramit : Force push here (I have edit the rights for this PR), as we need to fix master ASAP.

PR #2112, opened 2018-10-03
Adjust blits/postprocess for stereo y-flip fixes in 2019.1 [DO NOT MERGE]

nerites:
### Purpose of this PR
Fixes for stereo y-flip recently made it into Unity Editor trunk. The HDRP final blit/renderpostprocess was previously compensating for incorrect y-flip, and now that the fix is in trunk, stereo for HDRP is upside-down. This PR resolves the y-flip issue while maintaining back-compatability with older Editor versions. 

Although, since tip of SRP master already depends on API that's new with 2019.1, maybe there's no point in maintaining that back compatability.

---
### Release Notes
Handles y-flip consistently with the Editor fixes in @brandon-unity3d's Editor PR: https://ono.unity3d.com/unity/unity/pull-request/72904/_/xr/graphics/y-flip-fix

Dependent on shader changes in the following Editor PR: 
https://ono.unity3d.com/unity/unity/pull-request/75417/_/xr_blitcopy

---
### Testing status
**Katana Tests**: 
Unfortunately Katana tests will show failures due to reference image compression: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=srp_stereoyflip&unity_branch=xr_blitcopy&automation-tools_branch=add-platform-filter

Tested in test scene. 
image

Nothing.

Template test project should demonstrate this.
Currently we expect HDRP to work with single-pass double-wide VR, in forward-only mode. 

---
### Overall Product Risks
**Technical Risk**:
Low- changed the flip of the final blit only.

**Halo Effect**: 
None

---
### Comments to reviewers


sebastienlagarde:
So, if I get it, it also mean that most of our debgu mode will be incorrectly flip (as it take another path compare to postprocess path). Still not sure if many people use debgu mode at runtime when they ahve a VR headset :). But be careful with that, try for example the tile lighting debug mode to see if it is reversed in game mode. Then try a full screen debug mode like SSAO

sebastienlagarde:
Hey, so for debug we use this:

  if (ShouldFlipDebugTexture())
                {
                    // Texcoord are already scaled by _ScreenToTargetScale but we need to account for the flip here anyway.
                    input.texcoord.y = 1.0 * _ScreenToTargetScale.y - input.texcoord.y;
                }


// Debug rendering is performed at the end of the frame (after post-processing).
// Debug textures are never flipped upside-down automatically. Therefore, we must always flip manually.
bool ShouldFlipDebugTexture()
{
    #if UNITY_UV_STARTS_AT_TOP
        return (_ProjectionParams.x > 0);
    #else
        return (_ProjectionParams.x < 0);
    #endif
}

But not sure if it will apply in your case. We use this test because we can copy texture that are flipped and need to unflip them when displaying them in debug mode.

nerites:
Hi Seb, I got some help from Brandon on this issue, and a single-line change to some internal shaders allows us to use the same Blit that is used for non-stereo rendering. Unfortunately this gates this PR on an Editor PR, but in the spirit of cramming as many changes in as possible before 19.1 I think it should be ok?

PR #2116, opened 2018-10-04
Begin/End profiler section mismatch

ArchyInf:
CommandBuffer is executed before profiling section is closed.

### Purpose of this PR
Profiler throws errors due to profiling section not correctly being closed.

### DIFF
--- a/com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
+++ b/com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
@@ -232,13 +232,14 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
                 setupToUse.Setup(renderer, ref renderingData);
                 renderer.Execute(context, ref renderingData);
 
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -232,13 +232,14 @@ public static void RenderSingleCamera(LightweightRenderPipeline pipelineInstance
                 setupToUse.Setup(renderer, ref renderingData);
                 renderer.Execute(context, ref renderingData);
 
-                context.ExecuteCommandBuffer(cmd);
-                CommandBufferPool.Release(cmd);
-                context.Submit();
 #if UNITY_EDITOR
                 Handles.DrawGizmos(camera);
```
phi-lira: This should be moved as well to after the submit.

phi-lira:
Since there was no interaction on this PR I've done the changes in the review in this one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2323

I'll close this PR.

PR #2121, opened 2018-10-04
Remove FullscreenPassthrough path on camera

sebastienlagarde:
Remove FullscreenPassthrough path on camera
user should use additionalCameraData.ExecuteCustomRender(renderContext, hdCamera); instead


sebastienlagarde:
Seems that there is still use for this, so for now close the PR

PR #2124, opened 2018-10-05
Hdrp/staging

sebastienlagarde:
- Cleaning FrameSettings and HDRenderPipeline names and display (#2123)
- fix offset tool (#2127)


sebastienlagarde:
Merging HDRP/Staging as it contain a critical fix for artists but can't launch any test on Katana due to too much load currently

PR #2130, opened 2018-10-05
Revamp of XR/SRP hookup + Full screen blit fix for single-pass stereo

nerites:
### Purpose of this PR
Mike Chow made a convincing argument for preserving a single-point-of-contact programming model for XRSettings for as long as XRSettings exists. Because of this, XRGraphicsConfig has been converted to a read-only XRGraphics interface with XRSettings. 

When stereo is enabled, SRP will be overridden by analogous settings in XRSettings. For now, the only affected setting is RenderScale in LWRP. Whenever ViewportScale is added, that too will be overridden.

No other XRSettings should overlap with SRP controls. 

This PR also contains a shader change to HDRP's Blit shader. It adds a shader variant to correctly sample a double-wide texture for a single-pass double-wide Blit. This Blit is dispatched through the command buffer, which takes advantage of native Y-flip correction. 

---
### Release Notes
- Fixes Y-flip in HDRP by using HDRP's Blit shader, which has been adapted to work with single-pass double-wide stereo rendering
- Changed SRP's interface with XRSettings to XRSettings override of common functionality

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=stereoplumbing&unity_branch=trunk&automation-tools_branch=add-platform-filter

Manually tested single-pass double-wide (with and without postprocessing enabled) in the template workshop scene. 
image

Nothing

Just the template project should do it

---
### Overall Product Risks
Low- change is a straightforward shift to the simplest possible integration with XR.

Medium- changes touched many places across LWRP and HDRP, including UI. 

---
### Comments to reviewers
Hopefully this is the final iteration of the attempts to fix y-flip for HDRP. 
Requesting feedback on HDRP control of stereo rendering- in this changeset, I've basically removed any additional control HDRP would have had over stereo rendering, the idea being that StartMultiEye and StopMultiEye are all the control over that that's needed on the SRP side. Feel free to give feedback on this.


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Remove all the debug mode from SSR that are obsolete now
 - Expose frameSettings and Capture settings for reflection and planar probe
 - Update UI for reflection probe, planar probe, camera and HDRP Asset
+- Read/write XRGraphicsConfig replaced with read-only XRGraphics to improve consistency of XR behavior between legacy and SRP
```
Katstodian: Please replace with:
`XRGraphicsConfig` is no longer Read/Write. Instead, it's read-only. This improves consistency of XR behavior between the legacy render pipeline and SRP.

nerites:
Hi everyone, can I get some eyes on this? It'd be nice to be able to work off this as a baseline for some of my other stuff.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopSettings.cs
```diff
@@ -107,8 +107,8 @@ public LightLoopSettings Override(LightLoopSettings overridedFrameSettings)
             // In HD, MSAA is only supported for forward only rendering, no MSAA in deferred mode (for code complexity reasons)
             aggregate.enableFptlForForwardOpaque = aggregate.enableFptlForForwardOpaque && !aggregateFrameSettings.enableMSAA;
 
-            // disable FPTL for stereo for now
-            aggregate.enableFptlForForwardOpaque = aggregate.enableFptlForForwardOpaque && !aggregateFrameSettings.enableStereo;
+            //// disable FPTL for stereo for now
```
JulienIgnace-Unity: If not useful anymore, might as well remove it?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1306,9 +1311,18 @@ public override void Render(ScriptableRenderContext renderContext, Camera[] came
                             using (new ProfilingSample(cmd, "Blit to final RT", CustomSamplerId.BlitToFinalRT.GetSampler()))
                             {
                                 // This Blit will flip the screen on anything other than openGL
-                                if (srcFrameSettings.enableStereo && (XRGraphicsConfig.eyeTextureDesc.vrUsage == VRTextureUsage.TwoEyes))
+                                if (camera.stereoEnabled && (XRGraphics.eyeTextureDesc.dimension == TextureDimension.Tex2D))
                                 {
-                                    cmd.BlitFullscreenTriangle(m_CameraColorBuffer, BuiltinRenderTextureType.CameraTarget); // If double-wide, only blit once (not once per-eye)
+#if UNITY_2019_1_OR_NEWER
+                                    Material finalDoubleWideBlit = GetBlitMaterial();
```
sebastienlagarde: How is this path working when postprocessing are enabled (which is almost always the case?)

sebastienlagarde:
So @nerites do you want to merge this PR or are you preparing another one?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
nerites: Post processing has its own Y-flip mechanism in its PostProcessContext, so we're able to use PostProcess's Copy shader instead of the native Y-flipping. I modified the code in RenderPostProcess to update the PostProcessContext's yflip thing correctly.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopSettings.cs
nerites: I want to support tile in stereo eventually, but it doesn't work right now. I removed this to debug tile lighting but I didn't mean to commit it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopSettings.cs
```diff
@@ -107,8 +107,8 @@ public LightLoopSettings Override(LightLoopSettings overridedFrameSettings)
...
+            //aggregate.enableFptlForForwardOpaque = aggregate.enableFptlForForwardOpaque && !aggregateFrameSettings.enableStereo;
```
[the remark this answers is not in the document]
nerites: See my response to Julien (:

nerites:
Sorry for the late response! This is the PR I want to merge, I just re-added the tile disable for stereo.

sebastienlagarde:
ok, let's be sure katana test pass and we are good. Note: Seems like right now master 2018.3/staging is broken, I am on it. I will launch the katana test for this PR when it is good

nerites:
Thanks a lot! (:

mikecunity:
This PR still doesn't address the duplication of values in XRGraphics and XRSettings such as renderScale and viewportScale. Since these are read-only, we should avoid the confusion and remove them from XRGraphics and use XRSettings directly. 

RenderScale is also renamed in XR to eyeTextureResolutionScale to avoid people misusing it, and we shouldn't expose RenderScale in XRGraphics to add to the confusion. Since non-stereo doesn't support viewportScale the way XR does, viewportScale should be removed from XRGraphics and use XRSettings.renderViewportScale directly.

mikecunity:
Since XRGraphics settings are now coming directly from XRSettings, we need to use same variable names as XRSettings rather than more aliases to avoid ambiguity: need to change viewportScale to renderViewportScale and eyeTextureScale to eyeTextureResolutionScale.

phi-lira:
@shawnmcclelland should we have a info box explaining why renderscale is hidden when xr is enabled?

nerites:
Rename done. 
I tried to have a mouseover explaining why renderscale gets greyed out when XR is enabled but it looks like it doesn't appear.

@mikecunity rename complete, does it look ok?

shawnmcclelland:
> @shawnmcclelland should we have a info box explaining why renderscale is hidden when xr is enabled?

Yeah, there should probably be some information here to let the user know why render scale is being hidden. Even if we leverage it for the first release and then deprecate it once we feel it's solidified.

shawnmcclelland:
Changes seem good to me. @nerites can you work with @Katstodian to write up tooltips for Stereo Separation, Stereo Convergence, and Target Eye?

The only request I'd have is right now the Frame Settings Overrides feel like they're embedded under XR Settings when they really apply outside the XR context. @sebastienlagarde we should look at fixing that in another PR so the frame overrides sit at the root level of the Inspector.

image

sebastienlagarde:
@RSlysz Let's discuss about the remark of Shawn above. We will do a separate PR, this isn't related to XR settings change here.

com.unity.render-pipelines.core/Runtime/Common/XRGraphics.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        public enum StereoRenderingMode
+        {
+            None,
+            MultiPass,
+            SinglePassDoubleWide,
+            SinglePassInstanced,
+            SinglePassMultiView
+        };
+
```
mikecunity: For consistency, for UNITY_2018_3_OR_NEWER, XRGraphics' stereoRenderingMode should just use value from XRSetting.stereoRenderingMode.  For older versions, the enum for XRGraphics StereoRenderingMode should have same names and enum values as ones from XRSettings. 

public enum StereoRenderingMode
        {
            MultiPass = 0,
            SinglePass,
            SinglePassInstanced,
            SinglePassMultiview
        }

nerites: done (:

sebastienlagarde:
So seems we are all good now, just merge latest master and run kanata test now that they work and are green. Thanks

sebastienlagarde:
Katana have an error on aan HDRP that come from master (we are on this issue) and not related to this PR. We are good to go. Merging.

PR #2143, opened 2018-10-08
Fix Shader Graph pipeline specific nodes

Kink3d:
### Purpose of this PR

Fixes the following issues by making their node's return values render pipeline specific. This is the replacement for PR  #1911. LWRP specific light nodes will be added in a separate PR.

#### Light Probe node
- Change name to **Baked GI** node.
- Move to `Input/Lighting`
- Make return values pipeline specific baked GI sampling

#### Reflection Probe node
- Move to `Input/Lighting`
- Make return values pipeline specific (make HDRP-safe failure)

#### Ambient node
- Move to `Input/Lighting`
- Make return values pipeline specific (make HDRP-safe failure)

#### Fog node
- Make return values pipeline specific (make HDRP-safe failure)
- Make LWRP implementation use function instead of duplicated code

#### Object node
- Make return values for port `Position` pipeline specific
- Keep default return value (absolute world position from model matrix)
- HDRP specific implementation converts to AbsoluteWS

---
### Release Notes
See changelog

---
### Testing status
**Katana Tests**: Will run when the build load cools down...

**Manual Tests**: Checked all changed nodes with both pipelines, all 3 masters. @spaztronaut  / @zadi0648  to continue.

**Automated Tests**: None

---
### Overall Product Risks
**Technical Risk**: Low?

**Halo Effect**: Low?

---
### Comments to reviewers
N/A


zadi0648:
Looking at this now

com.unity.shadergraph/Documentation~/Ambient-Node.md
```diff
@@ -6,6 +6,11 @@ Provides access to the Scene's **Ambient** color values. When Environment Lighti
 
 Note: Values of this [Node](Node.md) are only updated when entering Play mode or saving the current Scene/Project.
 
+Note: The behaviour of this [Node](Node.md) is undefined globally. The executed HLSL code for this [Node](Node.md) is defined per **Render Pipeline**, and different **Render Pipelines** may produce different results. Custom **Render Pipelines** that wish to support this [Node](Node.md) will also need to explicitly define the behaviour for it. If undefined this [Node](Node.md) will return 0 (black).
```
Katstodian: Please replace with (and copy over to relevant Nodes):

**Note:** The behavior of this [Node](Node.md] is undefined globally. Shader Graph does not define the function of the node. Instead, each **Render Pipeline** defines what HLSL code to execute for this [Node](Node.md).  

Different **Render Pipelines** may produce different results. If you're building a shader in one **Render Pipeline** that you want to use in both, try checking it in both pipelines before production. A [Node](Node.md] might be defined in one **Render Pipeline** and undefined in the other. If this [Node](Node.md) is undefined, it returns 0 (black).

spaztronaut:
removing myself since zack is looking at this

zadi0648:
Update: Matt is looking at the lightmap UVs...

weirduvs

sebastienlagarde:
Isn't this good to move forward ?

PR #2148, opened 2018-10-09
[WIP] Scalarise buffer loads for LightData

FrancescoC-unity:
### Purpose of this PR

This removes lots of vector buffer loads, reducing waits on vector memory and making few uncoherent branches coherent.
On PS4, I measured a win around 10-15%. 

An example of change: 

**Scalar ALU**: 349 -> 423 
**Scalar Memory**: 181 -> 229 
**Vector Memory**:  123 -> 88
**Unconditional branches**:  13 -> 20

---
### Release Notes

Optimized the forward rendering pass by scalarizing vector buffer loads. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=scalarize-lightdata-loads&automation-tools_branch=add-platform-filter&unity_branch=2018.3%2Fstaging [Running]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Low

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
@@ -140,6 +140,7 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
 
     #ifdef LIGHTLOOP_TILE_PASS
         GetCountAndStart(posInput, LIGHTCATEGORY_PUNCTUAL, lightStart, lightCount);
+        lightStart = WaveReadFirstLane(lightStart);
     #else
         lightCount = _PunctualLightCount;
         lightStart = 0;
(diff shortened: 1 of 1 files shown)


EvgeniiG:
For deferred this is a no-op, since LightData is already scalarized.
For clustered, it will break things, because you cannot just use the first light since you can have a bunch of different clusters distributed along the depth range of the tile and they may have different lights in them.

FrancescoC-unity:
> For deferred this is a no-op, since LightData is already scalarized.
> For clustered, it will break things, because you cannot just use the first light since you can have a bunch of different clusters distributed along the depth range of the tile and they may have different lights in them.

I got mislead by the define mentioning tiles instead of clusters and it did work in forward w/ tiles, will have a look at the cluster case.

FrancescoC-unity:
I have a couple idea that'd work with cluster, but they'd be likely not like this one, so I am closing the PR and will open a new one when it is ready with same reviewers.

PR #2159, opened 2018-10-10
Moved StartMultiEye and StopMultiEye from ScriptableRenderContext to RenderingCommandBuffer [DO NOT MERGE]

nerites:
### Purpose of this PR

---
### Release Notes
Contingent on this PR: https://ono.unity3d.com/unity/unity/pull-request/75672/_/xr/srp/migrate2cmdbuf

---
### Testing status
**Katana Tests**:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=replacemultieye&unity_branch=xr%2Fsrp%2Fmigrate2cmdbuf&automation-tools_branch=add-platform-filter

Manually checked that stereo rendering still worked on LWRP (SPI and SPDW) and HDRP (only SPDW). 

None.

I just used the template workshop project.

---
### Overall Product Risks
**Technical Risk**: Low- there is technical risk on the Editor side, and I am having experienced XR folks review it. 

**Halo Effect**: Low- only affects VR, and changes should land in time for 19.1. 

---
### Comments to reviewers
I'm unsure why adding BeginXR/EndXR passes around the DepthOnlyPass doesn't seem to work. It works fine with StartMultiEye/StopMultiEye calls added inside the DepthOnlyPass. It might be fine, but I can investigate further just for consistency's sake.


com.unity.render-pipelines.lightweight/Runtime/Passes/BeginXRRenderingPass.cs
```diff
@@ -15,13 +16,16 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
...
+        const string k_StartMultiEyeTag = "Start MultiEye Rendering";
         public override void Execute(ScriptableRenderer renderer, ScriptableRenderContext context, ref RenderingData renderingData)
         {
             if (renderer == null)
                 throw new ArgumentNullException("renderer");
-            
-            Camera camera = renderingData.cameraData.camera;
-            context.StartMultiEye(camera);
+
+            CommandBuffer cmd = CommandBufferPool.Get(k_StartMultiEyeTag);
```
sebastienlagarde: I am not sure to understand why you need to create a command buffer for a single instruction like this, looks overkill. @JulienIgnace-Unity any though?

com.unity.render-pipelines.lightweight/Runtime/Passes/EndXRRenderingPass.cs
```diff
@@ -12,14 +13,17 @@ namespace UnityEngine.Experimental.Rendering.LightweightPipeline
...
         {
             if (renderer == null)
                 throw new ArgumentNullException("renderer");
-            
-            Camera camera = renderingData.cameraData.camera;
-            context.StopMultiEye(camera);
-            context.StereoEndRender(camera);
+
+            CommandBuffer cmd = CommandBufferPool.Get(k_StopMultiEyeTag);
+            cmd.StopMultiEye();
```
sebastienlagarde: Same remark here. I think we may have performance issue if we execute the command buffer at every pass

sebastienlagarde:
Adding Julien to get more feedback about the creation of command buffer in the pool for single instruciton, looks like overkill to me (I haven't noticed that before, so sorry if I missed it in previous PR).

com.unity.render-pipelines.lightweight/Runtime/Passes/EndXRRenderingPass.cs
nerites: I was following the example set by the Pass coding style- maybe @phi-lira or @stramit can comment?

com.unity.render-pipelines.lightweight/Runtime/Passes/BeginXRRenderingPass.cs
JulienIgnace-Unity: This is for LW :) @phi-lira will review this :)

com.unity.render-pipelines.lightweight/Runtime/Passes/EndXRRenderingPass.cs
sebastienlagarde: Indeed, my mistake :P

nerites:
After talking to Mike about the Editor code that this is dependent on, some concerns were raised about moving commands to the command buffer that would not work with the legacy pipeline. I will close this pending a conclusion on that. ):

PR #2171, opened 2018-10-11
Improve IAfterRender

stramit:
### Purpose of this PR
Using IAfterRender had issues as it would never have a valid camera target. This fixes the issue so that more can be done at this stage.

---
### Release Notes
*Fix Post to not blit from the same src -> destination
*Improve IAfterRender to have a valid src / destination
*If callbacks exist we need to go via an RT
*Combine post fx passes
*Add new test for camera callbacks

---
### Testing status
**Katana Tests**: 
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=IAfterRender-Improvements&unity_branch=trunk

**Automated Tests**: Enable test 52

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipelineCore.cs
```diff
@@ -49,7 +49,6 @@ public struct CameraData
         public int msaaSamples;
         public bool isSceneViewCamera;
         public bool isDefaultViewport;
-        public bool isOffscreenRender;
```
phi-lira: We should not remove this. Users might be depending on it.

com.unity.render-pipelines.lightweight/Runtime/Passes/TransparentPostProcessPass.cs
```diff
@@ -4,44 +4,52 @@
...
+    /// Perform post-processing using the given color attachment
+    /// as the source and the given color attachment as the destination.
     ///
     /// You can use this pass to apply post-processing to the given color
-    /// buffer. The pass uses the currently configured post-process stack,
-    /// and it copies the result to the Camera target.
+    /// buffer. The pass uses the currently configured post-process stack.
     /// </summary>
-    public class TransparentPostProcessPass : ScriptableRenderPass
+    public class PostProcessPass : ScriptableRenderPass
```
phi-lira: Please rename the filename as well

phi-lira: Please rename the filename as well, otherwise it's confusing.

com.unity.render-pipelines.lightweight/Runtime/Passes/TransparentPostProcessPass.cs
```diff
@@ -4,44 +4,52 @@
...
-            RenderTargetHandle colorAttachmentHandle,
-            RenderTargetIdentifier destination)
+            RenderTargetHandle sourceHandle,
+            RenderTargetHandle destinationHandle,
+            bool opaquePost,
+            bool flip)
         {
-            this.colorAttachmentHandle = colorAttachmentHandle;
-            this.destination = destination;
+            if (sourceHandle == destinationHandle)
```
phi-lira: This check should only be performed when there are exactly 1 postprocessing effects in the stack. Otherwise it's fine because postfx is creating temp textures and blitting from same source and dest saves us memory/resources.

com.unity.render-pipelines.lightweight/Runtime/DefaultRendererSetup.cs
```diff
@@ -151,16 +192,20 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
             renderer.EnqueuePass(m_RenderOpaqueForwardPass);
-            foreach (var pass in camera.GetComponents<IAfterOpaquePass>())
+            foreach (var pass in m_AfterOpaquePasses)
                 renderer.EnqueuePass(pass.GetPassToEnqueue(baseDescriptor, colorHandle, depthHandle));
 
             if (renderingData.cameraData.postProcessEnabled &&
                 renderingData.cameraData.postProcessLayer.HasOpaqueOnlyEffects(renderer.postProcessingContext))
             {
-                m_OpaquePostProcessPass.Setup(baseDescriptor, colorHandle);
+                m_CreatePostOpaqueColorPass.Setup(baseDescriptor, m_ColorAttachmentAfterOpaquePost, sampleCount);
```
phi-lira: if we have more than 1 effect we don't need to create this texture. With more than one effect we have:
blit(source, tempPostFXTexture)
blit(tempPostFXTexture, source)

This is adding a perf regression.

com.unity.render-pipelines.lightweight/Runtime/DefaultRendererSetup.cs
```diff
@@ -170,41 +215,70 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
+            // if we have additional filters
+            // we need to stay in a RT
+            if (afterRenderExists)
             {
-                m_TransparentPostProcessPass.Setup(baseDescriptor, colorHandle, BuiltinRenderTextureType.CameraTarget);
-                renderer.EnqueuePass(m_TransparentPostProcessPass);
+                // perform post with src / dest the same
+                if (!renderingData.cameraData.isStereoEnabled && renderingData.cameraData.postProcessEnabled)
+                {
+                    m_CreatePostTransparentColorPass.Setup(baseDescriptor, m_ColorAttachmentAfterTransparentPost, sampleCount);
```
phi-lira: Same as with opaque post processing, we shouldn't create a texture if there are more than one effect. Also, consider the case we have 1 opaque and 1 transparent post process. With current code we are creating 3 color textures (color, opaque post, transparent post). Instead we should have a buffer of 2 worker textures we switch.

This is also adding a perf regression.

com.unity.render-pipelines.lightweight/Runtime/DefaultRendererSetup.cs
```diff
@@ -93,53 +120,67 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
             {
-                bool additionalLightShadows = m_AdditionalLightsShadowCasterPass.Setup(AdditionalLightsShadowmap, ref renderingData, renderer.maxVisibleAdditionalLights);
+                bool additionalLightShadows = m_AdditionalLightsShadowCasterPass.Setup(m_AdditionalLightsShadowmap, ref renderingData, renderer.maxVisibleAdditionalLights);
                 if (additionalLightShadows)
                     renderer.EnqueuePass(m_AdditionalLightsShadowCasterPass);
             }
 
             bool resolveShadowsInScreenSpace = mainLightShadows && renderingData.shadowData.requiresScreenSpaceShadowResolve;
             bool requiresDepthPrepass = resolveShadowsInScreenSpace || renderingData.cameraData.isSceneViewCamera ||
-                                        (renderingData.cameraData.requiresDepthTexture && (!CanCopyDepth(ref renderingData.cameraData) || renderingData.cameraData.isOffscreenRender));
```
phi-lira: This will break test 046. Either copy the check for being offscreen camera must be here.

PR #2172, opened 2018-10-11
Shadow Fixes for XR

brandon-unity3d:
### Purpose of this PR
The purpose of this PR is to ensure shadows render correctly in XR.
The changes in this pr are

1. Don't allow the explicit binding of a multi-sampled depth buffer if auto resolve is enabled.
2. Don't draw the occlusion mesh in the depth buffer texture.
3. Ensure that the unity_StereoEyeIndex variable is populated correctly in the standard shaders to ensure we sample from the depth buffer correctly if it is a texture array.

---
### Release Notes
Shadows are now sampled correctly in XR with all stereo rendering methods.

---
### Testing status

I tested on a Pixel 2.

---
### Overall Product Risks
Low

Low


brandon-unity3d:
Hey @zeroyao!

I added you to the PR to review the two shader files I modified where I replaced UNITY_SETUP_INSTANCE_ID(input); with UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);

I did this as a performance optimization.  In XR, UNITY_SETUP_INSTANCE_ID will calculate the unity_StereoEyeIndex based off the unity_InstanceId.  The unity_InstanceId is also set as well.  However these values would have already been computed in the vertex shader so there isn't any reason to recompute them per fragment.  I don't believe the LWRP standard shaders use the unity_InstanceID so we use the UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX to assign the unity_StereoEyeIndex to the vertex output value.  We do this in the builtin standard shaders as well.

zeroyao:
> Hey @zeroyao!
> 
> I added you to the PR to review the two shader files I modified where I replaced UNITY_SETUP_INSTANCE_ID(input); with UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
> 
> I did this as a performance optimization. In XR, UNITY_SETUP_INSTANCE_ID will calculate the unity_StereoEyeIndex based off the unity_InstanceId. The unity_InstanceId is also set as well. However these values would have already been computed in the vertex shader so there isn't any reason to recompute them per fragment. I don't believe the LWRP standard shaders use the unity_InstanceID so we use the UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX to assign the unity_StereoEyeIndex to the vertex output value. We do this in the builtin standard shaders as well.

Hmm I'm not sure if it works in this way. If unity_InstanceID is not going to be used, should all the computation for it just be optimized out?

brandon-unity3d:
> > Hey @zeroyao!
> > I added you to the PR to review the two shader files I modified where I replaced UNITY_SETUP_INSTANCE_ID(input); with UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
> > I did this as a performance optimization. In XR, UNITY_SETUP_INSTANCE_ID will calculate the unity_StereoEyeIndex based off the unity_InstanceId. The unity_InstanceId is also set as well. However these values would have already been computed in the vertex shader so there isn't any reason to recompute them per fragment. I don't believe the LWRP standard shaders use the unity_InstanceID so we use the UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX to assign the unity_StereoEyeIndex to the vertex output value. We do this in the builtin standard shaders as well.
> 
> Hmm I'm not sure if it works in this way. If unity_InstanceID is not going to be used, should all the computation for it just be optimized out?

When single-pass instancing is enabled, instancing is used.  So the computation wouldn't be compiled out in that case.  Furthermore, I don't believe our standard shaders use unity_InstanceId in the fragment shader.

zeroyao:
@phi-lira Check out CoreRP/ShaderLibrary/UnityInstancing.cginc, there are other constants i.e. lightmap STs and SH coefficients

phi-lira:
Is katana green? If so I'll merge this PR. Also, since we branched off, let us know if you need this backported to 2018.3

phi-lira:
Katana is failing on HDRP test 2202_ReflectionProbes_Volume but that's fixed in trunk already. Merging.

PR #2173, opened 2018-10-11
Fix 2 bugs in ShaderSourceMap's binary search

marctem:
1. Off-by-one error that was never caught
2. All code was 0-based but file line numbers are 1-based

Added unit tests to verify source map interals going forward.

### Purpose of this PR
ShaderSourceMap was broken, causing exceptions during shader error handling.

---
### Testing status
**Katana Tests**: Running...
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=shadergraph%2Fbinary-search-fix&unity_branch=trunk

**Manual Tests**: Verified on project with broken shader graph node that used to throw exceptions and now it does not.

**Automated Tests**: New Test class with 4 tests to verify binary search is effective

---
### Overall Product Risks
**Technical Risk**: Low - it's an algorithm that was wrong, and now is right.

**Halo Effect**: None - totally contained in 1 file


marctem:
Katana tests got killed, can't reschedule new ones due to load. Will check throughout the day and then ping Chris or Matt to merge.

Kink3d:
Tests green

PR #2177, opened 2018-10-12
Add more gpu intrinsics for GCN consoles

FrancescoC-unity:
### Purpose of this PR
Add few more intrinsics for GCN consoles. Note that I don't have an immediate use case in mind for all of them, but since I needed to add a couple, I used to occasion to add few useful ones. 

I also had to make explicit type variants as the Xbox intrinsics specify them in function name and it is quite strict. For the same reason, I needed to add some explict casting in BuildProbabilityTables.compute .

This has release/2018.3 as base

---
### Release Notes
Add support for more GPU intrinsics on GCN consoles.

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=add-more-gpu-intrinsics&automation-tools_branch=add-platform-filter&unity_branch=2018.3%2Fstaging [Running]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Low

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -180,8 +180,8 @@
 #define LODDitheringTransition ERROR_ON_UNSUPPORTED_FUNC(LODDitheringTransition)
 #endif
 
-// On non-consoles we error on cross-lane operations
-#if !defined(SHADER_API_PSSL) && !defined(SHADER_API_XBOXONE)
+// On everything but GCN consoles we error on cross-lane operations
+#ifndef SUPPORTS_WAVE_INTRINSICS
 #define WaveMinInt ERROR_ON_UNSUPPORTED_FUNC(WaveMinInt)
 #define WaveMinUint ERROR_ON_UNSUPPORTED_FUNC(WaveMinUint)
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -180,6 +180,20 @@
 #define LODDitheringTransition ERROR_ON_UNSUPPORTED_FUNC(LODDitheringTransition)
 #endif
 
+// On non-consoles we error on cross-lane operations
+#if !defined(SHADER_API_PSSL) && !defined(SHADER_API_XBOXONE)
```
sebastienlagarde: instead of testing
#if !defined(SHADER_API_PSSL) && !defined(SHADER_API_XBOXONE)

Let's do a define like: SUPPORT_WAVE_INTRINSIC that is define for xbox on PSLL, 
then when we test the code we can do

#if SUPPORT_WAVE_INTRINSIC 
dsfdsf
#else 
...
#endif

FrancescoC-unity: Makes sense, doing it now

PR #2182, opened 2018-10-12
Volumetric height fog

EvgeniiG:
### Purpose of this PR
Adds the initial implementation of volumetric height fog


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -55,7 +55,16 @@ void EvaluateLight_Directional(LightLoopContext lightLoopContext, PositionInputs
     float  shadowMask = 1.0;
 
     color       = light.color;
-    attenuation = 1.0; // TODO: implement volumetric attenuation along shadow rays for directional lights
+    attenuation = 1.0;
+
+    // Height fog attenuation.
```
sebastienlagarde: should we test that fog (or volumetric fog?) is enabled at this point?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -55,7 +55,16 @@ void EvaluateLight_Directional(LightLoopContext lightLoopContext, PositionInputs
...
+    {
+        float cosZenithAngle = L.z;
+        float fragmentHeight = GetAbsolutePositionWS(posInput.positionWS).y;
```
sebastienlagarde: You should convert the fog to be camera relative and not convert the position to absolute here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
EvgeniiG: For consistent visuals we must volumetrically shadow all surfaces and volumes the same way.

EvgeniiG: If we want to claim our lighting is "unified".

EvgeniiG: I agree that there's extra cost, though.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
EvgeniiG: Done.

PR #2187, opened 2018-10-15
Shadergraph/error handling review

marctem:
### Purpose of this PR
Just for review of the new SG error-passing approach. Looking for feedback so I can hopefully wrap it up with a real PR tomorrow.

Known issues: Does not display errors on Master Node due to no source map to find the proper node. I tried a couple of different ways to get the nodes out, but didn't figure it out.

Validation Errors should be plumbed now but I haven't tested it yet, wanted to get your thoughts on the approach before I get it perfect.


com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -85,6 +84,9 @@ public PreprocessorIf(string conditional)
         }
     }
 
+    // An easy alias for error handling
+    public class ShaderMessageList : Dictionary<Identifier, List<ShaderMessage>> {}
```
pbbastian: Not a fan of this in C#. I think it obscures what class it is, and should the need for changing the arise, the new one could have a totally different interface anyway.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -677,6 +755,28 @@ PreviewRenderData GetRenderData(Identifier id)
...
+            return m_MessageChanges;
+        }
+
+        internal void ClearNodeMessageChanges()
+        {
+            foreach (var change in m_MessageChanges)
+            {
+                if (change.Value.Count == 0)
+                {
+                    m_CurrentMessages.Remove(change.Key);
```
pbbastian: Hmm, does this mean that the messages are split between `m_CurrentMessages` and `m_MessageChanges`? I think it makes more sense to store them all in one place, and then just refer to the new ones in `m_MessageChanges`.

marctem: Not split. Current is what's displayed in the UI. Changes is where I track what's different since the last shader compile until the UI update happens, then I clear it out. That enables a minimal refresh on only changed nodes, like you were saying.

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
marctem: I think it's semantically clearer. The underlying type should be obscured by the functionality IMO. What I don't like is that C# won't let me easily define a ShaderMessageList::iterator type to make the abstraction really work, so you have to know what the type is to iterate.

I just really don't like giant, double-generic type names in so many places. I'll change it though.

pbbastian: The issue is that this is a very leaky abstraction. If you do wrap the type, it definitely shouldn't be via inheritance, and wrapping the type and re-implementing `IDictionary<TKey,TValue>` is a lot of boilerplate. You can always define a struct `ShaderMessage`, and then return an `IEnumerable<ShaderMessage>` somehow.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
pbbastian: Ahh okay, that makes sense then :)

marctem:
Closing, will make a new one with major changes!

PR #2192, opened 2018-10-15
Fix compilation issue with on non editor build in HDRenderPipelineAsset.cs

FrancescoC-unity:
### Purpose of this PR

Moving the endif as it was hiding the brace that closed class definition in non-editor builds.  This fixes a compilation issue when launching a non-editor build. 

---
### Release Notes
Fix compilation error in non-editor builds. 

---
### Testing status

**Manual Tests**:  Launched on PS4 and Xbox One. 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None


sebastienlagarde:
Thanks for this fix!

AnisB:
Woopsie

PR #2195, opened 2018-10-16
Fix main light shadow when there is no any caster

Seongdae:
### Purpose of this PR

Fixed main light shadow problem when there is no any shadow caster in the main light.
If the main light is set with cascades shadow, I can't find any problem. but the case of single cascade shadow it will reproduce shadowing error out of all bounds of shadow casters, because the main light shadowmap texture is set as default "gray" texture.

---
### Release Notes
Fixed shadowing error in the single cascade shadow map when there is no any shadow caster.

---
### Testing status
**Katana Tests**: None

**Manual Tests**: 
- Set lightweight RP
- Set shadow cascades as no cascades
- Place terrain object
- Move camera far from the position of terrain in play mode

**Automated Tests**: None

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
I added the exception to main light cascade shadow map in MainLightShadowCastersPass.cs
Please check if it is suitable fix or better fix anyone can give


phi-lira:
Hi @Seongdae, thanks for proposing this fix. However, this issue was fixed properly last week by this commit and it's waiting on some katana fixes in order to create a PR for it: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/38540d01a0265012d5b74114bc39a6bf960fc9e1

The correct approach is to not execute the shadow passes (caster and collect) if there are no shadow casters. Also, the fix we have also fix possible problem with additional light shadows.

If you don't see any other issues, I'll close this PR.

Seongdae:
Hi @phi-lira, thank you for your comments. I checked your PR now. great. Let me close this PR

PR #2209, opened 2018-10-17
Scalarize buffer loads in Light Loop (2019.1)

FrancescoC-unity:
### Purpose of this PR
Scalarization of the lightData (lights, decals and env lights) loads to benefit forward passes (tiled or clustered). 
Forcing cluster on forward opaque I measured a decrease in cost of between 20 and 30%. 
On transparent materials the cost decreases around 20%. 
Overall the VGPR count was also down by around 10%. 

Area lights were a bit tricky to scalarize, so I have held off on doing that for now, planning on revisiting it later on. 

---
### Release Notes
Performance improvement through scalarization of light/decal data loads for light loop on GCN consoles. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=scalarize-forward-lightloop&automation-tools_branch=add-platform-filter&unity_branch=2018.3%2Fstaging [Running] 

**Manual Tests**:  Ran on PC, PS4 and Xbox. On the latter two verified that the buffer loads are effectively scalar. 

---
### Overall Product Risks
**Technical Risk**:  Medium

**Halo Effect**:  Low

---
### Comments to reviewers

Sorry it took me a bit since last iteration, but iteration time on console is quite slow and I had a nasty and not very fun to debug GPU hang 😄 

***Important***: I haven't finished testing all possible configuration, just tested forward cluster. I will add a comment on the review when I verified it all, until then we should hold off from merging.
Also, I need to test final iteration on xbox.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+            uint s_lightIdx = v_lightIdx;
+#if SCALARIZE_LIGHT_LOOP
+            if (!fastPath)
+            {
+                // If we are not in fast path, v_lightIdx is not scalar
+                s_lightIdx = min(WaveMinUint(v_lightIdx), max(_PunctualLightCount - 1, 0));
+            }
+            // Note that the WaveReadFirstLane should not be needed, but the compiler might insist in putting the result in VGPR.
+            // However, we are certain at this point that the index is scalar.
+            s_lightIdx = WaveReadFirstLane(s_lightIdx);
```
FrancescoC-unity: This should be redundant but it is not. I assume it is because since I am using a bool for fastPath, vcc gets involved and everything turns into VGPR even though the VCC bits should be all equal. 
If my guess is right, I can work around this by not using a bool and maybe a scalar int, but it sounds less robust.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
         uint lightCount, lightStart;
-
+        bool fastPath = false;
     #ifdef LIGHTLOOP_TILE_PASS
         GetCountAndStart(posInput, LIGHTCATEGORY_PUNCTUAL, lightStart, lightCount);
-    #else
+#if SCALARIZE_LIGHT_LOOP
+        // Fast path is when we all pixels in a wave are accessing same tile or cluster.
+        uint lightStartLane0 = WaveReadFirstLane(lightStart);
+        fastPath = all(Ballot(lightStart == lightStartLane0) == ~0);
```
EvgeniiG: Please use the HLSL 6 naming convention for intrinsic abstraction:
https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12#shading-language-intrinsics

FrancescoC-unity: I will for ballot, but for the Min version unfortunately the xbox intrinsics make the type they act on explicit in the name so I can't use the same name.

FrancescoC-unity:
> Why not just write a loop over all clusters?

As started to discuss offline, that won't change much. On the other hand, moving the loop to operate per cluster rather than per light type like it already is will require a substantial restructure of how the light loop is now (and really, it is orthogonal to this change)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
     if (featureFlags & LIGHTFEATUREFLAGS_PUNCTUAL)
     {
         uint lightCount, lightStart;
-
+        bool fastPath = false;
```
EvgeniiG: I don't see how the "fastPath" variable makes things faster.
It should probably be removed.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
     #ifdef LIGHTLOOP_TILE_PASS
         GetCountAndStart(posInput, LIGHTCATEGORY_PUNCTUAL, lightStart, lightCount);
-    #else
+#if SCALARIZE_LIGHT_LOOP
+        // Fast path is when we all pixels in a wave are accessing same tile or cluster.
+        uint lightStartLane0 = WaveReadFirstLane(lightStart);
+        fastPath = all(Ballot(lightStart == lightStartLane0) == ~0);
+#endif
+#else
```
EvgeniiG: This preprocessor code should be properly indented and commented, otherwise it's almost unreadable.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: It avoids calling WaveMin which involves a bunch of swizzles and mins.

FrancescoC-unity: And since whether to take a fast path is per wave, all the threads will happily branch over the WaveMin

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: If you remove fastPath you can also remove this one.

FrancescoC-unity: As said below, fast path had a non-negligible positive impact (and btw, I believe that's what Doom does as well)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+            // Note that the WaveReadFirstLane should not be needed, but the compiler might insist in putting the result in VGPR.
+            // However, we are certain at this point that the index is scalar.
+            s_lightIdx = WaveReadFirstLane(s_lightIdx);
+#endif
+            LightData s_lightData = FetchLight(s_lightIdx);    
 
-            if (IsMatchingLightLayer(lightData.lightLayers, builtinData.renderingLayers))
+            // If current scalar and vector light index match, we process the light
+            // The v_lightListIdx for current thread is increased.
+            if (s_lightIdx >= v_lightIdx)
```
EvgeniiG: Is >= a typo? Should be ==?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: I agree on the indentation, VS just messes up with them everytime I save the file. 

W.r.t. the comment, what could be more clear than SCALARIZE_LIGHT_LOOP and the subsequent comment?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineAsset.cs
```diff
@@ -303,6 +303,6 @@ public void EvaluateSettings()
             }
 #endif
         }
-    }
 #endif
+    }
```
FrancescoC-unity: NOTE: I already submitted this in master.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: As said above, fast path had a non-negligible positive impact (and btw, I believe that's what Doom does as well)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,32 +139,74 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+        uint v_lightIdx = lightStart;
+        while (v_lightListIdx < lightCount)
+        {
+            v_lightIdx = FetchIndex(lightStart, v_lightListIdx);
+            uint s_lightIdx = v_lightIdx;
+#if SCALARIZE_LIGHT_LOOP
+            if (!fastPath)
+            {
+                // If we are not in fast path, v_lightIdx is not scalar
+                s_lightIdx = min(WaveMinUint(v_lightIdx), max(_PunctualLightCount - 1, 0));
```
EvgeniiG: Why clamp here?

EvgeniiG: I guess because some lanes are disabled?

EvgeniiG: If you use WaveActiveMin(), I suppose you don't need to clamp.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: I can do that, but I'd have to do it in the Xbox file. The problem here is that Xbox intriniscs are called 
__XB_WaveMin_I32, __XB_WaveMin_U32 etc. 

If we are fine with having such templates declared in platform specific files, I'll do that. Probably in a separate PR though as there are other intrinsics I modified this way.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: Is WaveMin is so expensive, perhaps it's better to go the lane masking route instead.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: Couldn't this lead to redundant evaluations? I am not really comfortable with this test.

EvgeniiG: If WaveMin() on Xbox is bad, we can always reimplement it with the existing intrinsics.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: For the sake of other reviewers: I am about to start experimenting with the lane masking technique a-la CoD rather than Doom. Will report on the results when I have them

FrancescoC-unity:
After receiving an answer from central tech it is clear what's happening and what the fix should look like. I have added the final proper fix for the hang. 

TL;DR: I was comparing a (vector)value from an helper lane with the value from WaveActiveMin that ignores helper lanes. 

Longer explanation: 
In Pixel shaders on entry, due to WQM (whole-quad mode), the initial exec mask was modified to enable few helper lanes. These helper lanes will execute the shader. In one pixel quad, one helper lane will survive while the other lanes (including the "really active" one) die. As a result, my vector value for v_lightIdx is from an helper lane, *but* WaveActiveMin considers helper lanes as inactive and in my case that meant that no lanes were active at all for the sake of the Min, hence I got a 0xfffffff for my scalar value.  
When comparing the scalar value (0xfffffff) for equality with the vector value coming from an helper lane (!= 0xffffffff), it always gave me false, resulting an infinite loop as the index governing the loop was never increased.

EvgeniiG:
Status update on this PR?

FrancescoC-unity:
> Status update on this PR?

All is done on my side*, waiting on other reviewers now that we are in the 2019.1 window 🙂  

* Except intrinsic renaming, but I'll do that in a separate PR since it touches other things.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,36 +139,87 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+        {
+            v_lightIdx = FetchIndex(lightStart, v_lightListOffset);
+            uint s_lightIdx = v_lightIdx;
+#if SCALARIZE_LIGHT_LOOP
+            if (!fastPath)
+            {
+                // If we are not in fast path, v_lightIdx is not scalar, so we need to query the Min value across the wave. 
+                s_lightIdx = WaveMinUint(v_lightIdx);
+
+                // If WaveMinUint returns 0xffffffff it means that all lanes are actually dead, so we can safely ignore the loop and move forward.
```
EvgeniiG: Need to update the comment with info about helper lanes.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: Doesn't this work?

s_lightIdx = fastPath ? lightStartLane0 : s_lightIdx;

Should be able to avoid WaveReadFirstLane() here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: Note that Ballot() returns 0xffffffff only if all lanes are active.
We should probably hit the fast path if ballot=exec_mask rather than 0xffffffff.
This can be done in a separate PR.

EvgeniiG: Also, ballot returns a ulong (64-bit) rather than a uint (32-bit). Are we correctly handling this?

FrancescoC-unity: Ha, you are right about the active lanes not voting for the ballot, not sure why I haven't thought of it. Will change as soon as this goes in. 

W.r.t. its return type, ulong doesn't really exist in hlsl, it gets converted as uint2 and that's why the `all` , checking both the upper and lower 32bits of the uint64 that are stored in separate uints

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: Well the lightStartLane0  is the ligth start, while s_lightIdx  could be any of the other subsequent lights in the cell, so that wouldn't work. 

The WaveReadFirstLane shouldn't really be needed at all, but the compiler really tries hard sometimes to put s_lightIdx in SGPR, the WaveReadFirstLane  should be a moderately cheap failsafe even in case the code does get slightly modified in the future.

EvgeniiG: OK, nevermind what I just said. The compiler is correct to requite WaveReadFirstLane(), though.

FrancescoC-unity: Well... kinda.

In the case of `!fastPath` then it shouldn't be needed as the value is coming from `WaveMin`. 

If `fastPath`  then the start index is scalar and since the wave goes in lockstep, given that all thread have the same starting index, all threads walk through the same set of indices at the same time.  

Since  `fastPath` is a result of an op on SGPR, it could be smart enough to go ahead and not require the `WaveReadFirstLane`. But yeah, maybe I am asking too much from it :)

sebastienlagarde:
Please, rebase this PR on master (we will then backport it)

FrancescoC-unity:
> Please, rebase this PR on master (we will then backport it)

I tried to do it via git and it ended up like a bit of a mess, I'll manually port it into a new branch with master as base first thing on Monday 🙂

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
FrancescoC-unity: I gave this another thinking today and I think we still need the >= operator and not ==. 

Assume a helper lane holds the smallest v_lightIdx and this index is only associated with the helper lane in current wave. Now, since WaveMin will ignore it, all the active lanes will go forward and exit the loop as we will meet s_lightIdx == v_lightIdx sooner or later.  However, this will never be the case for the helper lane as  s_lightIdx will never assume the value from the helper lane and the helper lane will be stuck in the loop. 

It is quite an unlikely scenario, but not impossible. 

I will update the code and comment on Monday as I move the base to master.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -137,36 +139,88 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+            }
+            // Note that the WaveReadFirstLane should not be needed, but the compiler might insist in putting the result in VGPR.
+            // However, we are certain at this point that the index is scalar.
+            s_lightIdx = WaveReadFirstLane(s_lightIdx);
+#endif
+            LightData s_lightData = FetchLight(s_lightIdx);    
 
-            if (IsMatchingLightLayer(lightData.lightLayers, builtinData.renderingLayers))
+            // If current scalar and vector light index match, we process the light. The v_lightListOffset for current thread is increased.
+            if (s_lightIdx == v_lightIdx)
```
FrancescoC-unity: I gave this another thinking today and I think we still need the >= operator and not == for 
 this check. 

Assume a helper lane holds the smallest v_lightIdx and this index is only associated with the helper lane in current wave. Now, since WaveMin will ignore it, all the active lanes will go forward and exit the loop as we will meet s_lightIdx == v_lightIdx sooner or later.  However, this will never be the case for the helper lane as  s_lightIdx will never assume the value from the helper lane and the helper lane will be stuck in the loop. 

It is quite an unlikely scenario, but not impossible. 

I will update the code and comment on Monday as I move the base to master.

sebastienlagarde:
close this PR as there is a new one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2384

PR #2234, opened 2018-10-19
Fix SSR for multiple views

EvgeniiG:
### Purpose of this PR
- Fix SSR for multiple views
- Fix bad SSR defaults causing no reflection to appear


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -359,23 +400,78 @@ public void Update(FrameSettings currentFrameSettings, PostProcessLayer postProc
...
 
             // Unfortunately sometime (like in the HDCameraEditor) HDUtils.hdrpSettings can be null because of scripts that change the current pipeline...
             m_msaaSamples = msaaSamples;
             RTHandles.SetReferenceSize(m_ActualWidth, m_ActualHeight, m_msaaSamples);
             m_HistoryRTSystem.SetReferenceSize(m_ActualWidth, m_ActualHeight, m_msaaSamples);
             m_HistoryRTSystem.Swap();
 
+            Vector3Int currColorPyramidBufferSize = Vector3Int.zero;
+
+            if (numVolumetricBuffersAllocated != 0)
```
sebastienlagarde: Looks like there is a typo here?
should be numColorPyramidBuffersAllocated

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -288,10 +273,8 @@ VBufferParameters ComputeVBufferParameters(HDCamera hdCamera)
             return new VBufferParameters(viewportResolution, vBufferDepthRange, vBufferDepthDistributionUniformity);
         }
 
-        public void InitializePerCameraData(HDCamera hdCamera)
+        public void InitializePerCameraData(HDCamera hdCamera, int bufferCount)
         {
-            // Note: Here we can't test framesettings as they are not initialize yet
```
sebastienlagarde: "// Note: Here we can't test framesettings as they are not initialize yet" you should let this comment as it is to prevent someone to be tempted to access framesettings here.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -359,23 +400,78 @@ public void Update(FrameSettings currentFrameSettings, PostProcessLayer postProc
 
             m_LastFrameActive = Time.frameCount;
 
-            Vector2 lastTextureSize = new Vector2(RTHandles.maxWidth, RTHandles.maxHeight);
+            // TODO: cache this, or make the history system spill the beans...
+            Vector2Int prevColorPyramidBufferSize = Vector2Int.zero;
+
+            if (numColorPyramidBuffersAllocated > 0)
```
sebastienlagarde: if this code here duplicated from the code below? do you forget to remove this code ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
EvgeniiG: Good catch!

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
EvgeniiG: Done.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
EvgeniiG: Good catch, thanks!

PR #2238, opened 2018-10-19
Expose "Receives SSR" option to Shader Graph

FrancescoC-unity:
### Purpose of this PR

Exposes the option "Receives SSR" to shader graph. Used to occasion to change how the stencil mask and ref was passed to the shader, making it a bit more robust to future changes, instead of hard coded values. It now also matches the logic of the non-graph shader w.r.t. the stencil ref/mask. 

In the image: left plane has a material with Receives SSR enabled, the right plane has it disabled.  
image

---
### Release Notes
Exposes the option "Receives SSR" to shader graph.

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=expose-receives-ssr-option-to-shader-graph&automation-tools_branch=master&unity_branch=2018.3%2Fstaging [Running]

**Manual Tests**:  Tried to turn on/off the option from a material created with the shader graph. 

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Low

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
@@ -454,6 +454,20 @@ namespace UnityEditor.ShaderGraph
         }
 
         [SerializeField]
+        bool m_ReceivesSSR = true;
+        public ToggleData receiveSSR
+        {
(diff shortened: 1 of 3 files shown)


sebastienlagarde:
This PR target release 2018.3 in order to pass katana test, but once validated we need to target master instead and backport to 2018.3

PR #2247, opened 2018-10-22
Expose "Receives SSR" option to Shader Graph

FrancescoC-unity:
### Purpose of this PR
Exactly the same as https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2238 but with base and target set as master instead of 2018.3

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=expose-receives-ssr-option-to-shader-graph&automation-tools_branch=master&unity_branch=2018.3%2Fstaging

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
@@ -675,11 +675,6 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
                 activeFields.Add("Decals");
             }
 
-            if (masterNode.receiveSSR.isOn)
-            {
-                activeFields.Add("SSR");
-            }
-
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitSubShader.cs
```diff
@@ -675,6 +675,11 @@ private static HashSet<string> GetActiveFieldsFromMasterNode(INode iMasterNode,
                 activeFields.Add("Decals");
             }
 
+            if (masterNode.receiveSSR.isOn)
+            {
+                activeFields.Add("SSR");
```
russoh: Where does the active field "SSR" get used? I didn't see any template hookup that reads this.

FrancescoC-unity: Ha, yes, I just followed what "Decals" was doing and forgot to follow up on where that was used. I can remove that 🙂

PR #2251, opened 2018-10-23
Fix for refraction plane model

FrancescoC-unity:
### Purpose of this PR
Fix for plane refraction model. It was using the wrong output direction. 

---
### Release Notes
Fix for plane refraction model. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix-plane-model-for-refraction&automation-tools_branch=master&unity_branch=2018.3%2Fstaging 

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: None


FrancescoC-unity:
After discussing with Fred, it's clear this was meant to be a "window" mode rather than plane.  I'll make a larger PR adding a third refraction mode and renaming the Plane we already have to "window"

PR #2259, opened 2018-10-24
Simple fix due to last changes in trunk

PaulDemeulenaere:
### Purpose of this PR
Simple fix build due to last changes in trunk


com.unity.render-pipelines.high-definition/Editor/Material/Fabric/ShaderGraph/FabricSubShader.cs
```diff
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
 using UnityEditor.Graphing;
 using UnityEditor.ShaderGraph;
-using UnityEngine.Experimental.Rendering;
+using UnityEngine.Rendering;
```
AnisB: Thanks for this

PR #2265, opened 2018-10-24
Fixing the fabricshadergraph conflict with specular occlusion

anisunity:
Fixing the fabricshadergraph conflict with specular occlusion
Adding a selection pass to the fabric shadergraph


sebastienlagarde:
This PR is required to fix master. (Current failure on master is due to this, so merging and launch katana test on master)

PR #2273, opened 2018-10-25
Refraction/Distortion bug fixes 

FrancescoC-unity:
### Purpose of this PR
This PR contains fixes/alleviation of the following bugs: 

https://fogbugz.unity3d.com/f/cases/1082733/ 
additiveblur

The following two especially are quite an hack and don't fully fix the issues, but they alleviate them. Not perfect at all, but cheap.  

https://fogbugz.unity3d.com/f/cases/1059310/
fetchoutside_before_after
2_prefetchoutside

https://fogbugz.unity3d.com/f/cases/1059312/
distortionhalo

https://fogbugz.unity3d.com/f/cases/1067061/
(This fix comes from an extract from @EvgeniiG 's PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2261/files ) 
4_offsetimage

For further reference, the pool scene is not regressing with these changes
7_pool_beforeafter

---
### Release Notes
Various distortion and refraction bug fixes. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=refraction-fixes&automation-tools_branch=add-platform-filter&unity_branch=2018.3%2Fstaging

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


EvgeniiG:
I feel like the first image looked better before. Is it just me?

FrancescoC-unity:
> I feel like the first image looked better before. Is it just me?

Personally I think the same, but the bug description mentioned the additive blur as the expected behaviour (first image is max blur)

sebastienlagarde:
I think we need to see in practice :), Neither Max nor Add make sense, but we don't have better blending mode. With add we will be blur too fast (which is what we see on the picture), with Max we don't blur enough. In addition the blur strength shouldn't be linear. Let's use add and we will see what artists say.

sebastienlagarde:
Katana green.

PR #2278, opened 2018-10-25
MSAA and Lit ShaderGraph Corrections

anisunity:
Correctly ouput normal during prepass for lit
Correct a bug in atmospheric scattering for MSAA
Adding tests for MSAA
Support MSAA for ShaderLitMode Both when FrameSettings is Forward

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix%2Fmsaa_and_lit_graph&automation-tools_branch=add-platform-filter&unity_branch=trunk#


sebastienlagarde:
Added @russoh for awareness about the fix of MSAA for shader graph

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/FrameSettingsUI.Drawers.cs
```diff
@@ -134,10 +134,12 @@ static void Drawer_SectionRenderingSettings(FrameSettingsUI s, SerializedFrameSe
                 area.Add(p.litShaderMode, litShaderModeContent, () => p.overridesShaderLitMode, a => p.overridesShaderLitMode = a,
                     () => !GL.wireframe && hdrpSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.Both,
                     defaultValue: defaultShaderLitMode);
+
+                bool assetAllowMSAA = hdrpSettings.supportedLitShaderMode != RenderPipelineSettings.SupportedLitShaderMode.DeferredOnly && hdrpSettings.supportMSAA;
+                bool frameSettingsAllowMSAA = p.litShaderMode.enumValueIndex == (int)LitShaderMode.Forward && p.overridesShaderLitMode || !p.overridesShaderLitMode && defaultShaderLitMode == LitShaderMode.Forward;
```
RSlysz: Just remembered why all boolean where concatenated in same line.
Here you have a small window between the evaluation of p.litShaderMode and the actual drawing of the enableMSAA FrameSettings parameter in inspector (which is drawn letter in the Draw function that will evaluate the lambda where boolean where initially). It means that you can have a wrongly drawed (enable or disabled) property for the frame where it change if the user is really quick.
Now it should be ok again at next editor refresh and the window should be really small. Depending on how events are handled on inspector, it could even not exist at all.

sebastienlagarde:
Katana green

PR #2280, opened 2018-10-25
Screen space refraction settings cleanup and share fade function with SSR

FrancescoC-unity:
### Purpose of this PR

This PR has few of changes in: 

- Cleaning up screen space refraction settings (removing old HiZ stuff) 
- Decoupling SSR and Screen Space Refraction settings (getting rid of ScreenSpaceLighting settings).
- Renaming the "Plane" Refraction model to "Box" as it is more intuitive. 
- Sharing of Fade function between SSR and Screen Space Refraction. 

---
### Release Notes

Remove unused refraction settings and use same fade function for refraction and reflection. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=refraction-cleanup-and-fade-sharing-with-ssr&automation-tools_branch=add-platform-filter&unity_branch=trunk

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceLightingCommon.hlsl
```diff
@@ -0,0 +1,6 @@
+﻿float EdgeFade(float2 coordNDC, float fadeRcpLength)
```
FrancescoC-unity: I added a new file since the already common one is dedicated to shader vars only. I am a bit unsure what's our convention on that, if we are fine with mixing uniform declaration and common functions, I'd move the uniform declaration in this file (or this function in the other file + modifying the name)

EvgeniiG: This functionality is not specific to screen-space lighting. I don't think we need another file. Maybe we should rename the function and then put it elsewhere.

EvgeniiG: To be more specific, this one should probably go into the shader library.

sebastienlagarde: I will suggest to move to commonMaterial.hlsl.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -1584,12 +1584,10 @@ IndirectLighting EvaluateBSDF_ScreenspaceRefraction(LightLoopContext lightLoopCo
 
     // Resolve weight and color
 
-    // Fade pixels near the texture buffers' borders
-    float2  weightNDC = clamp(min(hit.positionNDC, 1 - hit.positionNDC) * _SSRefractionInvScreenWeightDistance, 0, 1);
-            weightNDC = weightNDC * weightNDC * (3 - 2 * weightNDC);
     // TODO: Fade pixels with normal non facing the ray direction
```
sebastienlagarde: you can reemove these comment

EvgeniiG:
Do not put it into Common.hlsl. Unless it's used in a ton of places, no function should go into Common.hlsl. This file is included literally everywhere. We should keep it as small as possible.

FrancescoC-unity:
> Do not put it into Common.hlsl. Unless it's used in a ton of places, no function should go into Common.hlsl. This file is included literally everywhere. We should keep it as small as possible.

Where do you think it would fit better? This doesn't feel either material or lighting specific. If I had to chose, i'd go for material probably, but either way it doesn't fit that well :\

EvgeniiG:
Maybe ScreenSpaceLighting.hlsl.

EvgeniiG:
It's also literally 3 lines of dumb code, you could just inline it twice.

sebastienlagarde:
This PR cause 7 graphic test to fail, looks like refraction/distortion don't work at all

FrancescoC-unity:
> This PR cause 7 graphic test to fail, looks like refraction/distortion don't work at all

Uh! Odd, I certainly had it working locally, but that was pre merging. I'll check asap :)

PR #2289, opened 2018-10-26
Various fixes in the LIt Shader Graph and the Fabric Shader Graph

anisunity:
Removing obsolete Refraction code in the fabric master node (Fabric don't have refraction)
Adding "SSR control on material" code in the Fabric master node
Correcting the deferred pass in the Lit master node
Correcting the smoothness output in Lit master node and fabric, mean SSR wasn't working

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix%2Flit_and_fabric_graph&automation-tools_branch=add-platform-filter&unity_branch=trunk


sebastienlagarde:
Katana is green

PR #2298, opened 2018-10-29
force shadow qualities and restore them when exiting hdrp

RSlysz:
### Purpose of this PR
Fiw shadow not working in hdrp when a low quality settings was selected.

---
### Release Notes
Shadow will now always work even if low quality selected.

---
### Testing status
**Katana Tests**: 

**Manual Tests**: Tested locally (displaying locally the legacy field to check, and checking directly with shadow too)

**Automated Tests**: 

---
### Overall Product Risks
**Technical Risk**: Low
(few lines of code)

**Halo Effect**: None
(Should only affect shadow qualities when entering, exiting hdrp)

---


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added multiple supported LitShadermode and per renderer choice in case of both Forward and Deferred supported
 - Added custom specular occlusion mode to Lit Shader Graph Master node
 - Added separate editor resources file for those resources to not be taken in player builds.
+- Added shadow displayed in low quality settings.
```
sebastienlagarde: Move this to 5.1.0

sebastienlagarde: also the comment is wrong, it is force shadow qaultiy settings to all in HDRP

PR #2307, opened 2018-10-30
Made post-processing tests slightly less strict

Chman:
Post-processing tests are a bit too strict and can potentially fail when ran on different hardware. This PR slightly ups the error tolerance.


sebastienlagarde:
Merging without katana test as it this PR can't fail

PR #2317, opened 2018-10-30
Move NormalRNM node to an option on Normal Blend node

Kink3d:
### Purpose of this PR
NormalBlendRNM should not have been a separate node. This moves the functionality to an option of the Normal Blend node. Also solves a number of issues with naming and HLSL snippet conventions.

---
### Release Notes
Changelog

---
### Testing status
**Katana Tests**: Running

**Manual Tests**: None

**Automated Tests**: None

Any test projects to go with this to help reviewers? No

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High? Next to none

**Halo Effect**: None, Low, Medium, High? None

---
### Comments to reviewers


sebastienlagarde:
Adding  @anisunity to this PR as this will break our fabric graphic test, it will require an update on our side.

I am ok with the PR btw :)

Kink3d:
Ok, the node will upgrade automatically, youll just need to change the dropdown to "Reoriented" mode (we cant make it do that on upgrade afaik). If you could do that then merge this PR and its backport (#2318) that would be great, as I REALLY want to get in before we cut the 18.3 package tomorrow and dont want to mess with your tests and risk breaking something.

Thanks so much!

sebastienlagarde:
This change need to be in 2018.3 but katana is in trouble currently unable to launch any test. Change is small and we have checked with @anisunity it will not affect our test (this is not used in enabled test on our side). Merging

PR #2318, opened 2018-10-30
Backport: Move NormalRNM node to an option on Normal Blend node

Kink3d:
Backport #2317 to `release/2018.3`


sebastienlagarde:
Confirmed merging as explain in original PR.

PR #2333, opened 2018-11-02
Fix color pyramid MIP count computation & expose a problem with MIP chains and viewport scale

EvgeniiG:
### Purpose of this PR
Fix color pyramid MIP count computation
Expose a problem with MIP chains and viewport scale

### Comments to reviewers
Note that the RTHandle scaling system, in its current state, does not work correctly with (MIP level > 0) if the viewport scale is not 1. Please see the comment in the code for explanation.

## Katana
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix_color_pyramid_mip&automation-tools_branch=add-platform-filter&unity_branch=trunk


sebastienlagarde:
Failure of katana is due to LW test that is now fix on master. The failing test on HDRP was just a reference image missing. This have been update in a commit. Katana is green otherwise. Merging.

PR #2335, opened 2018-11-02
Fix instancing defines in HD ShaderGraph shaders

Kink3d:
### Purpose of this PR
Fixes FB#1095771 (but the issue was actually larger than this). Previously SG generated shaders were only checking one of the 3 instancing defines which lead to incorrectly defined instancing vars in attirbutes. This PR adds a new define that represents "any instancing path is enabled".

---
### Release Notes
Changelog

---
### Testing status
**Katana Tests**: Running...

**Manual Tests**: Tested with the FB case project, tested in some other basic environment.

**Automated Tests**: None

Any test projects to go with this to help reviewers? See FB case

---
### Overall Product Risks
**Technical Risk**: Low?

**Halo Effect**: Low

---
### Comments to reviewers
None


sebastienlagarde:
hmm.... I don't see any occurrence of UNITY_ANY_INSTANCING_ENABLED either in legacy unity (in original UnityInstancing.cginc) or in github SRP. So I expect this define is only for shader graph to correctly generate the instance ID value in input struct?

In this case the true fix shouldn't to add a new define that is use solely by shader graph but rather support the test of multiple define (That may be useful for other case?)

[Semantic("INSTANCEID_SEMANTIC")] [PreprocessorIf("UNITY_ANY_INSTANCING_ENABLED", "PROCEDURAL_INSTANCING_ON", "STEREO_INSTANCING_ON")] uint instanceID;

As I suppose this is more painful to implement, I am still ok with the PR :)

Kink3d:
@sebastienlagarde Yea we didnt really want to risk changing our preprocessor if code this late in the game and figured adding that extra define to core wasnt that big of a deal?

sebastienlagarde:
yes, I get it, I am fine with the PR :)

spaztronaut:
zack already tested this PR. removing mah self

sebastienlagarde:
Ping. This PR is ready. Do you want help to merge and backport it ?

Kink3d:
Tests are green against trunk

PR #2343, opened 2018-11-05
Fixing compilation error on the fabric shader graph

anisunity:
All in the title


sebastienlagarde:
Note: This PR has fixed the error in the log of the editor but not the problem with katana that appear on master itself. merging it, but this haven't solve the issue.

PR #2345, opened 2018-11-05
Move UnsafeTestUtils back to Core

Kink3d:
### Purpose of this PR
This makes `CoreRP_Tests` project compile again. Will run tests on it...

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix-move-unsafetestutils-to-core&automation-tools_branch=add-platform-filter&unity_branch=trunk#


sebastienlagarde:
Note: After a Katana run, this PR fix the error in Core log as intented, but of course it doesn't solve th e problem of katana. 
Another PR fix the error in HDRP log: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/1f25950d59e59a25625c96fa9305d99f574efb68 but it is also not the root cause of current breaking of HDRP in test editor. I merge this PR as it fix the problem it intent to solve.

PR #2346, opened 2018-11-06
Fix shader include path causing compilation error in BoatAttack

Wilfrid-Unity:
### Purpose of this PR
Fixes shader compilation error visible when opening github.com/Verasl/BoatAttack/commit/5f8123fe

---
### Release Notes
Fixed minor shader include error

---
### Testing status

**Manual Tests**: What did you do? Checked that BoatAttack project could run in the WindowsEditor after the fix

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low

---
### Comments to reviewers
Minor fix - thought updating Changelog.md was not required


phi-lira:
LGTM. I also agree this doesn't need a changelog as the issue is most likely because BoatAttack is doing a UsePass with this shader. Which will break the relative include path.

PR #2348, opened 2018-11-06
Added a debug window widget for bit fields

JulienIgnace-Unity:
### Purpose of this PR
Added a debug window widget for bit fields

Katana:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FDebugUI-BitField

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Core/Debugging/DebugUIDrawer.Builtins.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Core/Debugging/DebugUIDrawer.Builtins.cs
@@ -164,7 +164,7 @@ namespace UnityEditor.Experimental.Rendering
         public override bool OnGUI(DebugUI.Widget widget, DebugState state)
         {
             var w = Cast<DebugUI.BitField>(widget);
-            var s = Cast<DebugStateInt>(state);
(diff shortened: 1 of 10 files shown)


com.unity.render-pipelines.high-definition/Editor/Core/Debugging/DebugUIDrawer.Builtins.cs
```diff
@@ -158,6 +158,26 @@ public override bool OnGUI(DebugUI.Widget widget, DebugState state)
...
 
+    [DebugUIDrawer(typeof(DebugUI.BitField))]
+    public sealed class DebugUIDrawerBitField : DebugUIDrawer
+    {
+        public override bool OnGUI(DebugUI.Widget widget, DebugState state)
+        {
+            var w = Cast<DebugUI.BitField>(widget);
+            var s = Cast<DebugStateFlags>(state);
+
+            EditorGUI.BeginChangeCheck();
```
Chman: We have to use scopes now instead of Begin/End methods.

sebastienlagarde:
Katana fail on test on LW which is expected, so green.

PR #2355, opened 2018-11-07
Forward Opaque pass LoadActions

jonuuukas:
### Purpose of this PR
Introducing methods to pick LoadAction according to the Camera Clear Flags option in RenderForwardOpaquePass
---
### Release Notes
---
### Testing status
**Katana Tests**: Failing to build. Seems like a non-related issue

**Manual Tests**: Tested various combinations of cameras with different ClearFlag settings

---
### Overall Product Risks
**Technical Risk**: Low - not much being changed, minor risk of affecting performance as Load is more perf heavy than DontCare

**Halo Effect**: Medium - affects all Gfx Api's that support LoadActions

---
### Comments to reviewers


phi-lira:
Do we need to also change Load action for the depth prepass?

phi-lira:
We require engine changes for the DontCare to work as expected right?

jonuuukas:
> Do we need to also change Load action for the depth prepass?
i don't think that's needed as there is nothing to load there yet
> We require engine changes for the DontCare to work as expected right?
Yes and just to be exact - for Load to work as expected

ernestasKupciunas:
There were no problems related to this PR. I tried various settings of the camera and different Graphics APIs(Gles2, Gles3, Metal, Vulkan).
Verified with:
Version 2019.1.0a10 (84f9b5a9b69f)
Sat, 17 Nov 2018 19:19:42 GMT
Branch: trunk

Devices under testing:
VLNQA00175 Samsung Galaxy Note9, GPU: Mali-G72, OS: 8.1.0
VLNQA00216 Razer Phone 2*, GPU: Adreno (TM) 630, OS:8.1.0
000500 iPhone 7, SoC: A10, iOS: 10.2
VLNQA00212 iPhone Xs, SoC: 12, iOS: 12.0

phi-lira:
Tests pass. The only one failing is HDRP 8202_BentNormals which is fixed on master.

PR #2356, opened 2018-11-07
Evaluate fog using distance rather than depth

EvgeniiG:
### Purpose of this PR
- Use sphere-capped frustum for volumetrics as presented at Siggraph 2018
- Switch from clustered to Big Tile for volumetrics for scalarization (reduces VGPR pressure)
- Apply linear/exponential/volumetric fog using distance rather than depth
- Expose "Max Fog Distance" parameter for shading the far plane
- Solve FogBugz issue https://fogbugz.unity3d.com/f/cases/1098391
- Optimize the fog code

All fog types should be now compatible with rendering to cubemap.

### Katana
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sphere_capped_pyramid_frustum&automation-tools_branch=add-platform-filter&unity_branch=trunk


EvgeniiG:
Note that linear and exponential fogs were using linear depth values which were unrelated to the positionWS (the latter was modified at the far plane, and the former wasn't, and both were used). Since I fixed it, linear and exponential fogs produce different visuals. Hard to say whether they are better or not...

EvgeniiG:
After updating ref images, tests pass locally. Running Katana again.

EvgeniiG:
With the new ref images, Katana is now green.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild-clustered.compute
```diff
@@ -12,6 +12,7 @@
 
 #pragma kernel ClearAtomic
 
+#pragma enable_d3d11_debug_symbols
```
sebastienlagarde: remove it

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -213,10 +213,11 @@ CBUFFER_START(UnityGlobal)
 
     // TODO: put commonly used vars together (below), and then sort them by the frequency of use (descending).
     // Note: a matrix is 4 * 4 * 4 = 64 bytes (1x cache line), so no need to sort those.
-#if defined(USING_STEREO_MATRICES)
-    float3 _Align16;
-#else
+#ifndef USING_STEREO_MATRICES
     float3 _WorldSpaceCameraPos;
```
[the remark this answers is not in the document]
EvgeniiG: I have no idea. It was like this before.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/lightlistbuild-clustered.compute
EvgeniiG: Done.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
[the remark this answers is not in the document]
sebastienlagarde: ok so I think in our case it is useless, however it mean that we miss an update of this struct:

CBUFFER_START(UnityPerPassStereo)
float4x4 _ViewMatrixStereo[2];
float4x4 _ProjMatrixStereo[2];
float4x4 _ViewProjMatrixStereo[2];
float4x4 _InvViewMatrixStereo[2];
float4x4 _InvProjMatrixStereo[2];
float4x4 _InvViewProjMatrixStereo[2];
float4x4 _PrevViewProjMatrixStereo[2];
float4   _WorldSpaceCameraPosStereo[2];
#if SHADER_STAGE_COMPUTE
// Currently the Unity engine doesn't automatically update stereo indices, offsets, and matrices for compute shaders.
// Instead, we manually update _ComputeEyeIndex in SRP code. 
float _ComputeEyeIndex;
#endif
CBUFFER_END

I think you need to add float4   _WorldSpaceCameraPrevPosStereo[2];

and later the code use     #define _WorldSpaceCameraPos _WorldSpaceCameraPosStereo[unity_StereoEyeIndex].xyz

so you need to add the define here too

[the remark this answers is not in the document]
EvgeniiG: I will not add any VR code as a part of this PR.

sebastienlagarde:
@nerites : For your information, this PR add a previous relative world position camera : The change have been done here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2356/commits/cb729b81baccf59f576209c0c0f87fcc24108f5c. IIRC volumetric currently doesn't work with VR, but just in case there is this code that have been updated, until the work begin on it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLightingController.cs
```diff
@@ -2,9 +2,9 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
 {
     public class VolumetricLightingController : VolumeComponent
     {
-        [Tooltip("Near and far planes of camera's volumetric lighting buffers (in meters).")]
```
sebastienlagarde: FYI, when changing name like this you need to take care of serialization of previous data, here I have update the code with 

    public class VolumetricLightingController : VolumeComponent
    {
        [Tooltip("Range from the front to the back of the camera's volumetric lighting buffers (in meters).")]
        [FormerlySerializedAs("depthRange")]
        public FloatRangeParameter distanceRange = new FloatRangeParameter(new Vector2(0.5f, 64.0f), 0.01f, 10000.0f);
        [Tooltip("Controls the slice distribution: 0 = exponential (more slices near the camera, fewer slices far away), 1 = linear (uniform spacing).")]
        [FormerlySerializedAs("depthDistributionUniformity")]
        public ClampedFloatParameter sliceDistributionUniformity = new ClampedFloatParameter(0.75f, 0, 1);
    }

TestProjects/HDRP_Tests/Packages/manifest.json
```diff
@@ -1,10 +1,46 @@
 {
   "dependencies": {
```
sebastienlagarde: Caution: This file must not be update for graphic test, I will revert the change

TestProjects/HDRP_Tests/ProjectSettings/ProjectVersion.txt
```diff
@@ -0,0 +1 @@
+m_EditorVersion: 2019.1.0a9
```
sebastienlagarde: Caution: This file must not be update for graphic test, I will revert the change

sebastienlagarde:
Katana is green except one test in LW not related to this PR. Merging

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLightingController.cs
EvgeniiG: Thanks.

PR #2359, opened 2018-11-07
Hair Shader Graph and Fabric SG Fix

anisunity:
Initial Implementation of the Hair Shader Graph and fix in the fabric shader graph


pbbastian:
Please make sure to put this into the HDRP namespace, as the `UnityEditor.ShaderGraph` is reserved for the Shader Graph package :) There hasn't really been a precedent of doing this so far, but we're changing this with an upcoming PR. Also, make sure all the classes exposing Shader Graph stuff are internal, as they will otherwise fail to compile in the future.

sebastienlagarde:
>Please make sure to put this into the HDRP namespace
Will do that. Thanks

PR #2360, opened 2018-11-07
Fix terrain tangent frame

zeroyao:
### Purpose of this PR
Terrain tangent frame needs to be flipped because the bitangent vector is pointing the wrong direction.
That caused the normal textures from terrain layers to be flipped.
---
### Release Notes
Fixed the normal mapping for TerrainLit shader.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix_terrain_tangent&automation-tools_branch=add-platform-filter&unity_branch=trunk

**Manual Tests**: Created a terrain with TerrainLti and a plane with a Lit shader and compared the normal vector output to the g-buffer. Before the fix the normal is flipped along U-axis. After the fix they match.

**Automated Tests**: 

TerrainLit test scene should cover the case and make sure it doesn't regress over time, but it doesn't seem to run on Katana?

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---


zeroyao:
@sebastienlagarde How do I run graphics tests on Katana for SRP? They appear not to be included? Also I got red playmode tests:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix_terrain_tangent&unity_branch=trunk&automation-tools_branch=master
Not really sure why. Any hint?

sebastienlagarde:
Hey, so I haven't merge the PR yet, because it looks like more a workaround than a real fix of the root issue. Isn't the convention of terrain texture that is not following the unity one? i.e top-bottom (dx) instead of bottom up (OGL), or something like that? 

What I wondering is if there is a problem on UV / Tangent generation side of the terrain. Current patch looks like just: I setup "-1" because why not :).

Do you have more info?

Otherwise there is only one terrain test in current graphic test (don't know who ahve added it), but you should add more - And it is expected that this test fail with your change, the screenshot need to be update. You can run test locally by loading the HDRP_test project and then Windows -> General -> Test Runner. Then select Playmode and right click and Run the test.

Note I asked Sebastien Lachambre to perform some test (without the fix), just to validate  there is trouble.

debugging normals

so we confirm.

zeroyao:
Test 1602_TerrainLit_Normal is made to verify the normal automatically. @sebastienlagarde

PR #2365, opened 2018-11-08
Fix d3d warnings

FrancescoC-unity:
### Purpose of this PR

Fixed all the D3D warnings I could find by running some sample scenes and all graphics tests locally. 
There are a bunch: 

- Warning for outputing to a non bound RT in Depth Only.
- Warning for outputing to a non bound RT in Decal normal buffer pass.
- Error  for UAV type mismatch (missing unorm) for DecalNormalBuffer.
- Warning for outputing to a non bound RT in shadow clear pass.
- Error for UAV type mismatch (missing unorm) for DecalHTile.
- Warning for not binding an MSAA depth buffer for FPTL when MSAA enabled.
- Warning for outputing to a non bound RT in several stencil copy passes. 

---
### Release Notes

Fix D3D warning in HDRP.

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix-d3d-warnings&automation-tools_branch=add-platform-filter&unity_branch=trunk# [running, Mac passed already]

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Medium


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/CopyStencilBuffer.shader
```diff
@@ -17,9 +17,16 @@ Shader "Hidden/HDRenderPipeline/CopyStencilBuffer"
...
 
     int _StencilRef;
+
+    // Explicit binding not supported on PS4
+#ifdef SHADER_API_PSSL
     RW_TEXTURE2D(float, _HTile); // DXGI_FORMAT_R8_UINT is not supported by Unity
     RW_TEXTURE2D(float, _StencilBufferCopy); // DXGI_FORMAT_R8_UINT is not supported by Unity
-
+#else
+    RW_TEXTURE2D(float, _HTile) : register(u1); // DXGI_FORMAT_R8_UINT is not supported by Unity
```
JulienIgnace-Unity: I see that we have that (explicit binding) in several places. Maybe we could include that into the macro somehow?
Not sure if that would be better honestly (hides the fact that it's not supported everywhere... but that's already the case for other things so...eh)

FrancescoC-unity: > hides the fact that it's not supported everywhere 

Yeah, that was my reasoning for not doing it via Macro. But I see some merit in having the register included in the macro as well, so since we are at least two now, I'll do the change. 

I originally had in mind adding a macro variant a-la: 

`#define RW_TEXTURE2D(type, textureName,registerID)       RWTexture2D<type> textureName : register(registerID)
`

Did you have something different in mind? I would argue against a new macro name (e.g. RW_TEXTURE2D_WITH_BINDING) as a user that knows that explicit binding is not supported on PS4 might opt against it

sebastienlagarde: is macro overloading supported ? if it is the case then I am fine with it.

FrancescoC-unity: Apparently there is no variadic macro support in HLSL pre-processor, which makes the overloading not possible 🙁  

Given that, for the reasons above, I am not convinced that creating a differently named macro with explicit binding is a great way to go :/

sebastienlagarde: ok, then instead of  
#ifdef SHADER_API_PSSL
let's do a define for platform_support_rwbuffer_binding (or similar) as maybe other platform can have the issue.
Also add a comment for why we do explicit binding (i.E t remove warning XXX), as it is not obvious.
thanks

sebastienlagarde:
Katana is green - merging

PR #2368, opened 2018-11-08
float4 only in unityPerDraw

arnaudcarre:
### Purpose of this PR
our UnityPerDraw used some float3 in UnityPerDraw. Up coming SRP batcher change will check that we only use float4 or float4x4 in UnityPerDraw

---
### Release Notes

---
### Testing status

---
### Overall Product Risks

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -81,8 +81,8 @@ CBUFFER_START(UnityPerDraw)
     // z = Texel size on U texture coordinate
     float4 unity_ProbeVolumeParams;
     float4x4 unity_ProbeVolumeWorldToObject;
-    float3 unity_ProbeVolumeSizeInv;
-    float3 unity_ProbeVolumeMin;
+    float4 unity_ProbeVolumeSizeInv;
```
sebastienlagarde: please add a comment why it is move to float4 as this is different from legacy unity and people may wonder why

PR #2370, opened 2018-11-09
Fix script compilation error happening in stand-alone

Wilfrid-Unity:
### Purpose of this PR
Minor fix to script that could not compile in an IL2CPPstandalone platform

---
### Testing status
**Manual Tests**: 
Manually checked that after the fix BoatAttack could be built on Switch (Using Unity Editor revision aad884156933

https://github.com/Verasl/BoatAttack

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
I am actually not sure if error was related to IL2CPP or just caused by non-Editor-mode includes.


com.unity.render-pipelines.lightweight/Runtime/Data/LightweightRenderPipelineAsset.cs
```diff
@@ -483,7 +484,7 @@ float ValidateShadowBias(float value)
 
         int ValidatePerObjectLights(int value)
         {
-            return Math.Max(0, Math.Min(value, LightweightRenderPipeline.maxPerObjectLightCount));
+            return Mathf.Max(0, Mathf.Min(value, LightweightRenderPipeline.maxPerObjectLightCount));
```
jcs090218: Interesting that this is getting the compilation error. Using `Math` instead of `Mathf` seems to be the proper way to deal with in this particular case.

Wilfrid-Unity: Good catch jcs090218
I added a change to use System.Math.Max/Min instead

PR #2374, opened 2018-11-09
fix simple lit shader

sebastienlagarde:
Fix a missing update of GetPunctualShadowAttenuation for simple lit after PR for improvess PCSS quality

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP-fix-simple-lit%2Fmaster&unity_branch=trunk&automation-tools_branch=add-platform-filter#


sebastienlagarde:
Katana is green. merging

PR #2381, opened 2018-11-10
Error Handling for Shader Graph

marctem:
### Purpose of this PR
Display errors as badges on the nodes that caused them. This includes compilation errors and node validation errors.

This also removes the "minimal preview rendering" that we did back when we compiled a shader per node. Now we just draw each preview every frame (since that's cheap) and only recompile the preview shader when the graph changes.

---
### Release Notes
Show an error badge on each node that causes an error in the shader.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=shadergraph%2Ferror-handling&unity_branch=graphics%2Fshadergraph%2Ferror-bindings-2&automation-tools_branch=add-platform-filter

**Manual Tests**: Lots of creating nodes with errors, deleting, connecting, disconnecting, saving, etc.

**Automated Tests**: Added tests for ShaderSourceMap (which was broken before). Will add some for MessageManager when I update for PR feedback!

---
### Overall Product Risks
**Technical Risk**: Medium - no automated tests, had to make way more changes than I wanted because Shader errors were broken for us and our error handling was not developed at all. Also depends on a trunk change to expose one more function to C#.

**Halo Effect**: Low - Shader Graph only, but it reaches from the UI down to the Shader data layer.


zadi0648:
Dragging & dropping a newly created property creates an error badge that states "Error Found During Node Validation".

Repro:
1) Create a new ShaderGraph file.
2) Create a new Texture2D Property.
3) Drag the property into the graph view to create a node.
Result) Error

Newly created Texture2D:
texture2derror0

Each property I created would show an error badge until a new one was created, or enough "other things happened". Below the Texture2DArray is the newest property and this displays the error:
texture2derror1

zadi0648:
(I believe you may have brought this up before Marc and said we can't do anything about it but in case I'm remembering incorrectly) It appears only the first node displays the error badge. Deleting the first node doesn't transfer the error to the 2nd. Appears to reproduce regardless of how the nodes are connected.

onlyappliedtofirstnode0

zadi0648:
The last word of the warning often doesn't fit in the provided space depending on the zoom level.

textdoesntfitatcertainzoomlevels0

zadi0648:
It appears that some error badges will disappear when a new error comes up.

I'm not sure if this repro is great because I broke nodes inside the branch in order to test. It appears to reproduce regardless of how I break the nodes but I'm not 100% sure of that.

Repro:
1) Open up this project: https://drive.google.com/open?id=1sOC93ZSKuaPigxFEM1Z5odPZBju0nirz (will require fixing the project manifest)
2) Break the Random Range Node.
3) Break the Cosine Node.
4) Break the Polygon Node.
5) Open up the "PBR" ShaderGraph.
6) Notice that there are three error badges.
7) Break the Dither Node.
Result) Error Badges for Cosine & Polygon Node disappears. Only the top Dither node displays an error badge and Random Range.

marctem:
To respond to Zack's issues:

1. Property Node - fixed.
2. Dupe error nodes:
    a. Badge shows up on one - can't fix currently.
    b. Badge doesn't appear on the second after delete - something weird with serialization, will fix with a bugfix PR next week.
3. Warning tooltip doesn't fit - GraphView bug, they should be fixing it
4. Error badges disappear - Due to how code is generated, can't fix currently
5. Can't Save Stuck State - pretty sure is related to 2b, will fix with bugfix PR.

PR #2382, opened 2018-11-12
changing *LitInput.hlsl CBUFFER values to float precision

jonuuukas:
### Purpose of this PR
SRP Batcher on metal doesn't support half values in CBUFFER structs. Therefore, changing the precision enables user's to use SRP Batcher to increase perf and etc.

Changes were introduced in the Megacity mobile demo.

### Testing status
**Katana Tests**: running katana at the moment

**Manual Tests**: checked performance/visual correctness on megacity example

---
### Overall Product Risks
**Technical Risk**: low (might affect performance/mem usage a little bit

**Halo Effect**: low (most other platforms already use float)

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.lightweight/Shaders/LitInput.hlsl
+++ b/com.unity.render-pipelines.lightweight/Shaders/LitInput.hlsl
@@ -7,15 +7,15 @@
 
 CBUFFER_START(UnityPerMaterial)
 float4 _MainTex_ST;
-half4 _Color;
-half4 _SpecColor;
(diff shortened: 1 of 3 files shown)


phi-lira:
This is not the proper way to fix the issue, it was ok for MegaCity because the demo was CPU bound. The proper way is to add support to half precision on SRP batcher, which is being worked atm.

phi-lira:
Closing as the proper fix is coming for the SRP batcher.

PR #2393, opened 2018-11-12
Transparent MSAA Support

anisunity:
MSAA support for transparents and updating the msaa tests


sebastienlagarde:
KAtana is green except on the test from LW that fail on master. Merging

PR #2394, opened 2018-11-13
Transparent MSAA Support (#2393)

sebastienlagarde:
backport Transparent MSAA Support (#2393)


sebastienlagarde:
katana is green

PR #2397, opened 2018-11-13
Perform a small optimization of HDRP shader compile times [On Hold]

EvgeniiG:
### Purpose of this PR
- Add 'MaterialInstanceFlags' which are used for conversion of Shader Features into uniform branches
- Remove '_PIXEL_DISPLACEMENT_LOCK_OBJECT_SCALE' and '_DISPLACEMENT_LOCK_TILING_SCALE' as an example

## Testing status
**Katana Tests**: soon (tm)

### Comments to reviewers
Upgrading old materials needs to be tested.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
@@ -9,6 +9,8 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/CommonMaterial.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/EntityLighting.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl"
+
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/MaterialInstanceFlags.cs.hlsl"
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.hlsl"
 
 // Guidelines for Material Keyword.
--- a/com.unity.render-pipelines.high-definition/Editor/Material/Lit/BaseLitUI.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Material/Lit/BaseLitUI.cs
@@ -261,7 +261,7 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
         protected override void BaseMaterialPropertiesGUI()
         {
             base.BaseMaterialPropertiesGUI();
-            
+
             // This follow double sided option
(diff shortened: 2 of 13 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/Lighting.hlsl
```diff
@@ -7,34 +7,7 @@
...
-#endif
-
-#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl" // Depends on LightLoopDef and shadows
-
-// Volumetrics have their own light loop.
-#ifndef UNITY_MATERIAL_VOLUMETRIC
-    // LightLoop use evaluation BSDF function for light type define in Material.hlsl
-    #if defined(LIGHTLOOP_SINGLE_PASS) || defined(LIGHTLOOP_TILE_PASS)
-        #if HDRP_MATERIAL_TYPE_SIMPLE
-            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/SimpleLightLoop.hlsl"
```
sebastienlagarde: Be careful, VFX graph will break. We need to update VFX graph to it use the SimpleLightLoop.hlsl file correctly

EvgeniiG:
Should I manually upgrade test scenes? Do we want this PR at all?

com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
```diff
@@ -9,6 +9,8 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/CommonMaterial.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/EntityLighting.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl"
+
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/MaterialInstanceFlags.cs.hlsl"
```
sebastienlagarde: I don't like the fact that we have something specific to Lit/layeredlit shader, as something general. 
(Note: our new politics is to have everything as master node except the lit, layeredlit and AxF shader, so material instance is fully specific to lit shader).

sebastienlagarde:
>Should I manually upgrade test scenes? Do we want this PR at all?

Let's discuss about it. This PR is a start for the optimization of lit/layeredLit variant issue (less variant mean also more SRP batcher friendly). I think we need to go further than this (but not now) and will require to develop a strong upgrade material process to not mess with current project. 

Here we talk about upgrading: ScriptableRenderPipelineData, Fontainebleau, VFX demo, FPS sample game, graphic test from HDRP_Test, and have a solid upgrader for external project

sebastienlagarde:
Closing this PR as we can't use it in current state anyway. Too old now.

PR #2401, opened 2018-11-13
Wave intrinsics renaming to match SM 6.0

FrancescoC-unity:
### Purpose of this PR

Changed name of intrinsics to match SM 6.0 and got rid of type specific name functions needed for xbox. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=wave-intrinsics-renaming&automation-tools_branch=add-platform-filter&unity_branch=trunk [Just re-kicked after some changes]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Low

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -246,7 +246,7 @@ void ToggleBit(inout uint data, uint offset)
 }
 #endif
 
-#ifndef INTRINSIC_WaveReadLaneFirst
+#ifndef INTRINSIC_WAVEREADFIRSTLANE
     // Warning: for correctness, the argument's value must be the same across all lanes of the wave.
     TEMPLATE_1_REAL(WaveReadLaneFirst, scalarValue, return scalarValue)
     TEMPLATE_1_INT(WaveReadLaneFirst, scalarValue, return scalarValue)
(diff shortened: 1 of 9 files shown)


com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -182,24 +182,20 @@
 
 // On everything but GCN consoles we error on cross-lane operations
 #ifndef SUPPORTS_WAVE_INTRINSICS
-#define WaveMinInt ERROR_ON_UNSUPPORTED_FUNC(WaveMinInt)
```
EvgeniiG: Do we really need this? I'd prefer to keep them undefined.
That way, you can do #ifdef WaveActiveMin {... use WaveActiveMin ...} #endif.

FrancescoC-unity: Well you can still do something similar with SUPPORTS_WAVE_INTRINSICS. 

#if SUPPORTS_WAVE_INTRINSICS { ... use WaveWhatever ... } #endif 

without having to #if for every case, I mean if one is supported, all of them are, at least for the platform we support :)

EvgeniiG: Good point.

PR #2402, opened 2018-11-13
Remove false dependency between all HDRP materials and shaders

EvgeniiG:
### Purpose of this PR
- Optimize the include structure to avoid creating a false dependency between every single material and shader (e.i. Lit.shader only includes Lit.hlsl, Fabric.shader only includes Fabric.hlsl, etc).

## Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Freorg-material-includes&automation-tools_branch=add-platform-filter&unity_branch=trunk

### Comments to reviewers
Better double check VFX- and ShaderGraph-related changes.


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
```diff
@@ -294,14 +293,25 @@ Shader "HDRenderPipeline/Lit"
...
 
+    // Can't include 'ShaderVariables.hlsl' here because of USE_LEGACY_UNITY_MATRIX_VARIABLES. :-(
+    // Same story for 'Material.hlsl' (above) which includes 'AtmosphericScattering.hlsl' which includes 'ShaderVariables.hlsl'.
+    // #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+    // #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.cs.hlsl"
     #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitProperties.hlsl"
 
+    // TODO:
+    // Currently, Lit.hlsl and LitData.hlsl are included for every pass. Split Lit.hlsl in two:
+    // LitData.hlsl and LitShading.hlsl (merge into the existing LitData.hlsl).
```
sebastienlagarde: >LitData.hlsl and LitShading.hlsl (merge into the existing LitData.hlsl).
I am agree that we could split lit.hlsl in two. however we must not merge it to litData. LitData can be either generated by the shader graph (see the template) or come from the inspector version (current litData.hlsl).  So it is a separate concept.

it is more something like LitSurfaceData (Prepare surface and builtin data), LitBSDFData (prepare BSDF data) and LitShading (do the light loop)

com.unity.visualeffectgraph/Shaders/RenderPipeline/HDRP/VFXLit.cginc
```diff
@@ -8,11 +8,36 @@
...
+
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl"
+
+    #if HDRP_MATERIAL_TYPE_SIMPLE
+        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/SimpleLit.hlsl"
+    #else
+        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl"
+    #endif
+
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl"
```
sebastienlagarde: it shoud be

#if HDRP_MATERIAL_TYPE_SIMPLE
        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/SimpleLit.hlsl"
        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/SimpleLightLoop.hlsl"
    #else
        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl"
        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl"
    #endif

sebastienlagarde:
After checking the PR I think we should do as you mention first, include GBufferMacro file in lit.hlsl (initially we were talking about moving macro in lit.hlsl, but if macro are inside a file and included then it is fine. And so we can remove all the macro include in various file (including the GBuffer one).

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
EvgeniiG: LitSurfaceData.hlsl and LitSurfaceShading.hlsl sounds good.
We already have a SurfaceShading.hlsl file. :-)
Note: not a part of this PR.

com.unity.visualeffectgraph/Shaders/RenderPipeline/HDRP/VFXLit.cginc
EvgeniiG: Will fix. My bad.

PR #2411, opened 2018-11-14
Wait on start fence only if it has been pushed on <19.1 as well

FrancescoC-unity:
### Purpose of this PR

Whoopsies :) I added the "push start fence" feature at a second time, and I forgot to add it to the <19.1 case as well. 

I am fairly certain this is the issue, but I will launch a build on 2018.3 after lunch. 
 
---
### Release Notes
Fix Async compute crash on 2018.3 

---
### Testing status
**Katana Tests**:  None will cover this. 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


sebastienlagarde:
Tested in Fontainebleau demo. Work fine

PR #2418, opened 2018-11-14
Core out of experimental

Chman:
### Purpose of this PR
Remove last vestige of Experimental from Core.


com.unity.render-pipelines.core/Editor/CoreEditorDrawers.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using UnityEditor.AnimatedValues;
+using UnityEditor.Rendering;
 using UnityEngine;
 
 namespace UnityEditor.Experimental.Rendering
```
stramit: is this correct?

Chman: Yes. This class is going to be moved in HDRP anyway.

Chman:
Katana is [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=core-out-of-experimental&automation-tools_branch=add-platform-filter&unity_branch=trunk).

PR #2424, opened 2018-11-14
Remove false dependency between all HDRP materials and shaders

sebastienlagarde:
This PR is a new version of this one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2402 but squased in one commit.

### Purpose of this PR
- Optimize the include structure to avoid creating a false dependency between every single material and shader (e.i. Lit.shader only includes Lit.hlsl, Fabric.shader only includes Fabric.hlsl, etc).

## Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP-reorg-material-includes&automation-tools_branch=add-platform-filter&unity_branch=trunk

### Comments to reviewers
Better double check VFX- and ShaderGraph-related changes.


sebastienlagarde:
PR come from another PR already approve. Katana is green with latest master that include VFX graph test. Merging

PR #2425, opened 2018-11-14
Backport: Remove false dependency between all HDRP materials and shaders (#2402)

sebastienlagarde:
Remove false dependency between all HDRP materials and shaders (#2402)


sebastienlagarde:
Katana is green, merging

PR #2426, opened 2018-11-14
Fix for [case 1098749] and addition of stereo macros to templates

parashar-unity:
### Purpose of this PR
Main fix is for issue #1098749.
Lesser inclusions are for the stereo macros in the templates.  Chances are that there are no meaningful use cases for adding the stereo macros in the fragment shaders (and there currently aren't any in our test cases), but it seemed appropriate to add given there is a corresponding macro in the vertex shaders already present in these same files.

---
### Release Notes

---
### Testing status
**Katana Tests**: SRP All tests run #1337 (very convenient number)
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-All%20Tests/builds/1337?ScriptableRenderLoop_branch=stereoFix&unity_branch=trunk&automation-tools_branch=master
Currently green on editor tests, playmode tests hitting assertions on Mac, all tests pass on Windows, but hits a post-test failure.  Retesting in progress.
**Manual Tests**: Verified that case #1098749 is fixed.

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: Low

---
### Comments to reviewers
Full list of changesets for the branch include some changes to the blit shaders because I was testing the effect locally, but the following changeset removed those changes so you shouldn't see their effect in the PR.


com.unity.render-pipelines.high-definition/Runtime/Sky/HDRISky/HDRISky.shader
```diff
@@ -27,17 +27,21 @@ Shader "Hidden/HDRenderPipeline/Sky/HDRISky"
...
         Varyings output;
+
+        UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(output);
         output.positionCS = GetFullScreenTriangleVertexPosition(input.vertexID, UNITY_RAW_FAR_CLIP_VALUE);
         return output;
     }
 
     float4 Frag(Varyings input) : SV_Target
     {
+        UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
```
sebastienlagarde: Currently HDRP don't support the single pass instancing but only single pass. Not sure we need to add this here at this step (particularly as we have also SkyProcedural.shader and it is not update. Any Thought @nerites ?

parashar-unity: Yeah, I actually tried running through all the samples and tried turning on/off some features, and so far, I don't think there's an actual use case in HDRP yet for which this could be meaningful.  The only reason I added them at all is because there was a corresponding setting of the stereoEyeIndex inside the vertex shader output for these files, so there should, in principle at least, be this one matching it.

But yeah, it's not a particularly valuable change at this time.  I see no problem in practice with just cherry-picking the bug fix.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/Blit.shader
```diff
@@ -27,7 +27,7 @@ Shader "Hidden/HDRenderPipeline/Blit"
 
         Varyings Vert(Attributes input)
         {
-            Varyings output;
```
cdxntchou: No change here?

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRSubShader.cs
```diff
@@ -346,8 +346,8 @@ public class HDPBRSubShader : IPBRSubShader
             },
             RequiredFields = new List<string>()
             {
-//                "FragInputs.worldToTangent",
-//                "FragInputs.positionRWS",
+                "FragInputs.worldToTangent",
```
cdxntchou: May want to put a comment on these guys -- to indicate why they are required (so we don't comment them out again)

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/Blit.shader
parashar-unity: Yeah.  There were changes at one point, but I removed them in a later changeset.  Looks like the whitespace difference was still detected as a change, though.

sebastienlagarde:
Please, merge master, address comment (i.e just add comment) and we are good to merge. Thanks

PR #2443, opened 2018-11-15
Set the tags in an pipeline agnostic way.

russoh:
### Purpose of this PR
Correct output of tags for each render pipeline.

---
### Release Notes

---
### Testing status

Checked lightweight and HD shader generation.

---
### Overall Product Risks
Low
Low
---
### Comments to reviewers


sebastienlagarde:
There is the stacklit master node now to add to the list :) (once your merge master)

russoh:
Apparently the current tags for lightweight look weird, but actually do the right thing, so this isn't urgent to merge, but should probably happen soon anyway, since it's definitely hokey.

sebastienlagarde:
Sorry for the dealy, haven't noticed the PR. Need to update Hair + Decal master :P

com.unity.render-pipelines.lightweight/Editor/ShaderPreprocessor.cs
```diff
@@ -50,7 +50,7 @@ bool StripUnusedShader(ShaderFeatures features, Shader shader, ShaderCompilerDat
             if (shader.name.Contains("Debug"))
                 return true;
 
-            if (shader.name.Contains("HDRenderPipeline"))
+            if (shader.name.Contains("HDRP"))
```
sebastienlagarde: Note: this is intented :)

sebastienlagarde:
Note: I update the PR for all HDRP master node and change the name to match C# convention to k_ShaderTagName

Note that I make an attempt to rename LightweightPipeline to LWRP and HDRenderPipeline HDRP as we have been agree with the doc team to use that everywhere but it broke the preview (No idea why ??)

Merging as this PR have already been approved and need to move forward (cc @phi-lira )

sebastienlagarde:
Note: I have fix the name to be the original one use in this PR for Lightweight here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2685 sorry about that (Friday late night :P)

PR #2460, opened 2018-11-16
HDRP stereo fixes for refraction, SSS and atmospheric scattering

fabien-unity:
### Purpose of this PR
Fix stereo rendering for a few effects (Refraction, SSS, atmospheric scattering)

---
### Release Notes
Tested with SP double wide for now as SPI is not yet ready.

---
### Testing status
**Katana Tests**: green
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp-xr-fixes&unity_branch=trunk

**Manual Tests**: tested with a few different scenes (HDRP tests for SSS, custom scene for volumetric fog and refraction)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers
This is my first PR on GitHub! Please be verbose :)


com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -38,6 +38,11 @@ Shader "Hidden/HDRenderPipeline/OpaqueAtmosphericScattering"
         {
             PositionInputs posInput = GetPositionInput_Stereo(input.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V, unity_StereoEyeIndex);
 
+#if defined(UNITY_SINGLE_PASS_STEREO)
+            // XRTODO: fixup and consolidate stereo code relying on _PixelCoordToViewDirWS
+            V = -normalize(posInput.positionWS);
```
sebastienlagarde: I am not sure to understand what is happening heree, can you explain what this is supppopose to fix ? Looks like a dirty patch hiding the root cause of the issue

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -1595,21 +1595,26 @@ IndirectLighting EvaluateBSDF_ScreenspaceRefraction(LightLoopContext lightLoopCo
...
     float refractionOffsetMultiplier = max(0.0f, 1.0f - preLightData.transparentSSMipLevel * 0.08f);
 
     // If the hit object is in front of the refracting object, we use posInput.positionNDC to sample the color pyramid
     // This is equivalent of setting samplingPositionNDC = posInput.positionNDC when hitLinearDepth <= posInput.linearDepth
     refractionOffsetMultiplier *= (hitLinearDepth > posInput.linearDepth);
 
+    float2 samplingPositionNDC = lerp(posInput.positionNDC, hit.positionNDC, refractionOffsetMultiplier);
+
+#ifdef UNITY_SINGLE_PASS_STEREO
+    samplingPositionNDC.x = 0.5f * (samplingPositionNDC.x + unity_StereoEyeIndex);
```
sebastienlagarde: There is the same code in apply distortion, so if you do that here, you need to do the same for the distortion.
Note: Refraction is for our PBR transparent object (IOR, beer's law etc...), Distortion is the same thing but for artists with just a vector and a smoothness.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -371,11 +371,7 @@ public void ConfigureStereoDependentSettings(Camera cam)
                 // TODO: The work will be implemented piecemeal to support all passes
                 enableMotionVectors = enablePostprocess && !enableMSAA;
                 enableDecals = false;
-                enableDistortion = false;
```
sebastienlagarde: See my comment above, Distortion will not work

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -371,11 +371,7 @@ public void ConfigureStereoDependentSettings(Camera cam)
...
-                enableRoughRefraction = false;
-                enableSSAO = false;
```
sebastienlagarde: Don't think SSAO is working?

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
fabien-unity: The issue comes from _PixelCoordToViewDirWS which is not stereo-friendly at the moment. I applied the same fix used by the sky shader to recompute the view direction. posInput.positionWS respects the stereo matrices.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
fabien-unity: SSAO looks fine when I tested it. Maybe QA can confirm?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
fabien-unity: I've setup a test with distortion and it looks like it's working already. Here's some results from RenderDoc
- the distortion buffer generation is ok for both eyes
- in RenderDistortion(...), a compute shader is dispatched on the entire image (double wide) in my case. This will cause issues because one eye could sample texels from the other eye but this will soon go away as we switch to SP-instancing.

image

image

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
fabien-unity: image

fabien-unity:
@mikesfbay good point, I just tested and it works fine.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: perfect :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
sebastienlagarde: good

sebastienlagarde:
Katana is green, Fabien have veryify the effect, merging.

sebastienlagarde:
ah forget to add that you need to update the changelog.md file with the change. I will do it for this PR in HDRP/Staging, next PR think to update it.

PR #2461, opened 2018-11-16
Basic single-pass instanced stereo support for HDRP

nerites:
### Purpose of this PR
Enable single-pass instanced stereo rendering with the following constraints:
- No MSAA
- No Vulkan
- Forward only
- No FPTL
- No SSAO

When VR is enabled, HDRP will now allocate textures using the RenderTextureDescriptor provided by XRSettings.
HDRP's compute shaders have been modified to dispatch per-eye for SPI and eliminate texture dimensionality errors.

---
### Release Notes
Dependent on the following trunk PR: https://ono.unity3d.com/unity/unity/pull-request/77058/_/graphics/setrendertarget/mrtwithsubtarget/clean

The above PR was merged into trunk in commit 8c2cce111344. 

Added:
- StereoInstancingTextureDefines.hlsl added to CoreRP to define SPI-compatible texture declarations and operations.

Remaining issues:
- lightlistbuild-clustered.compute disconnects from compute shader compiler for Vulkan with SPI+Texture2DArray (bug here: https://fogbugz.unity3d.com/f/cases/1103046/)
- HDRP's call to PP's MSVO component results in unexpected texture dimension errors. Need to either modify PP's MSVO compute shaders to support SPI, or do additional blits before and after SSAO to/from Tex2D. @mikesfbay what do you recommend? -> Plan is to fix MSVO for SPI by modifying shaders and dispatching multipass in a separate PR.
- MSAA is broken because the depth/color resolve passes are done with DrawProcedural calls to MRT, which for some reason doesn't populate SV_InstanceID. Further, enabling MSAA with SPI will result in a flood of the following error: "Multisampled textures are only supported with the 2D texture type. 2D array texture type is not supported either." Will resolve in a later PR, hopefully by referencing LWRP SPI.

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=allocfromdefault&unity_branch=graphics%2Fsetrendertarget%2Fmrtwithsubtarget%2Fclean&automation-tools_branch=add-platform-filter

Visually confirmed basic SPI functionality with HDRP:
image

Nothing

---
### Overall Product Risks
**Technical Risk**: 
Medium: Impacts allocation of all textures, adds kernels to many compute shaders

**Halo Effect**: 
High: All texture allocations, added a new include file to ShaderVariables.hlsl, modified most shaders to add SPI macros, modified API-specific .hlsl files to add texture array macros. 

---
### Comments to reviewers
This is a huge changeset so I figured it'd be good to get eyes on it sooner. Will circle back to add MSAA support for SPI in a different PR. 

@brandon-unity3d When skimming the LWRP code I noticed that they didn't seem to do anything special in declaring their textures but I assume I'm missing something. How are we binding tex2darrays in LWRP's shader code when SPI is set?


nerites:
Hi everyone! I finally got around to getting Katana green again. This is a pretty huge PR so feel free to skewer it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MipGenerator.cs
```diff
@@ -72,15 +81,19 @@ public void RenderMinDepthPyramid(CommandBuffer cmd, RenderTexture texture, HDUt
                 cmd.SetComputeIntParams(   cs,         HDShaderIDs._DstOffset,         m_DstOffset);
                 cmd.SetComputeTextureParam(cs, kernel, HDShaderIDs._DepthMipChain,     texture);
 
-                cmd.DispatchCompute(cs, kernel, HDUtils.DivRoundUp(dstSize.x, 8), HDUtils.DivRoundUp(dstSize.y, 8), 1);
+                for (int eye = 0; eye < numSlices; eye++)
+                {
```
mikesfbay: Change numSlices to numEyes since you are iterating over each eye.  NumEyes is more implementation independent. We may later implement some techniques in double-wide as multi-passes over each eye too in which case "numSlices" wouldn't make sense. XRGraphics.NumSlices needs to be changed to NumEyes too.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DBufferManager.cs
```diff
@@ -56,23 +56,23 @@ public void ClearAndSetTargets(CommandBuffer cmd, HDCamera camera, bool rtCount4
             Color clearColor = new Color(0.0f, 0.0f, 0.0f, 1.0f);
             Color clearColorNormal = new Color(0.5f, 0.5f, 0.5f, 1.0f); // for normals 0.5 is neutral
             Color clearColorAOSBlend = new Color(1.0f, 1.0f, 1.0f, 1.0f);
-            HDUtils.SetRenderTarget(cmd, camera, m_RTs[0], ClearFlag.Color, clearColor);
-            HDUtils.SetRenderTarget(cmd, camera, m_RTs[1], ClearFlag.Color, clearColorNormal);
-            HDUtils.SetRenderTarget(cmd, camera, m_RTs[2], ClearFlag.Color, clearColor);
+            HDUtils.SetRenderTarget(cmd, camera, m_RTs[0], ClearFlag.Color, clearColor, depthSlice: XRGraphics.DepthSlice);
+            HDUtils.SetRenderTarget(cmd, camera, m_RTs[1], ClearFlag.Color, clearColorNormal, depthSlice: XRGraphics.DepthSlice);
+            HDUtils.SetRenderTarget(cmd, camera, m_RTs[2], ClearFlag.Color, clearColor, depthSlice: XRGraphics.DepthSlice);
```
mikesfbay: There's no need to pass global depthSlice to all SetRenderTargets if you can intercept it in key SetRenderTarget functions. There are currently 26 such SetRenderTarget changes. By comparison, LWRP SPI only had 2 or 3 key places where depthSlice of -1 is set for texture arrays.

Also this was more clear in LWRP:  
if (isStereoEnabled && XRGraphics.eyeTextureDesc.dimension == TextureDimension.Tex2DArray)
    Set render target with -1 as depth slice;

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MipGenerator.cs
nerites: Oops, I didn't see that clicking "resolve conversation" doesn't post my comment.

This has been fixed in the latest commit.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DBufferManager.cs
nerites: In the latest commit, I modified HDUtils and CoreUtils SetRenderTarget methods to calculate DepthSlice depending on camera data (if available) and XRSettings. Since these changes are to code that is widely used, I'm kicking off Katana again and I'll check SPI functionality tomorrow.

com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
```diff
@@ -169,8 +169,18 @@ public static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier[] c
         }
 
         public static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier[] colorBuffers, RenderTargetIdentifier depthBuffer, ClearFlag clearFlag, Color clearColor)
+        {   // TODO VR: Modify/overload this if there is ever a situation where:
+            // - XRGraphics.UsingTexArray is true, but
+            // - We want to render to non-arrayed MRT
+            if (XRGraphics.UsingTexArray)
+                cmd.SetRenderTarget(colorBuffers, depthBuffer, 0, CubemapFace.Unknown, XRGraphics.DepthSlice);
+            else
```
mikesfbay: Remove global XRGraphics.DepthSlice and pass -1 as depth slice since you are already testing for texArray in the if statement. Use the LWRP way which is more clear:

if (isStereoEnabled && XRGraphics.UsingTexArray)
Set render target with -1 as depth slice;

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
```diff
@@ -420,6 +435,365 @@ void Resize(int width, int height, MSAASamples msaaSamples, bool sizeChanged, bo
...
+        }
+        public RTHandle AllocFromDefault(
+            Vector2 scaleFactor,
+            FilterMode filterMode,
+            RenderTextureFormat colorFormat,
+            bool sRGB,
+            string name)
+        {
+            RenderTextureDescriptor overrideTextureDescriptor = s_DefaultDescriptor;
+            overrideTextureDescriptor.sRGB = sRGB;
```
mikesfbay: In RTHandlesystem.cs, It's not necessary to create various versions of AllocFromDefault when all we need from eyeTextureDesc are the width/height and dimension.  Instead of changing all Alloc calls to AllocFromDefault (65 changes) and adding a bunch of duplicated code in AllocFromDefault overloads, we should try to modify Alloc to account for eyeTextureDesc.

nerites: The reason all those overloads are needed are so that the eyeTextureDesc's fields don't get overwritten by default values when they aren't specified. 

When I talked to Brandon, he said that we need things in the eyeTextureDesc that aren't necessarily visible from C#- in particular, a yflip flag that is not exposed outside of native code. That's why AllocFromDefault copies from eyeTextureDesc and avoids creating a descriptor from scratch. 

In my most recent commit I refactored AllocFromDefault to try to make it less error-prone, since there were some bugs in it that caused assertions to fire on Playmode startup.

com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
nerites: Done in commit 7ae0885!

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
nerites: You know what, I'll check with Brandon to see if all the fields the Alloc() methods are overwriting are ok in terms of the things he was thinking of. If so then we're okay to clobber everything C# has access to, then I'll just add some if (XRGraphics.enabled) logic for height/width/depth/vrUsage.

mikesfbay: Allocating render textures for XR is already a solved problem. I would see how it is handled in PostProcessRenderContext.cs and search for GetDescriptor() and m_sourceDescriptor. In your case, I would make proper changes in each Alloc function to call GetDescriptor(). This can safely remove all AllocFromDefault and undo all Alloc api call changes.

nerites: Got it working on latest commit! Rerunning SRP tests.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -334,7 +349,7 @@ public static void BlitCameraTexture(CommandBuffer cmd, HDCamera camera, RenderT
             RenderTargetIdentifier colorBuffer,
             MaterialPropertyBlock properties = null, int shaderPassId = 0)
         {
-            CoreUtils.SetRenderTarget(commandBuffer, colorBuffer);
+            CoreUtils.SetRenderTarget(commandBuffer, colorBuffer, depthSlice: XRGraphics.DepthSlice);
```
mikesfbay: This should be changed to explicit if-then test rather than doing logic in XRGraphics.DepthSlice accessor. In fact, the global XRGraphics.DepthSlice should be removed for clarity. 

if (stereoEnabled && UsingTexArray)
  CoreUtils.SetRenderTarget(commandBuffer, colorBuffer, -1);
else 
   CoreUtils.SetRenderTarget(commandBuffer, colorBuffer);

mikesfbay:
Should fix this error when running with SPI and FPTL is enabled, "Compute shader (lightlistbuild): Property (g_depth_tex) at kernel index (1) has mismatching texture dimension (expected 2, got 5)"

com.unity.render-pipelines.core/Runtime/Common/XRGraphics.cs
```diff
@@ -159,6 +159,40 @@ public static int eyeTextureHeight
...
+            {
+                if (!enabled)
+                    return 0;
+                if (UsingTexArray)
+                    return -1;
+                else
+                    return 0;
+            }
+        }
+
```
mikesfbay: Global DepthSlice and NumSlices can be removed as discussed.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
nerites: Done!

com.unity.render-pipelines.core/Runtime/Common/XRGraphics.cs
nerites: Done!

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandles.cs
```diff
@@ -29,6 +30,7 @@ public static class RTHandles
             bool bindTextureMS = false,
             bool useDynamicScale = false,
             VRTextureUsage vrUsage = VRTextureUsage.None,
+            bool renderStereoIfVRActive = false,
             RenderTextureMemoryless memoryless = RenderTextureMemoryless.None,
```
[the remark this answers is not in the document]
nerites: Done!

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.shader
```diff
@@ -28,7 +28,7 @@ Shader "HDRenderPipeline/Decal"
 
     #pragma target 4.5
     #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
-    //#pragma enable_d3d11_debug_symbols
+    #pragma enable_d3d11_debug_symbols
```
mikesfbay: Undo debug mode for shader

mikesfbay:
Also be sure to test in Standalone-player mode on HMD (Oculus/Vive) using HD template scene. QA will need to pass on those.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/Decal.shader
nerites: Done!

nerites:
> Should fix this error when running with SPI and FPTL is enabled, "Compute shader (lightlistbuild): Property (g_depth_tex) at kernel index (1) has mismatching texture dimension (expected 2, got 5)"

Done in recent commit!

nerites:
When built (win64, dx11, developer mode), the app fails to render any objects with LOD applied: in the template scene, this includes the Construction Light prefab and the Workbench prefab. However, this seems to also occur on my branch with single-pass double-wide. I need to test and see if this happens on master as well.

sebastienlagarde:
Hi, sorry, haven't get the time to review this PR yet. 
I see that the C++ part finally reach trunk, so please @JulienIgnace-Unity, @anisunity , @EvgeniiG  have a look as it touch a lot of file.

Also a warning, this PR here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2562 is the Post Process V3 that fully replace PPV2 and that will be merge before this one. You will need to update this PR accordingly to be compatible. From my understanding there will be conflict mainly on HDRenderPipeline on the flip of the screen.
(poking @fabien-unity as well).

nerites:
Thanks for the heads-up, @sebastienlagarde! Is there... any chance at all of getting this PR in before the PPv3 PR? I talked with @chman and it sounds like PPv3 doesn't have any support for SPI, so it would be good to get this PR in with a version of HDRP that is compatible with PPv2. But maybe that doesn't make sense at all given that we're targeting PPv3 for 2019.1, and the API that this PR is dependent on only comes out in 2019.1...

Anyways, yesterday I confirmed that the LOD objects not being rendered also occurs on master with SPDW... So the issue doesn't seem to be specific to this branch. I've filed a bug here: https://fogbugz.unity3d.com/f/cases/1106682/ 

I also launched @fabien-unity's autotester, and it seems like relative to SPDW, SPI is missing: 
- Decals
- Specular occlusion
- Distortion
- Emission
- Subsurface scattering
Most of which are issues that @fabien-unity fixed for SPDW after I entered the SPI hyperbolic time chamber, but some of which (emission?) might hopefully be fixed once I adapt, for example, Deferred.compute to have... double the kernels... to support SPI.

sebastienlagarde:
> I talked with @Chman and it sounds like PPv3 doesn't have any support for SPI,
This is indeed the case. But we don't plan any  PPV2 for 2019.1 we fully replace it with PPV3 and we need to do it quickly as various part of the lighting depends on it (pre-exposure etc...) + camera (multiviewport and dynamic resolution). Note: the next 2 weeks HDRP can be a bit unstable given the large change we push in it (We have push new reflection system, we need to update the main RenderLoop (due to recursive call of Render() for reflection), we update to PPV3 which mean a full transition to RT handle system. Expect problem...

com.unity.render-pipelines.core/ShaderLibrary/StereoInstancingTextureDefines.hlsl
```diff
@@ -0,0 +1,59 @@
...
+CBUFFER_END
+#define unity_StereoEyeIndex _ComputeEyeIndex
+#endif
+#endif
+
+//////////////////////////////////////////////////////
+// Texture declarations and samplers for instancing
+//
+
+#if defined(UNITY_STEREO_INSTANCING_ENABLED) || defined(UNITY_STEREO_MULTIVIEW_ENABLED)
```
EvgeniiG: I -really- don't like adding more and more nested macros, especially with such ambiguous names.
If we sometimes need an array texture, I recommend that we always declare an array texture (and not invent new types), and vary the array size from C#.
This somewhat increases the pressure on the texture address generation unit, but it is rarely the bottleneck.
When it is (which we can determine after a profiling session), we can optimize that particular spot by not using an array texture, and in that case the increased code complexity will be worth it.

nerites: This works for me, and addresses a recurring issue I've seen where the scene view attempts to use a texture array as a shader resource but errors due to expecting a Tex2D and finding a Tex2DArray bound instead. It should also reduce the number of shader variants and compute kernels, which is a win.

The only thing is that now, even for projects that don't use XR, we will always be using tex2darrays of volumedepth 1. If you all are fine with that, I can begin making the changes today.

nerites: @Chman @mikesfbay if we go this route, PPv3 should be simplified too- it would need to be updated to expect Tex2DArray everywhere, and we still need to DispatchCompute per eye, but all compute shaders can be the same as for monoscopic.

sebastienlagarde: The very first thing to do before deciding anything is to evaluate the cost of this.
What I would like is that we do simple measurement on base PS4 / XBone where we replace texture by texturearray with index0 and measure the impact globally on a frame. (Note: memory or size is not a concern, the bandwidth is the same, what I want to know if the overhead of texArray fetch against tex). @fabien-unity any though about cost ? What was the solution on Frosbite ? 
Do you have access to any devkit to do some measurement ? othersise we will do it.

fabien-unity: We should measure perf and check the shader disassembly to confirm. Frostbite had no stereo-instancing support when I was there so I cannot compare.
I don't have console devkits at the moment though.

nerites: Is there any way to test at a smaller scale? If not, I've already started converting all textures in HDRP to tex2darray in another branch, so we can do a comparison once it's ready.

sebastienlagarde: @nerites perfect, once the branch is ready we can perform test here.
I think we need to to the test at full scale, otherwise it doesn't make sense.

EvgeniiG: For post-processing, can't we NOT allocate stereo (array) textures at all?
This would allow us to avoid potential issues with array texture perf, and save half the RT memory.
To be clear, we would write a "for (int i = 0; i < numEyes; i++)" loop in C#, and only allocate buffers (pyramids, etc) for one eye at a time. The reason for using arrays is stereo instancing, so once we are done rendering all the objects, we don't have to use arrays anymore (except for the final copy for present).

nerites: For PPv2, @mikesfbay repurposed the blit into PP to blit from stereo to a single texture, which circumvents needing to do anything stereo specific, aside from MSVO which HDRP invokes in the middle of rendering rather than as part of the PP stack.

For PPv3, we will either have to do per-eye dispatches for each effect in the stack (as you've described), OR add a blit from texarray to tex2d. The latter is not as attractive for PPv3 because there is not normally a blit before PPv3.

nerites: Ok @sebastienlagarde @EvgeniiG the texarray branch is ready: https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/hdvr_spi_texarrays

If you want to test with PostProcessing, you'll need the following branch: https://github.com/Unity-Technologies/PostProcessing/tree/v2_hdtexary

If you don't want to test with PostProcessing, just comment out line 2401 of HDRenderPipeline.cs and it should compile/enter playmode fine with unmodified PPv2.

mikesfbay: For texture arrays, we should allocate based on whether a render texture needs single-pass instancing rather than making all render textures texture array. In fact, with implementation of XR Interface for postprocessing v2, none of postprocessing effects use texture arrays in single-pass instancing, and most effects remain unchanged using tex2D. For V3, we'll try to maintain XR interface idea so to avoid XR changes to post effects.  See XR Interface design doc for the motiviation:
https://docs.google.com/document/d/1hANbhKCRIJs6ww7XoAIXbX3ArdAs7OBOTfZL1MqgtPI

As for the texture macros, they are similar to unity engine's HLSLSupport.cginc for stereo instancing so it's a proven idea. XR has been using single-pass instancing and double-wide texture macros since 2017.2. They are well documented for developers to write once and work across different XR platforms (PC, console, mobile, and new HMD's) and avoid plastering #ifdef all over their shader code.  We should be consistent with the macro usage so HD will continue to work when more platforms are added.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -632,17 +632,17 @@ uint DecodeFromGBuffer(uint2 positionSS, uint tileFeatureFlags, out BSDFData bsd
     // Isolate material features.
     tileFeatureFlags &= MATERIAL_FEATURE_MASK_FLAGS;
 
-    GBufferType0 inGBuffer0 = LOAD_TEXTURE2D(_GBufferTexture0, positionSS);
-    GBufferType1 inGBuffer1 = LOAD_TEXTURE2D(_GBufferTexture1, positionSS);
-    GBufferType2 inGBuffer2 = LOAD_TEXTURE2D(_GBufferTexture2, positionSS);
+    GBufferType0 inGBuffer0 = LOAD_TEXTURE(_GBufferTexture0, positionSS);
+    GBufferType1 inGBuffer1 = LOAD_TEXTURE(_GBufferTexture1, positionSS);
+    GBufferType2 inGBuffer2 = LOAD_TEXTURE(_GBufferTexture2, positionSS);
```
mikesfbay: Need to undo these GBuffer changes and fix the Katana deferred failures (since this PR is forward only).

nerites: The changes to the deferred shaders protect against build failures when SPI is enabled. In order for HDRP to compile with SPI enabled, either these macros need to be in or we need to go the full blit-depth-to-double-wide method that we discussed. I'm putting together another branch with that implementation now but I will preserve this branch as a building and katana-passing to both the Texarray branch and the BlitDepth branch in the meantime.

mikesfbay: The SPI changes are forward only and should not change deferred path. This is trying to mask a underlying katana test failure in the deferred path. We should find the cause of katana failure rather than put in a hack to mask away the problem. The katana image failures sound like a SPI change 'leaked' into the deferred path, which would be a bug in the PR that need to be fixed.

nerites: Removed GBuffer changes and replaced depth load with a dummy assignment for deferred in latest commits.

com.unity.render-pipelines.core/ShaderLibrary/StereoInstancingTextureDefines.hlsl
sebastienlagarde: Thanks for the update @nerites . We will check the perf. Sorry for the delay

nerites: Thanks @sebastienlagarde! What's the process for checking perf? Mike came up with an alternate design for SPI where we multipass all render passes that do not use scene geometry for rendering. It's been a lot trickier to implement than the texarray idea, but eventually we'll probably need to evaluate that for perf as well.

mikesfbay: The "XR Interface/multipass" way (see link above) is less intrusive to HDRP than forcing XR changes all over HDRP code. This also avoids changes to current and future compute shaders. The single-pass instancing really only improves CPU traversal time and doesn't help with any compute or GPU bound tasks. For most of compute shaders (such as ones with depth as input, such as ssao), we can iterate over each eye to compute per eye result (see doc for why performance is same or better).

There another advantage of this "multipass"/XR interface way: Multi-gpu's. With HD on multi-gpu's, each eye pass can be dispatched to separate gpu's to compute results of each eye. We can't do this if everything is done with instancing on single gpu.

EvgeniiG: I like this idea: "multipass all render passes that do not use scene geometry for rendering". A bit more typing, but much more clear what's going on, and with potential render target memory savings. :-)

nerites: Okay, the first draft of the hybrid MP/SPI implementation is ready here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/hdrp-master...hdvr_mpspi

I still need to clean up some errors (particularly color/depth buffer dimension mismatches when dropping out of playmode) but at least we can see what it looks like relative to the others.

EvgeniiG: > Yes, my Editor PR has made it into latest trunk already- what kind of errors are you seeing?

image

I will fix these locally.

EvgeniiG: Okay, looks like it was me who was too far behind. :-) Sorry about that.

EvgeniiG: After updating to the latest trunk from today, these issues have been fixed. Here are the new ones:
image

nerites: - Deferred is not supported for VR- please switch everything to Forward Only
- I think we can remove SV_InstanceID from v2f structs, we just need to have UNITY_VERTEX_OUTPUT_STEREO. 

I left many non-breaking errors in the branch because it was supposed to be just a proof of concept- if arraying all textures tanked the performance too hard, I didn't want to have already spent all that time refining it.

EvgeniiG: I need to test whether these changes make our deferred path slower. I am not testing it in VR. We have a demo which shipped on the PS4, so I am testing that for regressions.
We can't remove SV_InstanceID because this would break instanced draw calls.
I will try to patch the code on my end. It's of course important that everything works correctly, otherwise the comparison is not fair.

com.unity.render-pipelines.core/ShaderLibrary/API/PSSL.hlsl
```diff
@@ -77,6 +77,12 @@ bool WaveActiveAnyTrue(bool expression)
 #define TEXTURECUBE_ARRAY(textureName)        TextureCubeArray textureName
 #define TEXTURE3D(textureName)                Texture3D textureName
 
+// Only RW_ versions of MS textures are available in PSSL (https://ps4.siedev.net/resources/documents/SDK/6.000/PSSL-Specification/0016.html#__document_toc_00000060)
+#define TEXTURE2DMS(textureName)              RW_MS_Texture2D textureName
+#define TEXTURE2DMS_ARRAY(textureName)        RW_MS_Texture2D_Array textureName
+#define TEXTURE2DMS_TYPE(textureName)              RW_MS_Texture2D<type> textureName
+#define TEXTURE2DMS_ARRAY_TYPE(textureName)        RW_MS_Texture2D_Array<type> textureName
+
```
EvgeniiG: That's incorrect.
Both MS_Texture2D and MS_Texture2D_Array exist.

com.unity.render-pipelines.core/ShaderLibrary/StereoInstancingTextureDefines.hlsl
nerites: Oh okay- sorry for the delayed responses, I've been working on this other branch in the meantime, put together according to @mikesfbay's description earlier in the thread: https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/hdrp-master...hdvr_mpspi

I'll try and clean up the texarray branch a bit also.

EvgeniiG: OK, after spending several days fixing compiler errors, I was able to launch the Fountainbleu demo on the PS4 with the tex array branch. I was not able to test post-processing, as it has broken visuals with the tex array branch. Most of the time, there is zero performance difference. Only some passes which mostly perform texture fetches with little ALU (e.g. depth pyramid) see a small regression (around 5%).
I would say that using arrays is fine, as long as we use them where they are needed. It may be worth exploring multi-pass rendering for passes that happen after lighting, though (e.g. post-processing), depending on whether we can save memory or clean up the code.

nerites: Thanks for running those tests, Evgenii! I'm pretty sure PP is broken because it's still expecting tex2Ds and getting tex2DArrays. Sorry the errors took so long to clean up- could you push those changes back to the branch so that I can see about replicating them on the hdvr_mpspi branch?

I think we're trending towards using the hybrid multipass/SPI solution. The performance isn't great for VR, since right now we're doing some blits (and allocating additional textures) to translate between tex2darrays and tex2d, but theoretically we can optimize that away at some later point? 

I am still fond of the idea of using texarrays for everything... it just seems a lot more streamlined, especially if it doesn't significantly effect monoscopic rendering performance. I just don't have the time to polish up more than one of these implementations, especially if we're trying to get PPv3 involved...

mikesfbay: On VR performance, the hybrid multipass/SPI method is doing SPI on depth prepass, forward opaque and any geometry passes. So, we are maxing out on SPI to avoid CPU scene traversal times.  The depth array blit to 2D depth textures are still faster than doing multipass on depth prepass, so it's still a win.  We can try to optimize out the blits later too.

The advantage of hybrid multipass is that most of compute shaders remain unchanged and will just work.  MSVO compute, for example, takes in 2D depth textures and works as before.  Another advantage is multi-GPU is easier since we can dispatch each eye compute separately to each GPU.

com.unity.render-pipelines.core/ShaderLibrary/API/PSSL.hlsl
nerites: Thanks a lot! I don't think I was able to locate the exact info I needed in the playstation docs.

sebastienlagarde:
closing as it have been handled in another PR

PR #2471, opened 2018-11-19
Add more shader stripping opportunities 

FrancescoC-unity:
### Purpose of this PR
Add more options and opportunities to allow for shader variant stripping. 
Added a bit more stripping with already available options.
Added a variant stripping log. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=variant-stripping-pass&automation-tools_branch=add-platform-filter&unity_branch=trunk [Just kicked]

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**:  Medium

---
### Comments to reviewers
I tested this on few sample scenes, with both shadergraph and non materials, but if you could pay extra attention to the change it'd be great, it felt like a slippery area of code :-)


com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
```diff
@@ -15,9 +15,25 @@ public class LitShaderPreprocessor : BaseShaderPreprocessor
...
+            bool isDistortionPass = snippet.passName == "DistortionVectors";
+            bool isTransparentForwardPass = isTransparentPostpass || isTransparentBackface || isTransparentPrepass;
+
+            if (isDistortionPass && !hdrpAsset.renderPipelineSettings.supportDistortion)
+                return true;
+
+            if (isTransparentBackface && !hdrpAsset.renderPipelineSettings.supportTransparentBackface)
+                return true;
+
+            if (isTransparentPrepass && !hdrpAsset.renderPipelineSettings.supportTransparentDepthPostpass)
```
sebastienlagarde: There is a typo here: supportTransparentDepthPostpass

com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
```diff
@@ -28,33 +44,39 @@ public override bool ShadersStripper(HDRenderPipelineAsset hdrpAsset, Shader sha
...
                 if (isTransparentForwardPass)
                     return true;
 
-                // TODO: This check is disabled currently as it doesn't work. We have issue with lit VFX from VFX graph not working correctly, mean we are too agressive on
-                // removal. Need to check why.
-                // When we are in deferred (i.e !hdrpAsset.renderPipelineSettings.supportOnlyForward), we only support tile lighting
-                //if (!hdrpAsset.renderPipelineSettings.supportOnlyForward && inputData.shaderKeywordSet.IsEnabled(m_ClusterLighting))
-                //    return true;
+                // If Lit uses deferred only
+                if(shader.name.Contains("HDRenderPipeline/Lit") && hdrpAsset.renderPipelineSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.DeferredOnly)
```
sebastienlagarde: did it miss the test when you are forwardOnly? and then can discard GBuffer pass

FrancescoC-unity: Uhm, I have it in my notes, I must have deleted by mistake. Will absolutely put it back  in.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
```diff
@@ -28,33 +44,48 @@ public override bool ShadersStripper(HDRenderPipelineAsset hdrpAsset, Shader sha
...
                     return true;
 
-                // TODO: This check is disabled currently as it doesn't work. We have issue with lit VFX from VFX graph not working correctly, mean we are too agressive on
-                // removal. Need to check why.
-                // When we are in deferred (i.e !hdrpAsset.renderPipelineSettings.supportOnlyForward), we only support tile lighting
-                //if (!hdrpAsset.renderPipelineSettings.supportOnlyForward && inputData.shaderKeywordSet.IsEnabled(m_ClusterLighting))
-                //    return true;
+                if(shader.name.Contains("HDRenderPipeline/Lit"))
+                {
+                    if (hdrpAsset.renderPipelineSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.DeferredOnly)
```
sebastienlagarde: here we should be able to discard WRITE_NORMAL_BUFFER if we are deferredOnly

FrancescoC-unity: This one I did not have, you are right, didn't think of it!

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
```diff
@@ -56,6 +59,8 @@ public BaseShaderPreprocessor()
             m_DirectionalMedium = new ShaderKeyword("DIRECTIONAL_SHADOW_MEDIUM");
             m_DirectionalHigh = new ShaderKeyword("DIRECTIONAL_SHADOW_HIGH");
             m_WriteNormalBuffer = new ShaderKeyword("WRITE_NORMAL_BUFFER");
+            m_WriteMSAADepth = new ShaderKeyword("WRITE_MSAA_DEPTH");
+            m_SubsurfaceScattering = new ShaderKeyword("OUTPUT_SPLIT_LIGHTING");
```
sebastienlagarde: This one is almost useless
The only case where we have a multi compile for this (i.e  #pragma multi_compile _ OUTPUT_SPLIT_LIGHTING), is for deferred.shader which is just a "debug" shader (in the future we may remove it). It doesn't eat any time. Other definition are always a "define"

You can let it but add a comment to say it is only for deferred shader and have low impact

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
```diff
@@ -40,6 +42,7 @@ public BaseShaderPreprocessor()
             // FOG_LINEAR, FOG_EXP, FOG_EXP2
             // STEREO_INSTANCING_ON, STEREO_MULTIVIEW_ON, STEREO_CUBEMAP_RENDER_ON, UNITY_SINGLE_PASS_STEREO
             // INSTANCING_ON
+            m_ShadowMask = new ShaderKeyword("SHADOWS_SHADOWMASK");
```
sebastienlagarde: Not needed. Maybe add a comment here to say that that SHADOWS_SHADOWMASK is done buy builtin Unity and not need to be handled here (unless the log say the opposite? :) )

FrancescoC-unity: Yup, it is useless, I added the initialization as it was defined but not initialized and then proceeded to test whether the log showed otherwise re the built in (and it didn't). 

I'll remove the initialization *and* the definition.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
```diff
@@ -28,33 +44,48 @@ public override bool ShadersStripper(HDRenderPipelineAsset hdrpAsset, Shader sha
...
+
+                        // If we use deferred only, MSAA is not supported.
+                        if (inputData.shaderKeywordSet.IsEnabled(m_WriteMSAADepth))
+                            return true;
+
+                        if (isForwardPass && !inputData.shaderKeywordSet.IsEnabled(m_DebugDisplay))
+                            return true;
+                    }
+
+                    if (hdrpAsset.renderPipelineSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.ForwardOnly)
```
sebastienlagarde: Currently I noticed that this code is already declare above:)

            // When using forward only, we never need GBuffer pass (only Forward)
            if (hdrpAsset.renderPipelineSettings.supportedLitShaderMode == RenderPipelineSettings.SupportedLitShaderMode.ForwardOnly && isGBufferPass)
                 return true;

so it is redundant

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
FrancescoC-unity: I just went through all the multi-compile just as a safeguard in this case really. But good point on the comment, I'll mention it.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
sebastienlagarde: Note: Only Lit shader (and oinherit) can have a Gbuffer pass, so it should be fine to define it outside and this will apply to shader graph too.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/LitShaderPreprocessor.cs
```diff
@@ -28,33 +44,48 @@ public override bool ShadersStripper(HDRenderPipelineAsset hdrpAsset, Shader sha
...
                 // If opaque, we never need transparent specific passes (even in forward only mode)
                 if (isTransparentForwardPass)
                     return true;
 
-                // TODO: This check is disabled currently as it doesn't work. We have issue with lit VFX from VFX graph not working correctly, mean we are too agressive on
-                // removal. Need to check why.
-                // When we are in deferred (i.e !hdrpAsset.renderPipelineSettings.supportOnlyForward), we only support tile lighting
-                //if (!hdrpAsset.renderPipelineSettings.supportOnlyForward && inputData.shaderKeywordSet.IsEnabled(m_ClusterLighting))
-                //    return true;
+                if(shader.name.Contains("HDRenderPipeline/Lit"))
```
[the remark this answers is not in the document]
sebastienlagarde: In the future we will need to find a way to apply this to all shader graph inherit from litmasternode but currently we can't.

sebastienlagarde:
Katana is green except on a LW test but it was from master becaues this PR is on an old master. Merging

sebastienlagarde:
Katana is green except an instability in shader graph on mathnode not related to this PR

PR #2514, opened 2018-11-27
Switch support

kaychang-unity:
### Purpose of this PR
This PR optimizes ColorPyramid pass for Switch, and possibly other platforms (to be tested).
On Switch, the compute shader version of the pass runs at 5.0ms on a 720p render-target RGBA16F (undocked mode). The new fragment shader version runs at 1.7ms.

The compute shader is also currently broken on Switch, due to the generated GLSL code expecting a RGBA32F texture to be sampled, but given an RGBA16F (incorrect sampling from RWImage2D).

---
### Release Notes

---
### Testing status
**Katana Tests**: Katana is likely not setup to run Switch

**Manual Tests**:
Comparing the reference result from the (locally fixed) compute shader with the new fragment shader implementation, the blurred mipmap chain look the same.

**Automated Tests**: No automated test was run.

---
### Overall Product Risks
**Technical Risk**: Low
Other platforms that Switch are not touched. Previous implementation is broken on Switch too.

**Halo Effect**: Low

---
### Comments to reviewers
This optimization may benefit other platforms too, I only tested Geforce GTX 1050 on my laptop (it was a 33% win on Vulkan platform).
More saving are possible and will be committed later:
- render-targets should be R11G11B10, this would move the cost down to 1.35ms.
- copy of the base mipmap should be skipped (cost 0.6ms out of the 1.7ms)
- it may possible to reorganize the sub-passes (downsample, horizonal blur, vertical blur) and remove most of the GPU barriers, possibly saving a further  an additional ~0.2ms. This will however remove accumulation of blur though the mipmap chain, so further discussion is needed.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [5.3.0-preview] - 2018-11-xx
 
+### Changed
+- ColorPyramid compute shader passes is swapped to pixel shader passes on platforms where the later is faster (Nintendo Switch).
+
 ## [5.2.0-preview] - 2018-11-27
 
 ### Added
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStringConstants.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStringConstants.cs
@@ -384,7 +384,10 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
         public static readonly int _Source = Shader.PropertyToID("_Source");
         public static readonly int _Destination = Shader.PropertyToID("_Destination");
         public static readonly int _Mip0 = Shader.PropertyToID("_Mip0");
+        public static readonly int _SourceMip = Shader.PropertyToID("_SourceMip");
         public static readonly int _SrcOffsetAndLimit = Shader.PropertyToID("_SrcOffsetAndLimit");
+        public static readonly int _SrcScaleBias = Shader.PropertyToID("_SrcScaleBias");
+        public static readonly int _SrcUvLimits = Shader.PropertyToID("_SrcUvLimits");
         public static readonly int _DstOffset         = Shader.PropertyToID("_DstOffset");
         public static readonly int _DepthMipChain = Shader.PropertyToID("_DepthMipChain");
 
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/ColorPyramidPS.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/ColorPyramidPS.shader
@@ -1,0 +1,78 @@
+Shader "ColorPyramidPS"
+{
+    HLSLINCLUDE
+
+        #pragma target 4.5
+        #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
+        #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+        #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+
+        TEXTURE2D(_Source);
+        SamplerState sampler_LinearClamp;
+        uniform float4 _SrcScaleBias;
+        uniform float4 _SrcUvLimits; // {xy: max uv, zw: direction of blurfor 1 texel }
+        uniform float _SourceMip;
+
+        struct Attributes
+        {
+            uint vertexID : SV_VertexID;
+        };
+
+        struct Varyings
+        {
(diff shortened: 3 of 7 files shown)


EvgeniiG:
Very nice improvements!

What was the bottleneck of the compute path on Switch? Shared memory? Occupancy? Barriers? Scheduling? Something else?

If you have any info, you can post it here, or Slack me or @Chman.

Thanks!

Chman:
Ok so I ran more tests on PS4 and did a few local changes to the compute version. I got it to run at the same speed as the fragment path which bares the question: do we actually want to keep the compute version or not? Given that we'll never run this pass in async, the compute code is much more complicated and fails on some Intel and old Nvidias... We may as well use the fragment path for every platform if the gain is better compatibility across platforms & ease of maintenance.

What do you think @kaychang-unity? AFAIK @sebastienlagarde and @EvgeniiG agree.

kaychang-unity:
On Switch, compute shaders have more potential bottlenecks than fragment shaders, so they are likely to run slower. For a simple downsampling shader, using 32 threads group size (a Switch wavefront is 32 threads instead of 64) and using RWTexture for output is another bottleneck. The fragment shader version uses ROP for the output and has no such bottleneck on Switch.
As I mention, we should be avoiding the full mip0 copy altogether as it is wasteful.

As if we should give up the compute shader version for PS4, I would say: Yes for now. But the question becomes irrelevant if we implement the following optimization instead:
At the moment, we do the following for both compute shader and fragment shader versions:
downsample to mip1 -> horizontal blur mip1 -> vertical blur mip1 -> downsample to mip2 -> horizontal blur mip2 -> vertical blur mip2 -> downsample to mip3 -> horizontal blur mip3 -> vertical blur mip3 -> ....
GPU barriers are needed between each subpass because they are dependent operations. Also, when processing the smaller mips, the GPU will be under-utilized because there isn't enough pixels to process all the wavefronts.
Instead, we could perform all the downsampling subpasses together, followed by all horizontal blurs, followed by all vertical blurs. This would become:
downsample to mip1 -> downsample to mip2 -> downsample to mip3  -> horizontal blur mip3 -> horizontal blur mip2 ->  horizontal blur mip1 -> vertical blur mip3 -> vertical blur mip2 -> vertical blur mip1
This will reduce the GPU barriers to a few (downsampling can still output to several mips at a time using compute shader, removing more GPU barriers). Finally, by scheduling all small mips blur operations before the larger mips, they can run concurrently on the GPU and will occupy more GPU cores.

That optimization implies that we will not accumulate gaussian blurs through downsampling mips. However, the gain of accumulating gaussian blur is not as significant as one would expect (from wikipedia, the product of 2 gaussian blur of radii a and b is an equivalent gaussian blur or radii sqrt(a² + b²)).
If we look at the gaussian blur result at each mip using current implementation and consider they are attempts at approximation of gaussian blurs on mip0:
mip0: original
mip1: equivalent to gaussian kernel 18 on mip0 (because gaussian kernel 9 but applied mip1)
mip2: equivalent to gaussian kernel 40 (=sqrt(18² + 36²)) on mip0
mip3: equivalent to gaussian kernel 82 (=sqrt(40² + 72²)) on mip0

If we look at the gaussian blur result at each mip using proposed implementation:
mip0: original
mip1: equivalent to gaussian kernel 18 on mip0
mip2: equivalent to gaussian kernel 36 on mip0
mip3: equivalent to gaussian kernel 72 on mip0

Thus, the final amount of blur generated is less using the proposed implementation, but I think still acceptable considering the GPU gain.

Chman:
> As I mention, we should be avoiding the full mip0 copy altogether as it is wasteful.

The reason we need the full copy instead of doing the pyramid in-place in the original color buffer is because at some point in HDRP it needs to read & write from/to mip0 (and you can't read & write the same target on most platforms). It happens in the distortion effect. We could force distortion to do its own copy of mip0 and do it in-place which would result in a gain when distortion is disabled (but AFAIK it's always-on right now).

As for the proposed optimization... I'm fine with it if it doesn't look that much different.

Chman:
Note: with all of that said, distortion is the latest step in HDRP before postfx kicks in, and postfx doesn't make use of the pyramid (due to other concerns)... So we could do the pyramid in-place anyway and have distortion output to another target that gets sent to postfx, thus removing the need for the mip0 copy.

sebastienlagarde:
You are forgetting that the color pyramid for distortion is reuse for SSR the next frame.

It is a bit that we don't reuse the pyramid for bloom as well - but I understand why. Remember @Chman we were discussing about having quality control for bloom and on low, we could reuse this color pyramid to save time.

Note that there is 2 color pyramid currently.

One for refraction (done after pre-refraction pass) to allow refract object, then one for distortion + SSR. Distortion allow to distort everything including the object itself.

It is true that currently we can't disable it but we will add option to do it if there is no SSR or distortion. 
Alrenatively we could reuse the bloom target for SSR maybe ?

kaychang-unity:
I started running this katana build: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=switch-support&automation-tools_branch=add-platform-filter&unity_branch=trunk

PR #2515, opened 2018-11-27
Use ERROR_ON_UNSUPPORTED_FUNCTION instead of ERROR_ON_UNSUPPORTED_FUNC

hkr:
ERROR_ON_UNSUPPORTED_FUNC is not defined, API headers only define ERROR_ON_UNSUPPORTED_FUNCTION.
This is also a workaround for infinite recursion in the Unity shader compiler/preprocessor that doesn't properly support self-referencing macros.

### Purpose of this PR
Fix/workaround for https://issuetracker.unity3d.com/issues/mobile-os-restarts-because-of-high-memory-usage-when-compiling-shaders-for-opengles2

Shader compilers that use mojoshader preprocessor don't handle self-referencing macros like:
```
#define MACRO 1+MACRO
```
or (if ERROR_ON_UNSUPPORTED_FUNC is undefined)
```
#define LODDitheringTransition ERROR_ON_UNSUPPORTED_FUNC(LODDitheringTransition)
```
This generates infinite recursion, which is not trivial to fix in Unity's shader preprocessor.

---
### Release Notes
Workaround for out-of-memory of Unity Editor when trying to compile a shader that uses an unsupported function.
---
### Testing status

Didn't run yet

Created a shader

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers


phi-lira:
Thanks for looking into this. Please add release notes to changelog.md under 5.3.0. if there's no Fixed category in 5.3.0 yet. Add it below other categories (added or changed)

hkr:
@phi-lira Updated changelog

phi-lira:
Added @Katstodian for doc review. I think you should also add notes to lwrp changelog. Once she approves I'll merge.

com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [5.3.0] - 2018-11-xx
+### Fixed
+- Fixed uses of the ERROR_ON_UNSUPPORTED_FUNCTION() macro. This works around out-of-memory in Unity's shader compiler.
```
Katstodian: Can this be changed to:

Fixed incorrect use of ERROR_ON_UNSUPPORTED_FUNC(). When it's out of memory, the Unity shader compiler now correctly uses the ERROR_ON_UNSUPPORTED_FUNCTION macro.


?

phi-lira:
Backported on 18.3 on e81fa66e6c25e6599f17955214f52912029ee5fd

PR #2524, opened 2018-11-28
Fix for Mac : return float4 in frag shader function even for depth pr…

iTris666:
Backport if vfx/fix/float4-depth-prepass


julienf-unity:
Please backport

PR #2536, opened 2018-11-29
Fixed CopyStencilBuffer pass for Switch.

kaychang-unity:
### Purpose of this PR
Fix copy stencil buffer pass on Switch.
There is an internal bug on Switch where if the shader does not output to a color buffer, the RwImage2D write operation will also fail. As advised, we use pass 1 of the CopyStencilBuffer.shader instead to output to the color buffer directly. Avoiding RwImage2D write is also a speedup on Switch (25% faster).

---
### Release Notes
Fix copy stencil buffer pass on Switch.

---
### Testing status
**Katana Tests**: 
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=switch-support&unity_branch=trunk&automation-tools_branch=add-platform-filter

**Manual Tests**:
Using Switch GPU debuffer shows that the output of the pass is now correct and identical as PC D3D11 output.

**Automated Tests**:
Didn't run any test.

---
### Overall Product Risks
**Technical Risk**:
Low. Switch specific fix.

**Halo Effect**:
Low. The code is Switch specific and use Pass 1 of the CopyStencilBuffer.shader that no other platforms use.

---
### Comments to reviewers
-


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1167,8 +1167,17 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
                             // For material classification we use compute shader and so can't read into the stencil, so prepare it.
                             using (new ProfilingSample(cmd, "Clear and copy stencil texture", CustomSamplerId.ClearAndCopyStencilTexture.GetSampler()))
                             {
-                                HDUtils.SetRenderTarget(cmd, hdCamera, m_SharedRTManager.GetStencilBufferCopy(), ClearFlag.Color, CoreUtils.clearColorAllBlack);
+#if UNITY_SWITCH
+                                // Internal bug on Switch where the shader will not write into RwImage2D without a color render-target bound.
```
FrancescoC-unity: Is this comment still relevant? This pass now doesn't write to RwImage2D anymore. I would perhaps update it so that it says we are running pass 1 because of the perf benefit on switch

PR #2540, opened 2018-11-30
Hdrp terrain details shader, now we have pink  grass 

mikele92:
When placing grass on terrain under hdrp the grass appears pink.
Suspecting that there is no shader for terrain details.

Also speed tree shaders are pink. And cant aproach wind


0lento:
@mikele92 You are supposed to report bugs elsewhere, PR's are not the right place. 

First lines from this repos readme:
**NOTE: We have migrated reported issues to FogBugz. You can only log further issues via the Unity bug tracker.**

sebastienlagarde:
As advice above, please use FogBugz to report this kind of issue. Thanks

PR #2541, opened 2018-11-30
Terrain details appears pink under HDRP

mikele92:
When painting grass details on terrain under hdrp the grass appears pink.
I think you need shader for terrain details under hdrp
And also cant use speed tree materials they are also pink and there is no wind
And cant automatically convert them using the option convert materials to hdrp.


sebastienlagarde:
please use FogBugz to report this kind of issue. Thanks

PR #2542, opened 2018-11-30
blackboard in SG remember expanded states

martint-unity:
### Purpose of this PR
Fixing case https://fogbugz.unity3d.com/f/cases/1102332/
Now the blackboard remembers the state it is in when undoing. This is per Unity session.

---
### Release Notes

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Fblackboard-remember-unfolds&automation-tools_branch=add-platform-filter&unity_branch=trunk

Open up a shadergraph and add properties into the blackboard. unfold some of them and restart the graph and they should now be remembered. 
It will only remember during the session. SO if you close Unity it will no longer remember the state.

---
### Overall Product Risks
Low

Low

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
@@ -51,6 +51,14 @@ namespace UnityEditor.ShaderGraph
             get { return m_MovedProperties; }
         }
 
+        [NonSerialized]
+        Dictionary<IShaderProperty, bool> m_ExpandedProperties = new Dictionary<IShaderProperty, bool>();
+
+        public Dictionary<IShaderProperty, bool> expandedProperties
+        {
+            get { return m_ExpandedProperties; }
+        }
+
         [SerializeField]
         SerializableGuid m_GUID = new SerializableGuid();
 
@@ -241,6 +249,7 @@ namespace UnityEditor.ShaderGraph
             m_AddedProperties.Clear();
             m_RemovedProperties.Clear();
             m_MovedProperties.Clear();
+            m_ExpandedProperties.Clear();
         }
 
         public virtual void AddNode(INode node)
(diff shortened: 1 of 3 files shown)


com.unity.shadergraph/Editor/Data/Graphs/AbstractMaterialGraph.cs
```diff
@@ -51,6 +51,14 @@ public IEnumerable<IShaderProperty> movedProperties
             get { return m_MovedProperties; }
         }
 
+        [NonSerialized]
+        Dictionary<IShaderProperty, bool> m_ExpandedProperties = new Dictionary<IShaderProperty, bool>();
```
pbbastian: Since it isn't serialised here anyway, this should be stored in BlackboardProvider instead.

martint-unity:
@pbbastian This should now been fixed with the feedback

martint-unity:
@zadi0648 please confirm that the fix is working :)

com.unity.shadergraph/Editor/Drawing/Blackboard/BlackboardProvider.cs
```diff
@@ -253,6 +266,15 @@ void AddProperty(IShaderProperty property, bool create = false, int index = -1)
                 m_Section.Insert(index, row);
             m_PropertyRows[property.guid] = row;
 
+            try
+            {
+                m_PropertyRows[property.guid].expanded = SessionState.GetBool(property.guid.ToString(), false);
+            }
+            catch (Exception e)
+            {
+                Debug.LogException(e);
```
pbbastian: Do you expect to regularly get an exception here or?

martint-unity:
@zadi0648 Thanks for finding that. It was the default state of the properties always start in a expanded state. And I set the initial bool to be false which is completely wrong. Now those states will be the same. So please re-test :)

PR #2549, opened 2018-11-30
Fixed PP flip logic for VR

nerites:
### Purpose of this PR
When hack for multicamera support was reverted, logic to flip the scene when VR is active got added. This logic should only apply pre-2019.1 and is removed in this PR.

---
### Release Notes
Fixed flip for postprocessing with HDVR.

---
### Testing status
**Katana Tests**: 
In progress: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp_ppvrfix&unity_branch=trunk&automation-tools_branch=add-platform-filter

Manually confirmed flip is correct in HDVR with single-pass double-wide. 

None.

---
### Overall Product Risks
Low

Low

---


sebastienlagarde:
@fabien-unity @nerites we need to think about how to add graphic test for VR as currently none of this break change can be catch :(

PR #2550, opened 2018-12-02
Pipeline hooks for terrain details shaders in lightweight.

russoh:
### Purpose of this PR
Terrain detail rendering doesn't currently work in legacy pipeline if lightweight pipeline is installed, as the shaders are replaced. This change provides an SRP hook for lightweight.

Fixes this bug: https://fogbugz.unity3d.com/f/cases/1090634/

This PR is dependent on the following trunk change and should not be merged before it has landed:
https://ono.unity3d.com/unity/unity/pull-request/79232/_/graphics/terrain-details-srp-hooks

---
### Release Notes

---
### Testing status

Tested all 3 detail rendering shaders in lightweight and legacy pipeline.

---
### Overall Product Risks
Low

Low

---
### Comments to reviewers


com.unity.render-pipelines.lightweight/Runtime/Data/LightweightRenderPipelineResources.cs
```diff
@@ -9,6 +9,10 @@ public class LightweightRenderPipelineResources : ScriptableObject
         [FormerlySerializedAs("ScreenSpaceShadowShader"), SerializeField] Shader m_ScreenSpaceShadowShader = null;
         [FormerlySerializedAs("SamplingShader"), SerializeField] Shader m_SamplingShader = null;
 
+        [SerializeField] Shader m_TerrainDetailLitShader = null;
+        [SerializeField] Shader m_TerrainDetailGrassShader = null;
+        [SerializeField] Shader m_TerrainDetailGrassBillboardShader = null;
```
phi-lira: Serialize in the Editor resources instead. If done here they will always be included in the build. Check how autodesk shaders were added for reference.

com.unity.render-pipelines.lightweight/Runtime/Data/LightweightRenderPipelineAsset.cs
```diff
@@ -442,6 +442,21 @@ public override Shader autodeskInteractiveMaskedShader
         }
 #endif
 
+        public Shader terrainDetailLitShader
```
phi-lira: Should accessing editorResources and ifdef'ing UNITY_EDITOR here.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainDetailLit.shader
```diff
@@ -1,4 +1,4 @@
-Shader "Hidden/TerrainEngine/Details/Vertexlit" 
+Shader "Hidden/TerrainEngine/Details/LightweightPipeline-Vertexlit"
```
phi-lira: Minor/Nitpicking. For consistency. Do: LightweightRenderPipeline/VertexLit. Same for the ones below.

com.unity.render-pipelines.lightweight/Runtime/Data/LightweightRenderPipelineResources.cs
russoh: This PR is just intended as a bug fix, keeping the existing behavior of always including those shaders. We don't actually serialize a shader or material reference in the terrain data currently. That might be a suitable change, I agree that smaller builds are good, but it's definitely a change that could potentially have implications for script based logic, attempting to build stuff dynamically at runtime, instead of edit time. I simply don't know if that's a real issue, but I'd like to err on the side of caution until I've checked properly, and I think we should merge this bug fix PR ASAP since it's affecting users currently.

phi-lira: Ok. I guess that's fine for now. Just to make sure, in this case we are only including the LWRP terrain shaders in the build right?

russoh: Are you asking if the standard legacy builtin terrain resources are removed if lightweight render pipeline is active when a build is made? I.e. all of these:

    // Terrain engine
    RegisterShader(10490, "TerrainShaders/Splats/Terrain-Utilities.shader", "Hidden/Nature/Terrain/Utilities", false);
    RegisterShader(10500, "TerrainShaders/Details/VertexLit.shader", "Hidden/TerrainEngine/Details/Vertexlit", false);
    RegisterShader(10501, "TerrainShaders/Details/WavingGrass.shader", "Hidden/TerrainEngine/Details/WavingDoublePass", false);
    RegisterShader(10502, "TerrainShaders/Details/WavingGrassBillboard.shader", "Hidden/TerrainEngine/Details/BillboardWavingDoublePass", false);
    RegisterShader(10507, "TerrainShaders/Trees/BillboardTree.shader", "Hidden/TerrainEngine/BillboardTree", false);
    RegisterShader(10513, "TerrainShaders/Trees/CameraFacingBillboardTree.shader", "Hidden/TerrainEngine/CameraFacingBillboardTree", false);
    RegisterShader(10504, "TerrainShaders/Splats/DiffuseBase.shader", "Hidden/TerrainEngine/Splatmap/Diffuse-Base", false);
    RegisterShader(10506, "TerrainShaders/Splats/DiffuseBaseGen.shader", "Hidden/TerrainEngine/Splatmap/Diffuse-BaseGen", false);
    RegisterShader(10503, "TerrainShaders/Splats/AddPass.shader", "Hidden/TerrainEngine/Splatmap/Diffuse-AddPass", false);
    RegisterShader(10505, "TerrainShaders/Splats/FirstPass.shader", "Nature/Terrain/Diffuse");

phi-lira: Yes. I'm concerned about this PR adding more shader variants. (Including both builtin and lwrp shaders).

russoh: I don't think there's any stripping of builtin terrain resources based on active render pipelines at build time. ctchou is also not aware of anything attempting to strip stuff out. It's a fair consideration for build optimization, but will need some attention to check across all systems. Thankfully these resources are pretty small and not large image resources, but I'll make a note of this in our task tracking so that it can be properly considered moving forwards.

phi-lira: It's not about stripping based on the active pipeline, it's about stripping unreferenced shaders. If a shader is not referenced anywhere in the project it *will* not be included in the build. This is what happens for built in standard shaders, lwrp shaders and so on. I truly hope terrain does not always include shaders. If this is the case it needs to be fixed by terrain system asap. 

We are talking about multiple shaders with possible 10k variants per API. This is possible a change that adds dozens of MBs to device system memory plus minutes of build time. 

The issue with your PR is that it adds a reference to these shaders in a file that's always included in the build. This will make these shaders *always* be included in the build even if they are not used. That's my concerned. 

Could you please do the same that was done for Autodesk shaders? They are referenced in an editor resource file which does not get included in the build.

phi-lira:
Adding @Verasl so we check this PR doesn't increase shaders and built time in the template scene.

cdxntchou:
Maybe in the interest of getting this bug fixed ASAP -- we can simply remove the LWRP detail shader?

Then we can add it back after we have the full fixes for referencing it properly.

russoh:
Only issue I know of is that the detail renderer doesn't currently serialize a shader reference in editor, so requires additional changes to make this work. I've gone ahead and made the required change to support this, but it's meant that I've had to make a new trunk change iteration and it will probably take a few days to turn it around again. In terms of getting the bug fixed, I think we should have settled on a trivial solution to get it resolved asap, and then scheduled the "correct" fix as a follow up. I'm choked up with other stuff this week, and can't devote any more time to fixing this bug, so if there are other changes needed, we may need to do as Chris suggested in the short term and remove the lw shaders until the proper fix has landed. Thanks.

phi-lira:
I'm ok with the solution to disable detail/grass terrain shaders for now until we have a proper fix.

russoh:
OK, I believe the current solution works correctly as you suggested with the editor build time dependency. No need to remove the shader if the current PRs go through successfully. I'll post here again once the engine side of this is merged.

sebastienlagarde:
Katana is green except on HDRP test but which is due to master not about this PR. Merging.

PR #2551, opened 2018-12-02
HDRP metal fixes

aet:
enabling HD Metal tests still a work in progress, but bring few elementary fixes in + a test script for local testing

### Purpose of this PR

---
### Release Notes

---
### Testing status

---
### Overall Product Risks

---
### Comments to reviewers


aet:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp-metal-fixes&unity_branch=trunk

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/HDShadow.hlsl
```diff
@@ -18,7 +18,7 @@
 
 float GetDirectionalShadowAttenuation(HDShadowContext shadowContext, float2 positionSS, float3 positionWS, float3 normalWS, int shadowDataIndex, float3 L)
 {
-    return EvalShadow_CascadedDepth_Blend(shadowContext, _ShadowmapCascadeAtlas, sampler_ShadowmapCascadeAtlas, positionSS, positionWS, normalWS, shadowDataIndex, L);
+    return EvalShadow_CascadedDepth_Blend(shadowContext, _ShadowmapCascadeAtlas, s_linear_clamp_compare_sampler, positionSS, positionWS, normalWS, shadowDataIndex, L);
```
sebastienlagarde: is it part of the fix or just a simplification of the code ?
because here you assume that the sampler is a compare (i.E for PCF), but we are currently implementing moment sahdow map which is a regular sampler, so this code wouldn't work.

Just trying to understand the motivation see if we have an alternative

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
```diff
@@ -1623,8 +1623,8 @@ IndirectLighting EvaluateBSDF_Env(  LightLoopContext lightLoopContext,
 
     // TODO: We need to match the PerceptualRoughnessToMipmapLevel formula for planar, so we don't do this test (which is specific to our current lightloop)
     // Specific case for Texture2Ds, their convolution is a gaussian one and not a GGX one - So we use another roughness mip mapping.
-#if !defined(SHADER_API_METAL)
     float   IBLMipLevel;
+#if !defined(SHADER_API_METAL)
     if (IsEnvIndexTexture2D(lightData.envIndex))
```
sebastienlagarde: btw, isn't this stuff not fixed ?

#if !defined(SHADER_API_METAL)
    if (IsEnvIndexTexture2D(lightData.envIndex))	    if (IsEnvIndexTexture2D(lightData.envIndex))

Because currently metal still don't match windows for planar reflection (except that currently planar reflection test are disabled due to recent API changed).

aet: Apple's shader compiler got broken in 10.13.6 for this code, and it's working in 10.14. Unless we change HD to require 10.14, we may need to keep this workaround in.. there won't be any 10.13.x upgrades anymore for GPU stuff

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/HDShadow.hlsl
aet: in metal generated code, the sampler_ShadowmapCascadeAtlas/sampler_ShadowmapAtlas declarations ended up being actually unused in code, as the codegen decided to use internal declaration of depth sampler iirc. Switching to inline sampler declaration avoid hitting the sampler limitations

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.hlsl
sebastienlagarde: So I  suppose what is failing is the dynamic branch:
bool IsEnvIndexTexture2D(int index) { return (index & 1) == ENVCACHETYPE_TEXTURE2D; }

I will test an alternative storage, see if this solve it. Because current workaround don't match

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/HDShadow.hlsl
EvgeniiG: This is fine IMHO.

com.unity.render-pipelines.core/ShaderLibrary/API/Metal.hlsl
```diff
@@ -16,6 +16,7 @@
 #define CBUFFER_END };
 
 #define PLATFORM_SUPPORTS_EXPLICIT_BINDING 1
+#define PLATFORM_NEEDS_UNORM_UAV_SPECIFIER 1
```
EvgeniiG: How is this used?

sebastienlagarde: This is already there and used with normal buffer code.

PR #2558, opened 2018-12-03
Fix blit pass for stereo rendering

fabien-unity:
### Purpose of this PR
I found out that all passes (4x) of the shader were executed. I've set it up to use pass 1 (bilinear, triangles). We should probably be using point filter (unless dynamic resolution might need it) but this code should go away once SPI is working.

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=xr%2Fhdrp%2Fblit-fix&unity_branch=trunk)

**Manual Tests**: HDRP test scene with VR on Windows 10 with Oculus Rift (Oculus and OpenVR).

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None (stereo only)


nerites:
When I tested this on Vive, I confirmed that the image was right-side-up in GameView, but it the image was upside-down in the HMD. What recent changes in trunk are you referring to? I should make sure my Editor aligns with yours.

nerites:
Also, for what it's worth, it looks like the built-in Blit shader has the same behavior as the updated Blit.shader in HDRP. If we flip it to be right-side-up in the final blit to CameraTarget (in this case, HMD eye texture), then the XR.MirrorToGameView() call to the internal Blit shader will flip it upside-down again in GameView (which is probably what you were seeing). So that's... unfortunate.

fabien-unity:
Alright, I'll test on actual hardware tomorrow and look for a real solution.

fabien-unity:
I just tested with 2 headsets and It's correct with the Rift but not with the Vive :/

fabien-unity:
I've restored the blit, the fix for the pass should go in master though.
Ready again for review.

PR #2559, opened 2018-12-04
Hdrp stacklit better dualnormalmaps

slunity:
### Purpose of this PR
StackLit : Fix punctual/directional lighting in dual normal map mode.
Also fix a UI issue for the master node where transmission would not open a diffusion profile slot (and required SSS to assign it)

---
### Release Notes

---
### Testing status

---
### Overall Product Risks

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Material/StackLit/StackLit.hlsl
```diff
@@ -2876,34 +2978,114 @@ float3 EvaluateTransmission(BSDFData bsdfData, float3 transmittance, float NdotL
...
+}//EvaluateLights_GetDirections
+
+
 //-----------------------------------------------------------------------------
 // EvaluateBSDF_Directional
 //-----------------------------------------------------------------------------
 
 DirectLighting EvaluateBSDF_Directional(LightLoopContext lightLoopContext,
-                                        float3 V, PositionInputs posInput, PreLightData preLightData,
+                                        float3 inV, PositionInputs posInput, PreLightData preLightData,
```
sebastienlagarde: avoid to change parameter names (from V to inV), we all have subjective preference but the goal here is to be homogeneous though all files, even if it is not the best choice or best naming. 

Same for float3 mainN;, float3 mainL, float3 mainNDotL. the "main" prefix isn't required.

(Note: I will still merge with this currently.)

slunity: yes sorry will change that, I had renamed those when doing initial test of the idea when working within the original function, should have renamed back after factoring the code out.

PR #2564, opened 2018-12-04
Simple lit and vfx fixes

anisunity:
Skipping the probe contribution for the simple lit in case we do not have env lighting
Skipping refraction in the simple lit
Re-enabling the test and updating the reference
Re-compile the simple lit shadow vfx
Correcting the forward pass name for the simple lit (forward only)

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=vfx%2Ffix%2Fsimple_lit&automation-tools_branch=add-platform-filter&unity_branch=trunk#


sebastienlagarde:
Katana is green except the 2 test failing on master unrelated to this PR. merging

PR #2568, opened 2018-12-04
Fixed hdrp shader preprocessor.

JulienIgnace-Unity:
Fix hdrp stripper for builtin lit shaders.

There is no Katana here as the stripper isn't test by Katana, only when building a player.


FrancescoC-unity:
Oopsie :) Sorry about that!

PR #2597, opened 2018-12-10
Add WaveIsFirstLane and WaveGetLaneIndex intrinsic

FrancescoC-unity:
### Purpose of this PR
All in the title.  Adding them as there is something I'd like to try and notice we didn't have these. 

---
### Testing status
**Katana Tests**:  Pointless

---
### Overall Product Risks
**Technical Risk**:  Low, 

**Halo Effect**: Low

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/API/XBoxOne.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/API/XBoxOne.hlsl
@@ -79,6 +79,11 @@ bool WaveIsFirstLane()
     return (__XB_MBCNT64(WaveActiveBallot(true))) == 0;
 }
 
+uint GetWaveID()
+{
+    return __XB_GetWaveID();
+}
+
 #define INTRINSIC_MINMAX3
 GENERATE_INTRINSIC_VARIANTS_3_ARGS(Min3, __XB_Min3_, a, b, c);
 GENERATE_INTRINSIC_VARIANTS_3_ARGS(Max3, __XB_Max3_, a, b, c);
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.core/ShaderLibrary/API/XBoxOne.hlsl
```diff
@@ -69,6 +69,20 @@ bool WaveActiveAnyTrue(bool expression)
...
+}
+
+bool WaveIsFirstLane()
+{
+    return (__XB_MBCNT64(WaveActiveBallot(true))) == 0;
+}
+
+uint GetWaveID()
+{
+    return __XB_GetWaveID();
```
EvgeniiG: Isn't this basically (dispatchThreadId / 64)? What's the use case? Pixel shader? Would be nice to keep parity between PS4 and XB1 intrinsics.

EvgeniiG: Also, should probably call it WaveGetId() for consistency.

FrancescoC-unity: The parity here is preserved by the name itself, it is called GetWaveID() in PS4. And yes, it is dispatchThreadId / 64 but I like the more explicit name and the fact that it could be used for PS as well. 

 I thought of calling it WaveGetId(), but I didn't because all the Wave* are cross lanes ops, this is actually a wave constant. It's a subtle difference, so I am happy to rename it if needed :)

FrancescoC-unity: I mean I don't really need it tbh, I had it put there by mistake the first time around and I thought to leave it there because of the above reasons.

EvgeniiG: Wave_() is for any wave ops. :-) And you would certainly use this one in the context of wave ops. My 2c.

FrancescoC-unity: It is fair enough! Will change name and add the PS4 version

PR #2600, opened 2018-12-11
Adding support to LWRP for dynamic resolution scaling. 

davidrogers-unity:
Requires some changes on the engine side as well. https://ono.unity3d.com/unity/unity/pull-request/79744/_/graphics/dynamic-resolution/lwrp

### Purpose of this PR
Adding support for enabling dynamic resolution scaling to the lightweight render pipeline. More or less we need to be able to set the flag for enabling from scripting (engine changes) and then set it if the user has checked the box(which also gets added in this PR).

---
### Release Notes
Dynamic Resolution Scaling is now compatible with LWRP. It may be used in exactly the same way as with the non scriptable rendering pipeline.

---
### Testing status

**Manual Tests**: Testing with some projects that use LWRP, enabled DRS on the main camera and attached a script for controlling the scale.

Yes, I have a couple that I have prepared.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Medium

---


phi-lira:
Adding @kat for release notes review.

phi-lira:
Please let me know when this can be merged and if it should be backported to 18.4

Katstodian:
Please add an entry in the relevant Changelog.md, that I can review. :)

davidrogers-unity:
@phi-lira It affects Android/iOS/tvOS/XB1/PS4/Switch, I think that's all the platforms that support LWRP. It's not enabled by default and is designed to allow the user to improve performance. 

After talking with JC it turns out that this does not need to be back ported, only for 2019.1 and beyond.

@Katstodian I'll update the changelog in a few minutes. 

Also right now the tests are failing because the trunk support has not landed yet.

Katstodian:
@davidrogers-unity - I edited the changelog.md entry to:
LWRP now supports dynamic resolution scaling. The target platform must also support it.
:)

phi-lira:
@davidrogers-unity Ok. Just let me know when it can be merged.

davidrogers-unity:
@phi-lira Support has landed in trunk so this is ok to be merged now.

PR #2610, opened 2018-12-12
Shader graph add shader to project

sebastienlagarde:
### Purpose of this PR
Add dummy C# file and asm file to be able to broswer hlsl files from shader graph project.

No Katana test, tested manually.


sebastienlagarde:
thanks for the typo fix :)

PR #2562, opened 2018-12-04
HDRP/Post-processing

Chman:
### Purpose of this PR
Tightly integrated post-processing for HDRP. Requires 2019.1.
Once this PR goes in we should be able to move color buffers to R11G11B10 instead of fp16 and enjoy a nice speed boost thanks to pre-exposition.

Currently missing:
- Async compute support
- Curves (color grading)
- User LUT (color grading)
- Motion blur
- XR support

This is still in a (slightly above) WIP state. More details [in this doc](https://docs.google.com/document/d/1l7Ej-d3yzASux-tqRDN12ld7m1zXJrGhTZBm498mDjQ/edit#).

---
### Testing status
Mostly untested right now. Developed on Nvidia/DX11, Intel/DX11 and PS4. Seems to work on OSX/Metal as well. Other platforms haven't been tested. Will run Katana once the other PRs this depends on will get merged.

---
### Overall Product Risks
**Technical Risk**:
Medium.
Pre-exposition required modifications to the Lit, Sky and Volumetrics shader families.

**Halo Effect**:
High.
Breaks XR support in HDRP. Breaks post-processing setup in current demos for obvious reasons.
Adds a lot of stuff to resources, so it'll probably break as well.

---
### Comments to reviewers
- Due to the amount of files changed, Github auto-collapse most of them in the change list. Make sure you don't skip these.
- This PR currently has unrelated post-processing changes in it because of another PR that got merged into it and hasn't reached master yet. Once the original PR has been merged to master, this PR will get smaller.
- Most of the post-processing logic happens in `PostProcessSystem.cs`.
- Add more reviewers if you think it's needed.

PR #2617, opened 2018-12-13
Deprecate CoreEditorUtils.GetContent

RSlysz:
### Purpose of this PR
Removing CoreEditorUtils.GetContent that is duplicate of now exposed EditorGUIUtility.TrTextContent

---
### Release Notes

---
### Testing status
**Katana Tests**: 

**Manual Tests**: tested locally

**Automated Tests**: 

---
### Overall Product Risks
**Technical Risk**: High
(high number of files impacted)

**Halo Effect**: Low
(Affect only labels of inspector. Will spread warning (obsolete) in other package using CoreEditorUtils.GetComponent)

---


Chman:
Note: this will impact the #2562. I don't see anyone from LW in the reviewer section, was this function used in LW at all? cc @phi-lira

Chman:
If you do it on every frame then yes, it's a reference type and will generate garbage. That's why we use this function to cache them.

PR #2633, opened 2018-12-18
metal amd gfx-tests

aet:
### Purpose of this PR

As we're currently not maintaining GPU vendor specific template images, this PR now splits non-HD related tests to keep running on Intel GPUs to ease server load and run HD/VisualGraph tests on AMD

- VisualGraph is still disabled on a Mac, as a last minute issue was found that requires HLSLcc fix first
- HD tests against AMD can be triggered separately, but it's not yet a mandatory build step to succeed
- Updates some outdated templates / rebuild vfx so they'll work outside of tests runner

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=metal-fixes&unity_branch=trunk

---
### Release Notes

---
### Testing status

---
### Overall Product Risks

---
### Comments to reviewers


PaulDemeulenaere:
@aet  "that’s causing a faulty shader to be generated." <= Is it a compilation error with unsigned integer ?

aet:
Found the generic issue from HLSLcc, takes a while to fix.. will fix opening on scenes like HDRP from VFX but it's not related to 06_LineOutput, 08_Shadows and 10_SortPriority. Once fixed, should still make automated testing stable by not having shader compilation error

julienf-unity:
I took a look at bad gfx tests and I'm not sure issues come from us. I need to investigate more.

PaulDemeulenaere:
By the way, the change of all ".vfx" asset was optional. If you want to reduce your changeset and avoid any conflict, you can revert them (they are automatically recompile while running test anyway).

PR #2644, opened 2018-12-19
Fix Shader Graph window not repainting in the background

pbbastian:
### Purpose of this PR
If Unity does not have focus the Shader Graph window does not repaint. But we are receiving `Update` events, and thus we are actually re-drawing all the previews, even though they don't get used for anything. I've put in a `Repaint` call at the end of the `Update` function to fix this.

---
### Testing status

**Manual Tests**: Verified that before the change no repaints happen when Unity does not have focus, and that afterwards it does. I also looked into the rate of `Update` events with and without focus. Without focus we get half as many `Update` events, which can also be seen after the changes as the previews run at half speed.

**Automated Tests**: Nothing

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None


pbbastian:
Requestion docs review again as I forgot to include change log commit

pbbastian:
Awesome, thanks! :) There shouldn't be a performance gain, so that's all fine. It's more that previously we were drawing all of those previews in the background, and then not using them for anything.

PR #2652, opened 2018-12-19
Add right eye bounds and volumes to lightlist (stereo rendering)

fabien-unity:
### Purpose of this PR
Right eye was missing some bounds and volumes information for decals and volumetrics.

---
### Testing status
**Katana Tests**: [green on Windows](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=xr%2Fhdrp%2Flightlist&unity_branch=trunk)

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


sebastienlagarde:
Shoudn't this be done only if VR is enabled ? Allocating extra memory and tabs all the time for all camera isn't good.

fabien-unity:
Good point, I'll look into it.

PR #2664, opened 2018-12-20
Single-pass Instanced Stereo Support (hybrid multipass implementation)

nerites:
### Purpose of this PR
Enable single-pass instanced stereo rendering with the following constraints:
- No Vulkan
- No MSAA
- Forward only
- No FPTL

More info (incl. todo and future work): https://docs.google.com/document/d/12zqlUCcEluPe5h4ucRxupnm13evaqDxVWH7CjARJct4/edit?usp=sharing

---
### Release Notes
This implementation has been partially integrated with PPv3. For some reason postprocessing only applies to the right eye: when debugging the C#, it looks like the correct target from the TargetPool is picked for the left eye, but when viewed in RenderDoc, the incorrect target from the TargetPool is bound instead. Hoping to get some help from @Chman on that front; maybe it's something to do with how I'm pooling the temporary textures? Anyways, latest renderdoc here: 
https://drive.google.com/open?id=1XDxp3Pl6cwkWnL6Z5pdZkw_eqDkkgBHN

image

Remaining issues:
- lightlistbuild-clustered.compute disconnects from compute shader compiler for Vulkan with SPI+Texture2DArray (bug here: https://fogbugz.unity3d.com/f/cases/1103046/)
- Errors in SubsurfaceScattering.compute: "All kernels must use same constant buffer layouts; '_ScreenSize' property found to be different", also flags _ShapeParams, _TexturingModeFlags, _WorldScales, _ZBufferParams

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdvr_mpspi_ppv3&unity_branch=trunk&automation-tools_branch=add-platform-filter

Visually confirmed SPI worked in the HTC Vive headset with OpenVR. 

None yet!

---
### Overall Product Risks
High- added some additional textures, changed how all textures allocate in the RTHandleSystem

High- Touches all texture-allocating managers and refactored a good bit of HDRenderPipeline::Render

---
### Comments to reviewers
This is replacing PR #2461


nerites:
Todo based on previous discussions:
- Re-add legacy UnityInstancing macros for uniformity across legacy and SRP
- Replace #if UNITY_STEREO_INSTANCING_ENABLED code with macros as needed; should be much fewer in this branch
- Correct the PSSL TEXTURE2DMS macros- also re-add macros for multisampled textures? 
- Refactor texture selection to have an overall (typically 0-deep) "VRPass" array-of-structures and iterate over that instead. Near-term refactor would only include textures, but in the future could also include camera matrices, viewports, resolution scaling, etc...

Chman:
I haven't had the time to fully go through all of it, but looking at the render doc capture:
- The color grading LUT is broken, the texture 3D has a depth of 1 (should be the same as width & height to form a cube). I haven't looked at the cause, but it probably has to do with the changes to RTHandles. That's why it looks wrong.
- Dynamic exposure is computed independently for each eye, I don't think it's a good idea. We should probably just pick one of the eye and use that for all the other eyes to avoid weird exposure discrepancies.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ShaderVariablesScreenSpaceLighting.hlsl
```diff
@@ -8,6 +8,7 @@
     TEXTURE2D(_DepthPyramidTexture);
     // Ambient occlusion texture
     TEXTURE2D(_AmbientOcclusionTexture);
```
sebastienlagarde: Why not a texture array here ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -865,14 +866,27 @@ public void SetupGlobalStereoParams(CommandBuffer cmd)
             cmd.SetGlobalVector(HDShaderIDs._TextureWidthScaling, textureWidthScaling);
         }
 
-        public RTHandleSystem.RTHandle GetPreviousFrameRT(int id)
+        public RTHandleSystem.RTHandle GetPreviousFrameRT(int id, int stereoPass = 0)
         {
-            return m_HistoryRTSystem.GetFrameRT(id, 1);
+            int idx = stereoPass;
+            if (id == (int)HDCameraFrameHistoryType.ColorBufferMipChain)
```
sebastienlagarde: do you have more information / commen here ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1075,103 +1093,121 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
...
-                    // Caution: We require sun light here as some skies use the sun light to render, it means that UpdateSkyEnvironment must be called after PrepareLightsForGPU.
-                    // TODO: Try to arrange code so we can trigger this call earlier and use async compute here to run sky convolution during other passes (once we move convolution shader to compute).
-                    UpdateSkyEnvironment(hdCamera, cmd);
+                        if (shouldRenderMotionVectorAfterGBuffer)
+                        {
+                            // See the call RenderObjectsVelocity() above and comment
+                            RenderObjectsVelocity(cullingResults, hdCamera, renderContext, cmd);
+                        }
 
+                        RenderCameraVelocity(cullingResults, hdCamera, renderContext, cmd);
```
sebastienlagarde: UpdateSkyEnvironment(hdCamera, cmd); have disappear ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ShaderVariablesScreenSpaceLighting.hlsl
nerites: To minimize shader changes, we are only dispatching instanced draw calls for passes that traverse scene geometry (so the depth prepass and the forward pass for now). All other passes, including the gpucopy that populates the pyramids, are encapsulated in a per eye for loop instead. This follows the XR Interface implementation paradigm recommended by @mikesfbay

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
nerites: It’s just been moved out of the multipassed eye loop since it is dispatched as an instanced draw call as well

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
nerites: Since we have a separate color pyramid per eye, indexing into the history buffer becomes numEyes * numHistoryFrames + historyFrameIndex. I’ll add a comment noting this

sebastienlagarde:
Hey, so we have done a review with Julien of a big chunk of the PR.

Here is various thought:

- RTHandle system is design to be our low level texture allocation system, we must not do code like stereo:
if (XRGraphics.enabled && requireStereoSupport)
inside the low level, it should be outside. For example we should check value of XRGraphics.eyeTextureDesc outside and provide it to the RTAlloc call, in stereo mode. We have already a "slices" parameters to provide dimension for the texture array etc...

- HDRP is already very complex to understand and maintain, growing complexity is a serious concern. We have already a whole set of macro for texture declaration etc... as discussed in previous PR and after performance test on PS4 we think we should simply use texture2DArray everywhere and not allocate array of texture. I can see in the PR that a lot of memory is allocated, for VR, for MSAA, and it is hard to track it.

- With the approach of doing loop over most of the passes we waste a lot of performance. If I take the example of subsurface scattering, we do a full loop that render HTile, reset all parameters etc... nothing is share compare to using texture array. I get that the PR is very huge and it was simpler to do it like this for now, but it is not good to merge it like that. 

- We still have this issues with compute shader not supporting multicompile. Sadly there is no solution here... If one day we need to make VR work with deferred, we will be doomed, as it will not be possible to make more variant in term of workflow. Nothing we can change for this PR, but to have in mind because currently we can't strip VR variant of compute shader, so they will be compile even on switch...

- The PR is too big as it is (note we are aware that we are guilty about it too :) ), we need to go more step by step, it is hard to track all the change particularly as on our side we continue to develop HDRP (Recently we merge DXR PR...). This is also our fault as we don't review quickly enough...

We have discuss this also with Fabien Houlmann that will continue your work. Here is what we suggests.
We keep this PR as reference. And we try step by step to integrate part of it.
We being with a PR that just convert everything to texture array (no extra macro)
Then disable all features (no distortion, no sss etc...) and enable one by one after conversion of the shader / compute shader to instanced version (no multiple pass except for postprocess maybe)
- only do single pass stereo for HDRP, drop other variant - platform using HDRP should support it (and maybe multipass as it is very simple to implement).

Please @nerites check with @fabien-unity but I guess it could handle the work for you, so you can focus on DXR :)

nerites:
Okay so I've numbered your bullet points so that I can reply to them more clearly...

> Hey, so we have done a review with Julien of a big chunk of the PR.
> 
> Here is various thought:
> 
> 1. RTHandle system is design to be our low level texture allocation system, we must not do code like stereo:
>   if (XRGraphics.enabled && requireStereoSupport)
>   inside the low level, it should be outside. For example we should check value of XRGraphics.eyeTextureDesc outside and provide it to the RTAlloc call, in stereo mode. We have already a "slices" parameters to provide dimension for the texture array etc...
> 2. HDRP is already very complex to understand and maintain, growing complexity is a serious concern. We have already a whole set of macro for texture declaration etc... as discussed in previous PR and after performance test on PS4 we think we should simply use texture2DArray everywhere and not allocate array of texture. I can see in the PR that a lot of memory is allocated, for VR, for MSAA, and it is hard to track it.
> * With the approach of doing loop over most of the passes we waste a lot of performance. If I take the example of subsurface scattering, we do a full loop that render HTile, reset all parameters etc... nothing is share compare to using texture array. I get that the PR is very huge and it was simpler to do it like this for now, but it is not good to merge it like that.
> 3. We still have this issues with compute shader not supporting multicompile. Sadly there is no solution here... If one day we need to make VR work with deferred, we will be doomed, as it will not be possible to make more variant in term of workflow. Nothing we can change for this PR, but to have in mind because currently we can't strip VR variant of compute shader, so they will be compile even on switch...
> 4. The PR is too big as it is (note we are aware that we are guilty about it too :) ), we need to go more step by step, it is hard to track all the change particularly as on our side we continue to develop HDRP (Recently we merge DXR PR...). This is also our fault as we don't review quickly enough...
> 
> We have discuss this also with Fabien Houlmann that will continue your work. Here is what we suggests.
> We keep this PR as reference. And we try step by step to integrate part of it.
> We being with a PR that just convert everything to texture array (no extra macro)
> Then disable all features (no distortion, no sss etc...) and enable one by one after conversion of the shader / compute shader to instanced version (no multiple pass except for postprocess maybe)
> 
> 5. only do single pass stereo for HDRP, drop other variant - platform using HDRP should support it (and maybe multipass as it is very simple to implement).
> 
> Please @nerites check with @fabien-unity but I guess it could handle the work for you, so you can focus on DXR :)

1. The reason the code is inside RTHandles was to try and preserve the flags from the XRGraphics eyeTextureDescriptor. If we can add an Alloc() override which takes in a descriptor, then we can move the XR-specific logic (and looping) into a helper function (probably in XRUtils.cs). 

2. I've made a branch for that, and I can clean it up if that's the way we want to do this, but when I talked with @mikesfbay about it he brought up the point that converting to tex2DArrays isn't going to be compatible with a multi-GPU future. There's a lot of interest in multi-GPU rendering from the XR side, and from the DXR docs I've seen that there's a lot of interest in multi-GPU rendering from the Graphics team side, so setting up a multi-pass, array-of-structures type design is probably the most future-proof, but this PR was sort of a compromise- a refactor for multipass, but without the refactoring that would be needed to treat textures (and matrices/viewports?) as an array of structures. The memory usage of maintaining tex2DArrays that are redundant with multiple tex2Ds is concerning, but is there a way to deallocate intermediate tex2DArrays once their contents are copied out (for consumption by shaders that are expecting tex2Ds)?

3. As I understood it, the main benefit of using a texArray/instancing in general is to share scene geometry data. I thought that for compute passes, the difference between instancing and multipass is negligible... and anyways even in the version of my branch that passed tex2Ds to the compute shaders, the compute shaders were multipassed so that we could set up the correct unity_StereoEyeIndex. 

4. One of the key benefits of this hybrid multipass/instanced PR was that fewer of the compute shaders would need to be modified/have kernels added. With multipass, we can simply dispatch compute shaders twice without adding kernels for using/outputting tex2DArrays. We can't completely evade the problem (as the CameraColor buffer and the DepthPrepass output are both tex2DArrays), but fewer compute shaders had to have kernels added in this branch than in my pure instancing branch.

5. Once again, the incentive for setting up multi pass is to prepare for multi GPU in the future. If this isn't a priority, then I'm fine with it.

Either way, it seems unlikely that I'll be able to get something in by the end of this week.... so for reference, here are my branches:

hdvr_mpspi_ppv3 (this one): Multi-pass/instanced hybrid design, with PPv3 integration
hdvr_mpspi: Multi-pass/instanced hybrid design, PPv2 only
allocfromdefault: Instanced design, PPv2 only
hdvr_spi_texarrays: Instanced design + convert *everything to* texarrays. Evgenii should push his changes to this branch so that it builds for PS4

mikesfbay:
@sebastienlagarde @fabien-unity @nerites We discussed these points with XR team earlier: 

- Most HDRP shaders are pixel bound and don't benefit from much CPU geometry instancing. Doing separate dispatch compute per eye has same performance for pixel bound scenario. This avoids changes to compute shaders.

- As example of this, for postprocessing we have implemented the per eye pass approach and showed how all effects and shaders can remain unmodified by XR and stay performant. It also simplifies future code maintenance.  

- Current per-eye hybrid multipass implementation isn't perfect. Currently, we do instanced draw for z-prepass and opaque. We iterate per eye pass for all compute shaders before opaque pass. This allowed compute shaders such as ambient occlusion MSVO to work without XR modifications. There's a blit from z-prepass texarray to 2D depth texture, which we can get rid of later (note, this still cheaper than traversing scene twice). Subsurface scattering function can be optimized by avoid function call per eye with compute dispatch per eye. 

- Keeping per eye pass for compute shaders allow dispatch to multiple gpu's. TexArray resource is bound to one GPU only and can't be shared among gpus.

Let's keep this thread open as we discuss these issues.

sebastienlagarde:
We have merged another PR that handle SPI.

PR #2668, opened 2018-12-21
Make the light delegate function in LightweightRenderPipeline private

egomeh:
### Purpose of this PR

Makes the light delegate function in LWRP private. Having the function public is not necessary.

---
### Release Notes

The lightmap delegate is no longer public in LWRP.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Flights-delegate-private&unity_branch=trunk

**Manual Tests**: I have run a test scene that uses baked lights without issues.

---
### Overall Product Risks
**Technical Risk**: Very low

**Halo Effect**: Low. If someone already depends on delegate being public, this might cause issues.


egomeh:
Yes, @phi-lira, I rebaked the lighting in one of the test scenes.

phi-lira:
Katana is green.

PR #2671, opened 2018-12-21
Motion blur wip

FrancescoC-unity:
This is oh so WIP, but it has a base line that could be tested already. However keep in mind: 

 - Needs further optimizations
 - Will probably need fixes here and there. 
 - Might change as problems are found. 
 - Contains *a lot* of temp code and hacks and TODOs

Essentially I wasn't expecting to merge this in this branch now so it is not that ready, but still we can do a first pass test. 

I've put all the settings under advanced, set the effect off by default and added a warning to specify it is still a WIP and should not be used in production.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Builtin/BuiltinData.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Builtin/BuiltinData.hlsl
@@ -42,7 +42,7 @@ void EncodeVelocity(float2 velocity, out float4 outBuffer)
 
 void DecodeVelocity(float4 inBuffer, out float2 velocity)
 {
-    velocity = inBuffer.xy;
+    velocity = inBuffer.x > 1.0f ? 0.0f : inBuffer.xy;
 }
 
 void EncodeDistortion(float2 distortion, float distortionBlur, bool isValidSource, out float4 outBuffer)
--- a/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/MotionBlur.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/MotionBlur.cs
@@ -1,0 +1,32 @@
+using System;
+using UnityEngine.Rendering;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+    public enum MotionBlurMode
+    {
+        Off,
+        UsePhysicalCamera,
+        Manual
+    }
+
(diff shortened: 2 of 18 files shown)


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/MotionBlur.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    [Serializable, VolumeComponentMenu("Post-processing/Motion Blur")]
+    public sealed class MotionBlur : VolumeComponent, IPostProcessComponent
+    {
+        public MinIntParameter sampleCount = new MinIntParameter(8, 2);
+
+
+        // Physical settings
+        public MinFloatParameter intensity = new MinFloatParameter(1.0f, 0.0f);
```
Chman: Should be 0.0 by default.

com.unity.render-pipelines.high-definition/Runtime/Material/Builtin/BuiltinData.hlsl
```diff
@@ -42,7 +42,7 @@ void EncodeVelocity(float2 velocity, out float4 outBuffer)
 
 void DecodeVelocity(float4 inBuffer, out float2 velocity)
 {
-    velocity = inBuffer.xy;
+    velocity = inBuffer.x > 1.0f ? 0.0f : inBuffer.xy;
```
sebastienlagarde: what is this ?

sebastienlagarde: ok, I see the comment, 

  // If forceNoMotion, we output an implausible velocity (>1) that will be taken care of at decoding time.
     // This is to explicitly flag this case for cases in which we manually recompute camera velocity (e.g. Motion Blur)

Then there is a typo here, it is inBuffer.xy, please fix a,d add the comment

sebastienlagarde: Also it assume that the texture format is float. I remember unreal moving to u16 as it was better.

FrancescoC-unity: Fixed up (I didn't use the y since if *any* coordinate was >1 it meant it was because of my flagging, so no need to do comparison on two channels)

JakubSmaga:
com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Line 206 isn't aligned with the rest.

PR #2676, opened 2018-12-21
Hdrp decal shader graph 3

Paulm-Unity:
### Purpose of this PR
- Add Master node for decal shader graph
- Update UI of decal
- Remove Albedo scale opacity

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp-decal-shader-graph-3&automation-tools_branch=add-platform-filter&unity_branch=trunk#


com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalPass.template
```diff
@@ -0,0 +1,163 @@
...
+        SurfaceDescription surfaceDescription = SurfaceDescriptionFunction(surfaceDescriptionInputs);
+
+        // setup defaults -- these are used if the graph doesn't output a value
+        ZERO_INITIALIZE(DecalSurfaceData, surfaceData);
+
+        // copy across graph values, if defined
+        $SurfaceDescription.Albedo:                     surfaceData.baseColor.xyz = surfaceDescription.Albedo;
+#if (SHADERPASS == SHADERPASS_DBUFFER_PROJECTOR)
+        $SurfaceDescription.Normal:                     float4x4 normalToWorld = UNITY_ACCESS_INSTANCED_PROP(matrix, _NormalToWorld);
+        $SurfaceDescription.Normal:                     surfaceData.normalWS.xyz = mul((float3x3)normalToWorld, surfaceDescription.Normal);
```
sebastienlagarde: #if (SHADERPASS == SHADERPASS_DBUFFER_PROJECTOR)
         $SurfaceDescription.Normal:                     float4x4 normalToWorld = UNITY_ACCESS_INSTANCED_PROP(matrix, _NormalToWorld);
         $SurfaceDescription.Normal:                     surfaceData.normalWS.xyz = mul((float3x3)normalToWorld, surfaceDescription.Normal);

For this PR let's do it like that, but have in mind that Normal is suppose to be in world space for master node, not tangent space  (Particularly when we use decals mesh, we can do what we want). To let the freedom to the users, we will need to add a specific path in the transform node to handle decals case. 

Let's postpone that in a next PR, but to not forget :)

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSubShader.cs
```diff
@@ -0,0 +1,230 @@
...
+            ShaderPassName = "SHADERPASS_DBUFFER_PROJECTOR",
+
+            CullOverride = "Cull Front",
+            ZTestOverride = "ZTest Greater",
+            ZWriteOverride = "ZWrite Off",
+            BlendOverride = "Blend 0 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 1 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 2 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 3 Zero OneMinusSrcColor",
+
+            Includes = new List<string>()
+            {
+                "#include \"Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalProperties.hlsl\"",
```
sebastienlagarde: you shouldn't require DecalProperties.hlsl, is there a reasons to include it or it is a mistake ?

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSubShader.cs
```diff
@@ -0,0 +1,230 @@
...
+            MaterialName = "Decal",
+            ShaderPassName = "SHADERPASS_DBUFFER_MESH",
+
+            ZTestOverride = "ZTest LEqual",
+            ZWriteOverride = "ZWrite Off",
+            BlendOverride = "Blend 0 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 1 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 2 SrcAlpha OneMinusSrcAlpha, Zero OneMinusSrcAlpha Blend 3 Zero OneMinusSrcColor",
+
+            Includes = new List<string>()
+            {
+                "#include \"Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalProperties.hlsl\"",
```
sebastienlagarde: same remark here: DecalProperties.hlsl

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalSystem.cs
```diff
@@ -541,6 +547,24 @@ public int DrawOrder
...
+            {
+                get
+                {
+                    if (m_IsHDRenderPipelineDecal)
+                    {
+                        return (int)this.m_Material.GetFloat("_MaskBlendMode");
+                    }
+                    else
+                    {
+                        // TODO
```
sebastienlagarde: is this a TODO?  I guess returning 0 is fine here no ? Maybe add a comment about why, explain there is a master node with only the correct pass genreated

sebastienlagarde:
Great job!

Do you still have test failing due to normal encoding ?

Paulm-Unity:
> Great job!
> 
> Do you still have test failing due to normal encoding ?
> 
> Add a katana

I started Katana 30 mins ago, so I will wait and see what happens.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSubShader.cs
Paulm-Unity: ShaderPassDBuffer.hlsl is shared between graph and non-graph version and needs access to _DecalMeshDepthBias, which is in DecalProperties... however depth bias is not even working for decal graph, so maybe add a #define DECAL_SHADERGRAPH and exclude it based on that?

sebastienlagarde:
@Paulm-Unity I have updated the PR to fix the katana test + update UI and expose DecalMeshBias.

Here is the list of task remaining to do in another PR:

- There is one TODO left in public int MaskBlendMode, not sure what was the intent
- Add graphic test for decal graph
- Make DrawOrder and DecalMesh Bias working (I haven't tested)
- Have better unit for DecalMesh Bias ? Currently we apply "input.vmesh.positionCS.z -= _DecalMeshDepthBias * input.vmesh.positionCS.w;" This mean bias may vary with distance. Is it intented ? do we want this.
We need to provide help for what typical value to put here, it is hard to know
- Metal, AmbientOcclusion and Smoothness Opacity must be separated in master node
It mean we have 5 opacity, thus why initally I was thinking about having two mode: one with a single opacity and one with the 5 opacity in the Settings, so advanced users can switch to it if they want, but by default regular users aren't afraid. What do you think ? Ask Rinaldo maybe?
- Have a Affect Base Color and Affect Normal. There is now no way to know what is the users intent, so we need to be explicit everywhere

- Need to move UNITY_ACCESS_INSTANCED_PROP(matrix, _NormalToWorld) in transform node. Here we have changed the convention of shader graph from world space normal to decal space normal, which isn't good for decal mesh.

- Note: I have remove Alpha scale, we already have opacity for base color, don't want to overload the UI)

sebastienlagarde:
Katana is green (once VFX graph test fail (timeline) but unrelated to this PR, instability on master)

Paulm-Unity:
> * Need to move UNITY_ACCESS_INSTANCED_PROP(matrix, _NormalToWorld) in transform node. Here we have changed the convention of shader graph from world space normal to decal space normal, which isn't good for decal mesh.

I know you mentioned this before, but I still do not understand this part. In DecalPass.template we have

#if (SHADERPASS == SHADERPASS_DBUFFER_PROJECTOR)
        $SurfaceDescription.Normal:                     float4x4 normalToWorld = UNITY_ACCESS_INSTANCED_PROP(matrix, _NormalToWorld);
        $SurfaceDescription.Normal:                     surfaceData.normalWS.xyz = mul((float3x3)normalToWorld, surfaceDescription.Normal);
#elif (SHADERPASS == SHADERPASS_DBUFFER_MESH)	
        // We need to normalize as we use mikkt tangent space and this is expected (tangent space is not normalize)
        $SurfaceDescription.Normal:                     surfaceData.normalWS.xyz = normalize(TransformTangentToWorld(surfaceDescription.Normal, fragInputs.worldToTangent));
#endif

So for meshes it also gets transformed into correct space. To test this, I put projector and mesh decals using the same material next to each other in a test scene and they look identical.

sebastienlagarde:
ok, I am sorry I say incorrect thing. I was thinking shader graph was taking normal in world space as input thus my question. So it is ok for this stuff.

However if a users want to make normal map manipulation in the shader graph between different space, then it may not work. but don't think this will happen in practice. Let's do it like that and we will see with usage. Sorry for the noise.

sebastienlagarde:
ok, I am sorry I say incorrect thing. I was thinking shader graph was taking normal in world space as input thus my question. So it is ok for this stuff.

However if a users want to make normal map manipulation in the shader graph between different space, then it may not work. but don't think this will happen in practice. Let's do it like that and we will see with usage. Sorry for the noise.

PR #2679, opened 2018-12-21
Fix Blackboard being restored to invalid positions

Antoine-Lassauzay:
### Purpose of this PR

Solving issues with the Shader Graph property going missing (fogbugz 1111305, fogbugz 1102285)

This was blocking me for reproducing another issue and it seemed easy enough to fix.

---
### Release Notes

Fixed the property blackboard going missing or very small.
---
### Testing status

**Manual Tests**: Verified this restored the blackboard to 10,10 when creating a project and that the blackboard position is correctly shared for all graph windows.
Also tested the behaviour or anchoring to right/left when resizing the window.

**Automated Tests**: Nothing

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Low

---


Antoine-Lassauzay:
@zadi0648 I can't see reproduce any issue with those steps unfortunately. If the steps are different, we should maybe log a fogbugz and look into separately.

zadi0648:
So in addition to the above mentioned bug, https://fogbugz.unity3d.com/f/cases/1102285/ does actually still seem to be reproducing for me. I'm using this branch with 2019.2.0a2 (d9648fe5f6b6) & HD. (no LW)

What version of Unity are you using? Are you using LW or HD?

zadi0648:
Blackboard goes missing during resizing of window
blackboardmissing
Creates the following error:
Pixels per point should be positive

jmaurice-unity:
@zadi0648 I was unable to repro this on latest master and Unity trunk branch.
Any more info would be greatly appreciated.

It might be related to the serialized window layout that is saved in ShaderGraph. Maybe it was corrupt in this case?! I'm not sure how I can get in that state...

zadi0648:
@jmaurice-unity
Still reproducing for me instantly. I'm using a brand new project, windows 10, 19.2.0a5 (20c8c112f207), sg/fix-blackboard.
Seems to only reproduce when creating a brand new project.

racarone:
I've implemented a similar fix in my project. The blackboard's layout eventually always goes NaN for me with master in 2019.1 & 2019.2 without a check.

zadi0648:
I updated https://fogbugz.unity3d.com/f/cases/1102285/'s steps to hopefully make this more consistent for people.

sebastienlagarde:
Sorry to poke everyone but can we please move forward with this PR and fix the issues. 
This is a critical part of the shader graph and this PR is staging for around 2 month. Think about the users...
@Kink3d @russoh @marctem @zadi0648 @cdxntchou . Even if it doesn't fix all the issue it fix the main one and allow artists to work. So let's merge it and solve other issue in another PR.

Thanks

zadi0648:
What if you check INF in addition to checking NaN ?

pbbastian:
Closed in favour of #2975

PR #2694, opened 2019-01-02
Lw/expose renderersetup in camera

phi-lira:
### Purpose of this PR
- Exposes custom renderers to LWRP. You can now choose in the pipeline a renderer type. Options at the moment are Forward Renderer and Custom. 
- Users can create a custom renderer by implementing a IRendererSetup and IRendererData. They can use that renderer for all cameras by setting the renderer to custom in the pipeline asset or overriding for a specific camera by settings to custom renderer in the Camera inspector.
pipelineasset
 
camerasettings

- All resources used by the ForwardRenderer are now stored in ForwardRendererData
---
### Release Notes

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fexpose-renderersetup-in-camera&unity_branch=trunk

**Manual Tests**: Tested myself with Graphics Tests scenes.

**Automated Tests**: Created custom renderer for test 045. Added editor tests to validate a new asset is created with a valid renderer.

---
### Overall Product Risks
**Technical Risk**: Medium. Changes how rendering data is serialized and makes it available for users to create custom renderers. 

**Halo Effect**: Medium. Should affect all rendering platforms.

---
### Comments to reviewers

IRendererData contains resources for the renderer and it's a Factory that creates renderers. 
IRendererSetup contains the logic to enqueue render passes. Each render pass now receives resources from the IRendererData. 

The renderer instance is stored in the pipeline asset. In the case of overriding a camera the renderer instanced is stored in the camera additional data. 

I've updated test 045 to use this new renderer. Also added Editor tests to validate a new pipeline asset contains valid references to editor resources and the renderer data. When creating a new pipeline asset the Forward renderer will be used by default.


com.unity.render-pipelines.lightweight/Runtime/Data/LightweightRenderPipelineAsset.cs
```diff
@@ -69,6 +69,12 @@ public enum ShaderVariantLogLevel
         AllShaders,
     }
 
+    public enum RendererType
```
rainsing: As this is used for serialization and we'll add more types of renderer (and conventionally 'Custom' should stay the last in the list), I think we should use explicit enum values here.

phi-lira: Good catch. I'll look into it.

aleks01010101: Why not make Custom the first one?

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererData.cs
```diff
@@ -0,0 +1,40 @@
+namespace UnityEngine.Experimental.Rendering.LWRP
+{
+    //[CreateAssetMenu()]
```
aleks01010101: Is this intentional?

com.unity.render-pipelines.lightweight/Runtime/LWRPAdditionalCameraData.cs
```diff
@@ -9,6 +10,12 @@ public enum CameraOverrideOption
         UsePipelineSettings,
     }
 
+    public enum RendererOverrideOption
+    {
+        Custom,
```
aleks01010101: Do I get this correct that selecting Custom here allows one to select the renderer on a per camera basis and this is not "I want a custom renderer regardless of what the pipeline settings say"?

com.unity.testing.srp.lightweight/Tests/Scenes/052_LWCallbacks/CameraCallbackTests.cs
```diff
@@ -96,51 +102,51 @@ class BlitPass : ScriptableRenderPass
 	{
         private RenderTargetHandle colorHandle;
         private RenderTargetHandle depthHandle;
+	    Material m_BlitMaterial;
```
aleks01010101: private Material?

aleks01010101: I mean, `private` is missing :)

com.unity.testing.srp.lightweight/Tests/Scenes/052_LWCallbacks/CameraCallbackTests.cs
```diff
@@ -20,8 +20,14 @@ public class CameraCallbackTests : MonoBehaviour
 	static RenderTargetHandle afterTransparent;
 	static RenderTargetHandle afterAll;
 
+    Material m_CopyDepthMaterial;
```
aleks01010101: is `private` missing?

com.unity.render-pipelines.lightweight/Runtime/LWRPAdditionalCameraData.cs
phi-lira: The pipeline seetings allow to select a default renderer to be used for all cameras. That can also be a custom renderer. This settings allows to override per-camera the renderer used for it.

com.unity.testing.srp.lightweight/Tests/Scenes/052_LWCallbacks/CameraCallbackTests.cs
phi-lira: private keyword is redundant and according to our guidelines should be ommited.

phi-lira:
Katana is passing on LWRP. With exception of Terrain test which should be fixed after merging master to it.

PR #2696, opened 2019-01-03
Fixed darker rendering when in gamma colorspace.

phi-lira:
### Purpose of this PR
 When in gamma color space objects were rendering darker. This was caused because when creating render texture sRGB flag was always being set. 

 This PR fixes it.

---
### Release Notes
 Added

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fbugfix%2Fdarker-rendering-ingammaspace&unity_branch=trunk

**Manual Tests**: Tested with scene rendering skybox and a quad.

**Automated Tests**: No automated tests added. We don't have yet support for gamma space tests.

---
### Overall Product Risks
**Technical Risk**: None. Single line change.

**Halo Effect**: Low. Affects all gamma space rendering but the change is safe.

---
### Comments to reviewers
 Checks active colorspace when creating the render texture.


phi-lira:
VFX Graph Test is failing. That's already fixed in master.

prodigga:
Am I allowed to comment on here..? This problem is visible in 5.2.3 LWRP, using Unity 2019.1b2. I can see that in ScriptableRenderer.cs, the function reads:

`        public static RenderTextureDescriptor CreateRenderTextureDescriptor(ref CameraData cameraData, float scaler = 1.0f)
        {
            RenderTextureDescriptor desc;
            float renderScale = cameraData.renderScale;

            if (cameraData.isStereoEnabled)
            {
                return XRGraphics.eyeTextureDesc;
            }
            else
            {
                desc = new RenderTextureDescriptor(camera.pixelWidth, camera.pixelHeight);
            }
            desc.colorFormat = cameraData.isHdrEnabled ? RenderTextureFormat.DefaultHDR :
                RenderTextureFormat.Default;
            desc.enableRandomWrite = false;
            desc.sRGB = true;
            desc.width = (int)((float)desc.width * renderScale * scaler);
            desc.height = (int)((float)desc.height * renderScale * scaler);
            return desc;
        }`

desc.sRGB is just being set to true. I think it is supposed to be `desc.sRGB = (QualitySettings.activeColorSpace == ColorSpace.Linear);`

prodigga:
capture

phi-lira:
This PR fix the issue by exactly doing that. 😁
This is already available in package 5.3.0

prodigga:
Interesting, I downloaded the Unity 2019.1b2, and selected the latest version available to me which is 5.2.x. Are we not supposed to have access to 5.3.0 yet? Sorry for the confusion!
image

PR #2699, opened 2019-01-03
Lw/dynamic resolution [DONT MERGE IT YET]

davidrogers-unity:
Requires some changes on the engine side as well. https://ono.unity3d.com/unity/unity/pull-request/79744/_/graphics/dynamic-resolution/lwrp

Also en route to 2019.1/staging.

https://ono.unity3d.com/unity/unity/pull-request/80294/_/2019.1/graphics/dynamic-resolution/lwrp

Purpose of this PR

Adding support for enabling dynamic resolution scaling to the lightweight render pipeline. More or less we need to be able to set the flag for enabling from scripting (engine changes) and then set it if the user has checked the box(which also gets added in this PR).

Release Notes

Dynamic Resolution Scaling is now compatible with LWRP. It may be used in exactly the same way as with the non scriptable rendering pipeline.

Testing status

Katana Tests: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fdynamic-resolution&unity_branch=2019.1%2Fstaging

Manual Tests: Testing with some projects that use LWRP, enabled DRS on the main camera and attached a script for controlling the scale.

Yes, I have a couple that I have prepared.

Overall Product Risks

Technical Risk: None

Halo Effect: Medium


davidrogers-unity:
Not rerunning tests on katana for the conflict. It was only in the changelog.

PR #2700, opened 2019-01-04
Added Support for multi viewport

JulienIgnace-Unity:
### Purpose of this PR
Added Support for multi viewport

---
### Release Notes
Added Support for multi viewport

---
### Testing status
Manual and Katana test run locally.
Currently test 3001_DebugView fails (image flipped)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -117,6 +118,8 @@ public Vector4 invProjParam
             }
         }
 
+        public bool isMainGameView { get { return camera.cameraType == CameraType.Game && camera.targetTexture == null; } }
```
Chman: Could be simplified, but not really important.
```
public bool isMainGameView => camera.cameraType == CameraType.Game && camera.targetTexture == null;
```

PR #2708, opened 2019-01-07
Added version define for LWRP.

phi-lira:
### Purpose of this PR

Adds define LWRP_5_3_0_OR_NEWER for both scripting runtime and shader.
 This ways users can check LWRP version. 

I've used the above format as it's also used by builtin Unity. https://docs.unity3d.com/Manual/PlatformDependentCompilation.html
 
 Started some discussion about it in slack.

---
### Release Notes
 Added

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fadd-version-define&unity_branch=2019.1%2Fstaging

**Manual Tests**: Tested defines myself.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---


com.unity.render-pipelines.lightweight/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - LWRP now uses the Unity recorder extension. You can use this to capture the output of Cameras.
 - You can now inject a custom render pass before LWRP renders opaque objects. To do so, implement an `IBeforeRender` interface.
+- Added version defines for both C# and shader in the format LWRP_X_Y_Z_OR_NEWER. F.ex 5.3.0 version defines LWRP_5_3_0_OR_NEWER.
```
JordanL8: LWRP now includes version defines for both C# and Shaders in the format of `LWRP_X_Y_Z_OR_NEWER`. For example, `LWRP_5_3_0_OR_NEWER` defines version 5.3.0.

phi-lira: updated.

RamiroOliva:
This is very useful however I can't find this macro define in latest versions of LWRP or URP.
Could you please explain or document how to use these macros?
LWRP_X_Y_Z_OR_NEWER for example, is not declared it seems.

seblagarde:
Hi, in latest version the macro have moved to Core package and are share between core , HD and UP.
And they have been updated to be easier to used:

#define SHADER_LIBRARY_VERSION_MAJOR 7
#define SHADER_LIBRARY_VERSION_MINOR 2

#define VERSION_GREATER_EQUAL(major, minor) ((SHADER_LIBRARY_VERSION_MAJOR > major) || ((SHADER_LIBRARY_VERSION_MAJOR == major) && (SHADER_LIBRARY_VERSION_MINOR >= minor)))
#define VERSION_LOWER(major, minor) ((SHADER_LIBRARY_VERSION_MAJOR < major) || ((SHADER_LIBRARY_VERSION_MAJOR == major) && (SHADER_LIBRARY_VERSION_MINOR < minor)))
#define VERSION_EQUAL(major, minor) ((SHADER_LIBRARY_VERSION_MAJOR == major) && (SHADER_LIBRARY_VERSION_MINOR == minor))

Guess it should be self explanatory now.

RamiroOliva:
Is this supposed to work in C#? I'm using Unity 2019.3 with latest URP package (on Mac) and the macros are not recognized.

seblagarde:
Oh those macro above are for hlsl, was thinking you were talking about shaders, sorry. Don't know if @phi-lira  have keep C# macro up-to-date

RamiroOliva:
Yeah, C# macro so custom ScriptableRenderPass classes can remain disabled until the correct package is imported. Otherwise, we need to move files out and inside the project to avoid compilation errors.

RamiroOliva:
And thanks for the prompt reply.
While I have you, one challenging question.
Is it possible to render a mesh to two different render targets from a ScriptableRenderPass?
Imagine I want to render an object to an off-screen target but at the same time I want to reuse the camera depth buffer so it gets clipped properly (using Configure or harsh SetRenderTarget).
It seems that the renderer.cameraDepth is always camera target but still it doesn't clip.

PR #2710, opened 2019-01-07
Lightweight Terrain shader feature upgrades

parashar-unity:
### Purpose of this PR
The goal of this PR is to bring a few additional features to the terrain shader in lightweight.  Specifically, it is focusing on the splat blending functionality to bring the lightweight terrain closer to parity with HD terrain.  This means adding the remap and tinting functionality as well as height-based blend, and density-based blend.  Also, the controls and UI are now basically the same as the HD version as well.  This also means that the same terrain layer assets can work more or less the same in both LWRP as well as HDRP (albeit with some limitations -- e.g. limited to 4 splats in single-pass in LW compared to 8 in HD).

This set of changes brings forth per-layer backwards compatibility between HDRP-style mask-based blending and the legacy styles of blending.
As a result, it fixes the following issues -
https://fogbugz.unity3d.com/f/cases/1110520/
https://fogbugz.unity3d.com/f/cases/1161116/    (this one is labeled HDRP, but also occurs in LW.  A separate PR for the HDRP fix is also in flight)
https://fogbugz.unity3d.com/f/cases/1150934/    (again, labeled HDRP, but first verified while working on this PR)

---
### Release Notes
Biggest file additions include the addition of a custom UI for the terrain shader/splat controls as well as the addition of a basemap generation shader (for bottom LOD).  The custom UI is more or less identical to the HDRP one with a few unnecessary bits taken out.

---
### Testing status
**Katana Tests**: Katana all green.
**Manual Tests**: Built small sample terrain test scene utilizing all the features in question.  Also did tests ensuring that converting the project from lightweight to HD meant that the same features still worked.

---
### Overall Product Risks
**Technical Risk**: Medium --  Is a relatively major change to the terrain shader in lightweight rendering pipeline.
Due to the fact that there are some differences in what some layer map channels meant (e.g. there was a different source for smoothness coming from the diffuse map rather than mask) means that there could be some assets that relied on old behaviors.
**Halo Effect**: Medium-- affects only terrain shader and not much else.  Does mean that there is more parity between lightweight terrain and HD terrain, and lightweight's terrain shader is no longer a copy of the built-in, but both are still backwards compatible since they are essentially a superset of the features.

---
### Comments to reviewers
Add Pass (multi-pass in general) cannot properly normalize or work properly with height-based or density-based blending because information about the layers beneath can't be known in the second pass.  There are tooltips and notes in the UI now to inform the user of these issues.  Some approaches are also written mindful of the fact that later PRs are going to account for high-quality mode support for single-pass 8-layer.

Two usability bugs mentioned below (see Chris' comment) are dependent on a corresponding trunk change.  These amount to 2 lines change, but it does have to happen in core.  This would come out as a trunk change with a 2019.1 backport.


com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitAdd.shader
```diff
@@ -61,7 +65,9 @@ Shader "Hidden/Lightweight Render Pipeline/Terrain/Lit (Add Pass)"
             #pragma multi_compile_instancing
             #pragma instancing_options assumeuniformscaling nomatrices nolightprobe nolightmap
 
+            #pragma shader_feature _TERRAIN_BLEND_HEIGHT
```
phi-lira: Shouldn't TerrainLitAdd.shader be removed now?

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -105,26 +121,50 @@ void SplatmapMix(VertexOutput IN, half4 defaultAlpha, out half4 splatControl, ou
...
-    mixedDiffuse += SAMPLE_TEXTURE2D(_Splat0, sampler_Splat0, IN.uvSplat01.xy) * half4(splatControl.rrr, alpha.r);
-    mixedDiffuse += SAMPLE_TEXTURE2D(_Splat1, sampler_Splat0, IN.uvSplat01.zw) * half4(splatControl.ggg, alpha.g);
-    mixedDiffuse += SAMPLE_TEXTURE2D(_Splat2, sampler_Splat0, IN.uvSplat23.xy) * half4(splatControl.bbb, alpha.b);
-    mixedDiffuse += SAMPLE_TEXTURE2D(_Splat3, sampler_Splat0, IN.uvSplat23.zw) * half4(splatControl.aaa, alpha.a);
+    mixedDiffuse += diffAlbedo[0] * half4(_DiffuseRemapScale0.rgb * splatControl.rrr, 1.0h);
+    mixedDiffuse += diffAlbedo[1] * half4(_DiffuseRemapScale1.rgb * splatControl.ggg, 1.0h);
+    mixedDiffuse += diffAlbedo[2] * half4(_DiffuseRemapScale2.rgb * splatControl.bbb, 1.0h);
+    mixedDiffuse += diffAlbedo[3] * half4(_DiffuseRemapScale3.rgb * splatControl.aaa, 1.0h);
 
 #ifdef _NORMALMAP
```
phi-lira: Is this function used in BasemapGen? If so, this code path will never be reached in that shader.

parashar-unity: The function itself is used in BasemapGen, and all we really care about there is how it affects diffuse color and the splat weights (which could be affected by heights).  The same code is used in the regular render pass, where it isn't unreachable, of course.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitAdd.shader
parashar-unity: Done.

parashar-unity:
Performance-wise, I can't say there has been any real measurement.  I do think it's a bit difficult to make a good apples-to-apples comparison because, simply put, there is now more that is supported in the shader than there previously was (for instance, the fact that we can add in baked AO now, which we couldn't before).  Given that quite a bit is wrapped inside #ifdefs and so on, there's a good chance that there is very little impact when you stay within the bounds of lesser features.

If you start going into the more feature-rich stuff and add in more maps to the layer data, I'd expect performance to degrade as a result of all those additional texture reads.  There's relatively tiny additional ALU cost, all else being equal, and I'm not aware of us being ALU-bound anyway, so I'd not expect significant impact from that.

Having said that, I am fully prepared to learn to savor the taste of the words I will soon be eating.

phi-lira:
Adding @Verasl for review on the impact of limiting terrain to 4 layers.

Verasl:
Is the limit of 4 layers, 4 splatmaps(4xrgba splatmaps) or 4 terrain layers(1xrgba splatmap)? because if it is 4 terrain layers this is a massive feature pullback as currently it supports many more.

phi-lira:
Let's add SH sampling in a separate PR.

phi-lira:
We need to somehow limit or inform users that there's a layer limit. 

Also should we consider a shader_feature keyword to add support to either 4 or 8 layers?

parashar-unity:
Well,

Support for more layers in lightweight was always just done with multi-pass
(hence the add pass shader), so I don't think there is a need for an
additional shader keyword.  The reason eliminating even came into
discussion was because the additonal blending complexity essentially means
it is impossible to properly normalize the mixing coefficients across the
two passes because you'd potentially need to read from all the layer splats
in order to solve the normalization.

You could still support it, but it is pretty well guaranteed to look wrong,
and artists would have to do some kind of compensation.

- Parashar K.

phi-lira:
I think we all agree that the multipass strategy has problems. On the other hand some in production demos for lwrp use more than 4 layers. That's why I am asking if we should support more than 4 layers. 

A shader keyword would allow 4 or 8 in a single pass.
@verasl also has some other ideas.

parashar-unity:
Okay.  I was assuming that lwrp was targeting a minspec that would make
single-pass 8 layers (like hdrp does) impossible.  Hence, multipass.
Seeing as how 8 layers could require up to 26 textures to sample in a
single pass, this made sense, too.  But if that is actually possible, that
pretty much would make it a pretty direct copy of the HD version.

parashar-unity:
Okay.  For now, I've restored the presence of the Add Pass, so at least we have multi-pass, and assuming that existing demos were using this for more than 4 layers, the "wrongness" should be consistent.  I'm assuming that those things didn't use features like height-based or density-based blends which would foul up the equation further (they couldn't have, after all, seeing as how those features didn't exist in lwrp until now).

phi-lira:
That's fine for now. However, in a future PR, I'd prefer to remove the AddPass and have the shader_feature that does either 4 or 8 layers in the same pass. 

Consider LWRP supports all platforms and this is how we scale rendering up if people are using LWRP with PC/Consoles that support the amount of texture for 8 layers.

cdxntchou:
Given LW's target platforms, it may even makes sense to have options for 2, 3, 4, 6, 8 materials...

cdxntchou:
And options to disable normal maps and spec maps, if they are targetting very low end platforms..

cdxntchou:
I was looking at a bunch of mobile games the last few days -- as far as I
can tell, most of them use 2-3 materials on their terrains, and no normal
maps...

phi-lira:
I agree. I'm just saying we have multiple ways to approach scaling quality, throught the LWRP shader quality defines I shared above or shader keywords.  

So, the question is should we do the layer/normal maps changes per platform in this PR or a following one? I'd be totally ok for a next PR to do it.

cdxntchou:
> Instancing seems to be broken, none of the tiles are being batched together, this was working at one point so not sure what is going on.

Terrain is instanced or not depending on the "Enable Instancing" checkbox -- pretty sure it doesn't use dynamic batching either way.

What is the issue you are seeing?

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
```diff
@@ -0,0 +1,286 @@
...
+            }
+            if (EditorGUI.EndChangeCheck())
+            {
+                optionsChanged = true;
+            }
+
+            bool enablePerPixelNormalChanged = false;
+            
+            // NB RenderQueue editor is not shown on purpose: we want to override it based on blend mode
+            materialEditorIn.EnableInstancingField();
```
Verasl: The terrain system seems to ignore the instancing property so it should be removed

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
```diff
@@ -0,0 +1,286 @@
...
+            if (materialEditorIn.IsInstancingEnabled() && enableInstancedPerPixelNormal != null)
```
Verasl: Since the terrain engine ignores the material editors GPUs instancing field it would be best to check if the `INSTANCING_ON` keyword is currently set, as this is set via the terrain component. Though this might not be ideal as since it lives in the material GUI it might never be enabled unless checked on the component then the material is viewed in the inspector.

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
parashar-unity: Done.

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
parashar-unity: For now, I've just included the control without any additional check (in the most recent push).  The shader itself anyway checks the `INSTANCING_ON` keyword before applying anything related to that field, so having the GUI element one way or another doesn't really cause problems.

Hopefully, that means the requested change can be cleared and this can go in.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitBasemapGen.shader
```diff
@@ -0,0 +1,208 @@
...
+        #define _METALLICSPECGLOSSMAP 1
+        #define _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A 1
+
+        #pragma shader_feature _TERRAIN_BLEND_HEIGHT
+        #pragma shader_feature _MASKMAP
+        
+        #include "TerrainLitInput.hlsl"
+        #include "TerrainLitPasses.hlsl"
+        
+        struct Attributes
```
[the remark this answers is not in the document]
parashar-unity: Yep... this one looks like the result of some broken merge conflict resolution.  In earlier versions, that stuff was defined in place, but that changed in the merged versions... I don't know why I got no errors on that on Windows.  Anyway, I'm fixing that along with a few others and the next update will have that.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLit.shader
```diff
@@ -66,7 +72,9 @@ Shader "Lightweight Render Pipeline/Terrain/Lit"
             #pragma multi_compile_instancing
             #pragma instancing_options assumeuniformscaling nomatrices nolightprobe nolightmap
 
+            #pragma shader_feature _TERRAIN_BLEND_HEIGHT
```
zeroyao: Use shader_feature_local.

zeroyao: How does density / height blending work with multi pass? Or do we not use them any more?

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitBasemapGen.shader
```diff
@@ -0,0 +1,194 @@
...
+            #else
+                masks[0] = half4(1.0h, 1.0h, 0.0h, mixedDiffuse.a);
+                masks[1] = half4(1.0h, 1.0h, 0.0h, mixedDiffuse.a);
+                masks[2] = half4(1.0h, 1.0h, 0.0h, mixedDiffuse.a);
+                masks[3] = half4(1.0h, 1.0h, 0.0h, mixedDiffuse.a);
+            #endif
+            
+                SplatmapMix(IN.uvMainAndLM, IN.uvSplat01, IN.uvSplat23, splatControl, weight, mixedDiffuse, normalTS);
+            
+                half4 defaultSmoothness = half4(_Smoothness0, _Smoothness1, _Smoothness2, _Smoothness3);
```
zeroyao: IIRC if _MASKMAP is not defined smoothness is taken from albedo.a?

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLit.shader
parashar-unity: Pretty much all forms of blending with multi-pass will be incorrect, and height/density being a non-linear operation means it will be horrifyingly so.  Doesn't really change if it's density/height or if its without.  So yeah, in multi-pass, it will be awful...  Strangely enough, there are actual tests that are designed to work around the awfulness and the canonicals assume the awfulness is a given, thereby transforming the bug into a feature.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitBasemapGen.shader
parashar-unity: That is done in the base pass, but in all the others, we've actually added smoothness controls...  which is a change from the older built-in shader, but this was meant to bring it closer to parity (or at least compatibility) with the HDRP version.  I know you've made changes to the HD version in the many moons which have passed since this was opened, so there will anyway need to be a later PR that gets that up to date among other things.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLit.shader
parashar-unity: Put in a change now to auto-disable height-based blending whenever multi-pass is on.  For now, the assumption is that >4 layers == multipass is on.  That logic will probably change slightly if we support 8-layer single-pass in the high-quality shading mode, but that will be a future PR and the corresponding change in the UI to address that will be a single-line change once I get to that point.

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
```diff
@@ -146,8 +159,10 @@ bool ITerrainLayerCustomUI.OnTerrainLayerGUI(TerrainLayer terrainLayer, Terrain
                 return false;
 
             // Don't use the member field enableHeightBlend as ShaderGUI.OnGUI might not be called if the material UI is folded.
-            bool heightBlend = terrain.materialTemplate.HasProperty(kEnableHeightBlend) && terrain.materialTemplate.GetFloat(kEnableHeightBlend) > 0;
-            
+            // heightblend shouldn't be available if we are in multi-pass mode, because it is guaranteed to be broken.
+            heightBlendAvailable = (terrainLayers.Length <= 4);
```
zeroyao: This function ITerrainLayerCustomUI.OnTerrainLayerGUI is only called when you select a Terrain Layer under the PaintTextureTools. The variable `heightBlendAvailable` will not be assigned if that's not the case. But in OnGUI there is no access to TerrainData because this class i.e. TerrainLitGUI basically serves only the purpose of rendering the GUI for all the material assets using TerrainLit shader (replace TerrainLit with the LW terrain shader name :))

It is a tricky problem because Terrain Layer / Terrain Material is separate from Terrain Data. I'd suggest do this:
1. Add a hidden shader property to the terrain shader:
[HideInInspector][PerRendererData] _HeightBlendShouldDisable ("Should Height Blend be disabled?", Float) = 0
2. In PaintTextureTools.cs (in trunk codebase), call:
terrainData.materialTemplate.SetFloat("_HeightBlendShouldDisable", terrainData.terrainLayerCount > 4);
inside OnInspectorGUI.
3. Finally in ShaderLitGUI, add:
MaterialProperty heightBlendShouldDisable = null; // somewhere in the declarations
heightBlendShouldDisable = FindProperty("_HeightBlendShouldDisable", props, false); // in FindMaterialProperties
... and then replace the "heightBlendAvailable" variable with the test of "heightBlendShouldDisable != null && heightBlendShouldDisable.floatValue == 0.0f"

What I'm trying to do with these changes is to store the result of testing terrainLayerCount > 4 inside the Material object (with a non-serialized property) and get that result back from the GUI code. It's quite hacky but I feel that's probably the easiest way to work around. An alternative less-hacky way is to have a new interface of "ITerrainMaterial" implemented on the TerrianLitShaderGUI and inside PaintTextureTools cast the shadergui to this interface and pass the TerrainData/Terrain objects to it.

zeroyao: But either way you'll need to change the trunk code and that requires a trunk PR...

zeroyao: The reason why HD TerrainLit shader doesn't need this is because TerrainLit shader doesn't have an AddPass shader. So even if you assign more terrain layers than it allows (which is 8), they will not be rendered. Therefore the Height Blend option is always available no matter how many layers are assigned. The option should also be hidden and preferrably display a warning message, if we can do what I suggested above, for TerrainLti shader as well.

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLit.shader
```diff
@@ -4,6 +4,7 @@ Shader "Lightweight Render Pipeline/Terrain/Lit"
     {
         [HideInInspector] [ToggleUI] _EnableHeightBlend("EnableHeightBlend", Float) = 0.0
         _HeightTransition("Height Transition", Range(0, 1.0)) = 0.0
+        [HideInInspector] [PerRendererData] _NumLayersCount ("Total Layer Count", Float) = 1.0
```
zeroyao: Please put a comment here explaining why are we doing this.

Katstodian:
Hi - van this please have a changelog entry when done, and be assigned to GFX Docs for review?

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -225,22 +280,96 @@ half4 SplatmapFragment(Varyings IN) : SV_TARGET
...
+//#endif
+    half4 hasMask = half4(_LayerHasMask0, _LayerHasMask1, _LayerHasMask2, _LayerHasMask3);
+    half4 hasNoMask = 1.0h - hasMask;
+    half4 defaultMetallic = half4(_Metallic0, _Metallic1, _Metallic2, _Metallic3);
+    half4 defaultOcclusion = 1.0h;
+    
+    half4 maskSmoothness = half4(masks[0].a, masks[1].a, masks[2].a, masks[3].a);
+    maskSmoothness *= half4(_MaskMapRemapScale0.a, _MaskMapRemapScale1.a, _MaskMapRemapScale2.a, _MaskMapRemapScale3.a);
+    maskSmoothness += half4(_MaskMapRemapOffset0.a, _MaskMapRemapOffset1.a, _MaskMapRemapOffset2.a, _MaskMapRemapOffset3.a);
+
```
cdxntchou: May be worth applying MaskMapRemapScale/Offset directly to the masks _before_ repacking them into per-property vectors..  it's the same cost on scalarized GPUs but will run more efficiently on older vectorized GPUs:

masks[0] = mask[0] * _MaskMapRemapScale0.rgba + _MaskMapRemapOffset0.rgba
masks[1] = mask[1] * _MaskMapRemapScale1.rgba + _MaskMapRemapOffset1.rgba
masks[2] = mask[2] * _MaskMapRemapScale2.rgba + _MaskMapRemapOffset2.rgba
masks[3] = mask[3] * _MaskMapRemapScale3.rgba + _MaskMapRemapOffset3.rgba

// then:
half4 maskSmoothness = half4(masks[0].a, masks[1].a, masks[2].a, masks[3].a);
half4 maskMetallic = half4(masks[0].r, masks[1].r, masks[2].r, masks[3].r);
half4 maskOcclusion = half4(masks[0].g, masks[1].g, masks[2].g, masks[3].g);

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
```diff
@@ -0,0 +1,327 @@
...
+                    min = maskMapRemapMin.w; max = maskMapRemapMax.w;
+                    EditorGUILayout.MinMaxSlider(s_Styles.smoothness, ref min, ref max, 0, 1);
+                    maskMapRemapMin.w = min; maskMapRemapMax.w = max;
+                }
+                else
+                {
+                    metallic = EditorGUILayout.Slider(s_Styles.metallic, metallic, 0, 1);
+                    // AO and Height are still exclusively controlled via the maskRemap controls
+                    // metallic and smoothness have their own values as fields within the LayerData.
+                    maskMapRemapMax.y = EditorGUILayout.Slider(s_Styles.ao, maskMapRemapMax.y, 0, 1);
```
cdxntchou: If you are switching back and forth from mask <--> nonMask case, the min and max can get reversed (and the UI goes a bit crazy).  Can add something like this to maintain min <= max:
maskMapRemapMin.y = min(maskMapRemapMin.y, maskMapRemapMax.y)

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/TerrainLitShaderGUI.cs
```diff
@@ -0,0 +1,327 @@
...
+                    if (heightBlend)
+                    {
+                        maskMapRemapMax.z = EditorGUILayout.FloatField(s_Styles.heightCm, maskMapRemapMax.z * 100) / 100;
```
cdxntchou: Same here, I would keep min <= max for sanity

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -225,22 +279,68 @@ half4 SplatmapFragment(Varyings IN) : SV_TARGET
...
+    half4 defaultSmoothness;
+    
+    half4 masks[4];
+    splatControl = SAMPLE_TEXTURE2D(_Control, sampler_Control, IN.uvMainAndLM.xy);
+    
+    masks[0] = 1.0h;    masks[0].b = _MaskMapRemapScale0.b + _MaskMapRemapOffset0.b;
+    masks[1] = 1.0h;    masks[1].b = _MaskMapRemapScale1.b + _MaskMapRemapOffset1.b;
+    masks[2] = 1.0h;    masks[2].b = _MaskMapRemapScale2.b + _MaskMapRemapOffset2.b;
+    masks[3] = 1.0h;    masks[3].b = _MaskMapRemapScale3.b + _MaskMapRemapOffset3.b;
+
```
cdxntchou: Since the masks[0].b are only used in HeightBasedSplatModify, I think it would make this more readable if you moved all of the masks[0].b = _MaskMapRemapScale0.b + _MaskMapRemapOffset0.b; statements into the #ifdef _TERRAIN_BLEND_HEIGHT block below.  Maybe add a comment that it reads the mask blue channel as height.

zadi0648:
Just an update: There is a few more things that I want to check but my brain is fried today. The bugs I found reproduce in trunk (going to report them), so I anticipate that this will be ready to be green'd tomorrow.

parashar-unity:
> Only two problems found:
> 
> 1. ❌ Height Blend: Height value not being set for layers that don’t have a Mask Map

This one should be fixed with the latest push.  Main this wasn't so much individual layers not having a mask map, but that _none_ of the layers had a mask map, which basically disabled the feature.

parashar-unity:
> 2. ❌ Density Blend: Checking Opacity as Density on textures that has “Alpha is Transparent” results in Black
And now this one is also fixed.  Normalization issue, basically.

phi-lira:
Hi, please double check changelog entries are in correct place and I'll merge. Also which version of Unity is required for this?

parashar-unity:
Okay.  The changelog entries are all lined up properly.  The version of
Unity that currently has all the fixes needed to support these changes is
2019.3.0a5.  There are also 2019.2 and 2019.1 backports of those core
changes in waiting.

- Parashar K.

PR #2711, opened 2019-01-08
Update shader stripping for decal

sebastienlagarde:
### Purpose of this PR
Improve shader variant stripping for decal.
use define instead of multicompile, and strip decal shader themselves.
Also remove unused wind.hlsl from shader graph shader.

---
### Testing status

**Automated Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp-add-decal-stripping&automation-tools_branch=add-platform-filter&unity_branch=trunk#

--

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
@@ -39,7 +39,7 @@ namespace UnityEditor.Experimental.Rendering.HDPipeline
             bool isTransparentPrepass = snippet.passName == "TransparentDepthPrepass";
             bool isTransparentPostpass = snippet.passName == "TransparentDepthPostpass";
             bool isTransparentBackface = snippet.passName == "TransparentBackface";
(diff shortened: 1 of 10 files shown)


sebastienlagarde:
Katana is green (except transparent test that fail due to slaves wtd-cpht-006 as reported in slack channel).

Paulm-Unity:
This only strips the shaders, not the passes right?  Because we have this code in DecalSystem.cs, so if passes order changes it will draw incorrectly. I always meant to rewrite that code to use a pass name lookup table so that it is independent of the pass order in the shader file. Maybe now is a good time? =)

shaderPass = perChannelMask ? MaskBlendMode : (int)Decal.MaskBlendFlags.Smoothness; // relies on the order shader passes are declared in decal.shader and decalUI.cs

cmd.DrawMeshInstanced(m_DecalMesh, 0, m_Material, shaderPass, m_DecalToWorld[batchIndex], kDrawIndexedBatchSize, m_PropertyBlock);

sebastienlagarde:
>This only strips the shaders, not the passes right? 
I have tested various scenario and it was working, I suppose the shader stripper simply don't compile the shader, but don't remove the "pass" declare in the shader, the order is still the same

PR #2712, opened 2019-01-08
 Small tweaks. 

phi-lira:
### Purpose of this PR
 Renamed Don't Inititalize -> Don't Care as this is more explicit in terms of what's happening. 
 RenderOpaques was being initialized with Load op when clear flags set to Skybox or Don't Initialize. We now do Don't Care whenever RenderOpaques is the first render pass to write to the camera color target. The only case it's not the first render pass is when there's an custom IBeforeRender render pass being executed. In that case the clear flag will be set to None. 

---
### Release Notes
 Updated Don't Initialize with Don't Care.

---
### Testing status

**Manual Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fsmall-tweaks&unity_branch=2019.1%2Fstaging

**Automated Tests**: We can't have tests on LOAD/STORE ops due to missing mobile tests.

Any test projects to go with this to help reviewers? LWGraphicsTest.

---
### Overall Product Risks
**Technical Risk**: Low. Small UI changes and change in renderpass. 

**Halo Effect**: Low. Should affect only mobile. 

---
### Comments to reviewers


com.unity.render-pipelines.lightweight/Runtime/Passes/RenderOpaqueForwardPass.cs
```diff
@@ -66,19 +66,13 @@ public override void Execute(ScriptableRenderer renderer, ScriptableRenderContex
             CommandBuffer cmd = CommandBufferPool.Get(k_RenderOpaquesTag);
             using (new ProfilingSample(cmd, k_RenderOpaquesTag))
             {
-                RenderBufferLoadAction colorLoadOp = (CoreUtils.HasFlag(clearFlag, ClearFlag.Color))
-                    ? RenderBufferLoadAction.DontCare
-                    : RenderBufferLoadAction.Load;
-                RenderBufferStoreAction colorStoreOp = RenderBufferStoreAction.Store;
+                // When ClearFlag.None that means this is not the first render pass to write to camera target.
+                // In that case we set loapOp for both color and depth as RenderBufferLoadAction.Load
```
jonuuukas: > loapOp

typo

PR #2719, opened 2019-01-09
Camera-relative support for stereo

fabien-unity:
### Purpose of this PR
Proper support of camera-relative code in stereo rendering. With this PR, VR should run out of the box with no more manual file edits.

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=xr%2Fhdrp%2Fcamera-relative&unity_branch=trunk)

**Manual Tests**: I've used test scene 9201 as reference. And the default HDRP template.

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
**Note 1:**
On my setup (Windows 10 with 1080 Ti), I encounter what looks like a shader compiler bug (annotated below). I'll run more tests to confirm. The observed behavior is punctual shadow flickering.

**Note 2:**
Far from the origin, the rendering is not stable due to micro-variations in the camera rotation from the HMD tracking. This can be simulated in non-VR too by adding some micro-noise to the camera rotation via script:
```
public class MicroNoise : MonoBehaviour
{
    public float noise = 0.001f; // match measured noise from Oculus Rift
    private Quaternion originalRotation;

    void Start()
    {
        originalRotation = transform.rotation;
    }

    void Update()
    {
        var r = originalRotation;
        for (int i = 0; i < 3; ++i)
            r[i] += Random.Range(-noise, noise);
        transform.rotation = r;
    }
}
```

Let me know if you have thoughts on this issue.

**Note 3:**
Once approved, I'll update https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki/VR-in-HDRP

**Note 4:**
While working on this, I encounter an engine bug related to shadow culling.
PR is up: https://ono.unity3d.com/unity/unity/pull-request/80697/_/graphics/culling/directional-shadow-precision
Grab this if you see shadow flickering.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -255,18 +255,22 @@ uint FetchIndexWithBoundsCheck(uint start, uint count, uint i)
...
 }
 
-LightData FetchLight(uint start, uint i)
+LightData FetchLight(uint index)
 {
-    int j = FetchIndex(start, i);
+    LightData lightData = _LightDatas[index];
+
+    // For camera-relative code to work in stereo, we need to translate position to be eye-relative
+    lightData.positionRWS =  StereoCameraRelativeCenterToEye(lightData.positionRWS);
```
fabien-unity: this line seems to trigger a shader compiler bug ... I'll dig more

fabien-unity: I can repro the same issue (flickering shadows) on a Mac Book Pro (AMD Radeon Pro 560) and 1080Ti. I'll look more into this problem.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -144,8 +144,27 @@ float2 TexCoordStereoOffset(float2 texCoord)
 {
 #if defined(UNITY_SINGLE_PASS_STEREO)
     return texCoord + float2(unity_StereoEyeIndex * _ScreenSize.x, 0.0);
-#endif
+#else
     return texCoord;
+#endif
+}
+
+float3 StereoCameraRelativeEyeToCenter(float3 pos)
```
sebastienlagarde: Add a comment here to explain what this function is doing

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
sebastienlagarde: I wondering something, if you do that, then you brake the optimization with scalar register ?
you should do this transform in Lightloop.cs instead of inside the shader, will be easier to handle

sebastienlagarde:
Note: Target hdrp-master for this PR please, currently it is our main branch of development (like for SPI branch) thanks

mikesfbay:
@fabien-unity When you say "Far from the origin, the rendering is not stable" are you referring to the camera or the shadow? Does it get worse the further out it gets?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
fabien-unity: done

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
fabien-unity: I slightly refactored to avoid this.

fabien-unity:
I've changed the base to hdrp-master and fixed the flickering shadows issue.
This PR is now fully ready to be reviewed and tested.

sebastienlagarde:
Note: Katana test on hdrp-master aren't green, I need to fix stuff, there is 11 test failing and I see on this katana it is the same thing, will first fix screenshots on hdrp-master before merging this one to be sure.

fabien-unity:
@sebastienlagarde I moved the offset code away to higher level functions so that it only applies to shadow computation (in EvalShadow_PunctualDepth() and EvalShadow_CascadedDepth_Blend()).
Usual lighting code should not be affected because it already relies on the proper camera position from the eye matrix. Shadows were problematic only because we render them once for both eyes and we need to account for the offset between the eye and the center of the HMD.

PR #2725, opened 2019-01-10
Graphics/terrain/gpu/terrain surface mask (C++ PR landed in trunk)

alekseiunity:
### Purpose of this PR

This is a scriptable render pipeline part of terrain surface mask (aka terrain holes) PR https://ono.unity3d.com/unity/unity/pull-request/79809/_/graphics/terrain/gpu/terrain-surface-mask
It enables terrain surface mask functionality for HDRP and LWRP terrain shaders.
---
### Release Notes

---
### Testing status
Only shader changes.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=graphics%2Fterrain%2Fgpu%2Fterrain-surface-mask&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

Performed manual painting of surface mask over terrain with different render pipeline settings.
No issues found.

Added native, editor and play mode tests as part of main surface mask PR.
Tests are passed.

No additional tests.
---
### Overall Product Risks
Low

Low

---
### Comments to reviewers
[HideInInspector] _TerrainSurfaceMaskTexture is a requirement for GI logic that uses that parameter to acquire surface mask texture for transmission calculations.


alekseiunity:
The unused variant will be stripped out from the build If shader_feature
used instead, correct?
Considering that user can change terrain surface mask at runtime and
TERRAIN_SURFACE_MASK_ENABLED will be enabled for the material this option
doesn't work I guess.
Correct me.

phi-lira:
Can we do a setting to control stripping similarly to *strip unused* fog/lightmaps/instancing then?
https://docs.unity3d.com/Manual/class-GraphicsSettings.html

Consider shader variants grows exponentially and we are already in a pretty bad situation. Worst scenarios with VR + Mobile we users reporting +1h build. Off course this is not only terrain issue but here we are doubling variants with that new keyword.

alekseiunity:
Correct me if I am wrong. Do I double size only for terrain shaders by
using this keyword?

phi-lira:
That's correct. Will double only terrain.

sebastienlagarde:
Yes. As felipe say multicompile double the number of variant. 

Question: why is it a multicompile? It is a feature that is enable globally ? Is it with surfacemask or without surfacemask for the whole game or is it only for the material?

If we talk about the material and each material can have different value it is a shader features not a multicompile. And all is fine. 

If it is a global mode then it mean we can remove extra variant in the stripper. This is not nice to add more variant on our already insane forward shader but doable. 

If it can be switch at runtime between one mode or the other then we are doomed and can't strip variant and build time will suffer a lot.

Now what I see is very dangerous. The surfacemask use a clip instruction. It mean that if you use the surfacemask and if it is global. You take a severe performance hit on all terrain material (no hierarchical z optimization any more. Which is bad).

Also last point a multicompile should be setup in all the pass that require it and not globally. Otherwise pass like metapass will generate several variant when this may not be necessary. But in your case as it seems to only control clip maybe indeed it must apply to all pass.

So if you can confirm if that should be have a shader feature and not a multicompile this will help. Otherwise we will need to be clever...

sebastienlagarde:
In case of hdrp. Doubling (again) the number of variant will be another build time hit (our forward is already heavy in variant). We need to add an hdrp asset option to be able to disable the feature if users don't want it (so we can strip variant in the shader stripper). But I don't think we have other choice if we want to support such features. We can't afford to have clip instrcution enabled on all terrain all the time . It kill the hi-z.

alekseiunity:
Question: why is it a multicompile? It is a feature that is enable globally
? Is it with surfacemask or without surfacemask for the whole game or is it
only for the material?

-- It is per material, as you know terrain consist in patches, if terrain
patch has surface mask it is rendered with one material, if not it is
rendered with another which doesn't use(sample and clip) surface mask.

If we talk about the material and each material can have different value it
is a shader features not a multicompile. And all is fine.

-- Like I said if the user doesn't have any holes in terrain at editor
time, it can call GetComponent<Terrain>().terrainData.SetSurfaceMask(...)
at runtime and create a hole in terrain. In that case shader_feature
doesn't work as I understand.

If it is a global mode then it mean we can remove extra variant in the
stripper. This is not nice to add more variant on our already insane
forward shader but doable.

If it can be switch at runtime between one mode or the other then we are
doomed and can't strip variant and build time will suffer a lot.

-- You are correct.

Now what I see is very dangerous. The surfacemask use a clip instruction.
It mean that if you use the surfacemask and if it is global. You take a
severe performance hit on all terrain material (no hierarchical z
optimization any more. Which is bad).

-- I am aware of that problem, but as I mentioned above, clip instruction
will be for shader variant that is used only by terrain patch that has
holes. NOT for all patches of all terrains. In most of the cases, amount of
patches with holes will be really low (I hope :) ). Moreover, I can't find
another way to cut correct holes in terrain without clip instruction. For
example, I can't cut holes using geometry as the resolution of a terrain
patch can be lower than the resolution of surface mask texture. (any advice
here?)

Also last point a multicompile should be setup in all the pass that require
it and not globally. Otherwise pass like metapass will generate several
variant when this may not be necessary. But in your case as it seems to
only control clip maybe indeed it must apply to all pass.

-- You are correct.

So if you can confirm if that should be have a shader feature and not a
multicompile this will help. Otherwise we will need to be clever...

alekseiunity:
So again, the main purpose of this keyword is to prevent calling clip
instruction for all patches of the terrain.

sebastienlagarde:
Hey. Sorry I have done an initial comment that I have remove after that because I get a better understanding of the feature so only second comment matter. And yes I get the features now and I think we have no choice. The only thing we can do is to allow users to disable it if they don't use it

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitSplatCommon.hlsl
```diff
@@ -33,6 +33,17 @@ TEXTURE2D(_TerrainHeightmapTexture);
...
 #endif
 
+#ifdef TERRAIN_SURFACE_MASK_ENABLED
+TEXTURE2D(_TerrainSurfaceMaskTexture);
+SAMPLER(sampler_TerrainSurfaceMaskTexture);
+
+void ClipSurfaceMask(float2 uv)
+{
+	float surfMask = SAMPLE_TEXTURE2D(_TerrainSurfaceMaskTexture, sampler_TerrainSurfaceMaskTexture, uv).r;
+	clip(surfMask == 0.0f ? -1 : 1);
```
sebastienlagarde: So there is one optimization that I would like to be supported with the terrain if possible.

In HDRP, any call to clip instruction must use the wrapper: DoAlphaTest() define in Material.hlsl

// This function must be use instead of clip instruction. It allow to manage in which case the clip is perform for optimization purpose
void DoAlphaTest(float alpha, float alphaCutoff)
{
    // For Deferred:
    // If we have a prepass, we  may want to remove the clip from the GBuffer pass (otherwise HiZ does not work on PS4) - SHADERPASS_GBUFFER_BYPASS_ALPHA_TEST
    // For Forward Opaque:
    // If we have a prepass, we may want to remove the clip from the forward pass (otherwise HiZ does not work on PS4) - SHADERPASS_FORWARD_BYPASS_ALPHA_TEST
    // For Forward Transparent
    // Also no alpha test for light transport
    // Note: If SHADERPASS_GBUFFER_BYPASS_ALPHA_TEST or SHADERPASS_FORWARD_BYPASS_ALPHA_TEST are used, it mean that we must use ZTest depth equal for the pass (Need to use _ZTestDepthEqualForOpaque property).
#if !defined(SHADERPASS_FORWARD_BYPASS_ALPHA_TEST) && !defined(SHADERPASS_GBUFFER_BYPASS_ALPHA_TEST) && !(SHADERPASS == SHADERPASS_LIGHT_TRANSPORT)
    clip(alpha - alphaCutoff);
#endif
}

The reasons is that we perform an optimization automatically for any object that have alpha test. 
An object with alpha test will be render WITH alpha test during depth prepass (we handle this with RenderQueue) and then during normal pass we don't perform the alpha test and do z equal test instead. This is handled with the function above, so all you have to do to support it is to use DoAlphaTest call (and check in renderdoc that there is no discard instruction in the shader :) ).

See  bool RenderDepthPrepass(CullingResults cull, HDCamera hdCamera, ScriptableRenderContext renderContext, CommandBuffer cmd) to see that we force a renderpass for renderqueue alpha test

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
```diff
@@ -65,13 +67,19 @@ Shader "HDRP/TerrainLit"
...
+	
     //-------------------------------------------------------------------------------------
     // Define
     //-------------------------------------------------------------------------------------
 
     #define SURFACE_GRADIENT
     #define HAVE_MESH_MODIFICATION
 
+	#ifdef TERRAIN_SURFACE_MASK_ENABLED
+	#define _ALPHATEST_ON
```
sebastienlagarde: It miss one step here related to my comment about our optimization, we need to add to GBuffer pass declaratoin

        #ifdef _ALPHATEST_ON
            // When we have alpha test, we will force a depth prepass so we always bypass the clip instruction in the GBuffer
            #define SHADERPASS_GBUFFER_BYPASS_ALPHA_TEST
        #endif

For forward this is already added (and for all type of terrain shader (2 of them).

If you don't use this optimization, defining _ALPHATEST_ON is useless

sebastienlagarde: See in Lit shader:

        // Caution: The outline selection in the editor use the vertex shader/hull/domain shader of the first pass declare. So it should not bethe  meta pass.
        Pass
        {
            Name "GBuffer"  // Name is not used
            Tags { "LightMode" = "GBuffer" } // This will be only for opaque object based on the RenderQueue index

            Cull [_CullMode]
            ZTest [_ZTestGBuffer]

            Stencil
            {
                WriteMask [_StencilWriteMask]
                Ref  [_StencilRef]
                Comp Always
                Pass Replace
            }

            HLSLPROGRAM

            #pragma multi_compile _ DEBUG_DISPLAY
            #pragma multi_compile _ LIGHTMAP_ON
            #pragma multi_compile _ DIRLIGHTMAP_COMBINED
            #pragma multi_compile _ DYNAMICLIGHTMAP_ON
            #pragma multi_compile _ SHADOWS_SHADOWMASK
            // Setup DECALS_OFF so the shader stripper can remove variants
            #pragma multi_compile DECALS_OFF DECALS_3RT DECALS_4RT
            #pragma multi_compile _ LIGHT_LAYERS

        #ifdef _ALPHATEST_ON
            // When we have alpha test, we will force a depth prepass so we always bypass the clip instruction in the GBuffer
            #define SHADERPASS_GBUFFER_BYPASS_ALPHA_TEST
        #endif


It was not done for terrain initialiy as terrain was not having alpha clip at this time :)

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitData.hlsl
```diff
@@ -39,6 +39,10 @@ void GetSurfaceAndBuiltinData(inout FragInputs input, float3 V, inout PositionIn
     }
 #endif
 
+#ifdef TERRAIN_SURFACE_MASK_ENABLED
+	ClipSurfaceMask(input.texCoord0);
```
sebastienlagarde: Random question, is input.texCoord0 enough ? isn't there any tiling option or thing like that ?

alekseiunity:
Thank you, Sebastien, for your comments. I fixed shaders. Now they work as
you described.

phi-lira:
@alekseiunity is it possible to add a `Strip Unused` setting like we have for fog modes/lightmap etc? Maybe we could expose this in the pipeline asset?

alekseiunity:
Hi, Sebastien mentioned "For the option and the stripping my team can add
it on HDRP side if you want.", if you can add this option that would be
much easy for me. Or I can do it myself.
Thanks.

sebastienlagarde:
I am good with that. @felipe  I suggest we add the option ourselves in our respective RP. I don't like current option of lightmaps for example as they are super hidden in legacy settings whereas we have our RP assets that already propose all our option for variant stripping. 

Are you good with the PR and can you handle the stripping case for LW? Otherwise I can help with it for LW too.

alekseiunity:
Thank you!

sebastienlagarde:
LAst link on the C++ PR to keep track: https://ono.unity3d.com/unity/unity/pull-request/81950/_/graphics/terrain/gpu/terrain-surface-mask

sebastienlagarde:
Note: THis PR here will conflict with the change made in current PR, you need to accept the code fromPR 2926 concerning the alpha test macro

https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2926

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit_Basemap.shader
```diff
@@ -16,6 +16,13 @@ Shader "Hidden/HDRP/TerrainLit_Basemap"
         [HideInInspector] _ZTestDepthEqualForOpaque("_ZTestDepthEqualForOpaque", Int) = 4 // Less equal
         [HideInInspector] _ZTestGBuffer("_ZTestGBuffer", Int) = 4
 
+		[HideInInspector] [ToggleUI] _AlphaCutoffEnable("Alpha Cutoff Enable", Float) = 1.0
+		[HideInInspector] _AlphaCutoff("Alpha Cutoff", Range(0.0, 1.0)) = 0.5 
+        [HideInInspector] _AlphaCutoffShadow("_AlphaCutoffShadow", Range(0.0, 1.0)) = 0.5
+        [HideInInspector] _AlphaCutoffPrepass("_AlphaCutoffPrepass", Range(0.0, 1.0)) = 0.5
+        [HideInInspector] _AlphaCutoffPostpass("_AlphaCutoffPostpass", Range(0.0, 1.0)) = 0.5
+		[HideInInspector] _TerrainSurfaceMaskTexture("Surface Mask Map (RGB)", 2D) = "white" {}
```
zeroyao: Is the "(RGB)" accurate?

alekseiunity:
Surface Mask texture is RGBA compressed or R channel Render Target.

spaztronaut:
Terrain seems to not be working correctly in master and in this branch though I don't believe it is related to these changes

sebastienlagarde:
Hey, so looks like this PR is ready? 
Don't forget that once we merge it we also need to work on shader striping for it. I suppose we want to wait that LW is out of preview before merging ? (@phi-lira ). 

In the mean time can you merge with latest master and re-run katana as there is conflict now with the change from @zeroyao

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
```diff
@@ -72,6 +79,12 @@ Shader "HDRP/TerrainLit"
     #pragma multi_compile_instancing
     #pragma instancing_options assumeuniformscaling nomatrices nolightprobe nolightmap
 
+	#pragma multi_compile_local __ TERRAIN_SURFACE_MASK_ENABLED
+	
+	#ifdef TERRAIN_SURFACE_MASK_ENABLED
+	#define _ALPHATEST_ON
```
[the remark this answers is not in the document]
alekseiunity: Fixed

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitTemplate.hlsl
```diff
@@ -42,7 +42,7 @@
     #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl"
 #endif
 
-#if SHADERPASS != SHADERPASS_DEPTH_ONLY || defined(WRITE_NORMAL_BUFFER)
+#if SHADERPASS != SHADERPASS_DEPTH_ONLY || defined(WRITE_NORMAL_BUFFER) || defined(TERRAIN_SURFACE_MASK_ENABLED)
```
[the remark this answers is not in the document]
alekseiunity: I actually fixed it in the following commit

com.unity.render-pipelines.lightweight/Editor/ForwardRendererDataEditor.cs
```diff
@@ -1,41 +1,41 @@
...
-                m_TransparentLayerMask.intValue = m_OpaqueLayerMask.intValue;
-            EditorGUILayout.Space();
-
-            serializedObject.ApplyModifiedProperties();
-            
-            base.OnInspectorGUI(); // Draw the base UI, contains ScriptableRenderFeatures list
-        }
-    }
-}
+using UnityEngine;
```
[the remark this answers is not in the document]
alekseiunity: I didn't change anything here and in a couple of other .cs files. It is just merge with master.

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
```diff
@@ -33,6 +33,13 @@ Shader "HDRP/TerrainLit"
 
         [ToggleUI] _EnableInstancedPerPixelNormal("Instanced per pixel normal", Float) = 1.0
 
+		[HideInInspector] [ToggleUI] _AlphaCutoffEnable("Alpha Cutoff Enable", Float) = 1.0
```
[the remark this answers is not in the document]
alekseiunity: I didn't change TerraibLitUI because I don't want to expose those properties to the user. These properties are needed only for terrain holes. They don't need to be changed by the user.

com.unity.render-pipelines.lightweight/Editor/ForwardRendererDataEditor.cs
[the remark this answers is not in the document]
alekseiunity: Reverted those files

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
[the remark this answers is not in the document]
zeroyao: If so then no need to declare them as properties. Properties AFAIK serves two purposes: serialization and providing default values.

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitTemplate.hlsl
```diff
@@ -16,6 +16,12 @@
     #define SHADERPASS_FORWARD_BYPASS_ALPHA_TEST
 #endif
 
+#if defined(TERRAIN_SURFACE_MASK_ENABLED)
+	#define _ALPHATEST_ON
```
zeroyao: By looking at the original comment:
// When we have alpha test, we will force a depth prepass so we always bypass the clip instruction in the GBuffer
That probably means HDRP detected alpha testing objects in the depth prepass so that later in GBuffer pass we don't need to do clip any more. Since TERRAIN_SURFACE_MASK_ENABLED is enabled only by terrain engine, big chance is that it's not detected by the depth prepass. Issue?
@sebastienlagarde

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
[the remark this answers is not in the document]
alekseiunity: What do you mean as properties? I thought that they are shader properties. No?

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitTemplate.hlsl
alekseiunity: I tested everything, no issues. When the surface mask is enabled, alpha test is enabled for depth prepass only, and depth function is 'Equals' for GBuffer pass. So we don't clip twice.

zeroyao: Can you check if depth prepass is forced when we only have surface mask terrain in the scene? (need to confirm that HDRP correctly detects terrain has alpha test so it forces a depth prepass. Make sure the depth prepass is not forced before you paint holes on terrain)

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit.shader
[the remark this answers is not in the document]
zeroyao: They are actually Material properties. The values declared here will be serialized with Material asset (or provide default values). As for shader uniforms they are found at the time a GPU program is compiled by using shader code reflection. If you do "Material.SetFloat(prop, value)" or "MaterialProperyBlock.SetFloat(prop, value)" or "Shader.SetGlobalFloat(prop, value)" the prop always gets its value even if it's not declared in the Properties block.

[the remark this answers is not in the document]
zeroyao: Oh the third purpose of a shader property: for GUI code to pick up :) but nothing has to do with actual shader uniforms.

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLitTemplate.hlsl
sebastienlagarde: hmm... Yao is right. So I am not sure to understand how this work correctly.

In HDRP we have the following:
- Object that are in the renderqueue opaque alpha-test are always render in depth prepass
- It is the BaseUnlitUI.cs that setup the renderqueue for the material (TerrainLitUI.cs derive from it)
- To know in Material inspector if we are alpha tested we do the following test:
bool alphaTest = material.HasProperty(kAlphaCutoffEnabled) && material.GetFloat(kAlphaCutoffEnabled) > 0.0f;
as the terrain only used #define _ALPHATEST_ON I am not sure how this can work. Maybe you setup in C++ the right render queue ?  But then it will not work with the code that setup the Equals:

            // Alpha tested materials always have a prepass where we perform the clip.
            // Then during Gbuffer pass we don't perform the clip test, so we need to use depth equal in this case.
            if (alphaTestEnable)
            {
                material.SetInt(kZTestGBuffer, (int)UnityEngine.Rendering.CompareFunction.Equal);
            }
            else
            {
                material.SetInt(kZTestGBuffer, (int)UnityEngine.Rendering.CompareFunction.LessEqual);
            }

sebastienlagarde: >and depth function is 'Equals' for GBuffer pass. So we don't clip twice.
Try to check in render doc that you don't have the discard instruction in the assembly as a sanity check.

So seems that it works for you but not sure to understand why :). Maybe the default blend state match by luck? 
Yao suggest a good test. Try in isolation and with others objects that have and don't have alpha test to see if there is an issue.

alekseiunity: Oh...You both were right. It worked because for DefautHDTerrainShader _AlphaCutoffEnable is enabled by default (for new created material is not) and full depth prepass was forced by decals. Thank you. Working on fix...

alekseiunity: I noticed that if I change _AlphaCutoffEnable property for any material in HD at runtime from a script, then obviously SetupMaterialKeywordsAndPass will not be called, because it is called only from UI. This means that I can't change alpha test properties at runtime. Am I correct?

sebastienlagarde: hmm... arf indeed....

_ALPHATEST_ON  is a shader features in our shaders, so it is expected that you don't change it at runtime...

So indeed, you can't call SetupMaterialKeywordsAndPass and you will need to do explicitly the call to set the blend state for the terrain.  i.e you need to do this:

if (required to switch terrain material)
{
material.EnableKayword(TERRAIN_SURFACE_MASK_ENABLED)
material.SetInt(kZTestGBuffer, (int)UnityEngine.Rendering.CompareFunction.Equal)}
}
else
{
material.disableKeyword(TERRAIN_SURFACE_MASK_ENABLED);
material.SetInt(kZTestGBuffer, (int)UnityEngine.Rendering.CompareFunction.LessEqual);
}

Side note: I may suggest to rename TERRAIN_SURFACE_MASK_ENABLED  as _ALPHATEST_ON  so future code that rely on keyword _ALPHATEST_ON  will work (like DXR code). I know it is bad in term of naming but it will work as keyword checking. Just a suggestion.

alekseiunity: So I renamed TERRAIN_SURFACE_MASK_ENABLED to _ALPHATEST_ON.
I also added SRP support in C++ side. Sebastien, I added you to the main PR, @zeroyao and @sebastienlagarde please review this changeset.
https://ono.unity3d.com/unity/unity/changeset/888eebbb1921a3b88f348e9351a404807d0aa4e9

alekseiunity: Going to write some graphics test too.

sebastienlagarde:
hey, just poking as I forget what is the state of this PR, where are we on this ?

alekseiunity:
The main PR is in trunk queue. As soon as it landed I will notify.

alekseiunity:
Hi. C++ part is landed in trunk. Please merge this PR into SRP. We agreed that you team can perform a shader stripping feature also, can you do that? I also added graphics tests for HDRP and LWRP.

sebastienlagarde:
Hi. Yes, HDRP team will handle the stripping on HDRP side. 
I don't remember, but I guess this PR can't go for 2019.2 as it (i.e there is no guardand with #ifdef).
So we will need to wait a bit before merging as we focus on 19.2 package right now for the next two weeks.

Also you will need to merge a recent master and relaunch katana for it. Thanks

alekseiunity:
Terrain holes is for 2019.3. I will merge recent master.
Thanks.

alekseiunity:
We decided to give name terrain holes instead terrain surface mask, I renamed everything. I also merged master. Please wait until C++ PR with the same renaming will land in trunk. https://ono.unity3d.com/unity/unity/pull-request/88544/_/graphics/terrain/terrain-holes

alekseiunity:
Hi. Do you have any updates about when this PR can be merged in master? Thanks.

sebastienlagarde:
Hi, we are finally ready to merge this PR. Please merge latest master and launch a Katana (sorry for asking again but thing have moved in master. @zeroyao can you accept the PR? do you have remaining concern?

sebastienlagarde:
Please use #pragma multi_compile and not multi_compile_local . From what I remember, multi_compile_local don't work well with stripper and we want to add an option to removed these variant in case users don't use the terrain hole feature.

Note for @phi-lira , this PR will double the number of variant of your Terrain shader, be aware of this.
On HDRP side we will add an option to allow to remove those extra variant if required.

sebastienlagarde:
ok. Ready to merge, we have been agree with Felipe. 
Waiting for Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=graphics%2Fterrain%2Fgpu%2Fterrain-surface-mask&automation-tools_branch=master&unity_branch=trunk#

alekseiunity:
I merged master and fixed multi_compile. I tried to run Katana, but there are some unknown and unrelated to graphics errors occurred.

sebastienlagarde:
yes, sadly both Katana and Yamato aren't working currently...

alekseiunity:
Ok, will rerun later.

sebastienlagarde:
I launched katana but the test 071 on LW don't match. Opened the project locally, picture don't match at all. Is there an issue with LW implementation ?

alekseiunity:
For which configuration they don't match? I just checked and everything is
fine.

seblagarde:
Graphic test. (win and Mac). It was failing on katana:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=graphics%2Fterrain%2Fgpu%2Fterrain-surface-mask&automation-tools_branch=master&unity_branch=trunk
the log:

for this build: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-Test%20PlayMode%20-%20Windows/builds/5231?ScriptableRenderLoop_branch=graphics/terrain/gpu/terrain-surface-mask&automation-tools_branch=master&unity_branch=trunk

sadly log have been removed :(

And now with VFX failing we can't see until VFX C# PR is merged.

When you say you have checked, you mean locally you have run the test?
On my side I have run test runner  for graphic test on Windows and get same result as katana.

alekseiunity:
I mean that I ran it locally on Windows without any errors. And I also
checked all reference images. Could you post images that you get please?
Reference and Actual?

sebastienlagarde:
are you sure you run Lightweight ? There is no problem with HDRP.

071_Terrain_Holes diff

071_Terrain_Holes expected

071_Terrain_Holes

sebastienlagarde:
Merged master with VFX fix and relaunch katana to get result on katana

alekseiunity:
Cool, lets wait what katana says. I attached a screenshot from my Unity
Editor.

sebastienlagarde:
:(, seems to fail on shader graph know... suppose there is a C++ breaking change from graphview...

Packages\com.unity.shadergraph\Editor\Drawing\Views\MaterialNodeView.cs(235,13): error CS0618: 'IStyle.borderColor' is obsolete: 'IStyle.borderColor is deprecated. Use left/right/top/bottom border properties instead.'
Packages\com.unity.shadergraph\Editor\Drawing\Views\MaterialNodeView.cs(240,13): error CS0618: 'IStyle.borderColor' is obsolete: 'IStyle.borderColor is deprecated. Use left/right/top/bottom border properties instead.'
Packages\com.unity.shadergraph\Editor\Drawing\Views\MaterialNodeView.cs(246,20): error CS0618: 'IResolvedStyle.borderColor' is obsolete: 'IResolvedStyle.borderColor is deprecated. Use left/right/top/bottom border properties instead.'

alekseiunity:
Yeah, I just went to build machine and verified that LW graphics test
worked fine if I ran it from the editor.

sebastienlagarde:
Hey, so with the latest trunk you have merge and the latest katana you can see that the LW test I was mentionning is still failing:
https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Windows/31050341_26_06_2019_19_31_01_%2B0000/logs/TestResults_playmode_LWGraphicsTest.xml

<test-case id="1051" name="071_Terrain_Holes" fullname="LWGraphicsTests.071_Terrain_Holes" methodname="Run" classname="LWGraphicsTests" runstate="Runnable" seed="42097604" result="Failed" start-time="2019-06-26 20:37:44Z" end-time="2019-06-26 20:37:46Z" duration="1.356821" asserts="0">

sebastienlagarde:
Katana is finally green :), merging. 
@phi-lira  I will add the strip change on HDRP side. I will poke you want this is done.

alekseiunity:
Perfect!

PR #2726, opened 2019-01-11
Prevent non-RTed cameras from loading from stale RT textures

nerites:
### Purpose of this PR
For a game camera that doesn't have a raytracing filter component, AO and area shadows that were rendered out during scene view rendering get used for rendering the game view. To avoid unnecessarily texture writes/loads, control flow was updated rather than clearing the textures every cycle. 

Minor refactor of RTAO control flow. 

---
### Testing status
Visual hand-testing only: 
image

---
### Overall Product Risks
Low

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
@@ -2606,11 +2606,7 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
 
         public void RenderDeferredLighting(HDCamera hdCamera, CommandBuffer cmd, DebugDisplaySettings debugDisplaySettings,
             RenderTargetIdentifier[] colorBuffers, RenderTargetIdentifier depthStencilBuffer, RenderTargetIdentifier depthTexture,
-            LightingPassOptions options
-#if ENABLE_RAYTRACING
-            , bool areaShadowsRendered
-#endif
-            )
+            LightingPassOptions options)
         {
             cmd.SetGlobalBuffer(HDShaderIDs.g_vLightListGlobal, s_LightList);
 
@@ -2641,13 +2637,6 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                     int numTilesX = hdCamera.camera.stereoEnabled ? ((w / 2) + 15) / 16 : (w + 15) / 16;
                     int numTilesY = (h + 15) / 16;
                     int numTiles = numTilesX * numTilesY;
-
-#if ENABLE_RAYTRACING
-                    if (areaShadowsRendered)
-                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 1);
-                    else
-                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 0);
-#endif
 
                     bool enableFeatureVariants = GetFeatureVariantsEnabled() && !debugDisplaySettings.IsDebugDisplayEnabled() && !hdCamera.camera.stereoEnabled; // TODO VR: Reenable later
 
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
@@ -2606,7 +2606,11 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
 
         public void RenderDeferredLighting(HDCamera hdCamera, CommandBuffer cmd, DebugDisplaySettings debugDisplaySettings,
             RenderTargetIdentifier[] colorBuffers, RenderTargetIdentifier depthStencilBuffer, RenderTargetIdentifier depthTexture,
-            LightingPassOptions options, bool areaShadowsRendered = false)
+            LightingPassOptions options
+#if ENABLE_RAYTRACING
+            , bool areaShadowsRendered
+#endif
+            )
         {
             cmd.SetGlobalBuffer(HDShaderIDs.g_vLightListGlobal, s_LightList);
 
@@ -2638,11 +2642,12 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                     int numTilesY = (h + 15) / 16;
                     int numTiles = numTilesX * numTilesY;
 
+#if ENABLE_RAYTRACING
                     if (areaShadowsRendered)
                         cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 1);
                     else
                         cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 0);
-
+#endif
 
                     bool enableFeatureVariants = GetFeatureVariantsEnabled() && !debugDisplaySettings.IsDebugDisplayEnabled() && !hdCamera.camera.stereoEnabled; // TODO VR: Reenable later
 
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
@@ -2606,7 +2606,7 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
 
         public void RenderDeferredLighting(HDCamera hdCamera, CommandBuffer cmd, DebugDisplaySettings debugDisplaySettings,
             RenderTargetIdentifier[] colorBuffers, RenderTargetIdentifier depthStencilBuffer, RenderTargetIdentifier depthTexture,
-            LightingPassOptions options)
+            LightingPassOptions options, bool areaShadowsRendered = false)
         {
             cmd.SetGlobalBuffer(HDShaderIDs.g_vLightListGlobal, s_LightList);
 
@@ -2637,6 +2637,12 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
                     int numTilesX = hdCamera.camera.stereoEnabled ? ((w / 2) + 15) / 16 : (w + 15) / 16;
                     int numTilesY = (h + 15) / 16;
                     int numTiles = numTilesX * numTilesY;
+
+                    if (areaShadowsRendered)
+                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 1);
+                    else
+                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 0);
+
 
                     bool enableFeatureVariants = GetFeatureVariantsEnabled() && !debugDisplaySettings.IsDebugDisplayEnabled() && !hdCamera.camera.stereoEnabled; // TODO VR: Reenable later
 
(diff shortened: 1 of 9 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2606,7 +2606,7 @@ public void RenderScreenSpaceShadows(HDCamera hdCamera, RTHandleSystem.RTHandle
 
         public void RenderDeferredLighting(HDCamera hdCamera, CommandBuffer cmd, DebugDisplaySettings debugDisplaySettings,
             RenderTargetIdentifier[] colorBuffers, RenderTargetIdentifier depthStencilBuffer, RenderTargetIdentifier depthTexture,
-            LightingPassOptions options)
+            LightingPassOptions options, bool areaShadowsRendered = false)
```
sebastienlagarde: Please, guardband the change like this with

#if ENABLE_RAYTRACING
#endif

Reasons is that it allow to track what is specific to RT, so I can refactor/clean the code in the future when we will add improve area light support for rasterization + it avoid to break compilation of regular path that shouldn't suffer from Raytrace prescence.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/AmbientOcclusion.cs
```diff
@@ -213,30 +213,12 @@ public void InitRaytracing(HDRaytracingManager raytracingManager, SharedRTManage
...
-                {
-                    m_RaytracingAmbientOcclusion.RenderAO(camera, cmd, m_AmbientOcclusionTex, renderContext);
-                    cmd.SetGlobalTexture(HDShaderIDs._AmbientOcclusionTexture, m_AmbientOcclusionTex);
-                    cmd.SetGlobalVector(HDShaderIDs._AmbientOcclusionParam, new Vector4(0f, 0f, 0f, settings.directLightingStrength.value));
-
-                    // TODO: All the push-debug stuff should be centralized somewhere
-                    (RenderPipelineManager.currentPipeline as HDRenderPipeline).PushFullScreenDebugTexture(camera, cmd, m_AmbientOcclusionTex, FullScreenDebugMode.SSAO);
-                }
-            }
+            if (m_Settings.supportRayTracing && rtEnvironement != null && rtEnvironement.raytracedAO)
```
sebastienlagarde: Note: for now we consider that Raytrace AO and Raytrace SSR replace SSAO and SSR from the regular pipeline, 
We haven't update UI yet to make this more clear (we should have in Frame Settings only AO and Reflection and have an option to say if it is screen space or raytrace), but in the mean time we still rely on camera.frameSettings.enableSSAO to say if the settings is enabled or not.

So in the test above if (m_Settings.supportRayTracing && rtEnvironement != null && rtEnvironement.raytracedAO)
we also need to check camera.frameSettings.enableSSAO

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -206,6 +206,8 @@ public int GetDecalAtlasMipCount()
         RenderTexture                   m_TemporaryTargetForCubemaps;
         Stack<Camera>                   m_ProbeCameraPool = new Stack<Camera>();
 
+        bool                            m_AreaShadowsRendered;
```
sebastienlagarde: use guardband #if ENABLE_RAYTRACING

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2422,13 +2425,13 @@ void RenderDeferredLighting(HDCamera hdCamera, CommandBuffer cmd)
                 // Output split lighting for materials asking for it (masked in the stencil buffer)
                 options.outputSplitLighting = true;
 
-                m_LightLoop.RenderDeferredLighting(hdCamera, cmd, m_CurrentDebugDisplaySettings, m_MRTCache2, m_SharedRTManager.GetDepthStencilBuffer(), depthTexture, options);
+                m_LightLoop.RenderDeferredLighting(hdCamera, cmd, m_CurrentDebugDisplaySettings, m_MRTCache2, m_SharedRTManager.GetDepthStencilBuffer(), depthTexture, options, m_AreaShadowsRendered);
```
sebastienlagarde: use #if ENABLE_RAYTRACING

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1479,16 +1482,16 @@ void ExecuteRenderRequest(RenderRequest renderRequest, ScriptableRenderContext r
...
-                    if (shadowsRendered)
+                    m_AreaShadowsRendered = m_RaytracingShadows.RenderAreaShadows(hdCamera, cmd, renderContext);
+                    if (m_AreaShadowsRendered)
                     {
                         cmd.SetGlobalInt(HDShaderIDs._RaytracedAreaShadow, 1);
                     }
                     else
+#endif
                     {
                         cmd.SetGlobalInt(HDShaderIDs._RaytracedAreaShadow, 0);
```
sebastienlagarde: Let _RaytracedAreaShadow inside guardband #if ENABLE_RAYTRACING

nerites: My worry is that for Deferred.compute, ENABLE_RAYTRACING is always set, even if raytracing support is disabled in asset. Currently the _RaytracedAreaShadow doesn’t prevent rendering errors if raytracing is disabled in render pipeline settings, but it is needed to prevent area shadows from the scene camera from being incorrectly rendered to cameras that don’t have a raytracing filter component.

nerites: I guess that’s an issue that should be resolved separately, though... tomorrow I’ll ifdef it as you say

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2638,6 +2638,12 @@ public void RenderScreenSpaceShadows(HDCamera hdCamera, RTHandleSystem.RTHandle
                     int numTilesY = (h + 15) / 16;
                     int numTiles = numTilesX * numTilesY;
 
+                    if (areaShadowsRendered)
+                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 1);
+                    else
+                        cmd.SetComputeIntParam(deferredComputeShader, HDShaderIDs._RaytracedAreaShadow, 0);
```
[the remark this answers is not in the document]
sebastienlagarde: Note: This code shouldn't be required, I don't understand it .We use 
cmd.SetGlobalInt(HDShaderIDs._RaytracedAreaShadow, 1); that is access by all shader including compute and it is correctly init if there is no CameraFilter.
See RenderAreaShadows

  // Try to grab the acceleration structure for the target camera
            HDRayTracingFilter raytracingFilter = hdCamera.camera.gameObject.GetComponent<HDRayTracingFilter>();
            RaytracingAccelerationStructure accelerationStructure = null;
            if (raytracingFilter != null)
            {
                accelerationStructure = m_RaytracingManager.RequestAccelerationStructure(raytracingFilter.layermask);
            }
            else if (hdCamera.camera.cameraType == CameraType.SceneView || hdCamera.camera.cameraType == CameraType.Preview)
            {
                // For the scene view, we want to use the default acceleration structure
                accelerationStructure = m_RaytracingManager.RequestAccelerationStructure(m_PipelineAsset.renderPipelineSettings.editorRaytracingFilterLayerMask);
            }
            // If no acceleration structure available, end it now
            if (accelerationStructure == null)
            {
                return false;
            }

so with a gameview and no AS, this return false and this is setup with SetGlobal for the current camera. Why is it require ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingAmbientOcclusion.cs
```diff
@@ -69,27 +69,31 @@ public void RenderAO(HDCamera hdCamera, CommandBuffer cmd, RTHandleSystem.RTHand
...
-            if (raytracingFilter != null)
+
+            bool missingRequiredResources = rtEnvironement == null || noiseTexture == null || bilateralFilter == null || aoShader == null || m_PipelineResources.shaders.reflectionRaytracing == null;
+            bool useEditorCam = (hdCamera.camera.cameraType == CameraType.SceneView || hdCamera.camera.cameraType == CameraType.Preview);
+            bool currCamRTAO = ((raytracingFilter != null) || useEditorCam) && hdCamera.frameSettings.enableSSAO; // If SSAO should not be raytraced for current camera
+
+            if (missingRequiredResources || !currCamRTAO) 
             {
-                accelerationStructure = m_RaytracingManager.RequestAccelerationStructure(raytracingFilter.layermask);
+                cmd.SetGlobalTexture(HDShaderIDs._AmbientOcclusionTexture, Texture2D.blackTexture);
```
anisunity: What i understood:  all the AO related changes are for adding this fallback here (no AS or no Filter) right?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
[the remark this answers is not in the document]
anisunity: Same, i don't see why this is required

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingAmbientOcclusion.cs
nerites: Most of it's just combining a bunch of conditional returns to hopefully make things more legible, but I did add the currCamRTAO logic, so that the a camera which does not have a filter component and is not the editor camera will not use leftover AOTextures rendered out from other cameras. This was to fix an issue where I was seeing SceneView AO applied to GameView.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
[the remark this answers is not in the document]
nerites: Sorry, I got the fix I made for AO confused with the change I made for raytraced area shadows. 

HDRenderPipeline.cs only does cmd.SetGlobalInt; is that supposed to apply to compute shaders as well? From the behavior I was seeing, it didn't seem like _RaytracedAreaShadow was being set correctly. See this clip: 

areashadowtoggle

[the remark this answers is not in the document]
sebastienlagarde: Yes Setglobalxxx must affect raytrace and compute shader. We use that a lot for compute shader and regular shader. For example all our camera properties are binded like that. If it doesn't work it mean there is an issue somewhere that we must fix :)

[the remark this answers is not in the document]
nerites: Ok, I managed to fix it by moving the declaration of _RaytracedAreaShadow from the deferred shaders to ShaderVariables.hlsl. Latest commit reflects this.

sebastienlagarde:
Hi.

PR is fine except for area shadow. It is not at all the right place for it (shader config isn't for this and .cs.hlsl file are generated file. So your code will be overwrite. 

So I am wondering if the issue is because arayraytraced is in the deferred.compute CB. Can you try in deferred.compute to put it outside the structure? This is what was done before I move inside the CB. Note: currently deferred.shader isn't use (unless you check some opti9n in framesettings) but I bet it was working. 

If it work outside the CB then it mean we have a serious issue to investigate.
Just test this. If it don't work I will handle it and move stuff at correct place. Thanks!

sebastienlagarde:
@anisunity want to merge this change in its current working branch, so he will make a new PR and squash and merge into it before merging to hdrp master. I will move the declaration stuff.

nerites:
Wait, shadervariables is generated?

anisunity:
Changes have been reported to this commit https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2770

PR #2738, opened 2019-01-11
Material validator hdrp

Wadarass:
### Purpose of this PR
Adding Albedo and Metalness material validation for HDRP pipeline

---
### Release Notes
Added material validator for albedo and metalness for HDRP Lit shaders

---
### Testing status
**Katana Tests**: Katana still running, but most tests are currently green.

**Manual Tests**: Tested validator in test level.

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -287,7 +291,7 @@ void ApplyDebugToSurfaceData(float3x3 worldToTangent, inout SurfaceData surfaceD
 }
 
 // This function is similar to ApplyDebugToSurfaceData but for BSDFData
-void ApplyDebugToBSDFData(inout BSDFData bsdfData)
+void ApplyDebugToBSDFData(inout BSDFData bsdfData, bool metallic, bool metallicWorkflow)
```
sebastienlagarde: You should use ApplyDebugToSurfaceData instead of ApplyDebugToBSDFData.
With ApplyDebugToSurfaceData  you will be able to access baseColor and metal without to provide them in argument. it is not good to change the prototype of ApplyDebugToBSDFData as it is apply for all material (AxF, staclikit, fabric, hair etc... that don't have relationship with mettalic workflow or thing like that. with ApplyDebugToSurfaceData  you know inside the function that you are a mettalic workflow

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -834,7 +849,11 @@ uint DecodeFromGBuffer(uint2 positionSS, uint tileFeatureFlags, out BSDFData bsd
     // perceptualRoughness can be modify by FillMaterialClearCoatData, so ConvertAnisotropyToClampRoughness must be call after
     ConvertAnisotropyToRoughness(bsdfData.perceptualRoughness, bsdfData.anisotropy, bsdfData.roughnessT, bsdfData.roughnessB);
 
-    ApplyDebugToBSDFData(bsdfData);
+    // Pass specular or metallic workflow infromation to debug bsdf it's needed for material validation
+    // When coming from g-buffer we don't know always know if the metallic value is in the g-buffer.
```
sebastienlagarde: See comment above, in ApplyDebugToSurfaceData  you know you are mettalic.

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugEditorViz.hlsl
```diff
@@ -0,0 +1,89 @@
+#ifndef UNITY_DEBUG_EDITOR_VIZ_INCLUDED
```
sebastienlagarde: This file being used for PBR validator, I will suggest to be explicit on the name, i.e not DEbugEditorViz.hlszl but PBRValidator.hlsl

com.unity.render-pipelines.high-definition/Runtime/Debug/LightingDebug.cs
```diff
@@ -14,6 +14,8 @@ public enum DebugLightingMode
         VisualizeShadowMasks,
         IndirectDiffuseOcclusion,
         IndirectSpecularOcclusion,
+        ValidateAlbedo,
```
sebastienlagarde: This enum is reserved to lighting related mode. Ideally we should display these mode on the Material panel. I can ask someone in our team to make the change if it is more convenient for you after we merge the PR.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -11,6 +11,10 @@
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/NormalBuffer.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/VolumeRendering.hlsl"
 
+#ifdef DEBUG_DISPLAY
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Debug/DebugEditorViz.hlsl"
```
sebastienlagarde: Add the include in DebugDisplay.hlsl not here

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/SimpleLit.hlsl
```diff
@@ -14,6 +14,10 @@
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/NormalBuffer.hlsl"
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/VolumeRendering.hlsl"
 
+#ifdef DEBUG_DISPLAY
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Debug/DebugEditorViz.hlsl"
```
sebastienlagarde: Not required, see previous comment

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
Wadarass: Yeah I was going back and forth on this.  If I do the validation with the surfaceData I would have to do the ComptueDiffuseColor an ComputeFresnel0 calls myself and this is already in the correct format in the bsdfData.  Wanted to reduce the bugs by using the already comptued values.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
Wadarass: If I use only the surfaceData I can move it.  it's just that BSDF data is different per material type so had to be included here.

com.unity.render-pipelines.high-definition/Runtime/Debug/LightingDebug.cs
Wadarass: @sebastienlagarde I think I should move this to full screen debug mode.  Then expose that in the material panel.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: This is less error prone to call ComptueDiffuseColor / ComputeFresnel0  from my point of view. 
Have in mind that ApplyDebugToBSDFData is call by all our Material and that most of them don't have mettalic workflow, so this is just pollution of prototype :)

com.unity.render-pipelines.high-definition/Runtime/Debug/LightingDebug.cs
sebastienlagarde: ok

Wadarass:
@sebastienlagarde  Updated the branch with your suggestions.

sebastienlagarde:
ok, I was not thinking about this when I say move to debug material.
I don't like too much that we need to add a GetPBRValidatorDebug() method all all materialsbut this can still ok. So let's accept it because currently I haven't better to propose :)

So I am good with the PR, just the name pbrMetalValidate (and for symetry Albedo)

Thanks

Wadarass:
@sebastienlagarde "So I am good with the PR, just the name pbrMetalValidate (and for symetry Albedo)"  What do you mean?

PR #2741, opened 2019-01-14
Modify GetPropertiesDeclaration to fix SRP batcher

laurenth-unity:
Move non exposed properties out of the CBUFFER "UnityPerMaterial" declaration.

### Purpose of this PR
This fixes SRP Batcher compatibility with global shader variables. Also non exposed properties are likely not to be per material.

---
### Testing status
**Katana Tests**: not tested on Katana

**Manual Tests**: Tested in a project that uses global shader variables for vertex position animation.
Before the fix the batcher would ignore vertex position animation (in play mode only), after the fix I can see the vertex position animation.

---
### Comments to reviewers
I don't know if this is how you want to do this. You could also test if the property is exposed and take that into account in order to determine if the property "isBatchable".


sebastienlagarde:
We need to move forward and I need to backport this to 2018.3, so merging.

PR #2747, opened 2019-01-15
Added multi-pass stereo rendering support to LWRP.

brandon-unity3d:
### Purpose of this PR
The purpose of this PR is to allow developers to choose multi-pass stereo rendering.  By doing so, developers will have an option of rendering in VR without requiring shader changes to support a single-pass stereo mode.  We also need to support this for various internal XR rendering refactoring.  For more information, ping me directly.

---
### Release Notes
LWRP now supports multi-pass stereo rendering.

---
### Testing status
I am running the katana tests now...

I tested with the Mock HMD.  I tested against all the built-in LWRP render passes and everything rendered as expected.

---
### Overall Product Risks
Low

Medium

Notes to Reviewers:
The multi-pass implementation shares shadow generation work between the two eyes.  Shadow maps, both for the main directional light and additional lights, are generated once per frame for both eyes.  Culling also happens once per frame for both eyes.

Some engine based changes need to be made to support this functionality.  The changes are in a separate ono based PR.

### DIFF
--- a/com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
+++ b/com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
@@ -7,13 +7,13 @@ namespace UnityEngine.Experimental.Rendering.LWRP
 {
     internal class ForwardRendererSetup : IRendererSetup
     {
-        private DepthOnlyPass m_DepthOnlyPass;
+        private DepthOnlyPass[] m_DepthOnlyPass;
+        private ScreenSpaceShadowResolvePass[] m_ScreenSpaceShadowResolvePass;
         private MainLightShadowCasterPass m_MainLightShadowCasterPass;
         private AdditionalLightsShadowCasterPass m_AdditionalLightsShadowCasterPass;
-        private SetupForwardRenderingPass m_SetupForwardRenderingPass;
-        private ScreenSpaceShadowResolvePass m_ScreenSpaceShadowResolvePass;
+        private SetupForwardRenderingPass[] m_SetupForwardRenderingPass;
         private CreateLightweightRenderTexturesPass m_CreateLightweightRenderTexturesPass;
-        private BeginXRRenderingPass m_BeginXrRenderingPass;
+        private BeginXRRenderingPass[] m_BeginXrRenderingPass;
         private SetupLightweightConstanstPass m_SetupLightweightConstants;
         private RenderOpaqueForwardPass m_RenderOpaqueForwardPass;
         private CreateColorRenderTexturesPass m_CreatePostOpaqueColorPass;
@@ -27,7 +27,7 @@ namespace UnityEngine.Experimental.Rendering.LWRP
         private CreateColorRenderTexturesPass m_createColorPass;
         private FinalBlitPass m_FinalBlitPass;
         private CapturePass m_CapturePass;
-        private EndXRRenderingPass m_EndXrRenderingPass;
+        private EndXRRenderingPass[] m_EndXrRenderingPass;
 
 #if UNITY_EDITOR
         private GizmoRenderingPass m_LitGizmoRenderingPass;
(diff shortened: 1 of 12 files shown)


phi-lira:
So, this becomes the fallback for VR in LWRP instead of Single Pass Instanced and we can optimize shader variant stripping?

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -229,11 +229,20 @@ public static void RenderSingleCamera(LightweightRenderPipeline pipelineInstance
                 InitializeRenderingData(settings, ref cameraData, ref cullResults,
                     renderer.maxVisibleAdditionalLights, renderer.maxPerObjectAdditionalLights, out var renderingData);
 
-                renderer.Clear();
-
                 var rendererSetup = setup ?? settings.rendererSetup;
-                rendererSetup.Setup(renderer, ref renderingData);
-                renderer.Execute(context, ref renderingData);
+
+                for (int eye = 0; eye < renderingData.totalEyes; ++eye)
```
phi-lira: It seems to me the only reason you require the allowViewIndepentSetup is because you do the eye loop here. Instead, if you do the eye loop in the ForwardRendererSetup then you can have all rendering steps/passes that are shared between eye (shadowmaps, setup light indices, etc) outside the loop. 

IMO that would make the code much cleaner and would require less changes.

com.unity.render-pipelines.lightweight/Runtime/Passes/AdditionalLightsShadowCasterPass.cs
```diff
@@ -50,6 +51,13 @@ public AdditionalLightsShadowCasterPass()
 
         public bool Setup(RenderTargetHandle destination, ref RenderingData renderingData, int maxVisibleAdditinalLights)
         {
+            m_CleanupAdditionalLightsShadowmapTexture = false;
+            if (!renderingData.allowViewIndepentSetup)
+            {
+                m_CleanupAdditionalLightsShadowmapTexture = true;
+                return true;
+            }
```
phi-lira: This is one of the examples. If you do the eye loop in the ForwardRendererSetup then you'll only call Setup and Execute for shadowmap passes once. Therefore this would not be required.

com.unity.render-pipelines.lightweight/Runtime/Passes/MainLightShadowCasterPass.cs
```diff
@@ -115,7 +123,7 @@ public override void FrameCleanup(CommandBuffer cmd)
             if (cmd == null)
                 throw new ArgumentNullException("cmd");
 
-            if (m_MainLightShadowmapTexture)
+            if (m_CleanupMainLightShadowmapTexture && m_MainLightShadowmapTexture)
```
phi-lira: Also this would not be required because the framecleanup would be called once both eyes are rendered.

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
```diff
@@ -49,18 +49,21 @@ internal class ForwardRendererSetup : IRendererSetup
         [NonSerialized]
         private bool m_Initialized = false;
 
-        private void Init()
+        private void Init(RenderingData renderingData)
```
phi-lira: Not used anymore

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
```diff
@@ -148,114 +149,70 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
-                    || Display.main.requiresBlitToBackbuffer
-                    || renderingData.killAlphaInFinalBlit;
+                        || m_AfterDepthpasses.Count != 0
+                        || m_AfterOpaquePasses.Count != 0
+                        || m_AfterOpaquePostProcessPasses.Count != 0
+                        || m_AfterSkyboxPasses.Count != 0
+                        || m_AfterTransparentPasses.Count != 0
+                        || m_AfterRenderPasses.Count != 0
+                        || Display.main.requiresBlitToBackbuffer
+                        || renderingData.killAlphaInFinalBlit;
```
phi-lira: This is not initialize at this point because you move the GetComponents calls further down.

com.unity.render-pipelines.lightweight/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -72,21 +74,41 @@ public override void Execute(ScriptableRenderer renderer, ScriptableRenderContex
...
+                        context.ExecuteCommandBuffer(cmd);
+                        ++eyeIndex;
+                    }
+                    else
+                    {
+                        for (int i = 0; i < 2; ++i)
+                        {
+                            Camera.StereoscopicEye eye = (Camera.StereoscopicEye)i;
+                            cmd.SetViewProjectionMatrices(camera.GetStereoViewMatrix(eye), camera.GetStereoProjectionMatrix(eye), true, i);
+                            context.ExecuteCommandBuffer(cmd);
```
phi-lira: I'm not sure I follow why this has to be done. Shouldn't beginXR setup all stereo matrices?

com.unity.render-pipelines.lightweight/Runtime/Passes/RenderOpaqueForwardPass.cs
```diff
@@ -83,6 +84,26 @@ public override void Execute(ScriptableRenderer renderer, ScriptableRenderContex
...
+                    {
+                        Camera.StereoscopicEye eye = (Camera.StereoscopicEye)eyeIndex;
+                        cmd.SetViewProjectionMatrices(camera.GetStereoViewMatrix(eye), camera.GetStereoProjectionMatrix(eye));
+                    }
+                    else
+                    {
+                        for (int i = 0; i < 2; ++i)
+                        {
+                            Camera.StereoscopicEye eye = (Camera.StereoscopicEye)i;
+                            cmd.SetViewProjectionMatrices(camera.GetStereoViewMatrix(eye), camera.GetStereoProjectionMatrix(eye), true, i);
```
phi-lira: Same here. We can't expect anyone that creates and injects a custom render pass to setup stereo matrices.

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
```diff
@@ -148,114 +149,70 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
-
-            foreach (var pass in m_AfterTransparentPasses)
-                renderer.EnqueuePass(pass.GetPassToEnqueue(baseDescriptor, colorHandle, depthHandle));
 
 #if UNITY_EDITOR
             m_LitGizmoRenderingPass.Setup(true);
-            renderer.EnqueuePass(m_LitGizmoRenderingPass);
 #endif
 
             bool afterRenderExists = m_AfterRenderPasses.Count != 0;
```
phi-lira: This will also break as you call GetComponent later.

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
```diff
@@ -265,60 +222,149 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
+
+                        renderer.EnqueuePass(m_FinalBlitPass);
+                    }
+                }
+
+                if (renderingData.cameraData.isStereoEnabled)
+                    renderer.EnqueuePass(m_EndXrRenderingPass);
+
+#if UNITY_EDITOR
+                renderer.EnqueuePass(m_UnlitGizmoRenderingPass);
```
phi-lira: render passes from this point this should be outside the eye loop

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererSetup.cs
```diff
@@ -265,60 +222,149 @@ public void Setup(ScriptableRenderer renderer, ref RenderingData renderingData)
...
+
+                if (renderingData.cameraData.isStereoEnabled)
+                    renderer.EnqueuePass(m_BeginXrRenderingPass);
+
+                if (requiresDepthPrepass)
+                {
+                    renderer.EnqueuePass(m_DepthOnlyPass);
+
+                    foreach (var pass in m_AfterDepthpasses)
+                        renderer.EnqueuePass(pass.GetPassToEnqueue(m_DepthOnlyPass.descriptor, m_DepthTexture));
```
phi-lira: It seems that GetPassToEnqueue should only be called once as well, just like setup. Otherwise this will create an instance of the renderpass per-eye.

sebastienlagarde:
Is this required for out of preview LW ? should we close the PR and do another one ?

sebastienlagarde:
Hey, is this PR still valid? should we close it?

phi-lira:
Closing this PR. Codebase is too different now and a new PR is recommended for this work. We can still search for it if required.

PR #2754, opened 2019-01-16
enable LIGHT_LAYERS keyword on deferred lighting and add it to forwar…

Seongdae:
### Purpose of this PR
Micro shadowing enabled with occlusion from bent normals, artifacts appeared. so I tried to enable light layers in asset, the specular occlusion was not transitioned to ambient occlusion. I found 'LIGHT_LAYERS' keyword was not enabled during LightLoop.

---
### Release Notes
Fix micro shadowing from ambient occlusion when 'LIGHT_LAYERS' is enabled.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fix_lightLayers&unity_branch=trunk&automation-tools_branch=master

**Manual Tests**:
1. Create any objects(in HDRP).
2. Set micro shadowing volume and enable it,
3. Replace codes with the below in Lit.hlsl
```
float ComputeMicroShadowing(BSDFData bsdfData, float NdotL)
{
#ifdef LIGHT_LAYERS
    return 1.0;
#else
    return 0.0;
#endif
}
```
4. check if the result is changed to white or black, when enable light layers in HDPipeline asset.

**Automated Tests**: None.

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers
I added keyword in forward pass and deferred shader and added kernels in compute shader.
It generates many shaders(in deferred.shader and deferred.compute).

Is it better to use ternary operation instead of using keyword? check please here.

for example
```
float ComputeMicroShadowing(BSDFData bsdfData, float NdotL)
{
#ifdef LIGHT_LAYERS
    return ComputeMicroShadowing(bsdfData.ambientOcclusion, NdotL, _MicroShadowOpacity);
#else
    // No extra G-Buffer for AO, so 'bsdfData.ambientOcclusion' does not hold a meaningful value.
    return ComputeMicroShadowing(bsdfData.specularOcclusion, NdotL, _MicroShadowOpacity);
#endif
}
```
to
```
float ComputeMicroShadowing(BSDFData bsdfData, float NdotL)
{
    float occlusion = _EnableLightLayers ? bsdfData.ambientOcclusion : bsdfData.specularOcclusion;
    return ComputeMicroShadowing(occlusion, NdotL, _MicroShadowOpacity);
}
```

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
@@ -43,6 +43,7 @@ Shader "Hidden/HDRP/Deferred"
             #pragma multi_compile _ OUTPUT_SPLIT_LIGHTING
             #pragma multi_compile _ SHADOWS_SHADOWMASK
             #pragma multi_compile _ DEBUG_DISPLAY
+            #pragma multi_compile _ LIGHT_LAYERS
 
             #define USE_FPTL_LIGHTLIST // deferred opaque always use FPTL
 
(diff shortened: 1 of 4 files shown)


sebastienlagarde:
Hi, 
Sorry, you are right, my initial fix was wrong, here is proper fix:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2757

We don't add a Light_LAYERS multicompile on purpose as it is extremely time consuming to compute all the extra variants and the deferred.compute is already too long to compile. In the case of deferred we rely on _EnableLightLayers instead of the define, and in the case of forward we always have LIGHT_LAYERS on, so we don't test for it.

I will close this PR as we don't want this behavior.

PR #2758, opened 2019-01-16
 These changelog entries were lost becase of 5.2.x branchoff.

phi-lira:
### Purpose of this PR
 Fix missing changelog entries that were lost because of 5.2.x branchoff.


phi-lira:
Upon speaking with Andre these changelog entries don't need to be on master as there were for patch releases on 5.2.0.

PR #2760, opened 2019-01-16
Updated default LWRP asset settings to favor optimization by default

phi-lira:
### Purpose of this PR
When creating a new LWRP asset it's now created with options that will favor performance by default in mobile. 

 * Disabled MSAA by default as this generally comes with a big cost in mobile.
 * Disabled shadow cascades by default as this forces resolving shadows in screen space, which in turn costs more bandwidth and requires a depth prepass.
 *  SRP batcher is enabled by default as this comes with huge performance wins in CPU bound games.
 * Disabled Dynamic Batching as this causes meshes to be batched in CPU every frame. 
 * Updated settings for the LWRP testing asset. 

 * TODO: We need to update templates. 
---
### Release Notes
 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=lw%2Fupdate-lwrp-default-settings&unity_branch=2019.1%2Fstaging

**Manual Tests**: Tested manually myself. 

**Automated Tests**: No new tests added.

---
### Overall Product Risks
**Technical Risk**: Low. Only affects users creating new lwrp asset.

**Halo Effect**: Low. Same as above.

---
### Comments to reviewers


phi-lira:
Katana is fialing on Windows tests due to timeout. This is a known issue.

PR #2766, opened 2019-01-16
[HDRP] Upgrade FrameSettings

RSlysz:
### Purpose of this PR
Refactor the FrameSettings workflow:
#### Less garbage collection:
Everything is made in struct now, except a small part when we want to use debug menu
This solve also the hacky workflow we used before with multiple instance of framesettings for runtime and serialized.
Now everything is quickly computed at the frame in most of the case.
If debug menu is in use, the history is kept in a collection to be able to display it in debug menu. This version add the equivalent of the previous runtime instance of the FrameSettings.
#### Easier to add new element in FrameSettings:
Now each inspector and debug menu are constructed the same way using the same source of information.
Almost every thing is generated from the enum FrameSettingsField.
In order to add a new entry in the FrameSettings, simply add an entry in this enum and add a [FrameSettingsField] attribute to specify how it should be displayed (order, aspect, indentation, etc.)
The only case that cannot be handled here is when we have specific restriction regarding the supported feature required to override it. In this case, you will still need to amend the info generated with AmmendInfo commend in FrameSettingsUI.Drawer.cs 
#### More element in FrameSettings:
It is now possible to go up to 128 elements. to add more, it only require to change the structure handling the data. The structure accepting 256 element (BitArray256) is already written. For more, you should write the new structure basing on previous ones.
This allow efficient quick computation of the aggregation but only handle booleans. It is still possible to add other kind of data in FrameSettings even if in this cleaning I have been able to remove every other type. But in this case, it could quickly become less efficient. Also, there is still no support for display for other type as before.
#### Various improvements
- It is now possible to multi edit FrameSettings (except on hdrp asset which have an issue on that that is beyond FrameSettings)
- It is possible to see the whole history of aggregation in debug menu to have additional hints in case of issues.
- Added DebuggerDisplay to show a humanized representation of the BitArray and visually see the content of the FrameSettings.
#### Side fixes:
- Fixed a reflection method for Camera or PlanarReflectionProbe preview where code have been upgraded recently in trunk (1aa572a) (2c67a99)
- Added auto retrieval of resources that are broken often by the serialisation  (0e6e09d)

---
### Release Notes
Upgraded FrameSettings workflow, DebugMenu and Inspector relative to it.

---
### Testing status
**Katana Tests**: 

**Manual Tests**: Tested locally, Build a player successfully, Tested test project before rebasing to lastest hdrp-master (it now have strange assembly issue after rebasing: error CS0234: The type or namespace name 'HDPipeline' does not exist in the namespace 'UnityEngine.Experimental.Rendering' (are you missing an assembly reference?))

**Automated Tests**: Added unit test on all BitArray structures.

---
### Overall Product Risks
**Technical Risk**: High (change numerous files)

**Halo Effect**: High (impact every renderer)

---


sebastienlagarde:
One test to validate:
- Open both scene and game view at the same time and try to manipulate HDRP asset and Debug menu value, is it reacting correctly?

RSlysz:
> * Open both scene and game view at the same time and try to manipulate HDRP asset and Debug menu value, is it reacting correctly?

Yes (disabling opaque remove opacu in both view), also tested adding-removing camera to scene, disabling-enabling them, modifying per renderer (camera or probe) FrameSettings.

However, since I updated trunk to last one, when I change something in HDRenderPipelineAsset (could be not linked to FrameSettings), I got this error without callstack:
Destroying GameObjects immediately is not permitted during physics trigger/contact, animation event callbacks or OnValidate. You must use Destroy instead.

sebastienlagarde:
>Fixed a reflection method for Camera or PlanarReflectionProbe preview where code have been upgraded >recently in trunk

be careful, is this in 2019.1/staging ? otherwise we need define here

RSlysz:
Just found issue when having several Scene View opened at the same time.

RSlysz:
> > Fixed a reflection method for Camera or PlanarReflectionProbe preview where code have been upgraded >recently in trunk
> 
> be careful, is this in 2019.1/staging ? otherwise we need define here

I'll check this. However, I think it is better to see if we can do a trunk pr to expose this instead on relying on easily breakable reflection.

com.unity.render-pipelines.high-definition/Editor/Core/Debugging/DebugUIDrawer.Builtins.cs
```diff
@@ -191,14 +290,36 @@ public override bool OnGUI(DebugUI.Widget widget, DebugState state)
...
             var s = Cast<DebugStateBool>(state);
-
+            
             EditorGUI.BeginChangeCheck();
 
-            bool value = EditorGUILayout.Foldout(w.GetValue(), EditorGUIUtility.TrTextContent(w.displayName), true);
+            Rect rect = PrepareControlRect();
+            bool value = EditorGUI.Foldout(rect, w.GetValue(), EditorGUIUtility.TrTextContent(w.displayName), true);
+
+            //bool value = EditorGUILayout.Foldout(w.GetValue(), EditorGUIUtility.TrTextContent(w.displayName), true);
```
Chman: Clean this out.

com.unity.render-pipelines.high-definition/Runtime/Core/Debugging/FrameSettingsFieldAttribute.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+    public static partial class StringExtention
+    {
+        public static string CamelToPascalCaseWithSpace(this string text, bool preserveAcronyms = true)
```
Chman: Why do we need this function? Can't we just use `ObjectNames.NicifyVariableName`?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -1,504 +1,371 @@
...
-            this.lightLoopSettings.CopyTo(frameSettings.lightLoopSettings);
-
-            frameSettings.Refresh();
+            get => bitDatas[(uint)FrameSettingsField.LitShaderMode] ? LitShaderMode.Deferred : LitShaderMode.Forward;
+            set => bitDatas[(uint)FrameSettingsField.LitShaderMode] = value == LitShaderMode.Deferred;
         }
 
-        public void ApplyOverrideOn(FrameSettings overridedFrameSettings)
+        /// <summary>Get stored data for this field.</summary>
+        public bool IsEnable(FrameSettingsField field) => bitDatas[(uint)field];
```
Chman: Should be `IsEnabled`

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -1,504 +1,371 @@
...
+        /// <summary>Set stored data for this field.</summary>
+        public void SetEnable(FrameSettingsField field, bool value) => bitDatas[(uint)field] = value;
```
Chman: `SetEnabled`

com.unity.render-pipelines.high-definition/Runtime/Core/Debugging/FrameSettingsFieldAttribute.cs
RSlysz: ObjectNames.NicifyVariableName is in UnityEditor but I need this in UnityEngine side for runtime debug menu UI

RSlysz:
@Chman Fixed all in fd3d368

RSlysz:
@sebastienlagarde Checked that same API modification have been done in 2019.1/staging. Thus no need for define.

sebastienlagarde:
I fixed issue with test, remaining test broken are the one on hdrp-master

PR #2772, opened 2019-01-17
Add SH to TerrainLit shader

Verasl:
### Purpose of this PR
This PR adds support for the terrain SH, sampling per vertex if no normal maps are used and per pixel if normal maps are being used.

---
### Release Notes
Added SH sampling on terrain when not light mapped.

---
### Testing status
**Katana Tests**: Ran locally and passed, don't have any specific tests that make use of this as we need a wider range of terrain tests

**Manual Tests**:Checked terrain with both normal maps and no normal maps, both had lighting from the ambient probe.

**Automated Tests**: Have a test that will be getting added to both graphics and performance projects.

---
### Overall Product Risks
**Technical Risk**: Low, there will be a slowdown on mobile when using normals as it will be additional calculations per pixel, if it is deemed too much a define can be added later but will add to variant count so for now this is the best course.

**Halo Effect**: None, only affects the TerrainLit shader, and only the forward pass.


com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -55,10 +56,12 @@ void InitializeInputData(VertexOutput IN, half3 normalTS, out InputData input)
     input = (InputData)0;
 
     input.positionWS = IN.positionWS;
+    half3 SH = 0;
```
phi-lira: hum. I think I've seen an issue related to some compilers not liking this. Can't be sure but it's safer to initialize it as half3(0, 0, 0).

com.unity.render-pipelines.lightweight/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -55,10 +56,12 @@ void InitializeInputData(VertexOutput IN, half3 normalTS, out InputData input)
...
 
 #if defined(_NORMALMAP) && !defined(ENABLE_TERRAIN_PERPIXEL_NORMAL)
     half3 viewDirWS = half3(IN.normal.w, IN.tangent.w, IN.bitangent.w);
     input.normalWS = TransformTangentToWorld(normalTS, half3x3(IN.tangent.xyz, IN.bitangent.xyz, IN.normal.xyz));
+    OUTPUT_SH(input.normalWS.xyz, SH);
```
phi-lira: OUTPUT_SH should be only called in vertex. This will call SampleSHVertex. If you want to sample SH fully per-pixel here do SampleSH inside if !defined(LIGHTMAP).

Verasl: Changed to SampleSH but left in #if defined(_NORMALMAP) since the per vertex is reliant on there being no normals

PR #2773, opened 2019-01-17
Update RTEnvironment checking

nerites:
The current RTEnvironment checking code tries to access raytracedObjects even for null rtEnvironments, which causes breaking errors for scenes without RTEnvironments. The diff here looks ugly but I just bracketed the rest of the code in an if statement that first checks if rtEnvironment exists, THEN checks raytracedObjects.


nerites:
I tried to change the base of this PR to hdrp-master since Anis's branch got merged in, but it looks like the changesets that should've been merged in from Anis's branch still appear in the diff for this PR, so I'm leaving as-is for readability.

PR #2788, opened 2019-01-22
Reduce inheritance and now-useless interfaces in core Shader Graph code

pbbastian:
### Purpose of this PR
`IGraph` and `INode` are relics from the past that no longer serve a purpose, so they have been removed in this PR. I also reduced `AbstractMaterialGraph`, `MaterialGraph`, and `SubGraph` into `GraphData`. There were 3 special cases that have been taken care of with branches.

`IShaderProperty` is no longer. `AbstractShaderProperty<T>` now has a non-generic base class `AbstractShaderProperty` that essentially replaces the interface.

Also, this bumps the import queue for Shader Graphs (not sub graphs), ensuring that they are always imported after sub graphs. 

---
### Release Notes
Nothing user facing :) 

---
### Testing status
**Katana Tests**: On the way :) 

**Manual Tests**: Create new graph, add node, open existing test graphs

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
Loooots of auto refactoring going on here, so this won't be possible to verify by hand, but do have a glance :)


marctem:
Definitely in favor of getting rid of INode since it only has one implementation.

I don't like how now IMasterNode and AMN are related but not in a compiler-enforceable way. Not sure how we could do that, but it feels icky to me as-is.

I'm fine getting this in asap before we get too many other conflicting PRs!

👍

pbbastian:
Great, I'll try to get this in today 👍 Agree about `IMasterNode`. We could introduce a base class `MasterNode` for `MasterNode<T>`, and then get rid of `IMasterNode` since there's no other implementations of `IMasterNode`.

PR #2724, opened 2019-01-10
Update for 32 threads in wavefronts on Switch.

kaychang-unity:
### Purpose of this PR
This PR improves performance on Switch for the light tile classification pass (saves 1.2ms).

---
### Release Notes
-

---
### Testing status
Didn"t run Katana tests. Switch specific code change.

**Manual Tests**: What did you do?
I ran a couple of test scenes from HDRP_Tests.

**Automated Tests**: What did you setup?
-

Any test projects to go with this to help reviewers?

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High?
Low.

**Halo Effect**: None, Low, Medium, High?
Low.

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #2793, opened 2019-01-22
Fix shader compilation issue 

FrancescoC-unity:
Not sure when this went in, however I just merged hdrp-master to my branch and it didn't compile for consoles. 
The way WaveGetLaneCount() was defined doesn't seem to compile. Moreover, when we are going to support SM 6.0, this won't have compiled anyway since we cannot use functions to define groupshared 
 mem array size.  

I changed that to a constant, but I kept the function  there anyway, in case we want to use it in the future 🙂


sebastienlagarde:
This have been broken in this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2724

PR #2799, opened 2019-01-22
remove condition that the material needs to be transparent to show sh…

laurenth-unity:
### Purpose of this PR
Shadergraph shaders using Hair shader in opaque mode with cutoff cannot  compile before this fix.
This fixes shadow cutoff threshold slot being missing in opaque mode.

---
### Testing status
**Katana Tests**: Small change didn't launch tests

**Manual Tests**: I checked my shadergraph shader and it compiles again now.


EvgeniiG:
Included into this PR:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2807

PR #2800, opened 2019-01-23
Fix HitT generation in AO

nerites:
### Purpose of this PR
Previously the HitT texture was always getting written to 0 since minDist was initialized to 0f.
Stencil still needs to be hooked up to raytrace shader but this at least puts meaningful data into HitT.


anisunity:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2802

PR #2805, opened 2019-01-23
Dynamic resolution v1

FrancescoC-unity:
### Purpose of this PR
This PR introduce a first version of Dynamic Resolution for HDRP. At the moment only the software version works. Seems to be fine, but a good QA pass would be nice.

The hardware version has some problems on C++ side as far as I understand. However, most of the code for HDRP is there, just waiting for the platform side of things to be resolved. 

---
### Release Notes
First implementation of dynamic resolution support for HDRP.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=dynamic-resolution-v1&automation-tools_branch=add-platform-filter&unity_branch=trunk

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using UnityEngine.Assertions;
 using UnityEngine.Rendering;
+using UnityEngine.Experimental.Rendering.HDPipeline;
```
Chman: I guess the RTHandle system is now HDRP-specific. cc @JulienIgnace-Unity

FrancescoC-unity: Or I can move the dynamic res handler outside of Rendering.HDPipeline

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FinalPass.shader
```diff
@@ -8,9 +8,14 @@ Shader "Hidden/HDRP/FinalPass"
         #pragma multi_compile _ FXAA
         #pragma multi_compile _ GRAIN
 
+        #pragma multi_compile NO_UPSCALE BILINEAR CATMULL_ROM_4 LANCZOS POINT
```
Chman: Remove `NO_UPSCALE` and use `_` instead, it's free and doesn't require another keyword.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    accumulation += float4(Bilinear(_InputTexture, float2(UV0.x, UV_1.y)), 1) * weight23.x * weight1.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV0.x, UV0.y)), 1) * weight23.x * weight23.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV0.x, UV2.y)), 1) * weight23.x * weight4.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV0.x, UV3.y)), 1) * weight23.x * weight5.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV2.x, UV_1.y)), 1) * weight4.x * weight1.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV2.x, UV0.y)), 1) * weight4.x * weight23.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV2.x, UV2.y)), 1) * weight4.x * weight4.y;
+    accumulation += float4(Bilinear(_InputTexture, float2(UV3.x, UV0.y)), 1) * weight5.x * weight23.y;
+
+    return accumulation /= accumulation.w;
```
Chman: Implicit trunc (float4 -> float3)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -0,0 +1,111 @@
...
+
+float3 Nearest(TEXTURE2D(_InputTexture), float2 UV)
+{
+    return Sample(_InputTexture, s_point_clamp_sampler, UV);
+}
+
+
+float3 Bilinear(TEXTURE2D(_InputTexture), float2 UV)
+{
+    return Sample(_InputTexture, s_linear_clamp_sampler, UV);
```
Chman: Unsafe, as you used `TEXTURE2D_ARGS` in the `Sample` function you have to use `TEXTURE2D_PARAM` in case the sampler part is stripped on a platform.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -0,0 +1,111 @@
...
+
+float3 Sample(TEXTURE2D_ARGS(_InputTexture, _InputTextureSampler), float2 UV)
+{
+    float2 ScaledUV = min(UV, 1.0f - 0.5f * _ScreenSize.zw) * _ScreenToTargetScale.xy;
+    return SAMPLE_TEXTURE2D_LOD(_InputTexture, _InputTextureSampler, ScaledUV, 0).xyz;
+}
+
+float3 Nearest(TEXTURE2D(_InputTexture), float2 UV)
+{
+    return Sample(_InputTexture, s_point_clamp_sampler, UV);
```
Chman: See the comment below.

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
sebastienlagarde: currently yes, RT handle is in HDRP/Core

com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -298,7 +298,7 @@ void ScreenSpaceReflectionsReprojection(uint2 positionSS : SV_DispatchThreadID)
     // TODO: it's important to account for occlusion/disocclusion to avoid artifacts in motion.
     // This would require keeping the depth buffer from the previous frame.
     float2 velocityNDC;
-    DecodeVelocity(SAMPLE_TEXTURE2D_LOD(_CameraMotionVectorsTexture, s_linear_clamp_sampler, hitPositionNDC * _ScreenToTargetScale.zw, 0), velocityNDC);
+    DecodeVelocity(SAMPLE_TEXTURE2D_LOD(_CameraMotionVectorsTexture, s_linear_clamp_sampler, min(hitPositionNDC, 1.0f - 0.5f * _ScreenSize.zw) * _ScreenToTargetScale.zw, 0), velocityNDC);
```
sebastienlagarde: is this something use at other place ? should we do a function for this or it is specific to SSR ?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -1875,6 +1894,17 @@ void ComputeSplitToning(out Vector4 shadows, out Vector4 highlights)
 
         #endregion
 
+        #region FXAA
```
sebastienlagarde: Don't use #region (it is in the norm doc of unity).
(not my choice :) )

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomUpsample.compute
```diff
@@ -28,14 +28,14 @@ CBUFFER_END
 void MAIN(uint2 dispatchThreadId : SV_DispatchThreadID)
 {
     PositionInputs posInputs = GetPositionInput(float2(dispatchThreadId), _TexelSize.zw, uint2(GROUP_SIZE, GROUP_SIZE));
-    float2 uv = posInputs.positionNDC * _ScreenToTargetScale.xy;
-
+    float2 maxCoord = (1.0f - _TexelSize.zw) * _ScreenToTargetScale.xy;
```
sebastienlagarde: as seen before, I think it should be a function.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FXAA.hlsl
```diff
@@ -0,0 +1,88 @@
+﻿#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl"
+
```
sebastienlagarde: don't we have already a FXAA. Haven't Thomas add one? or it is different ?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -0,0 +1,111 @@
...
+    return SampleTexture2DBicubic(TEXTURE2D_PARAM(_InputTexture, s_linear_clamp_sampler), UV * _ScreenToTargetScale.xy, bicubicWnd, (1.0f - 0.5f * _ScreenSize.zw) * _ScreenToTargetScale.xy);
+}
+
+
+
+// https://en.wikipedia.org/wiki/Lanczos_resampling
+// TODO: Revisit derivation.
+float3 Lanczos(TEXTURE2D(_InputTexture), float2 inUV)
+{
+    // Lanczos 3
```
sebastienlagarde: hmm... I think it should be something  to move to a shader library in filtering.hlsl or something like that. Also have you consider a bicubic?
http://vec3.ca/bicubic-filtering-in-fewer-taps/

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Chman: We use it in the post-processing file to keep the file organized (instead of doing what we generally do, that is "File.Subsection.cs" which is getting out of hand in HDRP right now). I know it's not allowed in the internal Unity conventions but... Here's to hoping it'll change for the "external" conventions.

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/ShadingModels/SimpleLitGUI.cs
```diff
@@ -1,156 +1,156 @@
...
-                    color = "_BaseColor";
-
-                var col = material.GetColor(color);
-                col.a = material.GetFloat("_Smoothness");
-                material.SetColor(color, col);
-            }
-        }
-    }
-}
+﻿using System;
```
sebastienlagarde: unexpected modifiaction ?

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/ShadingModels/BakedLitGUI.cs
```diff
@@ -1,26 +1,26 @@
...
-            }
-        }
-
-        public static void Inputs(BakedLitProperties properties, MaterialEditor materialEditor)
-        {
-            BaseShaderGUI.DrawNormalArea(materialEditor, properties.bumpMapProp);
-        }
-    }
-}
+﻿using System;
```
sebastienlagarde: unexpected modifiaction ?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/FXAA.hlsl
Chman: It's the same FXAA, the difference being that we do FXAA in the final pass when no dynamic resolution is used, else we need to do it before the final upsampling/downsampling pass (faster than trying to FXAA every sample needed for upsampling/downsampling in the final pass).

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
Chman: Bicubic is already in and used as well (see `CatmullRomFourSamples`).
As for moving Lanczos in Filtering we talked about it but we need to validate the implementation first.

com.unity.render-pipelines.lightweight/Editor/ShaderGUI/ShadingModels/SimpleLitGUI.cs
Chman: These changes (and in the following file) appear in all our PRs for some reason. Same issue as the ShaderGraph files (line ending stuff?).

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/BloomUpsample.compute
FrancescoC-unity: It can, but it'd need to accept so many params that it won't necessarily be much cleaner. 

ClampUV(float2 UV, float2 TexelSize, float numberOfTexels)

So we'd be turning 

min(UV, 1.0f - x * TexelSize) 

to ClampUV(UV, TexelSize, x) which might not necessarily look more clear. In particular, I'd get confused by the x parameter. 

I guess, since the most common case is by far half a texel clamp, I could make that a function and keep the other cases as explicit.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
sebastienlagarde: This is what everyone ask for :) But then it is not a convention anymore.

PR #2823, opened 2019-01-24
HDRP - Fix instancing macro usage for decals

fabien-unity:
### Purpose of this PR
Fix inconsistent naming in instancing macros for decals.

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp-fix-instancing-macro-decals&unity_branch=trunk)

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers
I found this issue while working on stereo instancing.
I also removed _DecalAtlas because it's not referenced anywhere.


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
```diff
@@ -21,6 +18,6 @@ TEXTURE2D(_DecalHTileTexture);
 
 UNITY_INSTANCING_BUFFER_START(Decal)
```
Paulm-Unity: So I guess BUFFER_START doesn't care about the name?

fabien-unity: Well yes it matters, UNITY_INSTANCING_BUFFER_START and UNITY_INSTANCING_BUFFER_END should match.

Paulm-Unity: I would've thought so too. I am surprised this actually worked and doesn't produce a shader compiler error...

PR #2824, opened 2019-01-24
Fix LOAD_TEXTURE2D_ARRAY_MSAA macro

fabien-unity:
### Purpose of this PR
Fix incorrect macro for MSAA texture arrays

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp-msaa-texarray-macros&unity_branch=trunk)

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers
[Reference HLSL doc](https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/sm5-object-texture2dmsarray-load)


fabien-unity:
To be clear, these macros are not currently in used (they will with stereo-instancing) so nobody noticed they did not work.

PR #2826, opened 2019-01-24
Rename SampleCameraDepth to LoadCameraDepth and SampleCameraDepth

fabien-unity:
### Purpose of this PR
The same name "SampleCameraDepth" was used with different inputs signature (uint2 or float2), making this function very unsafe to use because of automatic casting.
I've renamed it to be more explicit (SS vs UV) and also replace all occurrences of _CameraDepthTexture sampling to use these functions in order to make the code more maintainable and more friendly for stereo-instancing.

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrp-rename-samplecameradepth&unity_branch=trunk)

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers
No new features here, just cleanup and small refactor.


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -344,14 +344,14 @@ CBUFFER_END
 
 // Note: To sample camera depth in HDRP we provide these utils functions because the way we store the depth mips can change
 // Currently it's an atlas and it's layout can be found at ComputePackedMipChainInfo in HDUtils.cs
-float SampleCameraDepth(uint2 pixelCoords)
+float SampleCameraDepthSS(uint2 pixelCoords)
 {
     return LOAD_TEXTURE2D_LOD(_CameraDepthTexture, pixelCoords, 0).r;
```
EvgeniiG: The name is misleading because the Sample() function performs a Load() instead of a Sample() operation.

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -104,6 +104,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Micro-shadowing in Lit forward now use ambientOcclusion instead of SpecularOcclusion
 - Upgraded FrameSettings workflow, DebugMenu and Inspector part relative to it
 - Update build light list shader code to support 32 threads in wavefronts on Switch
+- Rename SampleCameraDepth to SampleCameraDepthSS and SampleCameraDepthUV
```
EvgeniiG: General comment (IMHO), not a problem with this implementation per se.

I would prefer not to have another layer of abstraction for texture sampling.
We already have one layer with macros.
With the proposed design, we would have to implement 2 functions per texture.
There are 3 issues with this.

1. For every new texture, 2 functions have to be written.
2. Before using a texture, one cannot follow the common pattern of LOAD_TEXTURE2D_ARRAY(textureName, unCoord2, EYE_INDEX), and instead has to find (or write from scratch) a function corresponding to this particular texture.
3. The functions hide implementation details, which is not something I am really fond of.

I understand the motivation, and the fact that the feasibility of using array textures for everything is still undetermined. With this in mind, personally, I think that we should probably wait with changes like this until we can figure out if we can just use arrays for everything stereo-related (which would be my preferred approach).

fabien-unity: As discussed on slack, here's the proposed design to handle texture array:https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/hdrp-xr-texture2dx

Note that you don't need to write 2 functions to use a new texture, it's only the case for depth and color because it's a very common pattern.

I agree that the macros hide some implementations details though, but if we want to be able to support tex array (as optional), I don't see a clever way. We could be very explicit and add a ifdef/else for every textures but the code would become more polluted.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
fabien-unity: I've renamed the functions to Load / Sample

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MSAA/ColorResolve.shader
```diff
@@ -29,32 +29,37 @@ Shader "Hidden/HDRP/ColorResolve"
             return output;
         }
 
+        float4 SampleColorTexture(float2 pixelCoords, uint sampleIndex)
```
JulienIgnace-Unity: Shouldn't this be LoadCoorTexture for consistency?

fabien-unity: I agree! Done.

PR #2831, opened 2019-01-25
Switch support

kaychang-unity:
### Purpose of this PR
This PR adds support for fragment shader running in place of Deferred.compute, with exact same tiling/clustering rendering support.
This is achieved by replacing calls to DispatchIndirect() by DrawProceduralIndirect() and let the vertex shader generates the tile locations via instancing.

On platforms that support quad topology, the vertex shader will generate real quads. On other platforms, it will generate 2 triangles. Unity does not fully expose graphics capabilities via the C# interface, so this is hardcoded per platform instead (currently only Switch).

SSS (split-lighting codepath) is working using the fragment shader version.

I haven't been able to test VR support but the fragment shader code should be the same.

At the moment, the fragment shader path is activated only on Switch and in the editor if Switch platform is the active target.

---
### Release Notes

---
### Testing status
Katana test in progress https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=switch-support&automation-tools_branch=add-platform-filter&unity_branch=trunk#

I tested a couple of scenes in HDRP_Tests: 2002_Dynamic_Mix.unity, 1301_SubSurfaceScattering.unity

---
### Overall Product Risks
Low: no functional changes in other platforms.

None.

---
### Comments to reviewers
Should Deferred.shader become legacy? In that case, we may want to replace k_PreferFragment by a new bool FrameSettingsField.FragmentLightEvaluation?


kaychang-unity:
Katana tests are failing but this is due to terrain shaders (same known error as on master branch).

sebastienlagarde:
Please launch katana on 2019.1/staging instead of trunk, so we it will avoid the current issue

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -300,7 +321,7 @@ public void Allocate()
 
         const bool k_UseDepthBuffer = true;      // only has an impact when EnableClustered is true (requires a depth-prepass)
 
-#if UNITY_SWITCH
+#if !UNITY_EDITOR && UNITY_SWITCH
```
sebastienlagarde: why "!UNITY_EDITOR &&" ? isn't if switch sufficient here ? 
or you mean UNITY_SWITCH could be define in editor when you switch to this platform ? (which make no sense to me, for other console we could have trouble too);

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2400,17 +2452,37 @@ public void BuildGPULightListsCommon(HDCamera hdCamera, CommandBuffer cmd, Rende
                     cmd.DispatchCompute(buildMaterialFlagsShader, buildMaterialFlagsKernel, numTilesX, numTilesY, 1);
                 }
 
+                #pragma warning disable 0162 // disable unreachable code warning
```
sebastienlagarde: just a tough, when adding pragma like this, it become add to debug the code as the debugger for unkown reasons isn't able to return the correct line number. my suggestion is to not use a constant for k_PreferFragment but just a static bool (like we do in shader config.). If it is not a constant, no warning are trigger and we can still debug at correct line properly in C# in this files.

sebastienlagarde: And as we may extent the code later to other platform, it make sense

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
kaychang-unity: So the target platform is Switch and we hit the Play button in the Editor, UNITY_SWITCH will be defined in the C# script when the graphics backend is D3D. So actually, the group thread count will be 64 and not 32. The !UNITY_EDITOR is needed to make sure we are really running on target, instead as from the editor.

kaychang-unity:
No problem! deferredTile.shader should have the same variants as deferred.shader. The only difference may be setting the blend states. I may fix that in the next PR.

kaychang-unity:
I didnt see you latest cleanup code. So blend states are now the same and the code is completely similar. Cheers!

PR #2859, opened 2019-01-29
Allow for transparent objects to write to velocity [Wait C++ PR]

FrancescoC-unity:
### Purpose of this PR

With this PR we expose an option for transparent object to write into velocity buffer. The content of the velocity buffer for the relevant pixels will be overridden. 

Also, use the correct depth (containing transparent depth pre/post pass info) for post processing.  ([this line](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2859/files#diff-f12428c0ee2f09917335196b987521d7R1790)) 

---
### Release Notes
Expose option to allow transparent object to override the content of the velocity buffer. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Ftransparent-writes-velocity&automation-tools_branch=add-platform-filter&unity_branch=trunk

**Manual Tests**:  Tested both Lit via inspector, via shadergraph and hair via shadergraph. Also tested that nothing gets rendered when motion vectors are off globally. 

---
### Overall Product Risks
**Technical Risk**: Low, 

**Halo Effect**:  Low


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1781,6 +1786,9 @@ void ExecuteRenderRequest(RenderRequest renderRequest, ScriptableRenderContext r
 
                 if (hdCamera.frameSettings.IsEnabled(FrameSettingsField.Postprocess))
                 {
+                    // Set the depth buffer to the main one to avoid missing out on transparent depth for post process.
+                    cmd.SetGlobalTexture(HDShaderIDs._CameraDepthTexture, m_SharedRTManager.GetDepthStencilBuffer());
```
FrancescoC-unity: This is not strictly related, but it is needed regardless.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2594,7 +2599,13 @@ void RenderForward(CullingResults cullResults, HDCamera hdCamera, ScriptableRend
                 }
                 else
                 {
-                    HDUtils.SetRenderTarget(cmd, hdCamera, hdCamera.frameSettings.IsEnabled(FrameSettingsField.MSAA) ? m_CameraColorMSAABuffer : m_CameraColorBuffer, m_SharedRTManager.GetDepthStencilBuffer(hdCamera.frameSettings.IsEnabled(FrameSettingsField.MSAA)));
+                    bool renderVelocitiesForTransparent = hdCamera.frameSettings.IsEnabled(FrameSettingsField.MotionVectors) && hdCamera.frameSettings.IsEnabled(FrameSettingsField.TransparentsWriteVelocity);
+                    cmd.SetGlobalInt(HDShaderIDs._ColorMaskTransparentVel, renderVelocitiesForTransparent ? (int)UnityEngine.Rendering.ColorWriteMask.All : 0);
```
sebastienlagarde: Is this working?  You can set a value for a render state globally like that ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2606,6 +2617,12 @@ void RenderForward(CullingResults cullResults, HDCamera hdCamera, ScriptableRend
                     {
                         transparentRange = HDRenderQueue.k_RenderQueue_AllTransparent;
                     }
+
+                    if(renderVelocitiesForTransparent)
```
sebastienlagarde: if (

Convention :)

FrancescoC-unity: Ha, good spot, got distracted 🙂

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: Yup, works fine. I tried manually to enable/disable motion vector (and therefore this flag) and the results are as expected.

FrancescoC-unity:
Gah, my git config on this repo is all fucked up. Do not merge this, I will create a fresh branch when the C++ PR lands

PR #2864, opened 2019-01-30
Added guards so that MATERIAL_INCLUDE_SUBSURFACESCATTERING and MATERI…

kaychang-unity:
…AL_INCLUDE_TRANSMISSION can be commented out.

### Purpose of this PR
This PR fixes the shader code so that MATERIAL_INCLUDE_SUBSURFACESCATTERING and MATERIAL_INCLUDE_TRANSMISSION can be commented out and still compile.

---
### Release Notes
-

---
### Testing status
Didn't run Katana tests.

This is running in the editor and for Switch target.

-

---
### Overall Product Risks
None.

None

---
### Comments to reviewers
This will allow some platforms to easily comment out effects they do not want in their shader code.
At the moment, some instructions are still added to the shader code (deferred.compute, deferredTile.shader) even when subSurfaceScattering or transmission shader variants are not enabled. Completely removing these instruction saves a fraction of a ms on Switch (decreased register pressure).

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Deferred.shader
@@ -140,12 +140,14 @@ Shader "Hidden/HDRP/Deferred"
                 Outputs outputs;
 
             #ifdef OUTPUT_SPLIT_LIGHTING
+                #ifdef MATERIAL_INCLUDE_SUBSURFACESCATTERING
                 if (_EnableSubsurfaceScattering != 0 && ShouldOutputSplitLighting(bsdfData))
                 {
                     outputs.specularLighting = float4(specularLighting, 1.0);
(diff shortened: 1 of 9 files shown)


sebastienlagarde:
With recent change of @FrancescoC-unity (no light list when just dir light + extra variant with only std lit) , I don't think this PR is need. Or as I mentionned, if we don't have fast path for standard lit shader, it mean we have an issue. I let Francesco reply and close this PR if we are agree it is useless.

FrancescoC-unity:
I agree this should be needed at all, however, I'll check to be sure nothing funny is going on. Will update the PR when I have a result.

FrancescoC-unity:
Verified on xbox, both the compute and pixel shader version of deferred, and as expected couldn't notice any variation in VGPR. I am assuming the difference was for when we didn't have the extra variants.

I'll close this, please let me know if you have data that goes against what I found

PR #2876, opened 2019-02-01
Added a pass for rendering objects after Post-Process

JulienIgnace-Unity:
### Purpose of this PR
Added a pass for rendering objects after Post-Process
This implied several other changes in the render loop
- Removed separate blit from post process to better share code. When posts are disabled, FinalPass is still used
- In Dev builds, added an intermediate render target between posts and final blit. This allows us to render debugs with the same Y orientation and only flip during the last blit. In Non dev builds, flip is done in the post process final pass as usual.
---
### Release Notes
Added a pass for rendering objects after Post-Process.

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fafter-post-process


com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl
```diff
@@ -28,8 +28,15 @@
 
 float4 ApplyBlendMode(float3 diffuseLighting, float3 specularLighting, float opacity)
 {
+
+#ifndef _SURFACE_TYPE_TRANSPARENT
+    // When doing off screen rendering (low-res transparency or after post process pass) and rendering opaque objects, opacity needs to be set to zero for proper compositing.
+    if (_OffScreenRendering != 0)
+        opacity = 0;
```
sebastienlagarde: I don't understand why this is here, it should only be in forwar unlit pass .hlsl

PR #2878, opened 2019-02-01
Change TAA group size on Xbox

FrancescoC-unity:
The shader runs about 20% faster on Scorpio (1.25 -> 1.02) with this group size. On Durango and PS4 it makes no difference.


Chman:
[Testing on PS4 Pro, if it's faster there as well just switch everything to 16,16 instead of doing it only for Xbox]

sebastienlagarde:
update changelog

FrancescoC-unity:
> [Testing on PS4 Pro, if it's faster there as well just switch everything to 16,16 instead of doing it only for Xbox]

Nope is not :)

PR #2890, opened 2019-02-04
Rearrange FetchDepth functions

fabien-unity:
### Purpose of this PR
Rearrange FetchDepth functions to prepare for stereo-instancing.

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=add-platform-filter&ScriptableRenderLoop_branch=hdrpx-xr-fetchdepth&unity_branch=trunk)

**Manual Tests**: local test on Windows 10, with HDRP tests suite

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
This change is needed to handle stereo-instancing macros override.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/ShaderBase.hlsl
```diff
@@ -1,38 +1,44 @@
...
-    return zdpth;
-}
-
-float FetchDepthMSAA(Texture2DMS<float> depthTexture, uint2 pixCoord, uint sampleIdx)
-{
-    float zdpth = LOAD_TEXTURE2D_MSAA(depthTexture, pixCoord.xy, sampleIdx).x;
-#if UNITY_REVERSED_Z
-    zdpth = 1.0 - zdpth;
+#ifdef MSAA_ENABLED
+    Texture2DMS<float> g_depth_tex : register( t0 );
```
sebastienlagarde: I will say, let's clean this code and let's move this FetchDepth to shadervariables.hlsl after  "SampleDepth' definition and let's use _CameraDepthTexture instead of g_depth_tex. Thought ?

Also I guess we should rename function: LoadCameraDepth (inversion of Z) and LoadRawCameraDepth  (no inversion). Thought ? @JulienIgnace-Unity  @alelievr (regarding the shadergraph) ?

fabien-unity: So actually, I remember trying to include ShaderVariables.hlsl in the lightlist compute shaders at first but I've run into variables redefinition issues from the light loop (RW versus read-only).
We can workaround those issues but I'm not sure it's going to be cleaner.

sebastienlagarde: ok. Let's see for another PR then. Can you fix the changelog :)

fabien-unity: done! :) One day I'll learn.

PR #2912, opened 2019-02-07
Fix hair lighting

EvgeniiG:
### Purpose of this PR
Completely re-do hair shading.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Ffix-hair-lighting&automation-tools_branch=master&unity_branch=trunk

**Manual Tests**: Yibing & Morten.

---
### Overall Product Risks
**Technical Risk**: 0.

---
### Comments to reviewers
None.


sebastienlagarde:
Schedule to be tested by @laurenth-unity

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
```diff
@@ -33,7 +33,7 @@ Pass
     // Variant Definitions (active field translations to HDRP defines)
     //-------------------------------------------------------------------------------------
     $Material.KajiyaKay:                #define _MATERIAL_FEATURE_HAIR_KAJIYA_KAY 1
-    $Material.Transmission:             #define _MATERIAL_FEATURE_TRANSMISSION 1
+    /* Always on */                     #define _MATERIAL_FEATURE_TRANSMISSION 1
```
mmikk: I assume it's still possible to turn it off using the texture map?

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
```diff
@@ -244,6 +232,33 @@ $include("SharedCode.template.hlsl")
...
+        // there is a whole "circle" of normals corresponding to any given tangent vector.
+        // Since we use the Kajiya-Kay shading model,
+        // the way we compute and use normals is a bit complicated.
+        // We need 4 separate sets of normals.
+        // For shadow bias, we use the geometric normal.
+        // For direct lighting, we (conceptually) use the "light-facing" normal.
+        // For reflected GI (light probes and reflection probes), we use the normal most aligned
+        // with the view vector (the "view-facing" normal).
+        // We reflect this normal for transmitted GI.
+        // For the highlight shift hack (along the tangent), we use the user-provided normal.
```
mmikk: Keep in mind we'd ideally like a shader that will work the day we try out spline hair. At that point there isn't a user-provided normal. There's also no geometric normal for shadow bias. In both cases all you have is the view-facing normal.

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
EvgeniiG: Good question, currently the answer is no.
We don't have a transmission intensity map (we can only adjust the thickness for Disney), we should consider adding one.

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
EvgeniiG: Okay, so for the shadow bias we can use the light-facing normal (we'd need to add a #define for splines, though).

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
mmikk: Well The R channel is labeled transmission T(R) in shadergraph at least. Either way transmission is likely to result in some artifacts and false positives/leaking due to limitations in quality of shadowing and lack of proper multiple scattering so I think it's important in the context of hair that it works.

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
EvgeniiG: The user-provided normal is provided via the normal map, so I suppose we could just ignore it.

EvgeniiG: The big issue here is IBL (reflection probes, to be specific), because we really rely on the card's normal (which tells us something about the shape of the head) to make the stretch hack work. I don't think it's going to work with the view-facing normal.

EvgeniiG: I'll test it to make sure.

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
EvgeniiG: I fully agree. T(R) is thickness, though. I think we may want to use something hair-specific (and simpler) here instead.

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
EvgeniiG: OK, I was wrong, it still works.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
```diff
@@ -354,14 +354,24 @@ float3 ShiftTangent(float3 T, float3 N, float shift)
...
+    float dirAttn = saturate(TdotH + 1.0); // Evgenii: this seems like a hack? Do we really need this?
 
-    float dirAttn = saturate(TdotH + 1.0);
+    // Note: Kajiya-Kay is not energy conserving.
+    // I attempt at least some energy conservation by approximately normalizing Blinn-Phong.
+    // This is not the exact normalization factor.
+    // The exact one is (n + 2) * (n + 4) / (8 * Pi * (n + pow(2, -0.5 * n))).
+    float nHalf = specularExponent;
+    float n     = 2 * nHalf;
+    float norm  = (n + 7) * rcp(8 * PI);
```
mmikk: I am not following this part. Is it because it's specific to hair?

In hard surface context when you do the integration for NdotH^n the normalized version becomes ((n+2)/(2*pi))*NdotH^n

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
```diff
@@ -379,13 +389,16 @@ void BSDF(  float3 V, float3 L, float NdotL, float3 positionWS, PreLightData pre
...
+        // Balancing energy between lobes, as well as between diffuse and specular is left to artists.
         float3 hairSpec1 = bsdfData.specularTint * D_KajiyaKay(t1, H, bsdfData.specularExponent);
         float3 hairSpec2 = bsdfData.secondarySpecularTint * D_KajiyaKay(t2, H, bsdfData.secondarySpecularExponent);
 
         float3 F = F_Schlick(bsdfData.fresnel0, LdotH);
         specularLighting = F * (hairSpec1 + hairSpec2);
 
         // Diffuse lighting
+        // Note: this normalization term is wrong, we also need to divide by (2 * Pi)...
+        // However, this would make the contribution of non-area-lights too small.
```
mmikk: I think it's just Pi? 

Int dot(l,n) dl = Int_0^{2*pi} Int_0^{0.5*pi} sin(phi)*cos(phi) * dphi dtheta
                      = 2*pi * [ 0.5*sin^2(phi) ]_0^{0.5*pi} = pi

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: Depends on the exact formulation of Blinn-Phong.
Mine is with NdotL (which technically comes from the rendering equation):
http://www.thetenthplanet.de/archives/255
Reminder: we multiply by the NdotL outside. :-)

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: It looks like Lambert but it's not.
The reason being that the normal is not fixed but rather varies with the light direction.
I posted you my derivation on Slack (see the big pictures on Slack):
image
Maybe I messed up. :-)

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: Of course, all of this logic only applies assuming that the normal does not vary with the light direction. :-)

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: I think the least-worst short-term solution (while we stick to KK) is to pretend that we use the normal Lambert and Blinn-Phong and normalize those.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
mmikk: I've seen the formulation before but I think it is incorrect. I'll try and explain why the best I can in ascii :) So in the definition of a valid microfacet distribution function D(h) the integration Int D(h) dot(n,h) dh must be 1 when integrating over the half sphere. So this is what we must keep normalized. Now the rendering equation is a separate issue. The way we get from this to the form in the rendering equation is by using integration by substitution using dl = 4 * dot(h,l) * dh. As we do this substitution we instead get (1/4) * Int D(h) * Id(dot(n,h)>0) * dl. This is also where the division by 4 comes from in the classic bsdf formulation. It's the result of this substitution step.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
```diff
@@ -356,23 +340,50 @@ float3 ShiftTangent(float3 T, float3 N, float shift)
...
 }
 
 // This function apply BSDF. Assumes that NdotL is positive.
 void BSDF(  float3 V, float3 L, float NdotL, float3 positionWS, PreLightData preLightData, BSDFData bsdfData,
             out float3 diffuseLighting,
             out float3 specularLighting)
 {
-    float LdotV, NdotH, LdotH, NdotV, invLenLV;
-    GetBSDFAngle(V, L, NdotL, preLightData.NdotV, LdotV, NdotH, LdotH, NdotV, invLenLV);
+    diffuseLighting = specularLighting = 0; // Deprecated
```
sebastienlagarde: Given that the structure return by  EvaluateCBxDF , CBxDF is use like this:

        lighting.diffuse  = (cbxdf.diffR + cbxdf.diffT * bsdfData.transmittance) * (lightData.diffuseDimmer  * (color * attenuation));
        lighting.specular = (cbxdf.specR + cbxdf.specT * bsdfData.transmittance) * (lightData.specularDimmer * (color * attenuation));

I don't see the reason to not use BSDF() function above with

diffuseLighting = (cbxdf.diffR + cbxdf.diffT * bsdfData.transmittance);
specularLighting =  (cbxdf.specR + cbxdf.specT * bsdfData.transmittance)

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
```diff
@@ -356,23 +340,50 @@ float3 ShiftTangent(float3 T, float3 N, float shift)
...
+
+    float dirAttn = saturate(TdotH + 1.0); // Evgenii: this seems like a hack? Do we really need this?
 
-    float dirAttn = saturate(TdotH + 1.0);
+    // Note: Kajiya-Kay is not energy conserving.
+    // I attempt at least some energy conservation by approximately normalizing Blinn-Phong.
+    // This is not the exact normalization factor.
+    // The exact one is (n + 2) * (n + 4) / (8 * Pi * (n + pow(2, -0.5 * n))).
+    float n    = specularExponent;
+    float norm = (n + 7) * rcp(8 * PI);
```
sebastienlagarde: isn't this the approximation from Naty? maybe add a reference.

mmikk: Could we instead get rid of that thing and use the proper D formulation (n+2)/twopi normalization and then scale by 0.25 where we multiply by the fresnel as per conventional FGD brdf form.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: Transmittance should not be a part of the BSDF.

EvgeniiG:
The PR is now ready for review.
@mmikk @sebastienlagarde

EvgeniiG:
@mmikk Valid point. Fixed.

zero-radiance:
Katana is green.

EvgeniiG:
Katana is green.

com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
```diff
@@ -70,21 +70,18 @@ real MipmapLevelToPerceptualRoughness(real mipmapLevel)
...
-// Ref: Donald Revie - Implementing Fur Using Deferred Shading (GPU Pro 2)
-// The grain direction (e.g. hair or brush direction) is assumed to be orthogonal to the normal.
-// The returned normal is NOT normalized.
-real3 ComputeGrainNormal(real3 grainDir, real3 V)
+// T is the fiber axis (hair strand direction, root to tip).
+float3 ComputeViewFacingNormal(float3 V, float3 T)
 {
-    real3 B = cross(grainDir, V);
-    return cross(B, grainDir);
+    return Orthonormalize(V, T);
```
sebastienlagarde: The orthonormalize is different from original code. 
ComputeGrainNormal => ComputeViewFacingNormal is it expected ?
ComputeGrainNormal was explicitely not normalizing.

this is why you have all these screenshots updated

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
```diff
@@ -113,7 +107,9 @@ Pass
 
     #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl"
     #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl"
+    #define MATERIAL_INCLUDE_TRANSMISSION // temp hack
```
sebastienlagarde: I don't think it will be "temp" :)

com.unity.render-pipelines.high-definition/Editor/Material/Hair/ShaderGraph/HairPass.template
```diff
@@ -199,33 +194,14 @@ $include("SharedCode.template.hlsl")
...
-
-#ifdef _MATERIAL_FEATURE_SUBSURFACE_SCATTERING
-        surfaceData.materialFeatures |= MATERIALFEATUREFLAGS_HAIR_SUBSURFACE_SCATTERING;
-#endif
-
-#ifdef _MATERIAL_FEATURE_TRANSMISSION
-        surfaceData.materialFeatures |= MATERIALFEATUREFLAGS_HAIR_TRANSMISSION;
+        surfaceData.materialFeatures |= MATERIALFEATUREFLAGS_HAIR_KAJIYA_KAY;
 #endif
+        surfaceData.materialFeatures |= MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THIN_THICKNESS; // Remove
```
sebastienlagarde: Shoudln't this be removed ? or is it in your other PR that you handle it, right ?

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
```diff
@@ -356,23 +338,49 @@ float3 ShiftTangent(float3 T, float3 N, float shift)
     return normalize(T + N * shift);
 }
 
+// Note: this is Blinn-Phong, the original paper uses Phong.
 float3 D_KajiyaKay(float3 T, float3 H, float specularExponent)
```
sebastienlagarde: Let's move it to BSDF.hlsl

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
sebastienlagarde: ok, we will see in the next PR how we rename the function between EvaluateCBxDF and BSDF(). I get you have done this to not have to rename everythingeverywhere in the PR. And yes, CBxDF is a good struct. But why BxDF and not BSDF ?

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
sebastienlagarde: canbe done in next PR

sebastienlagarde:
Just wait answer of question below, need a merge with master and I will merge PR to main. Good job!

com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
EvgeniiG: This code does the same thing, at roughly the same cost. The only difference is the normalization - the direction is the same. I'd rather not have two versions since it's the same hack, plus I'd rather generate a normalized vector.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: Not in this PR for sure.

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.hlsl
EvgeniiG: Because fiber models also introduce BFSDF and BCSDF.

com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
seblagarde: But visually there is a difference. Thus why I ask, the difference on the screenshots aren't negligeable. Still as everything is hack I suppose it is fine :)

EvgeniiG: The difference most likely comes from the lack of normalization. Originally, we just performed a texture fetch from the cubemap, but now that the code has evolved, normalization actually matters. This is why I prefer to keep the normalized version.

PR #2917, opened 2019-02-07
Disable motion blur if game is not playing 

FrancescoC-unity:
Disabling motion blur if game is not playing as motion vectors are not reliable in that case and can cause some distracting "flashes" of blur.


FrancescoC-unity:
We got a better fix. Closing this.

PR #2919, opened 2019-02-07
Use sign bit to find the cache type instead of lowest bit.

fredpointzero:
### Purpose of this PR
Fixes a compile issue due to a bug in Intel's driver for OSX.

@RemyUnity can you validate this work on Metal ?  thanks
---
### Release Notes

---
### Testing status
[**Katana Tests**: ABV Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FFixOSXEnvProbeCache&automation-tools_branch=add-platform-filter&unity_branch=trunk#)

**Manual Tests**: Local manual testing on windows

**Automated Tests**: -

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---


aet:
LGTM, I haven't been looking into any Intel specifics a while, but the original reason for that #ifdef to exist was the last-minute regression on OSX 10.13.6 GPU driver update on AMD. On 10.13.5 and 10.14.x it's working as expected.

Katana AMD builder is running 10.13.6 so if the tests are passing, good for me to go.

@fredpointzero Seeing that the #if !defined(SHADER_API_METAL) logic has inherited itself into plenty of other shaders, should be ok to clean them up after this?

sebastienlagarde:
Cleaning all the test, launching katana

PR #2923, opened 2019-02-07
Improving sampling method for raytraced effects

anisunity:
Removing the old noise textures
Expose the radius for the spatial filter


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingSampling.hlsl
```diff
@@ -1,51 +1,39 @@
...
-    tangent = float3(1.0 + sign * n.x * n.x * a, sign * b, -sign * n.x);
-    bitangent = float3(b, sign + n.y * n.y * a, -n.y);
-}
+    // Scramble the value
+    value = value ^ scramblingValue;
+
+    // convert to float and return
+    float v = (0.5f + value) / 256.0f;
+    return v;
+}
```
anisunity: @eturquin

eturquin: All good (the couple clamp() could be replaced by min() since we know inputs are positive, and we are assigning uints anyway)!
In other files, I noticed "need only need" typo in comments.

PR #2379, opened 2018-11-09
Vfx graph/mesh volume pcache

RemyUnity:
### Purpose of this PR
Add a mesh volume to pCache bake tool.

---
### Release Notes
See above.

---
### Testing status
Editor only tool tested locally.

---
### Overall Product Risks
**Technical Risk**: Low: Worst case is that the tool fails baking.

**Halo Effect**: Low: Renamed some classes from the other mesh bake tool ('Mesh' -> 'Mesh Surface')

PR #2935, opened 2019-02-08
Vfx graph/mesh volume pcache 2

RemyUnity:
New version of https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2379

Removed commit https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2379/commits/c470e577f469119f93fea394c09166c9eb383f74 to don't touch stuff that isn't for the pCache baker.


VladNeykov:
@RemyUnity can you add a warning if no point cache is created? 
I was using a very small mesh which even with a voxel size of 0.1 didn't produce an asset.

Something like:
"Creating the pCache file has failed: the voxel size is too large, or the mesh does not have enough vertices." (not sure if there are any other scenarios)

VladNeykov:
Starting from the bottom, it is possible to set all axis to be the same when using Axis reorder. Maybe this can be disallowed, or a warning can be written in the console when trying to bake like this?

image

RemyUnity:
> @RemyUnity can you add a warning if no point cache is created?
> I was using a very small mesh which even with a voxel size of 0.1 didn't produce an asset.
> 
> Something like:
> "Creating the pCache file has failed: the voxel size is too large, or the mesh does not have enough vertices." (not sure if there are any other scenarios)
Message added here : https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2935/commits/b416ffa4a97985e4a121fecc8037af44dccf7ecf


> Starting from the bottom, it is possible to set all axis to be the same when using Axis reorder. Maybe this can be disallowed, or a warning can be written in the console when trying to bake like this?
> 
> image
> 
> Also first, second, and third axis are a bit arbitrary; shouldn't they be X axis, Y axis, and Z axis?
This should be fixed here : https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/2935/commits/624290ff71e6b71e0ecd80552244b5c38beebddc

PR #2949, opened 2019-02-12
Display estimated rays/sec in display stats

nerites:
### Purpose of this PR
Estimate total rays/s using AsyncGPUReadback and display it in the Display Stats tab of the debug window:
image

---
### Testing status
image
Confirmed that the readback value matched the value written in the shader. Cross-checked with PIX capture also.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrtr_master_rayspersec&unity_branch=trunk&automation-tools_branch=add-platform-filter 

---
### Overall Product Risks
**Technical Risk**: Low- added an asyncreadback to an otherwise proven ray count mechanism

**Halo Effect**: Low- bracketed in RAYTRACING_ENABLED, and then the new code is only active if "Count Rays" is enabled. 

---
### Comments to reviewers
Do you think it's worth adding additional readbacks for each of AO/reflections/etc. separately? I'm pretty sure you'd have to allocate a separate texture for each, with how AsyncGPUReadback works, so I didn't for now. 

In this PR I'm using Time.smoothDeltaTime, based on an example that didn't use command buffers at all, so I don't know how command buffers confound the timing. Is there any way to grab timing from the ProfilingSamples we have scattered around the code?


com.unity.render-pipelines.high-definition/Runtime/Debug/RayCountManager.cs
```diff
@@ -53,13 +79,55 @@ public int rayCountEnabled
...
             get
             {
-                return m_RayCountEnabled ? 1 : 0;
+                return m_DebugDisplaySettings.data.countRays ? 1 : 0;
+            }
+        }
+
+        public float latestRaysPerSecond
+        {
+            get
```
anisunity: Given that this is a full evaluation and not a variable get, it would be better to use a Function GetLatestRaysPerSecond or EvaluateLatestRaysPerSecond.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/DebugViewRayCount.shader
```diff
@@ -65,12 +65,12 @@ Shader "Hidden/HDRP/DebugViewRayCount"
...
             static const uint TotalName[MAX_STRING_SIZE] = {'T','o','t','a','l',':',' ','\0','\0','\0','\0','\0','\0','\0','\0','\0'};
             #define MRPF_STRING_SIZE 12
-            static const uint MGPFName[MAX_STRING_SIZE] = {' ', 'm','r','a','y','s','/','f','r','a','m','e','\0','\0','\0','\0'};
+            static const uint MGPFName[MAX_STRING_SIZE] = {' ', 'M','r','a','y','s','/','f','r','a','m','e','\0','\0','\0','\0'};
 
             // Helper function to write strings
             void WriteMegaraysPerFrame(uint targetWord[MAX_STRING_SIZE], uint stringSize, float3 fontColor, uint2 unormCoord, inout uint2 displayUnormCoord, inout float3 result)
             {
-                for(int i = 0; i < stringSize; ++i)
+                for(uint i = 0; i < stringSize; ++i)
```
anisunity: Not sure why this

com.unity.render-pipelines.high-definition/Runtime/Debug/RayCountManager.cs
```diff
@@ -30,6 +54,8 @@ public void Init(RenderPipelineResources renderPipelineResources)
             // UINT textures must use UINT32, since groupshared uint used to synchronize counts is allocated as a UINT32
             m_RayCountTex = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32G32B32A32_UInt, enableRandomWrite: true, useMipMap: false, name: "RayCountTex");
             s_TotalRayCountBuffer = new ComputeBuffer(3, sizeof(uint));
+            m_TotalRayCountTex = RTHandles.Alloc(1, 1, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32_SFloat, enableRandomWrite: true, useMipMap: false, name: "RayCountTex");
```
anisunity: This creates a (1,1) sized texture that get rescaled into a full-sized render texture by the handle system. You endup reading a sized texture for just the corner value

com.unity.render-pipelines.high-definition/Runtime/Debug/RayCountManager.cs
```diff
@@ -7,6 +9,7 @@ public class RayCountManager
     {
         // Ray count UAV
         RTHandleSystem.RTHandle m_RayCountTex = null;
+        RTHandleSystem.RTHandle m_TotalRayCountTex = null;
```
anisunity: Same remark as for the other PR, the correct pattern here would be to use a compute buffer

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/CountTracedRays.compute
```diff
@@ -49,5 +50,6 @@ void CS_CountRays(uint3 globalId : SV_DispatchThreadID, uint3 localId : SV_Group
         InterlockedAdd(_TotalRayCountBuffer[0], aoRayCountPerThreadGroup);
         InterlockedAdd(_TotalRayCountBuffer[1], reflectionRayCountPerThreadGroup);
         InterlockedAdd(_TotalRayCountBuffer[2], areaShadowRayCountPerThreadGroup);
+		_TotalRayCountTexture[uint2(0, 0)] = (float)_TotalRayCountBuffer[0] + (float)_TotalRayCountBuffer[1] + (float)_TotalRayCountBuffer[2];
```
anisunity: why float btw?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/DebugViewRayCount.shader
nerites: I was getting some shader warnings about casting uint/int mismatch between stringSize and i.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/CountTracedRays.compute
nerites: AsyncGPUReadback expects to read floats from textures. I'm not sure if this is true for computebuffers so I'll try to use uints there

anisunity: I asked @julienf and he says it works with compute buffers

com.unity.render-pipelines.high-definition/Runtime/Debug/RayCountManager.cs
nerites: The overload that takes in width and height just allocates it using RenderTexture without adding it to the list of autosized textures, but it's moot if I can get async readback from a computebuffer working

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/CountTracedRays.compute
nerites: Ah, I missed the part in my code where I was telling the readback to expect a float instead of a uint. I'll fix it when I convert to computebuffer

nerites: I'm getting an incorrect value when I try to do GpuAsyncReadback from uint computebuffers. I posted a question on devs-graphics, hopefully someone knows what's going on
image

anisunity:
the code itself looks fine, however the goal of using an async readback instead of binding the buffer directly to the shader was to use the debug menus that we have for the rest right @sebastienlagarde

anisunity:
I'lle be doing the corrections and updating the PR @nerites

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/CountTracedRays.compute
```diff
@@ -1,53 +1,90 @@
...
+    // Copy the local values to to kernel
+    gs_rayCount[threadIdx] = _InputRayCountTexture[sampleIdx];
+
+    // Make sure everyone has copied it value to the LDS
+    GroupMemoryBarrierWithGroupSync();
 
-    if (localId.x == 0 && localId.y == 0)
+    // do reduction in shared mem
+    UNITY_UNROLL
+    for(uint s = REDUCTION_TOTAL_THREADS / 2u; s > 0u; s >>= 1u)
```
anisunity: If you can review the reduction code it would be awesome @Chman!

nerites: wow what's this do? i've never seen it before 0:

PR #2960, opened 2019-02-13
Additional post-processing data

Chman:
### Purpose of this PR
Uses the "sample" system of the package manager to distribute additional post-processing data (lens dirt & spectral textures for chromatic aberration for now).

---
### Testing status
No testing needed, data only.


sebastienlagarde:
@Chman should we merge this now, is it still ok ?

Chman:
Yes, we can merge it. Let's keep it for 19.3, no need to backport to 19.2.

PR #2964, opened 2019-02-13
Projects/ShaderGraph

deodatopr:
### Shader Graph Depth Node

I would like to know if Shader Graph has something like this function/node or if could be added:

http://wiki.amplify.pt/index.php?title=Unity_Products:Amplify_Shader_Editor/Depth_Fade


sebastienlagarde:
Hi. Please, use forum for such request.

PR #2971, opened 2019-02-14
Add support for half types in shaders for Switch platform.

kaychang-unity:
### Purpose of this PR
This PR adds minimal support for half types in shaders. This affects Switch platform, and maybe mobile platforms (not officially supported).

For the minimal support, I only modified temporary structures inside HDRP shader code to use half instead of float. As seen in the PR, a lot of fields are left to float because lighting artefacts would appear otherwise. So there isn't actually that many values that can be safely moved to half.

On Switch, these minimal changes improve the shader execution by 15%.

We could move more code to use half, but the gain seems less significant (I suspect less than 5% further gain) but would touch more code everywhere. I may fix more code in the future to use half (if time allows).

---
### Release Notes

---
### Testing status
**Katana Tests**:
Running Katana tests now:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=switch-support-half&automation-tools_branch=add-platform-filter&unity_branch=trunk

I tested a couple of scenes in HDRP_Tests (2001_Dynamic_Directional, 2002_Dynamic_Mix, 2201_ReflectionProbes_Volume) and made sure rendering did not changes with half support.

-

-
---
### Overall Product Risks
Low. Excepted Switch, all other platforms define "real" type as "float". so the generated code is identical.

Low.
Mobile platforms could be affected by the change, but since original code would not compile with SHADER_API_MOBILE enabled, there is currently no official support for mobile platforms and this commit help fix it.

---
### Comments to reviewers
com.unity.render-pipelines.core/ShaderLibrary/Packing.hlsl use "real" types everywhere, but at least in 1 occurrence, this didn't give enough precision when "real"="half". I fixed the related functions (PackNormalHemiOctEncode, UnpackNormalHemiOctEncode). But I believe most of these packing functions should explicitly use float.


com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -64,7 +64,7 @@
 // If a function require to have both a half and a float version, then both need to be explicitly define
 #ifndef real
 
-#ifdef SHADER_API_MOBILE
+#if defined(SHADER_API_MOBILE) || defined(SHADER_API_SWITCH)
```
EvgeniiG: Are you certain you want to do it globally and not on a per-file basis?
That's of course simpler (you only change one line in this file), but more dangerous (since enable half for all files).
Just something to consider. :-)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs.hlsl
```diff
@@ -82,12 +82,12 @@ struct LightData
...
     float volumetricLightDimmer;
-    float angleScale;
-    float angleOffset;
+    real angleScale;
+    real angleOffset;
     float3 forward;
     int lightType;
     float3 right;
-    float range;
+    real range;
```
EvgeniiG: This file is auto-generated, so modifying it manually is not a good idea.
How does C# manage to write half into a structured buffer or a constant buffer? The C# side only exposes floats, doesn't it?

EvgeniiG: You can probably convert many more of these to half safely in the future.

kaychang-unity: On Switch platform, half type is used to generate fp16 instruction, but does not change the storage size of the variable if coming from memory (it will stay 32bits).
Note that Metal behaves differently in that regard, but I am not sure how HLSLcc compiler behaves for that case.

EvgeniiG: So it basically converts from FP32 to FP16 on the fly?

kaychang-unity: Yes, it loads 32bit float value into a 16bit register (HLSL, GLSL have same behaviour, only Metal differs).

jessebarker: That's really unfortunate because for platforms that support it, there would be bandwidth savings as well as ALU benefits.

sebastienlagarde: For the convertion you will indeed require to modify how we generate code, otherwise you will lose all your change at the next call to Generate includes files. For this I think you can rely on the code of packing added by Francesco (Remember the PR packing I have show you), then add a type half (or real ?) in this packing, add corresponding generated code and it should be fine. I can ask Francesco to do it if you want.

sebastienlagarde: >real x = (real)(light.range) * a + b;
Doing this everywhere will pollute the code a lot, better to put half in my opinion and work for other platform.

kaychang-unity: OK, I will have a look at Francesco PR again and make the changes needed in ChaderToHLSL.cs...

EvgeniiG: @sebastienlagarde While I agree, the fact is, the data structure itself is unchanged. It only contains floats. So I find it a bit surprising that the code actually works at all.
IMHO, we should do the following.
1. For data structures that we fill on the GPU (e.g. BSDFData), we can generate the layout using real, no problem.
2. For data structures that we fill on the CPU (e.g. LightData), since they actually contain floats, the layout should indicate that. We can then cast on access (or do something else?).

kaychang-unity: I think maintenance would be hard. It is already quite hard to know when half-types should be used in shaders.
How about use half/real type on all structures, and then fix up the upload of the data from CPU when needed (in our case, LightData would contains real 16bits floats on CPU on Metal Mobile platforms only).

sebastienlagarde: Maybe I remember an incorrect statement, but isn't half in constant buffer consider as a float in term of storage ? @arnaudcarre you are the expert here, iirc you said me that storage size in constant buffer don't change, even if you use half type.

arnaudcarre: Half usage is quite tricky. To me there is two orthogonal things:
1) the ALU usage ( are ALU computations done in 16 or 32 bits ) ( could improve shader speed)
2) the buffer memory layout ( is the half stored in GPU buffer as a 16 or 32bits) ( could improve bandwidth )
To my understanding, "Switch" is a bit special: when using "half", ALU use 16bits but memory load/store are still in 32bits ( with automatic shader 32->16 conversion ). You get benefit of (1) only.

Getting benefit of (2) is way more tricky. If your buffer contains 16bits, the CPU code ( C# code in that case) that fill up that struct should explicitly call a Float32ToFloat16 conversion routine for each "half" declared vars. So C# code should have two code path. I don't even speak of the C# HLSL struct generator that should handle new layout ( 2bytes alignment etc). 

Side note: I added half support recently in SRP Batcher, using dedicated c++ code

kaychang-unity:
@FrancescoC-unity I modified ShaderTypeGeneration.cs code to handle the new FieldPacking.Half and FieldPacking.Reaf. The solution may not be that elegant, so any advice to make this nicer is welcome.

The newly autogenerated files (BuiltinData.cs.hlsl, Lit.cs.hlsl, HDShadowManager.cs.hlsl, LightDeficition.cs.hlsl) generate more debug #define but s a side-effect of enabling `containsPackedFields = true`. I believe it is harmless.

kaychang-unity:
@sebastienlagarde , are you fine with setting the precision qualifier (real/half) as an extra parameter in [SurfaceDataAttributes] ?

kaychang-unity:
@FrancescoC-unity I pushed the new solution that uses SurfaceDataAttributes for review.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
kaychang-unity: Half support is necessary to get good performance on Switch and users already expect it to work on other game projects. I think it is simpler to make it available all the time and have the behavior identical as other platforms that supports half.

sebastienlagarde:
Sorry for the insanely long review. we are now ready to merge the PR :)

sebastienlagarde:
katana is green

PR #2974, opened 2019-02-15
Hdrp/shadergraph default values + fabric tweaks

laurenth-unity:
### Purpose of this PR

Fixes default values for shadergraph masternodes https://fogbugz.unity3d.com/f/cases/1117273/
Removes schlick fresnel from fabric shader
Remap fabric shader smoothness range to remove invalid values

---
### Release Notes

Existing content with Fabric shader will have lower smoothness
---
### Testing status
Ran graphics tests, updated reference image for fabric tests


com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalMasterNode.cs
```diff
@@ -109,7 +109,7 @@ public sealed override void UpdateNodeAfterDeserialization()
             // Albedo
             if (MaterialTypeUsesSlotMask(SlotMask.Albedo))
             {
-                AddSlot(new ColorRGBMaterialSlot(AlbedoSlotId, AlbedoDisplaySlotName, AlbedoSlotName, SlotType.Input, Color.white, ColorMode.Default, ShaderStageCapability.Fragment));
+                AddSlot(new ColorRGBMaterialSlot(AlbedoSlotId, AlbedoDisplaySlotName, AlbedoSlotName, SlotType.Input, Color.grey.gamma, ColorMode.Default, ShaderStageCapability.Fragment));
```
sebastienlagarde: Are we sure we to put grey for decal?

laurenth-unity: Yes, so by default it does something, also smoothness at 0.5 will have an effect by default

PR #2975, opened 2019-02-15
Fix blackboard disappearing

pbbastian:
### Purpose of this PR
Solving issues with the Shader Graph property going missing (fogbugz 1111305, fogbugz 1102285). I changed it such that the blackboard position is set via `FloatingWindowsLayout` always. When de-serialisation of `FloatingWindowsLayout` fails, we now initialise it to sensible defaults. Finally, I reset the blackboard position for everyone, just to avoid cases where someone has bad data saved.

---
### Testing status
**Katana Tests**: None of the stuff on Katana touches this code.

**Manual Tests**: Repro'ed using steps in the 2 fogbugz cases using ShaderGraph test project, and confirmed that I they don't repro after fix.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


sebastienlagarde:
I launch creation for 6.2.0 so I will merge this one. If there is issue @zadi0648 please raise it and we can create a new PR for solving it

PR #3003, opened 2019-02-18
Area light (fake) realtime shadows 

FrancescoC-unity:
### Purpose of this PR
Adding a first version of real time area light shadow support. Overall, this is essentially just a spot light shadow with some tweak parameters. 

With this PR, the filtering used will be the same used for spot light w.r.t. shadow quality. 

The feature is hardcoded off as this version is very VGPR hungry and we don't want to impact Lit that much. Can be enabled changing, in Lit.hlsl, the define RASTERIZED_AREA_LIGHT_SHADOWS. 

I also added two visual tests, then realised that the feature is off by default, so the tests are off by default as well :) 

This PR also adds support for Shadow masks and area lights. 

---
### Release Notes
Preliminary real time area light shadow support.

 
---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Farealight-shadow&automation-tools_branch=master&unity_branch=trunk [Just kicked]

**Automated Tests**:  Added two tests, but kept disabled by default. 

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Medium


FrancescoC-unity:
Note for now shadow mask support is only when RASTERIZED_AREA_LIGHT_SHADOWS is set to 1. I will change this as soon as we have an acceptable solution that can be on by default.

FrancescoC-unity:
Green katana on windows (red on Mac, is that known issue?)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -762,8 +799,6 @@ public static void InitDefaultHDAdditionalLightData(HDAdditionalLightData lightD
                 case LightType.Rectangle: // Rectangle by default when light is created
                     lightData.lightUnit = LightUnit.Lumen;
                     lightData.intensity = k_DefaultAreaLightIntensity;
-                    // Disable shadows for area lights as it's not yet supported
-                    light.shadows = LightShadows.None;
```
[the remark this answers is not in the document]
FrancescoC-unity: Nope, you are right. This is me making a wrong merge of the current master.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -350,6 +382,12 @@ public void LayoutShadowMaps(LightingDebugSettings lightingDebugSettings)
             if (m_CascadeAtlas != null)
                 m_CascadeAtlas.UpdateDebugSettings(lightingDebugSettings);
 
+            // TODO: This is required to avoid problems when we are using a cone to simulate the area light.
+            // This can be handled better.
+            lightingDebugSettings.clearShadowAtlas = true;
```
alelievr: I don't understand why you need to force to clear the whole atlas, which problems does it avoid ?
Also it makes the button in the UI basically useless.

FrancescoC-unity: Discussed offline, I'll do the change we discussed first thing tomorrow. 

It'd be nice at some point to do a thorough profile of what's the impact later on, but for now what we discussed is clearly the cleaner solution.

FrancescoC-unity:
Since I merged master again, I'll kick katana again

FrancescoC-unity:
Since I am getting ready to do a PR with the EVSM update (will be ready today), I will close this PR. It would be pointless to merge this now 🙂

PR #3015, opened 2019-02-19
Fixed volume explorer using the wrong profile reference

Chman:
### Purpose of this PR
The volume explorer used `profile` instead of `sharedProfile` which instantiate a custom volume profile instead of editing the asset itself.

Added @laurenth-unity to the PR as well for awareness.

Note: having a "volume explorer" inside of the "light explorer" is counter-intuitive, it should be renamed.

---
### Testing status
Manual local testing & the demo team has been using this fix with success.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low


sebastienlagarde:
Please add changelog

PR #3025, opened 2019-02-20
Fixed culling and shadows for Pyramid, Box, Rectangle and Tube lights

JulienIgnace-Unity:
### Purpose of this PR
Implemented correct culling sphere and shadow matrices for Pyramid, Box, Rectangle and Tube lights thus avoiding false negatives during culling for these lights.

---
### Release Notes
Fixed culling and shadows for Pyramid, Box, Rectangle and Tube lights

---
### Testing status
Manual testing: provoked false negative culling of shadows and light for all types and verified that new bounding spheres fixed the problem (and did not cause new ones).
Hard to add a graphics tests because it would require a specific camera for each light type and thus a separate test. Given the state of graphics tests right now that would probably be a bad idea.

Katana:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Ffix-custom-light-culling&automation-tools_branch=master&unity_branch=trunk
Also ran graphics test locally.


com.unity.render-pipelines.high-definition/Editor/Lighting/SerializedHDLight.cs
```diff
@@ -109,7 +109,7 @@ public SerializedHDLight(HDAdditionalLightData[] lightDatas, AdditionalShadowDat
                     affectSpecular = o.Find(x => x.affectSpecular),
                     nonLightmappedOnly = o.Find(x => x.nonLightmappedOnly),
                     lightTypeExtent = o.Find(x => x.lightTypeExtent),
-                    spotLightShape = o.Find(x => x.spotLightShape),
+                    spotLightShape = o.Find("m_SpotLightShape"),
```
sebastienlagarde: is it expected we access to the property like this ?

JulienIgnace-Unity: Yes. I needed to make the property private with a Setter in order to be able to correctly update the bounding box when shape is changed from script. Unfortunately the property finder needs property to be public so I needed to go with string instead.

PR #3029, opened 2019-02-20
Custom Function Node and Subgraph Output Node refactor

Kink3d:
### Purpose of this PR
- Adds Custom Function Node
- Refactors Subgraph Output Node to support all port types and renaming
- Adds ReorderableSlotList View to handle the above two changes
- Changes Convert To Subgraph to create Properties/out ports with names based on types for Blackboard consistency

**NOTE:** This PR contains an issue that will be fixed by a separate PR (@spaztronaut )
- Deleting a port that has attached edges creates a nullref and error state is not resolved. This reproduces on both Custom Function Node and Subgraph Output. This also reproduces in master with current Subgraph Output node.

---
### Release Notes
- See changelog

---
### Testing status
**Katana Tests**: Running

**Manual Tests**: What did you do?
- Used Custom Function Node a lot with various settings
- Converted various graphs to subgraphs
- Test to fail:
   - Slot/Prop name duplicate sanity
   - Tried to break preview shader with Custom Function node in invalid state
   - Tried to create Convert to Subgraph issues

**Automated Tests**:  Nothing, should be added to graphics tests (in PR)

Any test projects to go with this to help reviewers? Nope

**NOTE**: Make sure to test upgrading from previously serialized subgraphs and graphs (with subgraphs in). @spaztronaut 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
N/A

PR #3030, opened 2019-02-21
[Backport] Custom Function Node and Subgraph Output Node refactor

Kink3d:
### Purpose of this PR
Backport of #3029


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [5.6.0] - 2019-XX-XX
+### Added
+- Added Custom Function node. With this node you can define a custom HLSL function either via string directly in the graph, or via a path to an HLSL include file.
```
sharlenet: Added Custom Function node. Use this node to define a custom HLSL function either via string directly in the graph, or via a path to an HLSL include file.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+
+### Changed
+- Subgraph Output node is now called Outputs.
+- Subgraph Output node now supports renaming of ports.
+- Subgraph Output node now supports all port types.
+- Subgraph Output node now supports reordering ports.
+- When converting nodes to a Sub-Graph, the resulting properties and output ports generated in the Sub-Graph are named for their types by default.
```
sharlenet: I'm having trouble understanding what "named for their types" means. Does this mean they are named based on their types?

com.unity.shadergraph/CHANGELOG.md
sharlenet: Change to:

Added Custom Function node. Use this node to define a custom HLSL function either via string directly in the graph, or via a path to an HLSL file.

com.unity.shadergraph/CHANGELOG.md
sharlenet: Change to:

When you convert nodes to a Sub-Graph, the resulting properties and output ports, which are generated in the Sub-Graph, are named based on their types by default.

zadi0648:
Concern: plugging a matrix that is bigger than the subgraph output matrix type results in an unhelpful error "Error found during node validation". This does not reproduce / is applicable to other nodes, it is expected that the matrix gets truncated into the smaller type (for example multiplication of nodes of two different sizes results in a mult of the smaller matrix size).

MatrixConcern

PR #3032, opened 2019-02-21
Fix SSR mip count issue

FrancescoC-unity:
### Purpose of this PR
Fix issues with SSR max mips.

The max depth mip was 1 too many and the color max mip was wrong. 
  
---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Ffix-ssr-mip-issues&automation-tools_branch=master&unity_branch=trunk [ Just kicked, but I am not expecting any change since we don't have active SSR tests  )

- [x] Opened test project + Run graphic tests locally _(We don't have SSR related tests, I did check with a locally built scene)_
- [x] Built a player
- [ ] Checked new UI names with UX convention  _( No change )_
- [ ] Tested UI multi-edition + Undo/Redo _(Not UI change)_

---
### Overall Product Risks
**Technical Risk**:  Low, 

**Halo Effect**: Low


sebastienlagarde:
@EvgeniiG We need to produce a new package for template validation error, so I merge this PR. But feel free to comment and correct if you find any problem.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1828,7 +1828,7 @@ void Callback()
 #if ENABLE_RAYTRACING
                 m_RaytracingRenderer.Render(hdCamera, cmd, m_CameraColorBuffer, renderContext, cullingResults);
 #endif
-                RenderColorPyramid(hdCamera, cmd, false);
+                m_PrevFrameColorPyramidMips = RenderColorPyramid(hdCamera, cmd, false);
```
EvgeniiG: Does RenderColorPyramid() now return the correct value for the numbers of MIPs, accounting for RTHandle scaling and a couple of last MIPs being zero? It used to have a few problems.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -211,7 +211,7 @@ public int GetDecalAtlasMipCount()
         public Material GetBlitMaterial(bool useTexArray) { return useTexArray ? m_BlitTexArray : m_Blit; }
 
         ComputeBuffer m_DepthPyramidMipLevelOffsetsBuffer = null;
-
+        int m_PrevFrameColorPyramidMips = 0;
```
EvgeniiG: That's not a good location for this variable.
Remember, there is a color pyramid (or two) per HDCamera. They are allocated and managed by the HDCamera. Therefore, if you do need to store this value, please do it there.

EvgeniiG: What I mean is, if you have several cameras, you have several pyramids, and they could have different sizes for different frames.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2900,7 +2900,7 @@ void RenderSSR(HDCamera hdCamera, CommandBuffer cmd, ScriptableRenderContext ren
                     cmd.SetComputeFloatParam(cs, HDShaderIDs._SsrRoughnessFadeEnd,               roughnessFadeEnd);
                     cmd.SetComputeFloatParam(cs, HDShaderIDs._SsrRoughnessFadeRcpLength,         roughnessFadeRcpLength);
                     cmd.SetComputeFloatParam(cs, HDShaderIDs._SsrRoughnessFadeEndTimesRcpLength, roughnessFadeEndTimesRcpLength);
-                    cmd.SetComputeIntParam(  cs, HDShaderIDs._SsrDepthPyramidMaxMip,             info.mipLevelCount);
+                    cmd.SetComputeIntParam(  cs, HDShaderIDs._SsrDepthPyramidMaxMip,             info.mipLevelCount-1);
```
EvgeniiG: My bad. -.-

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: That's a good point, I got misled by the positioning of the         ComputeBuffer m_DepthPyramidMipLevelOffsetsBuffer = null;. 

That should probably moved as well, I'll do that tomorrow

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: It looked correct, I didn't try dramatic rescales tho. I'll try tomorrow, I don't have the code in front of me, but IIRC it looked like it was doing the correct thing in computing the count.

EvgeniiG: Maybe it's been fixed. I just looked at the math a few months ago and there were issues with it.

PR #3056, opened 2019-02-22
Real time area light shadows (Blurred EVSM)

FrancescoC-unity:
### Purpose of this PR

This PR introduces area light shadows for the real time rasterization pipeline. 

It uses EVSM to minimize VGPR pressure for everything else and to provide very soft look. 
It is possible to perform extra passes of blur if the user wants a more soft look. 

unity_if1pjryvcv

some improvements can be done (in terms of quality and perf.) but this is a solid start. 

---
### Release Notes
Add area light shadow support for real time rasterization pipeline. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Farealight-shadow-evsm-filtering&automation-tools_branch=master&unity_branch=trunk

- [x] Built a player
- Other: 

**Automated Tests**: Added a new test

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Medium

**NOTE: Some tests are failing locally, I am investigating them, but since is a biggish PR I thought of kicking the review early.**


FrancescoC-unity:
Ok, tests are fine (except for 1 SSS one but it fails for me on master as well in exactly the same way)

FrancescoC-unity:
So I have two tests failing on Katana. One is textured area lights which for some reasons didn't fail locally and my own :D I guess I forgot to update the reference image.

sebastienlagarde:
So from QA point of view, here is few thing to validate:
- Is shadow mask of rectangle light working ? I.e changing light in mixed mode, baking shadow mask, looking at with the debug view mode of shadow mask if it is ok (cautoin: compression of lightmap affect the quality of the result).
- When raytracing version: is the option "use raytrace shadow" correctly available and correctly replace the current area shadow (i.e disable the rendering of shadow in the atlas) ?
- When no raytracing: is the option "use raytrace shadow" correctly disable. And is "no shadow" correctly don't generate shadow in the atlas.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -306,17 +322,29 @@ public void ReserveShadows(Camera camera, HDShadowManager shadowManager, HDShado
...
-                m_ShadowRequestIndices[index] = shadowManager.ReserveShadowResolutions(viewportSize, allowResize);
+                m_ShadowRequestIndices[index] = shadowManager.ReserveShadowResolutions(viewportSize, shadowMapType);
         }
 
         public bool WillRenderShadows()
         {
             return m_WillRenderShadows;
         }
 
+        public static float GetAreaLightOffsetForShadows(Vector2 shapeSize, float coneAngle)
```
sebastienlagarde: put comment about what is happening here. Why we calculate offset

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
```diff
@@ -110,7 +110,13 @@ public struct LightData
         public int     cookieIndex;             // -1 if unused
         public int     tileCookie;              // (TODO: use a bitfield)
         public int     shadowIndex;             // -1 if unused (TODO: 16 bit)
-        public int     contactShadowIndex;      // -1 if unused (TODO: 16 bit)
+#if ENABLE_RAYTRACING
```
sebastienlagarde: I guess I will prefer to not put #ifdef inside this kind of struct that generate hlsl code after it, can cause problem. Let's just always have it and put -1 by default

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ShadowMoments.hlsl
```diff
@@ -29,6 +29,22 @@ float2 ShadowMoments_WarpDepth( float depth, float2 exponents )
     return float2( pos, neg );
 }
 
+float ShadowMoments_WarpDepth_PosOnly(float depth, float exponent)
```
sebastienlagarde: isn't there a mistake here, both functoin are the same

ShadowMoments_WarpDepth_PosOnly
ShadowMoments_WarpDepth_PosOnlyBaseTwo

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
FrancescoC-unity: Can do. Worth noting that the thing that is added here inside the ifdef is not a new field, but really just a couple of functions to aid the C# side, they are ignored regardless in the hlsl gen.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
FrancescoC-unity: 👍

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ShadowMoments.hlsl
FrancescoC-unity: Nope, the BaseTwo version uses exp2 and expects the log2(e) factor baked in the exponent. Can add comment to make it more explicit.

FrancescoC-unity:
> PR looks good. See few comment. Can you also send me an excel diff with Fred's tool with before/after of the deferred.compute ? Thanks

Will do tomorrow when I am back in office 🙂

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ShadowMoments.hlsl
sebastienlagarde: ah right :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
sebastienlagarde: Ah yes, sorry my bad. Ok, no need to change anything as it will not generate extra field

PR #3071, opened 2019-02-25
HDRP dynamic resolution fix for XR

fabien-unity:
### Purpose of this PR
Some platforms can't clear a partial render target so we directly draw black triangles for guard bands.

---
### Testing status
**Katana Tests**: [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-dynamic-resolution-fix&unity_branch=trunk)

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Medium

---
### Comments to reviewers
This code is more or less implemented on the C++ side but with extra complexity for stereo that ultimately clear the entire target in some cases ... it's now done entirely in C# to avoid any surprises.

If performance is somehow impacted on consoles, we could abstract this logic in a new ClearGuardBand() functions. Let me know if you think this is required.


FrancescoC-unity:
It looks fine to me. In this specific instance I think is fine to manually clear since there is no advantage to be taken with fast clears anyway since all pixels are going to be accessed. 

My only comment would probably be to rename the shader to something more generic (clearPS or something like that) as it is has nothing to do with guardbands in itself, it's just a write black shader :-)

Chman:
I agree with @FrancescoC-unity, I'd rather have the shader get a generic name given that it doesn't do anything specific. Aside from that it's good.

fabien-unity:
Alright, I've renamed the material to "ClearBlack" and merged master in this branch. Tests are running again.

PR #3073, opened 2019-02-25
Preliminary changes for XR deferred shading

fabien-unity:
### Purpose of this PR
Add XR macros and cleanup code before next PR

---
### Testing status
**Katana Tests**: [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-deferred-prep&unity_branch=trunk)

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [x] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [x] C# and shader warnings (supress shader cache to see them)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #3074, opened 2019-02-25
Fix HDRP deferred shading with stereo instancing

fabien-unity:
### Purpose of this PR
Fix HDRP deferred shading with stereo instancing
- handle light list buffer generation (first half is for left eye, second half is for right eye)
- handle indirect tile dispatch (eye index is encoded in 2 bits, leaving 15 bits for tileX and tileY)

---
### Testing status
**Katana Tests**: [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-deferred-fixes&unity_branch=trunk)

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**: High

**Halo Effect**: High

---
### Comments to reviewers
This branch also includes preliminary changes from [another PR](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3073). I recommend you diff the code against it ([like this](https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/hdrp-xr-deferred-prep...hdrp-xr-deferred-fixes)).


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
```diff
@@ -128,20 +128,32 @@ StructuredBuffer<uint> g_TileList;
 void SHADE_OPAQUE_ENTRY(uint2 groupThreadId : SV_GroupThreadID, uint groupId : SV_GroupID)
 {
     uint  tileIndex   = g_TileList[g_TileListOffset + groupId];
-    uint2 tileCoord   = uint2(tileIndex & 0x0000FFFF, tileIndex >> 16); // see builddispatchindirect.compute
+    uint2 tileCoord   = uint2(tileIndex & 0x7FFF, (tileIndex >> 15) & 0x7FFF); // see builddispatchindirect.compute
     uint2 pixelCoord  = tileCoord * GetTileSize() + groupThreadId;
 
+    UNITY_STEREO_ASSIGN_COMPUTE_EYE_INDEX(tileIndex >> 30);
```
JulienIgnace-Unity: Not super obvious until you look at builddispatchindirect.compute
Might be nice to have constants like stereoTileEyeIndexBit or something like that.

fabien-unity: I agree. I've added variables to LightLoop.cs and regenrated the HLSL headers

sebastienlagarde:
@fabien-unity we need an update of this PR with master. @mmikk any insight ? Fabien, feel free to poke Morten on slack ? :)

fabien-unity:
I've merged latest master in this branch.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -321,11 +321,8 @@ public static void Sanitize(ref FrameSettings sanitazedFrameSettings, Camera cam
 
             // We have to fall back to forward-only rendering when scene view is using wireframe rendering mode
             // as rendering everything in wireframe + deferred do not play well together
-            if (GL.wireframe || stereo) //force forward mode for wireframe
+            if (GL.wireframe || stereoDoubleWide)
```
fabien-unity: is there a way to default to forward in VR only, but still has the ability to enable it back in the asset?

sebastienlagarde: I think we need a HDVR template for this...

fabien-unity:
I've used the ShaderPerformanceTool on Deferred.compute to compare the impact of this PR and it does not affect the number of registers used.
There's a slight change in bytecode size (between -8 and +4) depending in the variants.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
fabien-unity: Ok, I'll work with @benjaminsunity to make a HDVR template that defaults to forward

sebastienlagarde:
@fabien-unity , @mmikk talk about lightlistbuild.compute, not deferred.compute as this changed doesn't affect deferred.compute. The build light list is done async so I will expect that it shouldn't be an issue in term of performance, but will be good to do a sanity check. try to run the ShaderPerformanceTool  on lightlistbuild.compute. Thanks

fabien-unity:
Right ... ! 
I've run the tool on lightlistbuild.compute and I can see an increase in SGPR used (from 35 to 38) for TileLightListGen and TileLightListGen_FeatureFlags.
However, SPGR count is still the same (40) and the occupancy is till at 7.

PR #3080, opened 2019-02-26
Fix errors with light explorer window

laurenth-unity:
Fix build errors with light explorer window open while building game.
Also fixes Planar reflection probe weight not displaying correctly

### Purpose of this PR

Fixes bug https://fogbugz.unity3d.com/f/cases/1130261/

---
### Release Notes

- Fix build errors if light explorer window is open
- Fix planar reflection probe weight being not displayed correctly in light explorer
---
### Testing status
Small fix did only local tests. Tested building on windows with light explorer window open and each tab one by one.


sebastienlagarde:
Please add changelog

RSlysz:
Replaced by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3092

PR #3084, opened 2019-02-26
Disable async compilation in all test project to have correct katana test

sebastienlagarde:
All in the title.


sebastienlagarde:
Merging as we need this to have a green katana.

phi-lira:
LGTM. Will it be backported? @Verasl fyi.

sebastienlagarde:
Yes. It have been backport

PR #3093, opened 2019-02-27
Area Shadows and Double Light contribution

anisunity:
Defining a IncludeIndirectLighting for HDRP/Unlit to avoid double contribution for area lights (reflections + analytic)
Fixing a bad merge related to non-raytraced area shadows


FrancescoC-unity:
👌

PR #3096, opened 2019-02-27
Fix issue with setting a not available RT when disabling motion vectors

FrancescoC-unity:
Same as https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3058 , just not reusing same rendertarget but another one.


FrancescoC-unity:
Close to retarget master

PR #3111, opened 2019-02-28
Recorder implementation for HDRP

JulienIgnace-Unity:
### Purpose of this PR
This PR implements Record for HDRP. Now every type of Recorder captures will work.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Frecorder-implementation&automation-tools_branch=master&unity_branch=trunk

- [ ] Built a player
- Other: Followed the test protocol described in the fogbugz issue and checked that everything was fine.

---
### Comments to reviewers
@phi-lira I added you to the PR because I moved CameraCaptureBridge out of LWRP into Core.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -949,5 +956,26 @@ void ReleaseHistoryBuffer()
         {
             m_HistoryRTSystem.ReleaseAll();
         }
+
+        public void ExecuteCaptureActions(RTHandleSystem.RTHandle input, CommandBuffer cmd)
+        {
+            if (m_RecorderCaptureActions == null)
```
Dominique-Unity: You may also early-out if m_RecorderCaptureActions is empty (but non-null), which may happen if a recorder has been active for this camera in the past.

Dominique-Unity:
Thanks for doing this! With LWRP, post-processing initially ended up using a code path that wasn't being recorded: did you have a chance to try it out? We can probably ask our QA if you can think of scenarios that would be worth looking at.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
JulienIgnace-Unity: I don't see any API to check if an IEnumerator is empty. Am I missing something?

JulienIgnace-Unity:
> Thanks for doing this! With LWRP, post-processing initially ended up using a code path that wasn't being recorded: did you have a chance to try it out? We can probably ask our QA if you can think of scenarios that would be worth looking at.

@Dominique-Unity I test both with and without post-processing and it worked fine. It's pretty safe in HDRP as now all path goes through the same code anyway.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -959,7 +959,7 @@ void ReleaseHistoryBuffer()
 
         public void ExecuteCaptureActions(RTHandleSystem.RTHandle input, CommandBuffer cmd)
         {
-            if (m_RecorderCaptureActions == null)
+            if (m_RecorderCaptureActions == null || !m_RecorderCaptureActions.MoveNext())
```
Dominique-Unity: Prefect. Thanks!

PR #3087, opened 2019-02-26
Add remaining ShaderProperty types to Blackboard

Kink3d:
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?
This PR adds the reamining Property types to the Blackboard (Matrix2, Matrix3, Matrix4, Gradient, Sampler State). This means users can now define uniforms of these types.

alt text

What this really adds:
- Can now set matrices via C#
- Can reuse uniforms to reduce graph clutter
- Can have these types as input slots to subgraphs (now supports all port types)

See "Comments for reviewers" for detailed info...

---
### Release Notes
Added:
- You can now add Matrix, Sampler State and Gradient properties to the Blackboard.

Changed:
- Blackboard fields that had no affect on Sub-Graph input ports have been removed from the Blackboard in Sub-Graphs.

---
### Testing status
**Katana Tests**: Running

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [X] Checked new UI names with UX convention
- [X] Tested UI multi-edition + Undo/Redo
- [x] C# and shader warnings (supress shader cache to see them)
- Other: 
   - Created graphs using these types as propeties/slots/nodes in as many combinations as I could
   - Tested upgrading Shader Graphs and Subgraphs with currently supported property types to this branch.

**Automated Tests**: What did you setup?
Looking into if this is needed.

Any test projects to go with this to help reviewers?

---
### Overall Product Risks
**Technical Risk**: Medium - Changes to Gradient and SamplerState required to support them going through subgraphs and single property definitions are substantial and deep.

**Halo Effect**: Medium - Mostly only affects the mentioned types. However, some changes to Subgraph code gen (see details below) added some halo risk. Test subgraphs heavily.

---
### Comments to reviewers
- This work required substantial refactors to Gradients to allow them to be inlinable, which is required to support subgraph inputs (@spaztronaut / @zadi0648  Make sure you test everything related to Gradient nodes/slots/properties).
- Made considerable changes to SamplerState properties (@spaztronaut / @zadi0648  Make sure you test everything related to SamplerState nodes/slots/properties).
- Made minor changes to matrix shader properties. This could affect default inputs for matrix slot types (@spaztronaut / @zadi0648  Make sure you test everything related to Matrix nodes/slots/properties).. 
- SubgraphNode no longer collects graph properties from the referenced graph, only from its nodes.
- Includes a change that makes all Subgraph nodes generate their referenced graphs `NodeCode` in `ForReals` mode at all times.
[...]

PR #3112, opened 2019-02-28
Nested sub graphs

pbbastian:
### Purpose of this PR
This PR enables nesting of sub graphs. It should also improve on the stability of sub graphs. We now bake everything we need to know about a sub graph at import time, such that Sub Graph Node no longer needs any access to the referenced graph. Likewise things like shader stage capability are also computed at import time. 

I also made some modifications to `GraphData.ValidateGraph`. Previously each `ValidateNode` call would recursively call `ValidateNode` on all children, meaning that each node would end up having `ValidateNode` called many times. This caused my error checking code in `SubGraphNode` to spam messages because it was being called multiple times. `GraphData.ValidateGraph` now ensures that all children of a node has had `ValidateNode` called before calling it on the parent node, and so each node doesn't have to call into child nodes anymore.

---
### Testing status
**Katana Tests**: On the way https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?unity_branch=trunk&ScriptableRenderLoop_branch=sg%2Fnested-subgraphs&automation-tools_branch=master

- [ ] Built a player
- [x] ~Checked new UI names with UX convention~ No new UI
- [x] ~Tested UI multi-edition~ (Not relevant) + Undo/Redo
- Verify that dependency tracking works for main and sub graphs using nested sub graphs.
- Verified that the recursion check works and triggers errors in the console.
- Verified that adding in a recursion in a nested sub graph, and then removing that recursion, causes the main graph to go pink and then back to working.

**Automated Tests**: Nothing new.

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Low

---
### Comments to reviewers
Most of the action happens in SubGraphDatabaseImporter and SubGraphNode. The former class is an importer for a new asset type, for which only 1 asset at a fixed path can exist. I've set SubGraphImporter to have an import queue offset of 1, SubGraphDatabaseImporter to have an offset of 2, and finally ShaderGraphImporter to have an offset of 3. This means that the new importer essentially acts as a callback in between sub graphs and main graphs. The SubGraphImporter now only imports a timestamp, which I use in SubGraphDatabaseImporter to determine whether a sub graph (and all sub graphs that are using it) should be re-processed.

Please do try to stress this as much as you can by nesting sub graphs many times and using them in lots of places.


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [6.5.0] - 2019-XX-XX
+### Added
+- You can now use sub graphs in other sub graphs.
```
sharlenet: According to the pinned thread in shadergraph_megateam, we're capitalising Sub Graph.

You can now use Sub Graphs in other Sub Graphs.

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -744,8 +729,53 @@ public void ValidateGraph()
...
+            var stack = StackPool<AbstractMaterialNode>.Get();
             foreach (var node in GetNodes<AbstractMaterialNode>())
-                node.ValidateNode();
+            {
+                stack.Push(node);
+            }
+            while (stack.Count > 0)
+            {
+                var node = stack.Pop();
+                if (permanentMarks.Contains(node.tempId.index))
```
Kink3d: Empty if statement. Continue instead?

com.unity.shadergraph/Editor/Data/Graphs/ShaderGraphRequirements.cs
```diff
@@ -1,9 +1,10 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using UnityEditor.Graphing;
 
 namespace UnityEditor.ShaderGraph
 {
+    [Serializable]
```
Kink3d: Given that we are completely changing how these requirements work isnt serializing this pretty dangerous? Is there another way?

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -281,253 +315,164 @@ private void ValidateShaderStage()
...
-                }
+                hasError = true;
+                owner.AddValidationError(tempId, "Sub Graph failed to load");
+            }
+            else if (subGraphData.isRecursive)
+            {
+                hasError = true;
+                owner.AddValidationError(tempId, "Recursive Sub Graphs are not allowed");
+            }
+            else if (!subGraphData.isValid)
```
Kink3d: Super minor but imo the least descriptive error (this one) should be least likely to display (last in statement)

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
```diff
@@ -281,253 +315,164 @@ private void ValidateShaderStage()
...
-            properties.AddRange(props);
-        }
-
-        private string SubGraphFunctionName(GraphContext graphContext)
-        {
-            var functionName = subGraphAsset != null ? NodeUtils.GetHLSLSafeName(subGraphAsset.name) : "ERROR";
-            return string.Format("sg_{0}_{1}_{2}", functionName, graphContext.graphInputStructName, GuidEncoder.Encode(referencedGraph.guid));
+            foreach (var property in subGraphData.properties)
+            {
+                properties.Add(property.GetPreviewMaterialProperty());
```
Kink3d: I think this is going to trip up #3087 should be checked locally one way or another.

com.unity.shadergraph/Editor/Data/SubGraph/SubGraphAsset.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using UnityEditor.Experimental.AssetImporters;
+using UnityEditor.Graphing;
+using UnityEngine;
+
+namespace UnityEditor.ShaderGraph
+{
+    class SubGraphAsset : ScriptableObject//, ISerializationCallbackReceiver
```
Kink3d: Remove comment

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
pbbastian: Good point, will change

com.unity.shadergraph/Editor/Data/Graphs/ShaderGraphRequirements.cs
pbbastian: This is only for serialising in the imported object (i.e. in the Library folder), so if we make any changes we can just bump the importer version

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
pbbastian: ahh yeah, good point

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
pbbastian: This is only for properties gathered via nodes in the sub graph. Maybe I should rename `properties` to something like `nodeProperties`.

com.unity.shadergraph/Editor/Data/Graphs/ShaderGraphRequirements.cs
Kink3d: Ah I did wonder that ignore then!

com.unity.shadergraph/Editor/Data/Nodes/Utility/SubGraphNode.cs
Kink3d: Ah these are individual node properties? Then yes a rename for clarity please.

zadi0648:
Testing in progress:
https://confluence.hq.unity3d.com/display/SFTD/%5B+ShaderGraph+%5D+Nested+Sub+Graphs

pbbastian:
@zadi0648 The issues have been fixed :)

zadi0648:
Approved, with one minor concern/bug left marked as fix later, maybe out of our control:
https://confluence.hq.unity3d.com/display/SFTD/%5B+ShaderGraph+%5D+Nested+Sub+Graphs#id-[ShaderGraph]NestedSubGraphs-[%E2%9D%97]MissingSubGraphasseterrorbadgeonlyshowsuponfileclose&reopen

PR #3127, opened 2019-03-01
[HDRP] add shader parameter map in DebugMenu

RSlysz:
### Purpose of this PR
Add shader parameter mapping in DebugMenu
Add scripting API to configure DebugData for DebugMenu

---
### Release Notes
- Added shader parameter mapping in DebugMenu
- Added scripting API to configure DebugData for DebugMenu

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fframepass&unity_branch=trunk&automation-tools_branch=master

- [x] Opened test project + Run graphic tests locally (unit test ok, graphic test ok)
- [x] Built a player
- [ ] Checked new UI names with UX convention (NA)
- [ ] Tested UI multi-edition + Undo/Redo (NA)
- Other: 

**Automated Tests**: 

---
### Overall Product Risks
**Technical Risk**: High (faire amount of files modified)

**Halo Effect**: Medium (Should only be concealed in DebugMenu usage)

---
### Comments to reviewers

Here is an example script to use the scriptable part of the API:
```C#
using UnityEngine.Experimental.Rendering.HDPipeline.Attributes;
using UnityEngine.Rendering;

namespace UnityEngine.Experimental.Rendering.HDPipeline
{
    public class ParameterMapTester : MonoBehaviour
    {
        void ChangeMaterial(MaterialSharedProperty prop)
            => (new FramePassSettings(FramePassSettings.@default)).SetFullscreenOutput(prop).FillDebugData((RenderPipelineManager.currentPipeline as HDRenderPipeline).debugDisplaySettings.data);

        [ContextMenu("Material_None")]
        void MaterialNone() => ChangeMaterial(MaterialSharedProperty.None);

        [ContextMenu("Material_Albedo")]
        void MaterialAlbedo() => ChangeMaterial(MaterialSharedProperty.Albedo);

        [ContextMenu("Material_Normal")]
        void MaterialNormal() => ChangeMaterial(MaterialSharedProperty.Normal);

        [ContextMenu("Material_Smoothness")]
        void MaterialSmoothness() => ChangeMaterial(MaterialSharedProperty.Smoothness);

        [ContextMenu("Material_AmbientOcclusion")]
        void MaterialAmbientOcclusion() => ChangeMaterial(MaterialSharedProperty.AmbientOcclusion);

        [ContextMenu("Material_Metal")]
        void MaterialMetal() => ChangeMaterial(MaterialSharedProperty.Metal);

        [ContextMenu("Material_Specular")]
        void MaterialSpecular() => ChangeMaterial(MaterialSharedProperty.Specular);

        [ContextMenu("Material_Alpha")]
        void MaterialAlpha() => ChangeMaterial(MaterialSharedProperty.Alpha);

        void ChangeLighting(LightingProperty prop)
            => (new FramePassSettings(FramePassSettings.@default)).SetFullscreenOutput(prop).FillDebugData((RenderPipelineManager.currentPipeline as HDRenderPipeline).debugDisplaySettings.data);

        [ContextMenu("Lighting_All")]
        void LightingAll() => ChangeLighting(LightingProperty.DiffuseOnlyDirectional | LightingProperty.DiffuseOnlyIndirect | LightingProperty.SpecularOnlyDirectional | LightingProperty.SpecularOnlyIndirect);

        [ContextMenu("Lighting_Diffuse")]
        void LightingDiffuse() => ChangeLighting(LightingProperty.DiffuseOnlyDirectional | LightingProperty.DiffuseOnlyIndirect);

        [ContextMenu("Lighting_Specular")]
        void LightingSpecular() => ChangeLighting(LightingProperty.SpecularOnlyDirectional | LightingProperty.SpecularOnlyIndirect);

        void ChangeLighting(DebugFullScreen prop)
            => (new FramePassSettings(FramePassSettings.@default)).SetFullscreenOutput(prop).FillDebugData((RenderPipelineManager.currentPipeline as HDRenderPipeline).debugDisplaySettings.data);

        [ContextMenu("FullScreen_None")]
        void FullScreenNone() => ChangeLighting(DebugFullScreen.None);

        [ContextMenu("FullScreen_Depth")]
        void FullScreenDepth() => ChangeLighting(DebugFullScreen.Depth);

        [ContextMenu("FullScreen_SSAO")]
        void FullScreenSSAO() => ChangeLighting(DebugFullScreen.ScreenSpaceAmbientOcclusion);

        [ContextMenu("FullScreen_MotionVectors")]
        void FullScreenMotionVector() => ChangeLighting(DebugFullScreen.MotionVectors);
    }
}
```


com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
```diff
@@ -261,50 +297,190 @@ void BuildDebugRepresentation()
...
+            }
+        }
         public int debugViewEngine { get { return m_DebugViewEngine; } }
         public DebugViewVarying debugViewVarying { get { return m_DebugViewVarying; } }
         public DebugViewProperties debugViewProperties { get { return m_DebugViewProperties; } }
         public int debugViewGBuffer { get { return m_DebugViewGBuffer; } }
 
-        int                             m_DebugViewMaterial = 0; // No enum there because everything is generated from materials.
-        int                             m_DebugViewEngine = 0;  // No enum there because everything is generated from BSDFData
+        const int kDebugViewMaterialBufferLength = 10;
```
JulienIgnace-Unity: Where does the 10 comes from?

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
```diff
@@ -261,50 +297,190 @@ void BuildDebugRepresentation()
...
+
+        //buffer must be in float as there is no SetGlobalIntArray in API
+        static float[] s_DebugViewMaterialOffsetedBuffer = new float[kDebugViewMaterialBufferLength + 1]; //first is used size
+
+        int[]                m_DebugViewMaterial = new int[kDebugViewMaterialBufferLength + 1]; // No enum there because everything is generated from materials.
```
JulienIgnace-Unity: Can we put a comment to explain what's in this array? Not super clear on first reading

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
```diff
@@ -50,8 +50,34 @@ public enum DebugViewProperties
...
+            AmbientOcclusion,
+            /// <summary>There is no equivalent for AxF, Fabric and Hair shaders.</summary>
+            Metal,
+            Specular,
+            Alpha,
+
+            //[Todo: see for particular properties like aniso...]
+        }
+
+        public class FramePassMaterialMappingAttribute : Attribute
```
JulienIgnace-Unity: Why FramePass? From what I understand it's mapping from a particular shader property to a "common" property, right?
Maybe something like MaterialDebugPropertyMapping?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
```diff
@@ -0,0 +1,118 @@
...
+using UnityEngine;
+using UnityEngine.Experimental.Rendering.HDPipeline.Attributes;
+using static UnityEngine.Experimental.Rendering.HDPipeline.MaterialDebugSettings;
+
+namespace UnityEngine.Experimental.Rendering.HDPipeline
+{
+    [Flags]
+    public enum LightingProperty
+    {
+        DiffuseOnlyDirectional = 1 << 0,
```
JulienIgnace-Unity: I suppose by Directional we mean Direct?
Directional is confusing with the light type.
Something like DirectDiffuseOnly and IndirectDiffuse only is probably clearer.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
```diff
@@ -0,0 +1,118 @@
...
+
+    public enum DebugFullScreen
+    {
+        None,
+        Depth,
+        ScreenSpaceAmbientOcclusion,
+        MotionVectors
+    }
+
+    public unsafe struct FramePassSettings
```
JulienIgnace-Unity: Not convinced by the FramePass name here.
I think "pass" is confusing with what is actually a "render pass".
From what I understand this structure defines what properties (in a wide sense) we want to output for a given frame? Then maybe something like FrameOutputSettings? More ideas would be welcome :)
Side note: we should probably not call this Frame either because during one frame we can render multiple cameras... this is also true for the FrameSettings structure. This is a discussion probably best kept for another PR.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
```diff
@@ -0,0 +1,118 @@
...
+            return ref *thisPtr;
+        }
+
+        // Usage example:
+        // (new FramePassSettings(FramePassSettings.@default)).SetFullscreenOutput(prop).FillDebugData((RenderPipelineManager.currentPipeline as HDRenderPipeline).debugDisplaySettings.data);
+        public void FillDebugData(DebugDisplaySettings.DebugData data)
+        {
+            data.materialDebugSettings.SetDebugViewCommonMaterialProperty(m_MaterialProperty);
+
+            switch (m_LightingProperty)
```
JulienIgnace-Unity: All of these seems mutually exclusive: setting LightingProperty.DiffuseOnlyDirectional | LightingProperty.SpecularOnlyDirectional will throw an exception which is weird from the user stand point as you expect this kind of thing to just work.
Why not make it a regular enum? (with an added "Everything" entry). No error can occur in this case and it's more understandable for users.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassForward.hlsl
```diff
@@ -113,17 +113,27 @@ void Frag(PackedVaryingsToPS packedInput,
...
-
         bool needLinearToSRGB = false;
+        float3 result = float3(1.0, 0.0, 1.0);
 
-        GetPropertiesDataDebug(_DebugViewMaterial, result, needLinearToSRGB);
-        GetVaryingsDataDebug(_DebugViewMaterial, input, result, needLinearToSRGB);
-        GetBuiltinDataDebug(_DebugViewMaterial, builtinData, result, needLinearToSRGB);
-        GetSurfaceDataDebug(_DebugViewMaterial, surfaceData, result, needLinearToSRGB);
-        GetBSDFDataDebug(_DebugViewMaterial, bsdfData, result, needLinearToSRGB);
+        for (int index = 1; index <= bufferSize; index++)
```
JulienIgnace-Unity: How does a loop work here? Result will be overwritten by the last iteration.

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
RSlysz: It is an arbitrary length as for now we support less then 10 different shaders, it is ok.
It is also the length of the array that we pass to the shader and its length cannot change in the Unity way to comunicate with shaders so we want to keep it short if possible.

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
RSlysz: Well sure. It is a reliquate from when we still wonder how to call it.
Is CommonMaterialProperty ok ? I feel like this could be used beyond debugging

JulienIgnace-Unity: CommonMaterialPropertyMapping agreed that we should probably avoid "Debug" in there.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
RSlysz: The other part of the feature is currently handled by @fredpointzero which is currently adding missing direct/indirect view mode.
The idea is to have 4 part that can be drawn or not together. They come from this two separation axis:
- Direct/Indirect
- Specular/Diffuse
One can see all the Diffuse, or all the Direct or only the IndirectSpecularPart regarding what he is debugging.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
RSlysz: Will be modified in next PR Fred is doing to not polute its merge

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/FramePass/FramePassSettings.cs
RSlysz: Will be modified in next PR Fred is doing to not polute its merge

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
RSlysz: I tryed to do as we said: having a generated index slot in the buffer per material. But each time I use the generator, it broke a shader kernel which lead to me cleaning the library/shadercache and not beeing able to see changes.
As it were only for reeadability, I fallbacked on heavily comment code in shader and do not change the length of the buffer here.

sebastienlagarde:
hey, can you merge master and re-run, the last katana was once month ago and I want to be sure that there is no conflict with new code. Thanks

PR #3130, opened 2019-03-01
Fix FB#1098255 - Position view -> position object transform does not work [Hold]

Kink3d:
### Purpose of this PR
Fixes Fogbugz case 1098255 - Position view -> position object transform does not work

Currently all graph operations that go through positionWS in HDRP are using absolute position WS, this PR changes that to relative WS. This change that functionality in all active HD shadergraphs (but it needs to happen). This fixes basically all space transformations in HD shader graphs as they all went though WS.

Access to absolute WS should be added in a separate PR, via another entry in Space enum or a separate node. This is the correct fix i nthis case IMO.

---
### Release Notes
Changelog

---
### Testing status
Will run when master isnt red

- Other: 

N/A

---
### Overall Product Risks

**Halo Effect**: None, Low, Medium, High? Low - Affects everything with world space but it was basically all guaranteed to be wrong previously

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fbugfix-1098255&automation-tools_branch=master&unity_branch=trunk#

---
### Comments to reviewers
As this bug is currently assigned to SG (and is high priority) we will handle the merging and backporting of this fix by default. If you want to take over and merge it, please remember to backport!


sebastienlagarde:
Arf, I just see this PR :(, this is indeed a big change and super important and we need to offer an AbsoluteWorldPosition node. And I fully agree. Launching katana as it seems previous katana result have disappeard

sebastienlagarde:
hmm... after reflection I think we need to discuss more about this. I will not merge this PR yet and rather discuss it with you and Chris when I will be at Seattle this week.

sebastienlagarde:
PR fixed by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3783

PR #3134, opened 2019-03-02
Fix FB#1132342 - Multiply Bug. Zeroing components in code generation

Kink3d:
### Purpose of this PR
This PR fixes Fogbugz Case #1132342 - Multiply Bug. Zeroing components in code generation
Previously when the default input values were changed for a DynamicValueMaterialSlot each matrix row was written back into the column of the same index. this obviously caused lots of errors when inputting values into this field. Now the values are written into each matrix component individually, and correctly.

---
### Release Notes
Changelog

---
### Testing status
Pass

- Other: 

N/A

N/A

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High? Low

**Halo Effect**: None, Low, Medium, High? Low (Only Multiply node uses this slot type)

---
### Comments to reviewers


spaztronaut:
i take back my approval. something is still off

PR #3137, opened 2019-03-02
Fix FB#1131755 - Drag-dropping normal map texture into a shader graph causes errors

Kink3d:
### Purpose of this PR
Currently if you drag a texture asset with type set to Normal Map on to a Shader Graph it fails to create a node and throws a nullref error. This is because Shader Graph tries to assign the texture type before the node has been created. The mutator it uses to do this forces validation, which fails as the node hasnt been created yet.

This Pr moves the type set call to after the node is created, which makes it function as intended.

---
### Release Notes
Changelog

---
### Testing status
Pass

- Other: 

N/A

N/A

---
### Overall Product Risks

**Halo Effect**: None, Low, Medium, High? None

---
### Comments to reviewers


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [6.5.0] - 2019-XX-XX
+### Fixed
+- You can now drag normal map type textures on to a Shader Graph to create Sample Texture 2D nodes with the correct type set.
```
sharlenet: Minor change - made '"onto" one word:

You can now drag normal map type textures onto a Shader Graph to create Sample Texture 2D nodes with the correct type set.

PR #3138, opened 2019-03-02
Fix FB#1131211 - Voronoi Node Serialization Issue

Kink3d:
### Purpose of this PR
During a recent refactor `VoronoiNode` class got renamed in error, causing all previously serialized voronoi nodes fail to deserialize. 

This PR fixes the name of the class, while maintaining the ability to deserialize previously serialized node instances (from both the previous broken versions, and the ones serialized with the wrong names).

---
### Release Notes
Changelog

---
### Testing status
Pass

- Other: 

This fixes a previously broken automated test (which is in PR #2845 )

See bug case

---
### Overall Product Risks

**Halo Effect**: None, Low, Medium, High? Low - Affects serialization, but fixes it.

---
### Comments to reviewers


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [6.5.0] - 2019-XX-XX
+### Fixed
+- Fixed an error when opening saved Shader Graphs containing one or more Voronoi nodes.
```
sharlenet: If the user is the one opening the Shader Graphs then:
Fixed an error that previously occurred when you opened saved Shader Graphs containing one or more Voronoi nodes.
Else:
Fixed an error that previously occurred when opening saved Shader Graphs containing one or more Voronoi nodes.

PR #3141, opened 2019-03-04
Fix mobile lighting attenuation

rjonaitis:
### Purpose of this PR
Fixes light attenuation differences between Mali and Adreno devices due to precision differences.
Fixes mobile light attenuation always using SHADER_HINT_NICE_QUALITY path.

attenuation

---
### Release Notes
Fix light attenuation precision on mobiles

---
### Testing status
Manually tested reproduction project on Mali and Adreno devices

Project attached to fogbugz case [1131387](https://issuetracker.unity3d.com/product/unity/issues/guid/1131387/)

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: low

---


com.unity.render-pipelines.lightweight/ShaderLibrary/Lighting.hlsl
```diff
@@ -143,7 +143,7 @@ Light GetAdditionalLight(int i, float3 positionWS)
     float distanceSqr = max(dot(lightVector, lightVector), HALF_MIN);
 
     half3 lightDirection = half3(lightVector * rsqrt(distanceSqr));
-    half attenuation = DistanceAttenuation(distanceSqr, distanceAndSpotAttenuation.xy);
+    float attenuation = DistanceAttenuation(distanceSqr, distanceAndSpotAttenuation.xy);
```
phi-lira: final attenuation don't need to the full precision. we should probably cast inside DistanceAttenuation function after computing lightAtten * smoothFactor?

phi-lira:
Good catch! Just a minor thing otherwise fine to merge.

phi-lira:
Also add release notes by adding a entry to [changelog.md](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.lightweight/CHANGELOG.md) file

com.unity.render-pipelines.lightweight/ShaderLibrary/Lighting.hlsl
rjonaitis: either attenuation has to be float, or it can be half, but then DistanceAttenuation and AngleAttenuation has to be multiplied before assignment.

rjonaitis: as in https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3141/commits/2616a343c3c0df6ccf2a57294388eab6c2b50a58

rjonaitis:
Also assuming the shader would define SHADER_HINT_NICE_QUALITY, constants setup considers only platform for passing parameter, so it might have mismatch
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.lightweight/Runtime/Passes/SetupLightweightConstanstPass.cs#L139

phi-lira:
I'm not sure how to run katana as it seems you forked the repo.

PR #3143, opened 2019-03-04
Move scene color to R11G11B10 

FrancescoC-unity:
### Purpose of this PR
Move the scene color format from RGBA16F to R11G11B10. This will save a significant amount of performance.

On the VFX holotable demo scene on Xbox One X, it saved around 2 milliseconds. 

Note that I had to update visual tests that exhibit two differences: 
 - Some HDRI skies had high values resulting in lighting differences (case of the hair and some distortion in which it was the reflection bit that was off)
 - Differences in banding. 

---
### Release Notes
Move color format from RGBA16F to R11G11B10. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Fmove-scene-color-to-111110&automation-tools_branch=master&unity_branch=trunk

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Medium


Chman:
LGTM.
@sebastienlagarde do we want this as an option in the HD asset instead? Right now most of the pipeline makes the assumption that we don't care about alpha, but we'll need it at some point (film/compositing/AR).

FrancescoC-unity:
image

FrancescoC-unity:
HDRP tests are all passing, VFX are not. Do we know if they are known to be failing or (more likely) they need to be updated with this change as well?

FrancescoC-unity:
@julienf-unity  There are indeed some differences in the vfx tests, some bright colors are less bright. I guess we need to discuss whether this is a dealbreaker on your side.

julienf-unity:
So VFX Test project is still in half at the moment ? I'll check the differences, it's possible some emissive values / overdraw in VFX tests are stupidly high too. I have no objection about the PR, saving half the bandwidth is valuable.

FrancescoC-unity:
The tests are not Half, are using 11-11-10 as that's the new default. I can move them to use 16F for now until you can have a look or I can update them to their 11-11-10 version

julienf-unity:
Ok so VFX tests are currently failing on this branch. You need to either change the setting to half or update the tests. As @PaulDemeulenaere  is currently moving/renaming some tests,  just moving the test project to half should be easier and avoid future conflicts. We can update it to 11_11_10 ourselves a bit later.

PR #3153, opened 2019-03-05
Hdrp/add emission color normalization

alelievr:
### Purpose of this PR
Add option in preferences to toggle the light emission normalization

image
image

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FAddEmissionColorNormalization&automation-tools_branch=master&unity_branch=trunk)

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -464,6 +464,12 @@ static void DrawEmissionContent(SerializedHDLight serialized, Editor owner)
             }
             else
                 EditorGUILayout.PropertyField(serialized.settings.color, s_Styles.color);
+            
+            if (HDRenderPipelinePreferences.lightColorNormalization)
```
JulienIgnace-Unity: Did you test with multiple selection? I don't think this will work in this case.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/EmissionNode.cs
```diff
@@ -117,6 +131,10 @@ public void GenerateNodeFunction(FunctionRegistry registry, GraphContext graphCo
                         intensitySlot.concreteValueType.ToString(precision));
                     using (s.BlockScope())
                     {
+                        if (normalizeColor.isOn)
+                        {
+                            s.AppendLine("ldrColor = ldrColor * rcp(Luminance(ldrColor));");
```
JulienIgnace-Unity: Doesn't this break for pure black?
Also, might be interesting to check if the compiler actually folds the computation for non exposed colors.
We should do it in the material property for exposed colors though, is that doable in the current state of shader graph?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
alelievr: Indeed, added a ChangeCheck guard to fix the issue

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/Nodes/EmissionNode.cs
alelievr: Fixed with a max on the Luminance.

PR #3161, opened 2019-03-06
Update git attributes to avoid asset CRLF conversion on windows

alelievr:
### Purpose of this PR
Update git attributes to avoid asset CRLF conversion on windows, as unity only serialize text assets using LF line endings it avoids to have asset changed.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FUpdateGitAttributesForAssets&automation-tools_branch=master&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Unity assets are always serialized using lf endings
+*.asset text eol=lf


RSlysz:
Forgot to green it, but green it was.

Also, I checked *.shadergraph to see how they handle this case to not cause issue. But they just do not end any line. (Which make they file hard to read but this is another issue)

PR #3176, opened 2019-03-07
Hide Shadow Quality settings from HDAsset UI

FrancescoC-unity:
### Purpose of this PR
This caused some confusions in the demo team and to be fair, I see it creating more confusion to users as well. 

The quality for deferred is hardcoded in the shader to Low, so exposing a varying option is misleading.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -132,6 +132,7 @@ public struct HDShadowInitParameters
             maxShadowRequests = k_DefaultMaxShadowRequests,
             shadowMapsDepthBits = k_DefaultShadowMapDepthBits,
             useDynamicViewportRescale = true,
+            shadowQuality = HDShadowQuality.Low,
```
JulienIgnace-Unity: How does this work if user change from deferred to forward, change quality and switch back? Does the old value still linger?

FrancescoC-unity: It actually does (as it is currently fwiw), let me force it at UI stage to Low

PR #3179, opened 2019-03-07
Per object occlusion probes

egomeh:
### Purpose of this PR

Add support for per-object occlusion probes

---
### Release Notes

Add support for occlusion probes.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=per-object-occlusion-probes&automation-tools_branch=master&unity_branch=trunk
A test scene for mixed lighting is added. The reference image for the added scene is added for Windows, but not for OS X. The OS X test will then for the first test fail.

- [ ] Built a player
    Fail, but due to file not changes in these changes.
   No new UI.
   No new UI.
- Other: 

**Automated Tests**: Add mixed lighting test scene to build settings

---
### Overall Product Risks
**Technical Risk**: Medium - changes lighting code.

**Halo Effect**: Low - Changes mostly the lighting code and should be isolated to that.

---
### Comments to reviewers


com.unity.render-pipelines.lightweight/ShaderLibrary/Lighting.hlsl
```diff
@@ -108,7 +108,11 @@ Light GetMainLight()
 {
     Light light;
     light.direction = _MainLightPosition.xyz;
-    light.distanceAttenuation = unity_LightData.z;
+    #if defined(_MIXED_LIGHTING_SUBTRACTIVE) && defined(LIGHTMAP_ON)
+        light.distanceAttenuation = unity_LightData.z;
+    #else
+        light.distanceAttenuation = unity_ProbesOcclusion.x;
```
phi-lira: this seems wrong. unity_LightData.z should always be part of distanceAttenuation. (might bad on not adding a comment here, this is used for light layer culling). Also, since now we have baked indirect the mixed_lighting_subtractive don't make sense here. 

Should be something like:
light.distanceAttenuation = unity_LightData.z;
if defined(LIGHTMAP_ON)
    light.distanceAttenuation *= unity_ProbesOcculsion.x;
endif

egomeh: Ah yes, I need to fix that for the additional lights as well.

PR #3195, opened 2019-03-11
Zero out stencil bit after they are not needed anymore

FrancescoC-unity:
### Purpose of this PR
Clean the stencil bits when we don't need them anymore so to facilitate the reusage of them. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP/free-stencil-bits-when-unneeded-anymore&unity_branch=trunk

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Low


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -21,13 +21,8 @@ private enum SMAAStage
...
         {
-            // We don't have bits left to use nor we want to reserve them for SMAA
-            // so we use the same mask as DoesntReceiveSSR as this is likely to not contain
-            // a lot of objects already marked. Worst case scenario, we are running SMAA on more pixel
-            // than needed, but it should still be faster than clearing the stencil.
-            // TODO: We can probably free up some bits while writing others, by leveraging the limited lifetime
-            // of some of the bits stored. But this require a change that is orthogonal to SMAA.
-            SMAABit = HDRenderPipeline.StencilBitMask.DoesntReceiveSSR
+            // DistortionVectors bit is not needed anymore at this stage.
+            SMAABit = HDRenderPipeline.StencilBitMask.DistortionVectors
```
sebastienlagarde: Please, use an explicit name HDRenderPipeline.StencilBitMask.SMAABit  but use the same value as for DistortionVectors instead. It is very error prone to do this here where people editing HDRenderPipeline.StencilBitMask can't see it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/MotionVectors/CameraMotionVectors.shader
```diff
@@ -74,10 +74,11 @@ Shader "Hidden/HDRP/CameraMotionVectors"
             // We will perform camera motion velocity only where there is no object velocity
             Stencil
             {
+                WriteMask 128
                 ReadMask 128
                 Ref  128 // StencilBitMask.ObjectVelocity
                 Comp NotEqual
-                Pass Keep
+                Fail Zero   // We won't need the bit anymore.
```
sebastienlagarde: Let's discuss

FrancescoC-unity:
Rekicking visual tests

FrancescoC-unity:
And new tests are green too

PR #3202, opened 2019-03-12
Fix for motion blur on PS4 by declaring multiple variants for the scattering version

FrancescoC-unity:
### Purpose of this PR
PS4 has problem with this shader on 19.1 since R32 Uint is not supported. This is needed for Megacity. 

[Note, this targets release/2019.1 as I needed to test this, I will port this to hdrp/staging immediately and run tests]

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fscattering-variants-for-mb&automation-tools_branch=master&unity_branch=2019.1%2Fstaging


sebastienlagarde:
KAtana is green, Failure doesn't come from this PR and is fixed in another that have been merged in release/2019.1

PR #3207, opened 2019-03-12
Fix SMAA for XR texture macros and temporarily disable stencil

FrancescoC-unity:
This PR fixes SMAA with XR texture macros but also temporarily disable stencil as I found an issue with it (and don't want to block the merging of staging) 

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fsmaa-xr-fix-and-disable-stencil&automation-tools_branch=master&unity_branch=trunk


FrancescoC-unity:
Note that I am ignoring API that are not the one we select in SMAA. 

Also, I found that I need to clear the buffers to make stencil work, so I'll renable stencil now in this same PR since katana hasn't started yet.

Chman:
Would be nice to have graphics tests for XR :)

sebastienlagarde:
>Would be nice to have graphics tests for XR :)
THis have already been discussed. Can't until update of the framework.

PR #3213, opened 2019-03-12
Vfx/fix/2019.1/floatfield focus

iTris666:
### Purpose of this PR
Fix the problem when we type things like ".5" in a float field. The root problem is how we detect if a floatfield has the focus (to not overrride it if it does). The fix remove a temporary hack previously needed for UIElement and that is no longer working..

---
### Testing status

tests with vfx samples, in the blackboard float field as well as in the node's.

---
### Overall Product Risks

---
### Comments to reviewers


PaulDemeulenaere:
This is the wrong target branch

PR #3229, opened 2019-03-13
Add more variants for static only lighting and skip build light list when not needed. 

FrancescoC-unity:
### Purpose of this PR
This PR introduces two variants for cases with only static lighting, one with Lit only and one with Lit and SSS+Transmission. 

Also, it skips the build light list if there is nothing that needs culling (i.e. we only have directional lights). 

It would be be probably interesting to investigate a faster path for only very few lights. 

---
### Release Notes
Optimize rendering when only static lighting is available or when only directional lights are present. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fstatic-ligthing-lit-variants&unity_branch=trunk [Running] 

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**:  Medium

**Halo Effect**: Medium


sebastienlagarde:
Added @kaychang-unity for awareness of this optimization that can help switch

PR #3248, opened 2019-03-15
Enable contact shadow for multiple lights [2019.2]

alelievr:
### Purpose of this PR
+ Activate contact shadows for multiple lights instead of the dominant one.
+ Added a fullscreen debug mode to display the fade factor of contact shadows
+ Added a slider on the contact shadow debug
This PR needs 2019.2 binary as the R32 UAV used for contact shadows is not available on all platforms on 2019.1

### Performances:
Fontainebleau demo (1 directional light casting contact shadows)

Default contact shadow algorithm (master):

Pass name | VGPR | Occupancy | Time (µs) 
-- | -- | -- | --
Screen Space Shadow | 24 | 70 | 372 
Full frame |   |   | 22844|   

Ccontact shadow FPTL + Scalarized (new):

Pass name | VGPR | Occupancy | Time (µs)
-- | -- | -- | --
Screen Space Shadow | 32 | 70 | 461
Full frame |   |   | 24370

ScreenSpaceShadow.compute 23% slower Mostly due to 8 more VGPR used in the new alogrithm

Diff stats:
[ScreenSpaceShadow.compute](https://docs.google.com/spreadsheets/d/1Ehni9F9zOVpL7MFLgdnsUTUEp1D1HwYss4u--VRLltA/edit?usp=sharing)
[Deferred.compute](https://docs.google.com/spreadsheets/d/1ROgwWk-10_9XX1kQ8StbQ_2ZJScl0nX0G2B3UgbbUIU/edit?usp=sharing)

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FEnableContactShadowForAllLights&automation-tools_branch=master&unity_branch=trunk)

- [x] Built a player
- Other: 
 Tested with a simple scene of 2 spots and 1 directional light with contact shadows enabled containing objects with standard and translucent surfaces.

**Automated Tests**:

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Medium


EvgeniiG:
Variant 0 of the deferred shader does not have any analytical lights.

FrancescoC-unity:
To expand on Evgenii point, could you run the shader perf tool? I am curious to see what impact we have across variants.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/ShaderVariablesLightLoop.cs.hlsl
```diff
@@ -40,4 +40,6 @@
     uint _CookieSizePOT;
     int _RaytracedAreaShadow;
     int _RaytracedIndirectDiffuse;
+
```
FrancescoC-unity: This file is not used, can revert

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
```diff
@@ -10,15 +10,21 @@
...
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/NormalBuffer.hlsl"
 
+#define USE_FPTL_LIGHTLIST // Use light tiles for contact shadows
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/Lighting.hlsl"
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl"
+
 #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
 
-//#pragma enable_d3d11_debug_symbols
+#define SCALARIZE_LIGHT_LOOP (defined(SUPPORTS_WAVE_INTRINSICS) && !defined(LIGHTLOOP_DISABLE_TILE_AND_CLUSTER) && SHADERPASS == SHADERPASS_FORWARD)
```
FrancescoC-unity: Since this is now copy-pasted in two different spots, I'd move this to LightLoopDef.hlsl

alelievr:
Yes, I will run it also to see why i don't have the same amount of VGPR on the variant 0, it should not change anything because we don't use them on variant 0.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
```diff
@@ -10,15 +10,21 @@
...
+
+// #pragma enable_d3d11_debug_symbols
 
 #ifdef ENABLE_MSAA
     TEXTURE2D_X(_CameraDepthValues);
 #endif
 
-RW_TEXTURE2D_X(float, _DeferredShadowTextureUAV);
+RW_TEXTURE2D_X(int, _DeferredShadowTextureUAV);
```
FrancescoC-unity: uint

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
```diff
@@ -170,15 +181,105 @@ void DEFERRED_CONTACT_SHADOW_GENERIC(uint2 groupThreadId : SV_GroupThreadID, uin
...
+            float3 direction = normalize(light.positionRWS.xyz - posInput.positionWS);
+            
+            // TODO simplify with light loop and load (0 must mean no contact shadow)
+            // We store the inverse of the contact shadow:
+            bool occluded = ComputeContactShadow(posInput, direction, globalFade);
+
+            // light.contactShadowMask contains one bit at the position of the contact shadow index that will
+            // be tested in the lightloop, so it insert 1 at the index of the contact shadow if there is a contact shadow
+            // we take full bits at one multiplied by contact shadow and filter the bit at the contact shadow index.
+            contactShadowMask |= (~0 * occluded) & light.contactShadowMask;
```
FrancescoC-unity: Am I missing something or this could simply be light.contactShadowMask * occluded?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
alelievr: Yes, we also now have the scalarized `GetCountAndStart` and `FetchLight` that is duplicated, maybe having a function like ScalarizedGetCountAndStart or something like that in LightLoopDef  would be good

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
alelievr: Indeed, fixed

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
alelievr: fixed

sebastienlagarde:
ping @FrancescoC-unity for validatoin. Also need to merge with master

EvgeniiG:
Why is the full frame 1.5 ms slower? Wrong timings?

alelievr:
I don't know, some stuff took more time like the color gaussian mip and the opaque atmospheric scattering, I can send you the captures if you want

EvgeniiG:
> I don't know, some stuff took more time like the color gaussian mip and the opaque atmospheric scattering, I can send you the captures if you want

I will let @FrancescoC-unity handle it. ;-)

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
FrancescoC-unity: I think we should do it then 🙂

FrancescoC-unity:
Also, have you profiled what happens with multiple lights (especially point/spot that go through the light list) casting contact shadows? To see how it scales

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
FrancescoC-unity: This doesn't look fixed?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
```diff
@@ -111,12 +111,12 @@ public struct LightData
...
 #if ENABLE_RAYTRACING
-        // We store the ray traced area shadow index as a negative value inside the contactShadowIndex.
+        // We store the ray traced area shadow index as a negative value inside the contactShadowMask.
         // Contact shadows are disabled for area lights and setting the index as negative allows for still
         // disabling contact shadows in the shader code (checks for => 0)
-        public int     rayTracedAreaShadowIndex { get => -contactShadowIndex; set => contactShadowIndex = -value; }
+        public int     rayTracedAreaShadowIndex { get => -contactShadowMask; set => contactShadowMask = -value; }
 #endif
-        public int contactShadowIndex;      // negative if unused (TODO: 16 bit)
+        public int     contactShadowMask;      // negative if unused (TODO: 16 bit)
```
FrancescoC-unity: Comment is now obsolete?

FrancescoC-unity: Also, makes me think, do we still have a reason why this is not a uint?

FrancescoC-unity: Also, we should probably move it to uint and use something else to set rayTracedAreaShadowIndex (or use the same, but use a different value to validate that the rayTracedAreaShadowIndex  contains an actual value

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
```diff
@@ -69,7 +69,7 @@ public struct DirectionalLightData
         public int     shadowIndex;             // -1 if unused (TODO: 16 bit)
 
         public Vector3 color;
-        public int     contactShadowIndex;      // -1 if unused (TODO: 16 bit)
+        public int     contactShadowMask;      // 0 if unused (TODO: 16 bit)
```
FrancescoC-unity: Same comment as below

FrancescoC-unity:
Added few comments here and there :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
FrancescoC-unity: Since you can't test raytracing stuff, I'll probably do this change myself at some point

FrancescoC-unity:
@EvgeniiG  The rest of the disparity was due to the fact that one capture uses 11-11-10, the other uses 16161616 as scene color format, so nothing to worry about there

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
```diff
@@ -53,7 +59,7 @@ float SampleDepth(float2 UV, bool HalfRes)
 }
 
 
-float ScreenSpaceShadowRayCast(float3 positionWS, float3 rayDirWS, float rayLength, uint2 positionSS)
+bool ScreenSpaceShadowRayCast(float3 positionWS, float3 rayDirWS, float rayLength, uint2 positionSS, out float fade)
```
sebastienlagarde: Why return bool here whereas it is used as float after ?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -1618,8 +1618,8 @@ DirectLighting EvaluateBSDF_Rect(   LightLoopContext lightLoopContext,
 
 #if SUPPORTS_RAYTRACED_AREA_SHADOWS
     // We are using the contact shadow index for area light shadow index in case of ray tracing.
```
sebastienlagarde: udpate the comment about contact shadow index

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -419,7 +419,7 @@ void InitializeRenderTextures()
 
             // TODO: For MSAA, we'll need to add a Draw path in order to support MSAA properly
             // Use RG16 as we only have one deferred directional and one screen space shadow light currently
-            m_ScreenSpaceShadowsBuffer = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R16_UNorm, enableRandomWrite: true, xrInstancing: true, useDynamicScale: true, name: "ScreenSpaceShadowsBuffer");
+            m_ScreenSpaceShadowsBuffer = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R32_UInt, enableRandomWrite: true, xrInstancing: true, useDynamicScale: true, name: "ScreenSpaceShadowsBuffer");
```
sebastienlagarde: Warning, I guess R32_UInt is only supported from 2019.2 right ? So we can't backport this PR to 2019.1. or we need to implement a version with RGBA8 for 2019.1

sebastienlagarde:
Adding @JulienIgnace-Unity for awareness for rendergraph change

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/ScreenSpaceShadow.compute
alelievr: it's to separate the occluded and fading part, before we were only returning the fading * occluded, but now we need both information so we can store it in the target

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -280,22 +261,20 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
-        fastPath = WaveActiveAllTrue(envLightStart == envStartFirstLane); 
-    #endif
-
 #else   // LIGHTLOOP_DISABLE_TILE_AND_CLUSTER
         envLightCount = _EnvLightCount;
         envLightStart = 0;
 #endif
 
+        bool fastPath = false;
+    #if SCALARIZE_LIGHT_LOOP
```
FrancescoC-unity: This inconsistent with how you dealt with this above (around line 157). I prefer this better (with the #if). 

Also, makes me see a bug in the original implementation, if fastPath is true, envLightStart needs to be set to envStartFirstLane

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -257,6 +258,46 @@ uint FetchIndex(uint lightStart, uint lightOffset)
...
+    // Fast path is when we all pixels in a wave are accessing same tile or cluster.
+    lightStartLane0 = WaveReadLaneFirst(lightStart);
+    return WaveActiveAllTrue(lightStart == lightStartLane0); 
+#else
+    lightStartLane0 = lightStart;
+    return false;
+#endif
+}
+
+bool IsFastPath(uint lightStart)
```
FrancescoC-unity: What this is used for?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
```diff
@@ -257,6 +258,46 @@ uint FetchIndex(uint lightStart, uint lightOffset)
...
+{
+    uint unusedLightStartLane0;
+    return IsFastPath(lightStart, unusedLightStartLane0);
+}
+
+uint ScalarizeLightIndex(uint v_lightIdx, bool fastPath)
```
FrancescoC-unity: I think at this point I wouldn't call this Light related. We use a similar loop for decal (I can update it afterward), so I'd call this something like ScalarizeElementIndex. 

In a sense, I am a bit conflicted by this. This alone would indeed scalarize the index, but won't make it proper in itself outside of  how the loop is structured. I would add a comment on top of the function specifying that this only work in the context of the lightloop scalarization strategy.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
alelievr: It was a helper function i used in the lightloop, not used anymore, i'll remove it

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
alelievr: Indeed, i'll change this using the #if version.
Regarding the bug you mentioned, envLightStart is set to envStartFirstLane but after we fetch the sky for the SSR, i though this was desired but if it's not i can move the assignation just after the IsFastPath

FrancescoC-unity: Oh no that's fine then, I was only looking at the diff and missed the bit

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl
alelievr: I renamed the function to `ScalarizeElementIndex` and added a comment on which context this function can be useful

PR #3250, opened 2019-03-15
[WIP] Refactor the shading framework

EvgeniiG:
### Purpose of this PR

Refactor the shading framework to make it more easy to support arbitrary materials. Most of the non-material-specific code is now shared. The new CBxDF function which replaces the BSDF function is more flexible and allows to fully define the behavior of up to 4 types of lobes (more lobes can be grouped by type). The resulting code is efficient (automatic early-outs), simple and flexible.

---
### Release Notes
Refactor the shading framework to make it more easy to support arbitrary materials.

---
### Testing status
**Katana Tests**: WIP
**Manual Tests**: WIP

What did you do?
- [ ] Built a player
- Other: 

**Automated Tests**: None.

Any test projects to go with this to help reviewers? No.

---
### Overall Product Risks
**Technical Risk**: High.

**Halo Effect**: High.

---
### Comments to reviewers
Has a dependency on my hair PR.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -348,10 +331,10 @@ void EvaluateLight_Punctual(LightLoopContext lightLoopContext, PositionInputs po
 
 #ifdef DEBUG_DISPLAY
     if (_DebugShadowMapMode == SHADOWMAPDEBUGMODE_SINGLE_SHADOW && light.shadowIndex == _DebugSingleShadowIndex)
-        debugShadowAttenuation = step(FLT_EPS, attenuation) * shadow;
+        debugShadowAttenuation = shadow;
```
sebastienlagarde: isn't this suppose to be step(FLT_EPS, shadow) instead ? (poking @alelievr )

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,102 @@
...
+                                          float diffuseDimmer, float specularDimmer)
 {
-    float3 transmittance = 0.0;
+    DirectLighting lighting;
+    ZERO_INITIALIZE(DirectLighting, lighting);
 
-#ifdef MATERIAL_INCLUDE_TRANSMISSION
-    if (MaterialSupportsTransmission(bsdfData))
+#ifndef DEBUG_DISPLAY
+    if (Max3(lightColor.r, lightColor.g, lightColor.b) > 0)
```
sebastienlagarde: We don't need to test for this. on the C# side we don't sent any light that is black. (or if we do, then we shouldn't :) )

sebastienlagarde: This is the test in lightloop.cs

            bool contributesToLighting = ((additionalLightData.lightDimmer > 0) && (additionalLightData.affectDiffuse || additionalLightData.affectSpecular)) || (additionalLightData.volumetricDimmer > 0);

Note that we still need to test for lightdimmer > 0 because we haven't done the optimization of having two separate light list for volumetric and regular.

sebastienlagarde: So looks like we don't test color, so we should add it in C# for safety, but IIRC this is done on C++ side alsready. Worse to check.

Otherwise we do this test: This is the test in lightloop.cs

            bool contributesToLighting = ((additionalLightData.lightDimmer > 0) && (additionalLightData.affectDiffuse || additionalLightData.affectSpecular)) || (additionalLightData.volumetricDimmer > 0);

so we still need to test for lightdimmer > 0 because we haven't done the optimization of having two separate light list for volumetric and regular.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -138,147 +105,92 @@ DirectLighting ShadeSurface_Directional(LightLoopContext lightLoopContext,
...
-
-                // Avoid double shadowing. TODO: is there a faster option?
-                light.shadowIndex = -1;
-
-                // Note: we do not modify the distance to the light, or the light angle for the back face.
-                // This is a performance-saving optimization which makes sense as long as the thickness is small.
-            }
-        }
-    }
+    /*********************
```
sebastienlagarde: Don't use /*******
use //-------------
:)

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -165,60 +165,79 @@ VoxelLighting EvaluateVoxelLightingDirectional(LightLoopContext context, uint fe
...
-        // Pass the light direction as the normal to avoid issues with shadow biasing code.
-        context.shadowValue = EvaluateRuntimeSunShadow(context, posInput, light, -light.forward);
-    }
-    else
-    {
-        context.shadowValue = 1;
+        float3 L = -light.forward;
+
+        // Is it worth sampling the shadow map?
+        if ((light.volumetricLightDimmer > 0) && (light.shadowDimmer > 0))
```
sebastienlagarde: >if ((light.volumetricLightDimmer > 0) && (light.shadowDimmer > 0))
Maybe use  (light.volumetricLightDimmer * light.shadowDimmer > 0) instead. Should be faster

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -165,60 +165,79 @@ VoxelLighting EvaluateVoxelLightingDirectional(LightLoopContext context, uint fe
...
-        float phase    = CornetteShanksPhasePartVarying(anisotropy, cosTheta);
-
-        // Note: the 'weight' accounts for transmittance from 't0' to 't'.
-        float intensity = light.volumetricLightDimmer * attenuation * weight;
-
-        // Compute the amount of in-scattered radiance.
-        lighting.radianceNoPhase  += intensity * color;
-        lighting.radianceComplete += phase * intensity * color;
+        // Is it worth evaluating the light?
+        if ((light.volumetricLightDimmer > 0))
```
sebastienlagarde: double ((

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -298,21 +317,23 @@ VoxelLighting EvaluateVoxelLightingLocal(LightLoopContext context, uint featureF
...
-                float  dist      = distSq * distRcp;
-                float  distProj  = dot(lightToSample, light.forward);
-                float4 distances = float4(dist, distSq, distRcp, distProj);
-                float3 L         = -lightToSample * distRcp;
+                float3 L;
+                float4 distances; // {d, d^2, 1/d, d_proj}
+                GetPunctualLightVectors(posInput.positionWS, light, L, distances);
 
-                ModifyDistancesForFillLighting(distances, light.size.x);
+                float4 lightColor = EvaluateLight_Punctual(context, posInput, light, L, distances);
```
sebastienlagarde: no more ModifyDistancesForFillLighting ?

com.unity.render-pipelines.high-definition/Runtime/Material/AxF/AxF.cs
```diff
@@ -103,6 +103,10 @@ public struct BSDFData
 
             [SurfaceDataAttributes(new string[] { "Geometric Normal", "Geometric Normal View Space" }, true)]
             public Vector3 geomNormalWS;
+
+            public uint    materialFeatures;
+            public Vector3 transmittance;
+            public float   ambientOcclusion;
```
sebastienlagarde: Why adding transmittance and AO ? They don't exist with AxF material

sebastienlagarde: ok It is because of the factorisation of the code of MicroShadowing... maybe for AO it is ok, but forcing having transmisstance term on all material isn't great. Wondering if we have other options. Users will be allowed to write material and it is painful if we need to add stuff like that.

sebastienlagarde: (even if it is only on BRDFData side)

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
```diff
@@ -400,17 +404,29 @@ void BSDF(  float3 V, float3 L, float NdotL, float3 positionWS, PreLightData pre
...
         float3 F = F_Schlick(bsdfData.fresnel0, LdotH);
 
-        specularLighting = F * DV;
+        specTerm = F * DV;
+        diffTerm = DisneyDiffuse(abs(NdotV), abs(NdotL), LdotV, bsdfData.perceptualRoughness);
+    }
+
+    // The compiler should optimize these. Can revisit later if necessary.
+    cbxdf.diffR = diffTerm * clampedNdotL;
+    cbxdf.diffT = diffTerm * flippedNdotL;
```
sebastienlagarde: ok, so normally based on the fact hat transmittance is 0, we expect compiler to not compute flippedNdotL. It should be fine if we have a clear static testing for the init of transmittance

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
```diff
@@ -400,17 +404,29 @@ void BSDF(  float3 V, float3 L, float NdotL, float3 positionWS, PreLightData pre
...
 
-        // Note: diffuseLighting is multiply by color in PostEvaluateBSDF
-        diffuseLighting = DisneyDiffuse(NdotV, NdotL, LdotV, bsdfData.perceptualRoughness);
+    // Probably worth branching here for perf reasons.
+    // This branch will be optimized away if there's no transmission.
```
sebastienlagarde: you mean if there is transmission, instead of 'no transmission" ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -138,147 +105,92 @@ DirectLighting ShadeSurface_Directional(LightLoopContext lightLoopContext,
...
-
-        if (surfaceReflection)
+        if (ShouldEvaluateThickObjectTransmission(V, L, preLightData, bsdfData, light.shadowIndex))
         {
-            float intensity = attenuation * NdotL;
-
-            lighting.diffuse  = diffuseBsdf  * (intensity * light.diffuseDimmer);
-            lighting.specular = specularBsdf * (intensity * light.specularDimmer);
+            // Replace the 'baked' value using 'thickness from shadow'.
+            bsdfData.transmittance = EvaluateTransmittance_Punctual(lightLoopContext, posInput,
```
sebastienlagarde: ok, so it is because we access bsdfData.transmittance here that we must put transmittance in all our material... 
Don't we have a better way of doing this ? ideally we shoudn't access to any bsdfData properties in shared code as it force material properties. And when you author your own material you don't want to be force to add properties.

sebastienlagarde: For access to materialFeatures I think it is acceptable, but for the other, I am not fan of it.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -693,12 +672,19 @@ uint DecodeFromGBuffer(uint2 positionSS, uint tileFeatureFlags, out BSDFData bsd
 
     // In deferred ambient occlusion isn't available and is already apply on bakeDiffuseLighting for the GI part.
     // Caution: even if we store it in the GBuffer we need to apply it on GI and not on emissive color, so AO must be 1.0 in deferred
+    /**************************************************************************
+     * TODO: WTF? SEBASTIEN? These are terrible hacks.                        *
```
sebastienlagarde: I don't understand what you refer too. Can you clarify. Is it the bsdfData.ambientOcclusion = 1.0; ?

>And what about the extra G-Buffer for light layers?  
Again, I don't understand. What is the issue with the extra GBuffer for Lightlayer ?

>We must set this variable, and it should work the same way 
Which variable ? ambientOcclusion ? We don't have the information, so we use 1.0. This is descirbe in the wiki about difference between forward and deferred.
remember, we aim at feature parity, not feature quality parity
https://github.com/Unity-Technologies/ScriptableRenderPipeline/wiki/Forward-And-Deferred-Rendering

Also this is why we must keep the ComputeMicroShadowing() function

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,102 @@
...
-            // in the context of transmission, but we keep it unaltered for performance reasons.
-            lighting.diffuse  = transmittance * (diffuseBsdf * (intensity * light.diffuseDimmer));
-            lighting.specular = 0; // No spec trans, the compiler should optimize
+            // TODO: the shadow code should do it for us. That would be far more efficient.
+            float3 sN  = GetNormalForShadowBias(bsdfData);
+                   sN *= FastSign(dot(sN, L));
+
+            // This code works for both surface reflection and thin object transmission.
+            lightColor.a *= EvaluateShadow_Directional(lightLoopContext, posInput, light, builtinData, sN);
+            lightColor.a *= ComputeMicroShadowing(bsdfData.ambientOcclusion,
```
sebastienlagarde: you must not do that.  ComputeMicroShadowing(bsdfData, NdotL) is not the same in Lit.hlsl
I see later that currently you complain about ambient occlusion setup in decodeFromGBuffer, this is related to the same comment. ComputeMicroShadowing must be handled by the material and it avoid to force all material to add bsdfData.ambientOcclusion

sebastienlagarde: Here we shouldn't call ComputeMicroShadowing(bsdfData.ambientOcclusion, abs(dot(bsdfData.normalWS, L)), but ComputeMicroShadowing(bsdfData, NdotL);

EvgeniiG:
@sebastienlagarde Please open and read LightEvaluation.hlsl and SurfaceShading.hlsl instead of just looking at diffs. Otherwise my changes won't make sense to you.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: Well, if you ask me, transmittance should not be a part of the bsdfData. It is for us because of the baked thickness hack which we employ. I can pass a separate variable here instead of overwriting it in place, if it helps.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: Note that the design of this features lightlayers is like this:
We expect users to use light layers for cinematics in game, so when they use cinematic they switch to 5 GBuffer, when they are in game they use 4 GB and no light layers to keep high framerate. The extra buffer can be dynamically switch off based on frame settings, so you can enable it only on cinematic camera for example.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: So this means that we have do copy-paste every function which uses AO for every material? That's pretty bad, especially considering I can't think of a material which shouldn't support AO. If you were to write a new material, think what's simpler, figure out a way to compute AO (or pass along a baked value), or... rewrite all these confusing functions without docs?

sebastienlagarde:
Hey, this is going the right way, the new code is much cleaner, good!
I still have concern about forcing naming on parameters in BRDF data + some comment here and there, but it is nice.

Poking @slunity about change coming with this PR as he will be impacted. Stephane, please have a look and comment. Thansk

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: I think we should define a base set of properties every material should have. AO can be one of them. It's not hard to set it to 1 if you absolutely must kill it.

sebastienlagarde: I don't get your point. We have an example of material that don't use AO. (AxF). And for the other, the only thing that need to be implement is ComputeMicroShadowing function.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
EvgeniiG: This looks like a hack so I removed it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: Maybe I should rename this variable, it's not what you think it is. I can call it incomingRadiance...

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
EvgeniiG: The compiler will use a max(), i prefer this option for clarity.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
EvgeniiG: I am referring to this code:

image

EvgeniiG: We should just properly populate the AO variable instead of resorting to these hacks.

EvgeniiG: Then we can use the same set of functions everywhere.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
EvgeniiG: I mean what I say. :-)

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
EvgeniiG: We will need to test codegen. This is just a refactor for the ease of maintenance.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -88,34 +97,21 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
-            if (evaluateShadows)
+            // Is it worth sampling the shadow map?
+            if ((light.lightDimmer > 0) && (light.shadowDimmer > 0) &&
+                IsNonZeroCBxDF(V, L, preLightData, bsdfData) &&
+                !ShouldEvaluateThickObjectTransmission(V, L, preLightData, bsdfData, light.shadowIndex))
             {
-                context.shadowValue = EvaluateRuntimeSunShadow(context, posInput, light, shadowBiasNormal);
+                // TODO: the shadow code should do it for us. That would be far more efficient.
+                float3 sN  = GetNormalForShadowBias(bsdfData);
+                       sN *= FastSign(dot(sN, L));
```
slunity: @EvgeniiG  I caught that too as I'm in the middle of trying some stuff to improve dual normal maps handling in StackLit, the old "shadowBiasNormal *= FastSign(NdotL);" could leak in forward right ?

EvgeniiG: I am mostly doing this for transmission through thin objects since we evaluate lighting for the back of the object. Back-facing normals at grazing angles should gradually become self-shadowed (with a decent shadow bias impl). And of course we should use the geometric normal for shadow bias (although the term "geometric" is a bit loose here since this is going to be a Phong-interpolated vertex normal).

EvgeniiG: But yes, the old version with the wrong NdotL could potentially behave poorly. I now use the new version for all lights.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
sebastienlagarde: Wasn't a hack. I guess it is to display pure white or pure black on the screen. But waiting for Antoine reply to elaborate.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: ah I see:

lightColor.rgb *= lightColor.a; // Composite

Yes, instead the name is misleading, it shouldn't be name lightColor as it is not.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
sebastienlagarde: so you mean the compiler convert if ((light.volumetricLightDimmer > 0) && (light.shadowDimmer > 0)) to max(light.volumetricLightDimmer, light.shadowDimmer) > 0 ? 
Currently using max(light.volumetricLightDimmer, light.shadowDimmer) > 0 is nice and readable.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
EvgeniiG: Ah, OK, I can add it back.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
sebastienlagarde: So I am not sure I understand, we talk about this branch

  if (NdotL > 0)
    {
        cbxdf.specR = specTerm * clampedNdotL;
    }	    }

So if there is transmission (so if there is only regular shading), we perfom evaluation of reflection, otherwise we don't ? (sorry not clear what is the code path with specR  and transmisison)

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: ah yes, I remember we discussed about transmistance in bsdfData. Yes, I think using a variable would make more sense here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
EvgeniiG: The compiler will generate a max(). But like I said, we will need to test codegen when everything is working. Most likely issue here is that since these are all floats they will go through the vector unit which is less efficient as we want to stay everything in scalar registers. We can revisit this later.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
EvgeniiG: If there is no transmission, we will not get this far because IsNonZeroCBxDF() already performs the NdotL test. Please see the full file for details.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: >We should just properly populate the AO variable instead of resorting to these hacks
The problem is that we can't.

In deferred we don't have AO, so can't rely on it, we pre-apply it on GI during Guffer pass.
In the past I was storing AO in the extra GBuffer use for light layer. But as AO is still apply on GI in deffered and must not affect emissive it was wrong. So I have remove it from the extra GBuffer.

But now I see that I have forget to update the ComputeMicroShadowing code.
Still the code should be:


float ComputeMicroShadowing(BSDFData bsdfData, float NdotL)
{
    float sourceAO;
#if (SHADERPASS == SHADERPASS_DEFERRED_LIGHTING)
    sourceAO = bsdfData.specularOcclusion;
#else
    sourceAO = bsdfData.ambientOcclusion;
#endif

    return ComputeMicroShadowing(sourceAO, NdotL, _MicroShadowOpacity);
}

which doesn't solve the current thing you try to do (i.E sharing code of ComputeMicroShadowing).

I don't think adding ComputeMicroShadowing function in all material is a big deal (but however I am agree that now we have a lot of small "helper" like this to implement for material (like the one for PBR validation) and this is annoying. But don't think forcing naming on properties is better.

EvgeniiG: We can always make an exception for one function, but it just signifies a design problem. And if someone wants to extend HDRP, it will probably blow their mind that the variable with AO doesn't actually store AO. It's also pretty bad that forward and deferred will have different visuals depending on what you use for spec occlusion. I really feel like there is a problem here and we are trying to sweep it under the rug.

EvgeniiG: It's probably worth taking a pause to think if there's a better solution instead of just hacking away at it.

sebastienlagarde: > but it just signifies a design problem. 
The problem is that we don't have AO in deferred.
But you have AO in Forward, thus why we have AO in BSDFData and we use AO correctly when we are in forward (no double darkening, no apply to emissive color). There is no h>ack, there is a missing information in deferred case.

>It's also pretty bad that forward and deferred will have different visuals depending on what you use for spec occlusion

Disagree, it will be bad to force forward to not do thing correctly whereas it is able to do it.
So I am good with trying to have the best visual depends on the mode.
The only issue we have is deferred and because we can't store AO. And it only affect the micro shadow function + we have double darkerning in SSAO.

>I really feel like there is a problem here and we are trying to sweep it under the rug.
Yes, the problem is that we don't have AO :), because we have no space in GBuffer.

>It's probably worth taking a pause to think if there's a better solution instead of just hacking away at it.

I am open, it is not a quick hack, it is something that have been revisited several time without any good solution. Current one is my best default. But I am open to proposition.

EvgeniiG: Can't we compute AO from spec occlusion?

EvgeniiG: Or how about modifying the micro shadowing function to work with spec occlusion.

sebastienlagarde: >Or how about modifying the micro shadowing function to work with spec occlusion.
That would be something interesting to try, but don't know if it make sense, spec occ is view dependent. But even if we success, it doesn't solve the issue. I don't want to have to put specular occlusion on all the material. That the purpose of local material MicroShadowing function. It allow to not handle it if your material can't support it.

Going from AO to SO is okish, but the reverse will be bad.

My own guess is that it is the MicroShadowing itself that is a bad thing :). Was hoping that contact shadow will solve the issue. Sadly artists tend to love MicroShadow and still enable both micro shadow + contact shadow :( - give them better result...

EvgeniiG: Also loading spec occlusion for micro shadowing ruins our deferred perf. Seems like a bad technique all around.

slunity:
> Hey, this is going the right way, the new code is much cleaner, good!
> I still have concern about forcing naming on parameters in BRDF data + some comment here and there, but it is nice.
> Poking @slunity about change coming with this PR as he will be impacted. Stephane, please have a look and comment. Thansk

Not completely done reading, but very good so far, exactly what I was discussing a few weeks ago with you. I like the pushing of the foreshortening cosine factor in the BxDF evaluation, the removal of the separate nondescript NdotL to early out in light evaluations, better to have IsNonZeroCBxDF, proper handling of shadowbias and sign for all light types and the splitting of light eval / shadow eval.

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
slunity: > If there is no transmission, we will not get this far because IsNonZeroCBxDF() already performs the NdotL test. Please see the full file for details.

In other words the branch is really there for when there's possible transmission at compile time so that the specTerm calculation (FVD) is moved inside the if (NdotL > 0)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
alelievr: Indeed, when we debug shadows we don't want attenuation because it makes it very hard to see at the limit of the light.
The step with the attenuation is required to avoid seeing the screen tiles at the end of lights because the attenuation always falls to 0 before the tile ends.

EvgeniiG:
Any progress with Stacklit and Simplelit?

slunity:
> Any progress with Stacklit and Simplelit?

Yes, I just got back to this, will drop something for StackLit next week.

sebastienlagarde:
close, use [WIP] Refactor the shading framework (bis) #3703 instead

PR #3260, opened 2019-03-19
Reworked resource management [2019.2]

Chman:
### Purpose of this PR

Resource management in LWRP is very limited. Right now it's in an asset and it relies on the default script reference system in the editor to auto-populate the asset on creation. This works okay in basic cases but it doesn't support much (array support for instance). It's also currently unsafe, as in if the asset breaks (serialization bug, failed package upgrade and so on) everything breaks.

HDRP recently moved to a safer, easier to use resource management system. This PR moves the interesting code bits to Core and changed resource handling in LWRP to use this new system. See the following files to get a grasp on how to use the new set of attributes:
- `com.unity.render-pipelines.lightweight/Runtime/ForwardRendererData.cs` (simple case)
- `com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs` (more complex cases)

---
### Testing status
**Katana Tests**: [Green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=resources-refactor&automation-tools_branch=master&unity_branch=trunk).

**Manual Tests**:
- Explicitly broke the ForwardRendererData and the editor resource asset by setting some of the assets to null and reloaded the pipeline.
- In developer mode, auto-populated the assets by clicking Reload All.
- Manually checked every reference & associated paths in the assets.

---
### Overall Product Risks
**Technical Risk**: High

**Halo Effect**: High


com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
```diff
@@ -1,21 +1,21 @@
...
     /// <summary>
     /// The resources that need to be reloaded in Editor can live in Runtime.
     /// The reload call should only be done in Editor context though but it
     /// could be called from runtime entities.
     /// </summary>
-    internal static class ResourceReloader
+    public static class ResourceReloader
     {
-        public static void ReloadAllNullIn(System.Object container)
+        public static void ReloadAllNullIn(System.Object container, string basePath)
```
RSlysz: As the method now is in User Land (public), add in-code documentation

com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
```diff
@@ -146,67 +146,52 @@ static void SetAndLoadIfNull(Array array, int index, string path, bool builtin)
...
     {
-        public enum Package { Builtin, CoreRuntime, CoreEditor, HDRPRuntime, HDRPEditor };
+        public enum Package { Builtin, Root };
 #if UNITY_EDITOR
         public readonly Package package;
         public readonly string[] pathes;
 #endif
 
-        public ReloadAttribute(string[] pathes, Package package = Package.HDRPRuntime)
+        public ReloadAttribute(string[] pathes, Package package = Package.Root)
```
RSlysz: As the method now is in User Land (public), add in-code documentation

com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
```diff
@@ -146,67 +146,52 @@ static void SetAndLoadIfNull(Array array, int index, string path, bool builtin)
...
         {
 #if UNITY_EDITOR
             this.pathes = pathes;
             this.package = package;
 #endif
         }
 
-        public ReloadAttribute(string path, Package package = Package.HDRPRuntime)
+        public ReloadAttribute(string path, Package package = Package.Root)
```
RSlysz: As the method now is in User Land (public), add in-code documentation

com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
```diff
@@ -146,67 +146,52 @@ static void SetAndLoadIfNull(Array array, int index, string path, bool builtin)
...
             : this(new string[] { path }, package)
         { }
 
         public ReloadAttribute(string pathFormat, int rangeMin, int rangeMax,
```
RSlysz: As the method now is in User Land (public), add in-code documentation

com.unity.render-pipelines.lightweight/Runtime/ForwardRendererData.cs
```diff
@@ -1,56 +1,44 @@
...
-        {
-            get => m_ScreenSpaceShadowShader;
-        }
+        [SerializeField] StencilStateData m_DefaultStencilState = null;
 
-        internal Shader samplingShader
+        protected override void OnEnable()
         {
-            get => m_SamplingShader;
+            ResourceReloader.ReloadAllNullIn(this, LightweightRenderPipelineAsset.packagePath);
```
RSlysz: On HDRP side, Reload call are only done in Editor context. Ensure it the case or that the code could run in a build bersion when resources are missing.

sebastienlagarde:
Note: This PR affect both LW and HDRP. Question: should be this backport on 2019.1 ? Or it is ok to only have it for 2019.2

Chman:
We probably don't want to backport this to 2019.1 so close to getting LWRP out of preview. I'd rather have it in 19.2/.3 and well tested, the risk is just too high.

PR #3267, opened 2019-03-20
Add emission, diffusion profile and HD Scene Color nodes doc

alelievr:
Add emission, diffusion profile and HD Scene Color nodes doc
Also moved the HDRP shader graph doc to a separate folder


sebastienlagarde:
Poke @JordanL8  for review

sebastienlagarde:
Merged in order to include into package @JordanL8 (better to have doc with typo than no doc) but please do correct and redo a PR for next package.

PR #3274, opened 2019-03-22
Remove use legacy unity matrix variables

alelievr:
### Purpose of this PR

Remove USE_LEGACY_UNITY_MATRIX_VARIABLES define to use the same matrices for shadow pass than the other passes in HDRP.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FRemoveUSE_LEGACY_UNITY_MATRIX_VARIABLES-&automation-tools_branch=master&unity_branch=trunk)

- [x] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
```diff
@@ -491,15 +491,13 @@ Shader "HDRP/LayeredLit"
...
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
 
     //-------------------------------------------------------------------------------------
     // variable declaration
     //-------------------------------------------------------------------------------------
 
-    // Can't include 'ShaderVariables.hlsl' here because of USE_LEGACY_UNITY_MATRIX_VARIABLES. :-(
-    // #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
-    // Same story for 'Material.hlsl' (above) which includes 'AtmosphericScattering.hlsl' which includes 'ShaderVariables.hlsl'.
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
```
sebastienlagarde: please move it to before FragInputs.hlsl now :) (same for all  files)

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
```diff
@@ -320,15 +320,13 @@ Shader "HDRP/Lit"
...
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
 
     //-------------------------------------------------------------------------------------
     // variable declaration
     //-------------------------------------------------------------------------------------
 
-    // Can't include 'ShaderVariables.hlsl' here because of USE_LEGACY_UNITY_MATRIX_VARIABLES. :-(
-    // Same story for 'Material.hlsl' (above) which includes 'AtmosphericScattering.hlsl' which includes 'ShaderVariables.hlsl'.
-    // #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
+    #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
```
sebastienlagarde: see lit.shader comment

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
```diff
@@ -360,7 +358,6 @@ Shader "HDRP/Lit"
             // We reuse depth prepass for the scene selection, allow to handle alpha correctly as well as tessellation and vertex animation
             #define SHADERPASS SHADERPASS_DEPTH_ONLY
             #define SCENESELECTIONPASS // This will drive the output of the scene selection shader
-            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl"
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
```
sebastienlagarde: you have forget to remove Material.hlsl here

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1732,8 +1732,6 @@ void Callback(CommandBuffer asyncCmd)
                     // This call overwrites camera properties passed to the shader system.
                     m_LightLoop.RenderShadows(renderContext, cmd, cullingResults, hdCamera);
 
-                    // Overwrite camera properties set during the shadow pass with the original camera properties.
-                    SetupCameraProperties(camera, renderContext, cmd);
```
sebastienlagarde: @fabien-unity I poke you about this change. This is not required now for regular rendering, and iirc this was used in shiruken, but for shadow it shoul be fined. However I don't know what's happen for VR and shadow here. I know there is some setup done, but in the context of shadow it should still be fine, could you validate ? thanks

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -1103,7 +1103,6 @@ public static GenerationResults GetShader(this GraphData graph, AbstractMaterial
                 finalShader.AppendNewLine();
 
                 finalShader.AppendLine(@"HLSLINCLUDE");
-                finalShader.AppendLine("#define USE_LEGACY_UNITY_MATRIX_VARIABLES");
```
sebastienlagarde: I am not sure to understand what was the purpose of this define at this location initially, is it for preview ? not sure how this was conflicting.

com.unity.shadergraph/ShaderGraphLibrary/ShaderVariables.hlsl
```diff
@@ -276,11 +276,7 @@ float4x4 OptimizeProjectionMatrix(float4x4 M)
     return M;
 }
 
-#ifdef USE_LEGACY_UNITY_MATRIX_VARIABLES
-    #include "ShaderVariablesMatrixDefsLegacyUnity.hlsl"
-#else
-    #include "ShaderVariablesMatrixDefsHDCamera.hlsl"
-#endif
+#include "ShaderVariablesMatrixDefsLegacyUnity.hlsl"
```
sebastienlagarde: why do we include legacy matrix here ? and not "ShaderVariablesMatrixDefsHDCamera.hlsl" ?

sebastienlagarde:
Removing Peter and Adding Russel for the question about the  usage of USE_LEGACY_UNITY_MATRIX_VARIABLES in the core file of "shader graph"

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
alelievr: As far as i know, this is used to generate node previews (not the main) and the `USE_LEGACY_UNITY_MATRIX_VARIABLES` was forced to include the `ShaderVariablesMatrixDefsLegacyUnity.hlsl` in the [shadergraph ShaderVariables.hlsl](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/65a566d34409edc48f965add7d1483fce3c927b4/com.unity.shadergraph/ShaderGraphLibrary/ShaderVariables.hlsl) because render pipelines does not affect the node previews. It's the same as in the editor GUI where we are forced to use the legacy unity matrices.
I removed this define because i changed the ShaderVariables.hlsl that uses these previews to only include the legacy unity matrices

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
fabien-unity: This should not impact the shadow rendering in VR.
The other call to SetupCameraProperties() is still required until we move to XR SDK.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -237,6 +238,7 @@ CBUFFER_START(UnityGlobal)
     float4 _ScreenParams;
 
     float4 _FrustumPlanes[6];           // { (a, b, c) = N, d = -dot(N, P) } [L, R, T, B, N, F]
+    float4 _ShadowFrustumPlanes[6];     // { (a, b, c) = N, d = -dot(N, P) } [L, R, T, B, N, F]
```
EvgeniiG: Let's not pollute the global CBuffer with stuff you don't need during most passes. We already have plenty of things which should be removed.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -189,6 +189,7 @@ CBUFFER_START(UnityGlobal)
     float4x4 _ProjMatrix;
     float4x4 _InvProjMatrix;
     float4x4 _ViewProjMatrix;
+    float4x4 _CameraViewProjMatrix;
```
EvgeniiG: Why do we need this?

EvgeniiG: Also, looking a few lines above, why do we still have these:

    #if !defined(USING_STEREO_MATRICES)
        float4x4 glstate_matrix_projection;	        float4x4 glstate_matrix_projection;
        float4x4 unity_MatrixV;	        float4x4 unity_MatrixV;
        float4x4 unity_MatrixInvV;	        float4x4 unity_MatrixInvV;
        float4x4 unity_MatrixVP;	        float4x4 unity_MatrixVP;
        float4 unity_StereoScaleOffset;	        float4 unity_StereoScaleOffset;
    #endif	    #endif

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
EvgeniiG: So you should either use _FrustumPlanes or set up your own constant buffer for the shadow pass.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -64,15 +64,16 @@ public enum HDShadowFlag
     public class HDShadowRequest
     {
         public Matrix4x4            view;
-        // Use device projection matrix for shader and projection for CommandBuffer.SetViewProjectionMatrices
+        // Use the y flipped device projection matrix as light projection matrix
+        public Matrix4x4            deviceProjectionYFlip;
```
EvgeniiG: Why do you need to store both?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitDataMeshModification.hlsl
```diff
@@ -111,8 +110,8 @@ float4 GetTessellationFactors(float3 p0, float3 p1, float3 p2, float3 n0, float3
     if (_TessellationFactorTriangleSize > 0.0)
     {
         // return a value between 0 and 1
-        // Warning: '_ViewProjMatrix' is not the same as UNITY_MATRIX_VP for shadow views!
-        edgeTessFactors *= GetScreenSpaceTessFactor( p0, p1, p2, _ViewProjMatrix, _ScreenSize, _TessellationFactorTriangleSize); // Use primary camera view
+        // Warning: '_ViewProjMatrix' can be the viewproj matrix of the light when we render shadows, that's why we use _CameraViewProjMatrix instead
+        edgeTessFactors *= GetScreenSpaceTessFactor( p0, p1, p2, _CameraViewProjMatrix, _ScreenSize, _TessellationFactorTriangleSize); // Use primary camera view
```
EvgeniiG: I am a bit worried this will break with multiple cameras... As we would need to render things in a very specific order. And what about VR where you sort of have two cameras?

JulienIgnace-Unity: This was already the way it worked before (because shadows were using their own matrices so _ViewProjMatrix was the camera matrix). This does not break with multiple camera because currently each camera will render all shadows.
The day we share shadows across camera we will need to change how this works though.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
alelievr: We need this because when we render shadows the `_ViewProjMatrix` is the view proj matrix from the light and `_CameraViewProjMatrix` is always the one from the camera. It is used in the tessellation which require the camera matrix.

I'll remove this as it shouldn't be use anywhere now:
> #if !defined(USING_STEREO_MATRICES)
float4x4 glstate_matrix_projection;
float4x4 unity_MatrixV;
float4x4 unity_MatrixInvV;
float4x4 unity_MatrixVP;
float4 unity_StereoScaleOffset;
#endif

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
alelievr: I agree, unfortunately I can't use `_FrustumPlanes` because for the tessellation triangle culling we need both frustum planes from the camera and from the light, i'll see if it can be moved to a separate cbuffer for shadows.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
alelievr: done

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
alelievr: The deviceProjectionYFlip was used for the `SetViewProjectionMatrices`, previously this matrix was computed in c++ during the `SetViewProjectionMatrices` from the device projection.
And the other device projection matrix is used for shadow data in the shader (in `_HDShadowDatas`), i think it's more efficient to store the value of y flipped device projection matrix here instead of computing it with multiple `GL.GetGPUProjectionMatrix()` anywhere we need it.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
EvgeniiG: Let's not put single-pass-specific variables in here. Please create a CBuffer for the shadow pass instead.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitDataMeshModification.hlsl
EvgeniiG: Ah, I see. Thanks!

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
EvgeniiG: You only need to multiply 1 column by -1, but you can do as you wish. :-)

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
sebastienlagarde: perfect

com.unity.shadergraph/ShaderGraphLibrary/ShaderVariables.hlsl
sebastienlagarde: ok got it

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
sebastienlagarde: I am ok with CB shadow. But please note that we have merge all CB into one because: 1) we reached the limit of 14 CB. 2) this doesn't make any change on perf, Unity perf are simply bad and doing this or not don't change cost.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -97,23 +63,6 @@ CBUFFER_START(UnityPerDraw)
 
 CBUFFER_END
 
-#if defined(USING_STEREO_MATRICES)
-CBUFFER_START(UnityStereoGlobals)
```
sebastienlagarde: @fabien-unity are you ok with this removal of stereo matrix ?

sebastienlagarde:
@alelievr can you merge master for this PR. Just waiting from @fabien-unity reply for stereo camera constant removal as I am not sure it is ok to do this.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
fabien-unity: all good! XR only cares about the ones in UnityPerPassStereo.
I've tested the branch on a few scenes and it looks good in VR.

PR #3278, opened 2019-03-24
Add _TERRAIN_BLEND_DENSITY macro

zeroyao:
### Purpose of this PR
The macro is for controlling whether density blend mode should be enabled. It's not currently configurable and will be enabled if height blending is not enabled, to retain the old behaviour.

---
### Release Notes
TBD.

---
### Testing status
**Katana Tests**: Running, but should be green as the change is trivial.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---


sebastienlagarde:
Don't forget changelog (but yes, for this one I am fine with no changelog)

PR #3281, opened 2019-03-25
Shader Graph bug fixes

pbbastian:
### Purpose of this PR
This PR fixes the following cases:

https://fogbugz.unity3d.com/f/cases/1135908/
Undo/redo breaks Shader Graph windows if the window is not visible. This could happen if the window is docked, but not the active tab. This happens because `Update` is only called if the window is visible. I first fixed this by triggering an `Update` on undo-redo, but this was obsoleted by the following bug fix.

https://fogbugz.unity3d.com/f/cases/1126627/
Performing undo or redo rapidly (multiple times per frame) breaks Shader Graph windows -- i.e. they throw exceptions and go into error state. This happens because each undo/redo causes a `ReplaceWith` in `GraphData`, and multiple of those per frame is apparently more than our change tracking system can handle. I fixed this by delaying the call to `ReplaceWith` until the next `Update`. `GraphObject` now tracks a current version and a serialised version of the graph, and will call `ReplaceWith` if they don't match. This also happened to be a much more elegant fix for the above bug, and also makes rapid undo/redo appear faster as multiple undo/redo per frame will be squashed into the same work.

_(Untracked)_
The new sub graph database importer fails if a sub graph refers to a sub graph guid that does not exist in the project. We already do handle that case in `SubGraphNode`, but there was an edge case in the recursion detection causing it to throw a `NullReferenceException`. I fixed it with a check of whether the referenced sub graph actually exists.

https://fogbugz.unity3d.com/f/cases/1126569/
Shader Graph doesn't work with references to sub assets. This happens because we only store GUIDs. I've fixed this by reverting to our old way of serialising references, but in a better way than before. GUID deserialisation is still supported because yay backwards compatibility, but we aggressively replace GUIDs with proper references. They only get replaced if the Shader Graph asset is re-saved though.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?unity_branch=trunk&ScriptableRenderLoop_branch=sg%2Fbug-fixes-peter&automation-tools_branch=master

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers


zadi0648:
UndoBugStill1

1126627 still reproduces. It's rapid undo & redo on nodes with an edge between them.

zadi0648:
UndoBugStill1

1135908 is better because the graph no longer crashes, however it now bakes all the changes into one single undo/redo step while all the surrounding undo/redo steps do nothing. Is this intentional? I wouldn't mind approving this part for now as it's an improvement over crashing the graph, as I would imagine fixing this and making it so you can step through changes on a window that isn't opened is a problem with Unity's undo system and out of our control?

zadi0648:
The other two cases look good 👍

pbbastian:
@zadi0648 I think I finally killed the issue. Is the second issue you described still happening for you?

pbbastian:
Hmm, is the second behaviour actually new?

PR #3283, opened 2019-03-25
Hidden Path Namespace for Subgraphs

alindmanUnity:
### Purpose of this PR
Adding a check to prevent certain subgraphs from populating the Create Node menu. This behaves exactly like "Hidden/" paths in HLSL shaders. 

---
### Release Notes
Only changes single file, Search Window Provider. 
Currently only filters out "Hidden" in Pascal case. "HIDDEN" or "hidden" will populate the menu. This behavior matches existing behavior for HLSL shaders. 

---
### Testing status
Katan Green

- Other: 

---
### Overall Product Risks
Low - small line change, one file only

Low - users shouldn't already have "Hidden" in paths to suddenly see subgraphs disappear. 

---
### Comments to reviewers


com.unity.shadergraph/Editor/Drawing/SearchWindowProvider.cs
```diff
@@ -21,6 +21,7 @@ class SearchWindowProvider : ScriptableObject, ISearchWindowProvider
         public bool nodeNeedsRepositioning { get; set; }
         public SlotReference targetSlotReference { get; private set; }
         public Vector2 targetPosition { get; private set; }
+        private static string s_HiddenFolderName => "Hidden";
```
pbbastian:
```suggestion
        private const string k_HiddenFolderName => "Hidden";
```

com.unity.shadergraph/Editor/Drawing/SearchWindowProvider.cs
```diff
@@ -21,6 +21,7 @@ class SearchWindowProvider : ScriptableObject, ISearchWindowProvider
         public bool nodeNeedsRepositioning { get; set; }
         public SlotReference targetSlotReference { get; private set; }
         public Vector2 targetPosition { get; private set; }
+        private const string k_HiddenFolderName => "Hidden";
```
zadi0648: Should be '='

com.unity.shadergraph/Editor/Drawing/SearchWindowProvider.cs
```diff
@@ -88,9 +89,10 @@ public List<SearchTreeEntry> CreateSearchTree(SearchWindowContext context)
                 {
                     AddEntries(node, new string[1] { asset.name }, nodeEntries);
                 }
-                else
+                
+                var title = node.subGraphData.path.Split('/').ToList();
+                if (title[0] != s_HiddenFolderName)
```
zadi0648: 's_HiddenFolderName' does not match 'k_HiddenFolderName'

zadi0648:
At first I thought I found a bug in master but I realized I was on your branch and this only reproduces on your branch: giving a sub graph no group results in it being in the creation tree twice. Once at level 1, a 2nd on at level 2 with the parent being "   ".
AlexHiddenPathBug0

com.unity.shadergraph/Editor/Drawing/SearchWindowProvider.cs
alindmanUnity: => is a new flag for assigning a get/set, I'm pretty sure it should stay that way

PR #3305, opened 2019-03-28
Hdrp/staging dxr 2

anisunity:
Bringing in all the changes that were made during the demo


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingShadowManager.cs
```diff
@@ -55,8 +55,13 @@ public void Init(HDRenderPipelineAsset asset, HDRaytracingManager raytracingMana
             m_GbufferManager = gbufferManager;
 
             // Allocate the intermediate buffers
+            m_AnalyticProbBuffer = RTHandles.Alloc(Vector2.one, filterMode: FilterMode.Point, colorFormat: GraphicsFormat.R16G16_SFloat, enableRandomWrite: true, useDynamicScale: true, useMipMap: false, name: "AnalyticProbBuffer");
```
sebastienlagarde: I think you can remove this one right ? it is the probe buffer (to do in a next PR)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRaytracingShadowManager.cs
```diff
@@ -68,52 +73,90 @@ public RTHandleSystem.RTHandle GetIntegrationTexture()
         public void Release()
         {
             RTHandles.Release(m_AreaShadowTextureArray);
-            RTHandles.Release(m_DenoiseBuffer0);
+            RTHandles.Release(m_RaytracingDistanceBuffer);
+            RTHandles.Release(m_RaytracingDirectionBuffer);
             RTHandles.Release(m_DenoiseBuffer1);
+            RTHandles.Release(m_DenoiseBuffer0);
+            RTHandles.Release(m_AnalyticProbBuffer);
```
sebastienlagarde: poke - related to comment above

PR #3312, opened 2019-03-29
Added documentation for Render Pass

JulienIgnace-Unity:
### Purpose of this PR
Added documentation for Render Pass


sebastienlagarde:
Merged in order to include into package @JordanL8 (better to have doc with typo than no doc) but please do correct and redo a PR for next package.

PR #3319, opened 2019-03-29
Fix editor resources asset refereces breaking when upgrading the HDRP asset

alelievr:
### Purpose of this PR
Fix editor resources asset refereces breaking when upgrading the HDRP asset

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FFixEditorResourcesWhenUpgrading&automation-tools_branch=master&unity_branch=trunk)

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


alelievr:
Yep i know, but the `CreatePipeline` is only supposed to be called only once when we load the pipeline (if everything goes well) so it's fine
Btw we may need to do something like this when a shader does not compile and HDRP tries to create itself at each frame eating all your memory and killing the editor.

PR #3322, opened 2019-03-30
[HDRP] Fix Stereo MotionVectors

racarone:
### Purpose of this PR
Fixes the previous frame's stereo uniforms being set to the current values every frame, breaking motion vectors.

Adds stereo instancing support to FullScreenDebug.shader as well.

---
### Testing status
**Katana Tests**: 

**Manual Tests**: 

**Automated Tests**:

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None


sebastienlagarde:
@fabien-unity @Chman what do you think ? Looks dangerous

fabien-unity:
1. The change to the shader is good.
2. Let's wait for the isFirstFrame problem because I have a branch in progress that will refactor some code and also should this.

PR #3325, opened 2019-04-01
Leave light lists in a clean state when we don't need to build them anymore 

FrancescoC-unity:
This is a fairly quick (but not so dirty) fix for the issue to go forward with the package. I'll try to get a nicer cleanup solution soon. 

Fix for https://fogbugz.unity3d.com/f/cases/1139761/ .

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fclean-light-list-when-not-building-it&unity_branch=trunk


FrancescoC-unity:
Oopsie, it fails on katana, I am on the fix now

PR #3330, opened 2019-04-01
Fix FB#1141553 - Sampler State properties are treated as duplicate when parameters are different

Kink3d:
### Purpose of this PR
This PR fixes Fogbugz case:
1141553: Sampler State properties are treated as duplicate when parameters are different

This also fixes automated test  `LWGraphicsTest - 050_Shader_Graphs`

This PR fixes by changing the following:
- The duplicate check tests the properties reference name, this cannot simply be changed
- Therefore the sampler type parameters are now written directly into the reference name, this requires changes across `SamplerStateShaderProperty`, `SamplerStateNode`, `PropertyNode` and `SamplerStateMaterialSlot`
- This means you cannot now change the reference name of a `SamplerState` type property. This is implemented via an `isRenamable` field in all shader property types to avoid hard coded changes to this property type in `BlackboardFieldPropertyView`
- Removing the ability to rename this property type does not negatively impact the user as this cannot be set via C# 

---
### Release Notes
Changelog

---
### Testing status
**Katana Tests**:  Running...

- Other: 

**Automated Tests**: What did you setup? N/A

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High? Low - Just moving a name append elsewhere mostly...

**Halo Effect**: None, Low, Medium, High? Low - Only touches sampler state properties

---
### Comments to reviewers


Kink3d:
Tests green

PR #3331, opened 2019-04-01
Backport: Fix FB#1141553 - Sampler State properties are treated as duplicate when parameters are different

Kink3d:
Backport #3330 

Katana running...


Kink3d:
Tests green

PR #3334, opened 2019-04-03
Bug Fixes

martint-unity:
### Purpose of this PR
This pr fixes all these Fogbugz cases

@Kink3d 
https://fogbugz.unity3d.com/f/cases/1143026/
When creating a subgraph it was created with a warning message regarding missing output. This fix creates a default vector4 output.

BEFORE:
create_subgraph

AFTER:
create_subgraph_fixed

---
https://fogbugz.unity3d.com/f/cases/1143024/
If creating an output that starts with a number it fails the generation of the shader. Added a fix for those cases where a user wants to start the name with a digit. 

BEFORE:
output_start_digit

AFTER:
output_start_digit_fixed

---
https://fogbugz.unity3d.com/f/cases/1143023/
A user was able to add nodes to subgraphs that shouldn't been allowed, like masternode, output node when converting to subgraph. 

BEFORE:
convert_to_subgraph

AFTER:
convert_to_subgraph_fixed

---
https://fogbugz.unity3d.com/f/cases/1143021/
A user was able to remove all Master nodes in a graph. Now there is always one master node that has Active flagged. Creating a new Master node and then right-click (contextual menu) to set it as active will change the active master node to that one. 

BEFORE:
deleting_all_masters

AFTER:
deleting_all_masters_fixed

---
https://fogbugz.unity3d.com/f/cases/1143028/
When setting a Master Node as active the main preview is now correctly showing the result of that master node.

AFTER:
main_preview_active_fixed

---
https://fogbugz.unity3d.com/f/cases/1139429/
Duplicate names in outputs was not working as intended. This has been fixed.

BEFORE:
DuplicateNameInSubgraphsCanFreezeSG

AFTER:
DuplicateNameInSubgraphsCanFreezeSGFixed

---
@marctem 
https://fogbugz.unity3d.com/f/cases/1141035/
The shadergraph graphview and main preview was always redrawing. Fixed that by removing redraw() method in MaterialGraphEditWindow and only redraw main preview when the Master node has been added to m_NodesToDraw

BEFORE:
always_redraws_ui

AFTER:
always_redraws_ui_fixed

---
### Release Notes

---
### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Fbug-fixes-martin&automation-tools_branch=master&unity_branch=trunk

- [ ] Built a player
- Other: 

---
### Overall Product Risks
Medium

Medium

---
### Comments to reviewers


martint-unity:
I have made an @ symbol where I would like your specific input on. Except for @pbbastian He can check everything so I am not doing something stupid :)

com.unity.shadergraph/Editor/Data/Implementation/NodeUtils.cs
```diff
@@ -287,7 +287,12 @@ public static string GetHLSLSafeName(string input)
         {
             char[] arr = input.ToCharArray();
             arr = Array.FindAll<char>(arr, (c => (Char.IsLetterOrDigit(c))));
-            return new string(arr);
+            var safeName = new string(arr);
+            if (char.IsDigit(safeName[0]))
```
marctem: I really don't like special-casing something fundamental like variable names. Disallowing port names that start with a number seems fine to me, but if we really want to fix it (and I get it's an artist thing =) ), then we should come up with a solution that's right for all variables. I'd be more OK with prepending "var_" everywhere than just in this one case.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -263,14 +280,15 @@ public void RenderPreviews()
         {
             UpdateShaders();
             UpdateTimedNodeList();
-            
+
+            m_NodesToDraw.UnionWith(m_TimedNodes);
```
marctem: The previous code was done with performance in mind - m_NodesToDraw is usually empty, and m_TimedNodes rarely actually updates so only propagating time in UpdateTimedNodes does a lot less work. It made an real difference on large graphs. I agree this implementation is simpler, but is it necessary?

pbbastian: I helped out with this change, so I'm chiming in :)

That's a good point, I can see an issue with the call being before `PropagateNodeList`. Speaking of that, didn't that previously keep around a set of visited nodes? It looks like it will currently visit nodes one time for each connection to it.

Anyway, this was also done with performance in mind. The method called previously ([`Union<TSource>(IEnumerable<TSource>, IEnumerable<TSource>)`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.union?view=netframework-4.8#System_Linq_Enumerable_Union__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEnumerable___0__)) will cause a GC allocation due to the fact that it returns an `IEnumerable<T>`. Since it works on `IEnumerable<T>`s it's not going to take advantage of the fact that it's a `HashSet<T>`, and probably needs to internally allocate a temporary `HashSet<T>` or similar to keep track of which elements have already been enumerated. The new code re-uses `m_NodesToDraw`, so it won't have to re-allocate very often because `Clear()` doesn't release memory.

I think the fix here is to move this one line down so that it's past `PropagateNodeList`, and then re-introduce `PropagateNodeList(m_TimedNodes, PropagationDirection.Downstream);` in `UpdateTimedNodeList()`.

com.unity.shadergraph/Editor/Data/Implementation/NodeUtils.cs
pbbastian: Yeah, that is true.. Maybe we should prefix all port variables with `port_`. That also makes the generated code a bit clearer :)

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
marctem: The last fix you mention works for me.

I take your point about Union(). The issue I have with UnionWith here is that NodesToDraw is almost always empty, while TimedNodes is not (if you have any time node), so the UnionWith in that direction is doing unnecessary work. But not much work, and we can't do it the other way since TimedNodes doesn't get cleared every frame. I don't have direct evidence of perf issues with that, though, it's probably fine, just annoying to the obsessed optimizer in my brain.

marctem: And it looks like you're right about losing the "visited nodes" set. My bad when I factored out the "GetConnectedNodes" function. @somebody should set up some tests around them and fix that!

pbbastian: Yeah, it does require a copy of the pointers, including look-ups in the HashSet for each. The point of the IndexSet was to make operations like this really fast, because it just has to OR all the ints from each, so the amount of work is 1/32, and is _very_ cache friendly.

The alternative would be to refactor the loop into a function, and then loop over each, but then we somehow need to make sure that the sets are disjoint to avoid drawing some nodes twice.

PR #3344, opened 2019-04-05
Fixed checkout of file when baking reflection probes

fredpointzero:
### Purpose of this PR
Fixed checkout of file when baking reflection probes (when using lighting window)

---
### Testing status
**Katana Tests**: [ABV Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2FProbeVCSCheckout&unity_branch=trunk)

- [ ] Built a player
- Other: 

**Automated Tests**: -

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers


aras-p:
FYI @sebastienlagarde for these "keep on retrying until succeed, since checkout is async" type of things, maybe it's better to actually wait for checkout to complete.

`Provider.Checkout` (and other APIs) return a `Task` that has a `Wait` method, so the code could just do a `Checkout(...).Wait()` and it would not need to do that crazy "try to copy file 20 times" loop

PR #3346, opened 2019-04-08
Lw/strip hw tier variants

joeldevahl:
### Purpose of this PR
Currently we generate up to three tiers of shaders from C++ side, even though tiers are actually deprecated by SRP. This PR adds a way of users to explicitly strip anything but tier one shaders by a checkbox in the LWRP asset.

---
### Release Notes
Added checkbox to LW Asset to allow for shader tier stripping.

---
### Testing status
**Katana Tests**: HDRP tests failed, LW tests passed (https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-Test%20PlayMode%20-%20Windows/builds/4041?automation-tools_branch=master&ScriptableRenderLoop_branch=lw/strip-hw-tier-variants&unity_branch=trunk)
This PR only touches LW code so HDRP fails is unrelated.

- [ x] Built a player
- [x ] Checked new UI names with UX convention
- Other: Build a player with custom tier settings, stepped through C# code to see variants where actually stripped.
---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

### DIFF
--- a/com.unity.render-pipelines.lightweight/Editor/LightweightRenderPipelineAssetEditor.cs
+++ b/com.unity.render-pipelines.lightweight/Editor/LightweightRenderPipelineAssetEditor.cs
@@ -50,8 +50,9 @@ namespace UnityEditor.Rendering.LWRP
             public static GUIContent srpBatcher = EditorGUIUtility.TrTextContent("SRP Batcher (Experimental)", "If enabled, the render pipeline uses the SRP batcher.");
             public static GUIContent dynamicBatching = EditorGUIUtility.TrTextContent("Dynamic Batching", "If enabled, the render pipeline will batch drawcalls with few triangles together by copying their vertex buffers into a shared buffer on a per-frame basis.");
             public static GUIContent mixedLightingSupportLabel = EditorGUIUtility.TrTextContent("Mixed Lighting", "Support for mixed light mode.");
-            
+
             public static GUIContent shaderVariantLogLevel = EditorGUIUtility.TrTextContent("Shader Variant Log Level", "Controls the level logging in of shader variants information is outputted when a build is performed. Information will appear in the Unity console when the build finishes.");
+            public static GUIContent stripGraphicsTierShaderVariants = EditorGUIUtility.TrTextContent("Strip Graphics Tier Shader Variants", "Controls if shaders of higher graphics tier than Tier1 should be stripped from the build or now/");
 
             // Dropdown menu options
             public static string[] mainLightOptions = { "Disabled", "Per Pixel" };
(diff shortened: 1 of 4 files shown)


phi-lira:
Hey Joel thanks for looking at this.
Adding @aleks01010101 as reviewer as he know how these tiers work on mobile. 
Please also add mobile QA before merging. 

I have a couple of questions:
1) Don't we need any hookup on engine/compiler frontend to force to build with Tier1 as well when necessary?
2) Why do we need to expose a setting for users instead of stripping all unused tiers by default?

joeldevahl:
>     1. Don't we need any hookup on engine/compiler frontend to force to build with Tier1 as well when necessary?

From what I gathered from talking to André M. and Christophe we always build tier1?

>     2. Why do we need to expose a setting for users instead of stripping all unused tiers by default?
Yeah ideally what should be done here is to do a C++ change and not do any tiers, and not add any keywords from tiers. This is not something we will be able to get into 19.2 though so we need some way of allowing users to opt in to strip tier shaders.

This implementation was done in discussion with André, if it's not good enough we should probably have had another plan of implementation all along.

phi-lira:
Understood. Seems reasonable enough.

aleks01010101:
There are several things here.

- No mobile device defaults to Tier1. GLES2 is Tier3, everything else is Tier2.
- Users can specify current tier via scripts. Any shader loaded after that will try to use the user-provided tier.
- Different tiers are not generated by the shader compiler if and only if Quality settings are the same for all three tiers.

I'm not sure what the effect will be of stripping the tiers that are supposed to be there. Please check if Tier1 shaders are still loaded.

com.unity.render-pipelines.lightweight/Editor/LightweightRenderPipelineAssetEditor.cs
```diff
@@ -50,8 +50,9 @@ internal class Styles
             public static GUIContent srpBatcher = EditorGUIUtility.TrTextContent("SRP Batcher (Experimental)", "If enabled, the render pipeline uses the SRP batcher.");
             public static GUIContent dynamicBatching = EditorGUIUtility.TrTextContent("Dynamic Batching", "If enabled, the render pipeline will batch drawcalls with few triangles together by copying their vertex buffers into a shared buffer on a per-frame basis.");
             public static GUIContent mixedLightingSupportLabel = EditorGUIUtility.TrTextContent("Mixed Lighting", "Support for mixed light mode.");
-            
+
             public static GUIContent shaderVariantLogLevel = EditorGUIUtility.TrTextContent("Shader Variant Log Level", "Controls the level logging in of shader variants information is outputted when a build is performed. Information will appear in the Unity console when the build finishes.");
+            public static GUIContent stripGraphicsTierShaderVariants = EditorGUIUtility.TrTextContent("Strip Graphics Tiers", "Controls if shaders of higher graphics tier than Tier1 should be stripped from the build or not.");
```
aleks01010101: I suppose it's better to rephrase this as "of graphics tiers other than Tier1"

phi-lira:
What's the status on this one?

joeldevahl:
Status is I got pulled on to other things...
We can most probably go ahead and use this as is if you figured out if the HDR mode is not used by anything.

PR #3352, opened 2019-04-10
RTHandle refactoring

JulienIgnace-Unity:
### Purpose of this PR
- Remove the need for HDCamera parameter in every SetRenderTarget/SetViewport functions in HDUtils. This will help make RTHandle system SRP agnostic (required by RenderGraph)
- Introduce RTHandleProperties structure that can be queried from a RTHandleSystem. This structure contains scales and sizes related to this system
- Cleanup usage of said scales everywhere (there are many places where we confuse "last frame of main RTHandle System"  with "history buffer RTHandle System"
- Rename _ScreenToTargetScale to _RTHandleScale to be more explicit about what it is.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fpre-rendergraph-refacto&automation-tools_branch=master&unity_branch=trunk)

- [ ] Built a player

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
@@ -8,6 +8,15 @@ namespace UnityEngine.Experimental.Rendering
 {
     public delegate Vector2Int ScaleFunc(Vector2Int size);
 
+    public struct RTHandleProperties
+    {
+        public Vector2Int previousFrameSize;    // Size set as reference at the previous frame
+        public Vector2Int currentFrameSize;     // Size set as reference at the current frame
+        // Scale factor from RTHandleSystem max size to requested reference size (referenceSize/maxSize)
+        // (x,y) current frame (z,w) last frame (this is only used for buffered RTHandle Systems
+        public Vector4 screenToTargetScale;
+    }
+
     public partial class RTHandleSystem : IDisposable
     {
         public enum ResizeMode
@@ -23,6 +32,9 @@ namespace UnityEngine.Experimental.Rendering
         HashSet<RTHandle>   m_AutoSizedRTs;
         RTHandle[]          m_AutoSizedRTsArray; // For fast iteration
         HashSet<RTHandle>   m_ResizeOnDemandRTs;
+        RTHandleProperties  m_RTHandleProperties;
+        public RTHandleProperties rtHandleProperties { get { return m_RTHandleProperties; } }
+
 
         int m_MaxWidths = 0;
         int m_MaxHeights = 0;
@@ -65,6 +77,8 @@ namespace UnityEngine.Experimental.Rendering
 
         public void SetReferenceSize(int width, int height, MSAASamples msaaSamples)
         {
+            m_RTHandleProperties.previousFrameSize = m_RTHandleProperties.currentFrameSize;
+
             width = Mathf.Max(width, 1);
             height = Mathf.Max(height, 1);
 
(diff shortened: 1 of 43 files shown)


com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
```diff
@@ -8,6 +8,15 @@ namespace UnityEngine.Experimental.Rendering
 {
     public delegate Vector2Int ScaleFunc(Vector2Int size);
 
+    public struct RTHandleProperties
+    {
+        public Vector2Int previousFrameSize;    // Size set as reference at the previous frame
+        public Vector2Int currentFrameSize;     // Size set as reference at the current frame
```
FrancescoC-unity: As far as I understand, this is the *viewport* size, no? 
I wonder if it'd be useful to also store the max size. I know it could be computed on the fly with currentFrameSize and rtHandleScale, but sounds useful as a pre-computed value?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2952,6 +2962,68 @@ public void RenderForward(Camera camera, CommandBuffer cmd, bool renderOpaque)
             }
         }
 
+        public static void CopyStencilBufferForMaterialClassification(CommandBuffer cmd, RTHandleSystem.RTHandle depthStencilBuffer, RTHandleSystem.RTHandle stencilCopyBuffer, Material copyStencilMaterial)
```
FrancescoC-unity: I am happy with this change, but is it really part of this PR?

JulienIgnace-Unity: Not really (I started with this and ended up needing the RTHandle refactoring) but I got lazy and did not want to make a second PR... (I figured it was small enough a change). I can remove it if needed.

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
```diff
@@ -8,6 +8,15 @@ namespace UnityEngine.Experimental.Rendering
 {
     public delegate Vector2Int ScaleFunc(Vector2Int size);
 
+    public struct RTHandleProperties
+    {
+        public Vector2Int previousFrameSize;    // Size set as reference at the previous frame
```
FrancescoC-unity: Isn't this information already stored in the HistoryRT system? I might be missing something?

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/RTHandleSystem.cs
JulienIgnace-Unity: Yes, this is the viewport. I will rename this to previous/currentViewport as it makes more sense indeed.
Max size could be useful indeed but I settled on putting only things that were currently needed.
Some places still compute their own size/scale etc so I figured we could add them then.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
FrancescoC-unity: Nah I think is fine, just wondered if it was a leftover from another branch elsewhere

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -15,9 +15,9 @@ public class HDCamera
         [GenerateHLSL(PackingRules.Exact, false)]
         public struct ViewConstants
         {
-            public Matrix4x4 viewMatrix;
+        public Matrix4x4 viewMatrix;
```
FrancescoC-unity: This all look shifted wrong now as you lost a tab/4 spaces. Same below

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -524,28 +493,28 @@ internal void UpdateAllViewConstants(bool jitterProjectionMatrix)
 
                 xrViewConstants = new ViewConstants[viewCount];
                 xrViewConstantsGpu = new ComputeBuffer(viewCount, System.Runtime.InteropServices.Marshal.SizeOf(typeof(ViewConstants)));
-            }
+                    }
```
FrancescoC-unity: Again, misalignment here and below

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -524,28 +493,28 @@ internal void UpdateAllViewConstants(bool jitterProjectionMatrix)
...
                     GetXrViewParameters(viewIndex, out proj, out view, out cameraPosition);
                     UpdateViewConstants(ref xrViewConstants[viewIndex], proj, view, cameraPosition, jitterProjectionMatrix);
 
                     // Compute offset between the main camera and the instanced views
                     xrViewConstants[viewIndex].worldSpaceCameraPosViewOffset = xrViewConstants[viewIndex].worldSpaceCameraPos - mainViewConstants.worldSpaceCameraPos;
-                }
-            }
-            else
-            {
+                        }
```
FrancescoC-unity: yup all shifted

JulienIgnace-Unity: P4Merge screwing with tabs :( Will reformat the file.

PR #3354, opened 2019-04-10
Fix diffusion profile upgrade crash with verifying asset enabled

alelievr:
### Purpose of this PR
+ Fix diffusion profile upgrade crash with verifying asset enabled: https://fogbugz.unity3d.com/f/cases/1142845/ (moved the default diffusion profile that was created each time the HDRP asset was built to an asset on the disk)
+ The default diffusion profile is now completly green (instead of broken) so it's easy to see when the profile is missing from the HDRPAsset / Volume override.

image

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FFixDiffusionProfileUPgradeCrashWithVerifyingAssetEnabled&automation-tools_branch=master&unity_branch=trunk

- [x] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


RSlysz:
Did we still have default diffusion profile in hdrpasset targergeting packages one instead of created instabce of them?

PR #3358, opened 2019-04-10
Making the raytracing work on the hdrp/staging branch

anisunity:
Enabling texture X for d3d12
Correcting a bug in the sampleprobevolume code
Correcting all the texture calls in raytracing and declaration so that it used texture_x functions
Declare the missing sample_texture calls in the raytracing macros
Move the code for the lit so that the macro is declared before the functions
Fixing the raytracing resources


com.unity.render-pipelines.high-definition/Runtime/Core/Textures/TextureXR.cs
```diff
@@ -16,6 +16,7 @@ public static bool useTexArray
                 switch (SystemInfo.graphicsDeviceType)
                 {
                     case GraphicsDeviceType.Direct3D11:
+                    case GraphicsDeviceType.Direct3D12:
```
anisunity: Did check this with @FrancescoC-unity

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/ShaderVariablesLightLoop.hlsl
```diff
@@ -40,7 +40,7 @@
     TEXTURE2D_ARRAY(_AreaShadowTexture);
```
anisunity: We will need to move this one to texture_array_x when it is supported

com.unity.render-pipelines.high-definition/Runtime/Core/Textures/TextureXR.cs
FrancescoC-unity: Texture X stuff was not supported on 12 when I added this check, I assume it is now?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
```diff
@@ -437,6 +438,7 @@ Shader "HDRP/Lit"
             // both direct and indirect lighting) will hand up in the "regular" lightmap->LIGHTMAP_ON.
 
             #define SHADERPASS SHADERPASS_LIGHT_TRANSPORT
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
```
sebastienlagarde: Need to be done for LayeredLit and Tessellation variant also

PR #3361, opened 2019-04-10
Fix exposure to support TEXTURE2D_X

fabien-unity:
### Purpose of this PR
Exposure code was broken since we switched to TEXTURE2D_X

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-exposure-texture2dx&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
_InputTexture is used in multiple different contexts in the shader code. I've added _SourceTexture to differentiate where we need to sample the source camera texture or an intermediate temp texture.


fabien-unity:
Note: this was reported by the A2 team and should be back-ported where it makes sense.

PR #3369, opened 2019-04-11
Remove shift of near plane for area light shadows

FrancescoC-unity:
Near plane value did not need to be shifted as position was shifted, this created issues with objects being clipped too soon. 

Validated this with Laurent too. 

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fremove-issue-with-area-light-shadow-clip&unity_branch=trunk  [Still running]


sebastienlagarde:
katana is failing on shadowed area light

FrancescoC-unity:
Yup, am on it.

FrancescoC-unity:
All green now

PR #3370, opened 2019-04-11
Fix FB 1143542 - Slot names give better error messages

alindmanUnity:
### Purpose of this PR
Nodes that accept custom text for inputs now validate for special characters that prevent the shader from compiling. 
SlotNameErrorMsg

---
### Release Notes
Validation check happens in `NodeUtils` and is called from nodes that need it.
Characters are valid for the display name are defined as A-Z, a-z, 0-9, _ ( ) and whitespace. Separators are for readability in the graph and get stripped before injected into the HLSL, and the validation will account for any non-Latin character edge cases now. 

---
### Testing status

- Other: 

---
### Overall Product Risks
Low

Low

---
### Comments to reviewers
bug: https://fogbugz.unity3d.com/f/cases/1143542/


zadi0648:
Adding Martin because of the reason described above

martint-unity:
So, yes I have a PR that fixes it starting with digits. Also I don't think it is the right solution to do a string with characters to check. I guess regex should be used. Otherwise you will have to check all different characters like my native language Swedish åäö should also be added to that string I guess and many more crazy native keyboard chars (double ss in German etc). Just my 2 cents. I will add the regex as a fix when I am back if it hasn't been added by then.

pbbastian:
Maybe not regex, but I think it should be turned on its head, so that there's a list of allowed characters, and then we check if it contains anything not in that list.

alindmanUnity:
Good points, I can remove the digit check if that's no longer an error we need to worry about and flip the invalid character check.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - When you rapidly perform an undo or redo, Shader Graph windows no longer break.
 - Sub Graphs that contain references to non-existing Sub Graphs no longer break the Sub Graph Importer.
 - You can now reference sub-assets such as Textures.
+- SubGraph Output and CustomFunctionNode now validate slot names and display appropriate error badge.
```
sharlenet: Sub Graph Output and Custom Function Node now validate slot names, and display appropriate error badge when needed.

alindmanUnity:
@zadi0648 HLSL won't compile with those special characters, so it wouldn't benefit anyone to let them use it in the graph.

PR #3403, opened 2019-04-17
Fix SMAA bug with RTHandle scale different than 1

FrancescoC-unity:
The offset in SMAASampleOffset was not scaled, only the base UV, causing issues whenever the RTHandle scale was different than 1. 

This allows the SMAA test in https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3396 to be enabled.


Chman:
Can you enable the test and generate the ref image?

PR #3410, opened 2019-04-18
HDRP/SMAA test

Chman:
### Purpose of this PR
Enabled the SMAA test on HDRP. Note that this will fail on the current branch as the fix is only available in `master` and `release/2019.2` as of now.


sebastienlagarde:
will merge master in the branch

PR #3413, opened 2019-04-18
Updated the license year

Chman:
### Purpose of this PR
Updated the license year per RM request.

`com.unity.testing.visualeffectgraph` is missing a LICENSE.md file.

Tagging for awareness: @stramit @julienf-unity @phi-lira @Kink3d @JordanL8


julienf-unity:
Seems like vfx and shader graph licence files are not the same as the others. Is this an issue or we dont care?

PR #3416, opened 2019-04-18
Fixed several issues when changing resolution

JulienIgnace-Unity:
### Purpose of this PR
This PR fixes several issues happening when changing resolution (which broke some tests and caused artefacts in the editor)
- Now Swap of History RTHandle is done before Resizing. This avoid invalidating history buffer whenever the buffers grow.
- Passed the correct frame id to HDCamera.Update to avoid discrepancies in the editor
- Update of previous view constants is now done every render instead of every frame otherwise it breaks when using the same camera more than once during a frame with different parameters.

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -714,7 +714,7 @@ public void VolumetricLightingPass(HDCamera hdCamera, CommandBuffer cmd, uint fr
 
                 GetHexagonalClosePackedSpheres7(m_xySeq);
 
-                int sampleIndex = (int)frameIndex % 7;
+                int sampleIndex = frameIndex % 7;
```
FrancescoC-unity: Why is this an int? I can only see it used to access an array

JulienIgnace-Unity: Original Time.frameCount is an int and also because C# makes uint a second class citizen and arrays don't actually work correctly with them.
Besides we were casting it to int in many places.

FrancescoC-unity: Oh well TIL about the indexing with int in C# :) Odd stuff, it was mostly my curiosity here

PR #3432, opened 2019-04-19
Fix HDCamera last active frame bug

racarone:
### Purpose of this PR
`m_LastFrameActive` was set too early, breaking `UpdateAllViewConstants()` frame count logic and previous frame constants.

---
### Overall Product Risks
**Technical Risk**:  Low
**Halo Effect**: Low


0lento:
I tried this already :) https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3420
but Unity has fixed this on https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3416 (not yet on the master)

racarone:
Ah I see, thanks. I’ll close this then

PR #3454, opened 2019-04-24
Fix camera-relative issue with punctual lights stereo instancing

fabien-unity:
### Purpose of this PR
Account for camera-relative during punctual light evaluation with XR single-pass instancing

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-fix-punctual-cr&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers
More details here: https://fogbugz.unity3d.com/f/cases/1136402/

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
@@ -218,7 +218,7 @@ void ModifyDistancesForFillLighting(inout float4 distances, float lightSqRadius)
 // distances = {d, d^2, 1/d, d_proj}
 void GetPunctualLightVectors(float3 positionWS, LightData light, out float3 L, out float3 lightToSample, out float4 distances)
 {
-    lightToSample = positionWS - light.positionRWS;
+    lightToSample = StereoCameraRelativeEyeToCenter(positionWS - light.positionRWS);
     int lightType = light.lightType;
 
     distances.w = dot(lightToSample, light.forward);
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -218,7 +218,7 @@ void ModifyDistancesForFillLighting(inout float4 distances, float lightSqRadius)
 // distances = {d, d^2, 1/d, d_proj}
 void GetPunctualLightVectors(float3 positionWS, LightData light, out float3 L, out float3 lightToSample, out float4 distances)
 {
-    lightToSample = positionWS - light.positionRWS;
+    lightToSample = StereoCameraRelativeEyeToCenter(positionWS - light.positionRWS);
```
sebastienlagarde: There is several other place where we use positionWS - light.positionRWS.
Like in screenspace sahdow or for directional light.
Question: Do we have a single light list for both eye or one list by eye ?

I guess we should look for a more generate solutoin rather than polluting the code with StereoCameraRelativeEyeToCenter when we have to deal with camera relative stuff.

If we have a single eye position for light list, we should be able to light.positionRWS that is relative to some kind of eye center on C# side, then we can just modify the computation of positionWS to be relative to eye center instead of camera relative.

what I mean here is that camera relative IS simply a translatoin at origin. If we perform a translation at Eye center instead everything should work correctly for both eye.

I guess that there is nowhere in the code where we assume that cameraWorldPos is 0 (as we can disable camera relative), so all should be fine if we chose this path, and we don't need to define _WorldSpaceCameraPosViewOffset anymore. Thought ?

fabien-unity:
I've reworked and created a new PR due to some conflict with master: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3501

PR #3456, opened 2019-04-25
Fix freeze on save with sub graph node

pbbastian:
### Purpose of this PR
Fixes issue #1140846
https://fogbugz.unity3d.com/f/cases/1140846

Instead of hiding `PortInputView`s we now destroy them instead. The freeze comes from hiding them, and while I'm not entirely sure why, this is better either way and seems to fix the issue.

---
### Testing status
**Katana Tests**: Running https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?unity_branch=trunk&ScriptableRenderLoop_branch=sg%2Ffix-freeze-on-save&automation-tools_branch=master&sort=1-asc

**Manual Tests**: 
- [x] Able to repro (using steps from FB) before changes, not able to after changes

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None


alindmanUnity:
question, is this different from what exists in #3343  ?

pbbastian:
@alindmanUnity Yes, I talked to @Kink3d about it, we're going with this fix :)

PR #3459, opened 2019-04-25
Hdrp/make transparent write motion vecs fit in ui

FrancescoC-unity:
Without this change the text is cut off. 

This still works well with shrinking the side bar to the minimum possible.


sebastienlagarde:
@FrancescoC-unity can you add changelof please. Thanks :)

PR #3479, opened 2019-04-29
Dock new Shader Graph windows

pbbastian:
### Purpose of this PR
Uses the new API in Unity from 2019.2+ to make an attempt at docking new windows. It will first try to dock next to an existing Shader Graph window, and if that fails it will try to dock next to the Scene View. If that fails a new floating window is created.

GIF demo

---
### Testing status
**Katana Tests**: Running https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?unity_branch=trunk&ScriptableRenderLoop_branch=sg%2Fdock-windows&automation-tools_branch=master

**Manual Tests**:
* Open SG window with with Scene View, but no existing SG windows
* Open SG window with existing SG windows and Scene View
* Open SG window with existing SG windows, but no Scene View
* Open SG window with no existing SG windows nor Scene View

**Automated Tests**: None

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


zadi0648:
I'm wondering if it would be better if it only did this to _only_ other Shader Graph windows, and otherwise new Shader Graphs open up in a undocked window? I think in most cases the user won't want to have the SG tabbed over the scene view. And this informs noobies that they can of course dock the SG anywhere (should be common knowledge, but who knows). Anyhow I'll have to think about it some more.

pbbastian:
This was mostly based on my own preferences, because I find it annoying that the window spawns kind of on top of everything. But it is worth pulling in some opinions about :) If we had a settings pane, this would totally go in there :p

zadi0648:
Hey so I've been out sick for a while, but here is my thoughts on this after having more time to think it over: https://unity.slack.com/archives/G7HQCPV71/p1557123726173700,
thoughts?

Kink3d:
@pbbastian Are we merging this?

PR #3482, opened 2019-04-30
Features and fixes for the 2D renderer

rainsing:
### Purpose of this PR
Brings a batch of 2D renderer changes to master.

Really sorry for the gigantic PR. After we ship the first preview, we'll to change our development strategy to produce smaller PRs, and do PRs more often.

Notable changes to LWRP and SG: 
- Added two SG master nodes: Sprite Lit and Sprite Unlit
- GetDefaultShader() is added to ScriptableRendererData so that it can override the default shader returned by the pipeline asset.
- Two parameters, clearBlitTarget and pixelRect, were added to FinalBlitPass to give the pass more customizability. Used by the Pixel Perfect Camera. 
- Editor analytics are added for the 2D renderer.

---
### Release Notes
- Added two Shader Graph master nodes: Lit Sprite and Unlit Sprite. They only work with the 2D renderer.
- The 2D renderer and Light2D component received a number of improvements and are now ready to try as experimental features.

---
### Testing status
**Katana Tests**: Green Katana at 2619ba1
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=2d%2Fstaging&unity_branch=trunk&automation-tools_branch=master&sort=1-asc

- [x] Built a player

GFX tests: 
056_2D_Lights:
056_2D_Lights
056_2D_Lights_Shader_Graph:
056_2D_Lights_Shader_Graph
Unit tests:
Under `UnityEngine.Rendering.LWRP.Tests`, a few tests are added for the Light2D component. Also we migrated the existing tests for Pixel Perfect Camera to LWRP.

---
### Overall Product Risks
**Technical Risk**: Medium. Simply because a big amount of code was added. But most of it is experimental.

**Halo Effect**: Low. Changes are mostly for the 2D Renderer.

---
### Comments to reviewers


com.unity.shadergraph/Editor/AssetCallbacks/CreateSpriteLitShaderGraph.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEditor.ProjectWindowCallback;
+
+namespace UnityEditor.ShaderGraph
+{
+    public class CreateSpriteLitShaderGraph : EndNameEditAction
+    {
+        [MenuItem("Assets/Create/Shader/2D Renderer/Lit Sprite Graph", false, 208)]
+        public static void CreateMaterialGraph()
+        {
+            ProjectWindowUtil.StartNameEditingIfProjectWindowExists(0, CreateInstance<CreateSpriteLitShaderGraph>(),
```
pbbastian: Use `GraphUtil.CreateNewGraph`

com.unity.shadergraph/Editor/AssetCallbacks/CreateSpriteUnlitShaderGraph.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEditor.ProjectWindowCallback;
+
+namespace UnityEditor.ShaderGraph
+{
+    public class CreateSpriteUnlitShaderGraph : EndNameEditAction
+    {
+        [MenuItem("Assets/Create/Shader/2D Renderer/Unlit Sprite Graph", false, 208)]
+        public static void CreateMaterialGraph()
+        {
+            ProjectWindowUtil.StartNameEditingIfProjectWindowExists(0, CreateInstance<CreateSpriteUnlitShaderGraph>(),
```
pbbastian: Use `GraphUtil.CreateNewGraph`

com.unity.shadergraph/Editor/Data/MasterNodes/SpriteLitMasterNode.cs
```diff
@@ -0,0 +1,85 @@
...
+                new[]
+            {
+                PositionSlotId,
+                ColorSlotId,
+                MaskSlotId,
+                NormalSlotId,
+            });
+        }
+
+        //public NeededCoordinateSpace RequiresNormal(ShaderStageCapability stageCapability)
```
pbbastian: Remove commented code. Does `NormalMaterialSlot` actually work without this?

com.unity.render-pipelines.lightweight/Editor/2D/HeaderModifier.cs
```diff
@@ -0,0 +1,48 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UIElements;
+using System;
+
+
+class HeaderModifier : IMGUIContainer
```
pbbastian: Needs a namespace

com.unity.render-pipelines.lightweight/Editor/2D/PixelPerfectCameraEditor.cs
```diff
@@ -0,0 +1,251 @@
+using System;
+using UnityEngine;
+using UnityEngine.Experimental.Rendering.LWRP;
+
+namespace UnityEditor.Experimental.Rendering.LWRP
+{
+    [CustomEditor(typeof(PixelPerfectCamera))]
+    public class PixelPerfectCameraEditor : Editor
```
pbbastian: Does this need to be `public`?

Verasl:
So there is a lot of little UX/Workflow issues, along with missing Docs/Tests but I know most of these cannot be solved in this PR alone, we can discuss more in the 2D Renderer meeting sync

rainsing:
> Sprite Lit, Sprite Unlit, and accompanying code should live in LWRP package instead

This is done in 5eff437.

> Looks like there's a lot of stuff in `UnityEditor.Experimental.Rendering.LWRP`. Is it intentional that it is in the experimental namespace?

Yes. The whole 2D renderer will be experimental when we first ship it.

> What's with all the types prefixed with `_`?

It's because the symbol names would start with a number (2DSomething) if we don't prefix them with _. I'll ask around in the coding convention channel about this.

com.unity.shadergraph/Editor/AssetCallbacks/CreateSpriteLitShaderGraph.cs
rainsing: Fixed in 1080ea3.

com.unity.shadergraph/Editor/AssetCallbacks/CreateSpriteUnlitShaderGraph.cs
rainsing: Fixed in 1080ea3.

com.unity.shadergraph/Editor/Data/MasterNodes/SpriteLitMasterNode.cs
rainsing: Removed commented code in b7c8a6d. @unity-cchu Do you have an idea why this was commented out in the first place?

com.unity.render-pipelines.lightweight/Editor/2D/HeaderModifier.cs
rainsing: Fixed in b7c8a6d.

com.unity.render-pipelines.lightweight/Editor/2D/PixelPerfectCameraEditor.cs
rainsing: No, it doesn't. Fixed in b7c8a6d.

rainsing:
> Getting asserts when a new 2D renderer is assigned directly to the pipeline asset and a material preview window is visible, this doesn't happen when only a camera override is used.

Tried to reproduce this but couldn't. More info about the steps or a repro project?

rainsing:
> So there is a lot of little UX/Workflow issues, along with missing Docs/Tests but I know most of these cannot be solved in this PR alone, we can discuss more in the 2D Renderer meeting sync

Yeah let's talk about UX during the meeting. Regarding docs, we do have a very detailed manual, just need to figure out where and how to publish it. And what tests do you think we're missing?

Verasl:
> > Getting asserts when a new 2D renderer is assigned directly to the pipeline asset and a material preview window is visible, this doesn't happen when only a camera override is used.
> 
> Tried to reproduce this but couldn't. More info about the steps or a repro project?

Actually this seemingly happens without the 2D renderer assigned as well(just not a frequent) so ignore this.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Render2DLightingPass.cs
```diff
@@ -49,37 +57,58 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                    cmd.name = "Render Lights - " + SortingLayer.IDToName(layerToRender);
+#endif
+                    RendererLighting.RenderLights(camera, cmd, layerToRender);
+                }
+                else
+                {
+                    RendererLighting.ClearDirtyLighting(cmd);
+                }
+
+                SetRenderTarget(cmd, colorAttachment, RenderBufferLoadAction.Load, RenderBufferStoreAction.Store, ClearFlag.None, Color.white, TextureDimension.Tex2D);
```
phi-lira: LWRP does auto target setup management now. If you implement ScriptableRenderPass.Configure in your render pass and inside it call ConfigureTarget(RenderTargetIdentifier colorAttachment) it will do this for. 

The advantage is that LWRP will propertly setup the load/store actions.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Utility/RendererLighting.cs
```diff
@@ -142,7 +152,7 @@ static private bool RenderLightSet(Camera camera, int lightOpIndex, CommandBuffe
             return renderedAnyLight;
         }
 
-        static private void RenderLightVolumeSet(Camera camera, int lightOpIndex, CommandBuffer cmdBuffer, int layerToRender, RenderTargetIdentifier renderTexture, List<Light2D> lights)
+        static private void RenderLightVolumeSet(Camera camera, int blendStyleIndex, CommandBuffer cmdBuffer, int layerToRender, RenderTargetIdentifier renderTexture, List<Light2D> lights)
```
phi-lira: This method has some duplicated code with RenderLightSet. Refactoring it to be more modular will improve maintanance cost / reduce risk of forgetting to updated in both places.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Utility/RendererLighting.cs
```diff
@@ -290,40 +329,33 @@ static public void RenderNormals(ScriptableRenderContext renderContext, CullingR
...
-                    cmdBuffer.EnableShaderKeyword(keyword);
-                else
-                {
-                    cmdBuffer.DisableShaderKeyword(keyword);
+                if (!s_BlendStyles[i].enabled)
                     continue;
-                }
 
-                string sampleName = "2D Lights - " + s_LightOperations[i].name;
+                string sampleName = "2D Lights - " + s_BlendStyles[i].name;
```
phi-lira: This is creating GC pressure.

com.unity.render-pipelines.lightweight/Runtime/2D/_2DRenderer.cs
```diff
@@ -6,22 +6,81 @@ namespace UnityEngine.Experimental.Rendering.LWRP
     internal class _2DRenderer : ScriptableRenderer
```
phi-lira: Still think this _ prefix is weird / against code styles. Not sure what to suggest though.

com.unity.shadergraph/Editor/Data/MasterNodes/SpriteLitMasterNode.cs
pbbastian: But your changes to `GraphUtil` makes it only generate normals if needed by a node. If you do not have this method, then your master node won't need the normal, and it might not be there. Have you tried whether it works in a graph with only the master node?

com.unity.render-pipelines.lightweight/Runtime/ScriptableRendererData.cs
```diff
@@ -47,6 +47,11 @@ internal virtual Material GetDefaultMaterial(DefaultMaterialType materialType)
         {
             return null;
         }
+
+        internal virtual Shader GetDefaultShader()
```
phi-lira: Should this be implemented in the forward renderer data?

com.unity.render-pipelines.lightweight/Shaders/2D/Sprite-Lit-Default.shader
```diff
@@ -82,7 +31,74 @@ Shader "Lightweight Render Pipeline/2D/Sprite-Lit-Default"
             #pragma multi_compile USE_SHAPE_LIGHT_TYPE_2 __
             #pragma multi_compile USE_SHAPE_LIGHT_TYPE_3 __
 
-            #include "Include/CombinedShapeLightPass.hlsl"
+			struct Attributes
+			{
+				float3 positionOS   : POSITION;
+				float4 color		: COLOR;
+				half2  uv			: TEXCOORD0;
```
phi-lira: Minor . Several identation issues

com.unity.render-pipelines.lightweight/Shaders/2D/Light2D-Shape.shader
```diff
@@ -21,8 +21,11 @@ Shader "Hidden/Light2D-Shape"
             #pragma vertex vert
             #pragma fragment frag
             #pragma multi_compile_local SPRITE_LIGHT __
+			#pragma multi_compile_local USE_NORMAL_MAP __
```
phi-lira: minor. some weird formatting issues.

com.unity.shadergraph/Editor/Data/MasterNodes/SpriteLitMasterNode.cs
unity-cchu: Yes, this works with just the master node.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Render2DLightingPass.cs
rainsing: In this particular case we can't use the automatic target management. This is inside a for loop where we need to set the RT for each sorting layer. But I've replaced `ScriptableRenderPass.SetRenderTarget()` with `CoreUtils.SetRenderTarget()` as the former is going to be removed.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Utility/RendererLighting.cs
rainsing: We'll take note of this and refactor it at a later time. Thanks for pointing it out.

com.unity.render-pipelines.lightweight/Runtime/2D/Passes/Utility/RendererLighting.cs
rainsing: Fixed in e192b5f.

com.unity.render-pipelines.lightweight/Runtime/2D/_2DRenderer.cs
rainsing: In d5c8ca5 and efe9c07, we've renamed all files and symbols whose name starts with "_2D" to "Something2D".  e.g. "_2DRenderer" is now "Renderer2D".

com.unity.render-pipelines.lightweight/Runtime/ScriptableRendererData.cs
rainsing: You don't need to. Forward Renderer Data will continue to work like before.

com.unity.render-pipelines.lightweight/Shaders/2D/Light2D-Shape.shader
rainsing: Fixed in 50b7791.

com.unity.shadergraph/Editor/Data/MasterNodes/SpriteLitMasterNode.cs
pbbastian: Then I do not understand why. The code you've added in `GraphUtil` makes it add the vertex normal input only if `graphRequiements.requiresNormal != NeededCoordinateSpace.None`, which seems like a separate fix. Where do you set `requiresNormal`?

rainsing: You are right, the changes in `GraphUtil` is a separate fix. We were confused ourselves. Sorry about that.

The true reason why our master node doesn't implement `IMayRequireNormal` was we don't want to calculate a tangent space normal from the geometry if nothing is connected to the Normal slot, as Sprite geometry doesn't have normal or tangent information. To make sure we have tangent space normal generated for `SurfaceDescriptionInputs`, in the SubShader C# code we actually ignored `ShaderGraphRequirements` and hardcoded a tangent space normal by doing something like:
`ShaderGenerator.GenerateSpaceTranslationSurfaceInputs(NeededCoordinateSpace.Tangent, InterpolatorType.Normal, surfaceDescriptionInputStruct);`

After you pointed this out, we gave it a second thought and came to the conclusion it's both easier and cleaner to just use `Vector3MaterialSlot` instead of `NormalMaterialSlot` for our Normal slot on the master node. When nothing is connected to that slot, the tangent space "up" vector (0, 0, 1) is used as the default value. Please take a look at this change in ee914ba.

PR #3486, opened 2019-04-30
Fixed resource issues in LW

Chman:
### Purpose of this PR
As per our slack conversation. Needs to be backported to 2019.2.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lwrp%2Fresources-fix&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)


sophiaaar:
While it looks bad that 2 out of 3 tests failed, this is actually very promising, because on SRP master, standalone tests don't even run. So it confirms that this change fixes that issue

com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
```diff
@@ -163,7 +163,7 @@ static string GetFullPath(string basePath, ReloadAttribute attribute, int index
                     path = attribute.paths[index];
                     break;
                 case ReloadAttribute.Package.Root:
-                    path = Path.Combine(basePath, attribute.paths[index]);
+                    path = basePath + "/" + attribute.paths[index];
```
sophiaaar: why was this part changed?

Chman: Because `Path.Combine` does weird things with path separators and seems to be platform dependent. For instance on windows it would return paths like this:
```
Packages/com.unity.render-pipelines.lightweight\Runtime/Shaders/Utils/Blit.shader
```
(note the `\`) So I just forced it to a forward slash. Just a safety measure.

stramit:
internally we will remap the '/' to the right thing anyway.

com.unity.render-pipelines.core/Runtime/Utilities/ResourceReloader.cs
phi-lira: Yep, I was just seeing an error about this path weirdness in my branch.

PR #3493, opened 2019-04-30
[backport 2019.1] Miscellaneous Shader Graph Fixes

zadi0648:
### Purpose of this PR
Backport of #3349 

For 20% I joined in on some bug smashing, this will be the first PR. Fixed the following bugs:

---

https://fogbugz.unity3d.com/f/cases/1127892/: Converting a Texture2D Array Property to Inline Silently Fails
Was caused by ToConcreteNode() returning Texture2DAssetNode instead of Texture2DArrayAssetNode.

Texture2DConvertToInline

---

https://fogbugz.unity3d.com/f/cases/1135978/: Gradient, dragging from Input results in nothing in node creation menu
The shader stage was set to the meaningless value of 0 which resulted in the menu not finding it because it wasn't compatible with any stage (acceptable values are 1, 2, 3); now uses the default which is all stages.

(see ^ fogbugz case for before)
GradientFix

---

https://fogbugz.unity3d.com/f/cases/1138033/: Having multiple Custom Function Nodes with multiple functions from the same source produces errors
Was caused by the string builder attempting to add the same include multiple times. Now provides Function Source in the registry to prevent it from being added twice.

CustomFunctionFix0

---

https://fogbugz.unity3d.com/f/cases/1139341/: Changing meshes resets the rotation of the mesh, until it is dragged in which case the previously value is used.
Implemented fix option 1: makes more sense for the rotation to start anew when changing the shape.

(see ^ fogbugz case for before)
PreviewRotationFix0

---

https://fogbugz.unity3d.com/f/cases/1143544/: Sub-graphs causes connections to be lost when outputs are reordered
Now uses a dictionary to stores the slot and the edges associated with that slot, and reconnects everything after everything has been removed and added.

(see ^ fogbugz case for before)
ReorderableListFix

---

### Testing status

**Manual**: Tested each bug and confirmed that it's fix (see gifs), will add Wyatt since people aren't allowed to qa themselves.

**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=2019.1%2Fsg%2Fmisc-minor-fixes&automation-tools_branch=master&unity_branch=2019.1%2Fstaging

---

### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**:  Low

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/Texture2DArrayShaderProperty.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/Texture2DArrayShaderProperty.cs
@@ -80,7 +80,7 @@ namespace UnityEditor.ShaderGraph
 
         public override AbstractMaterialNode ToConcreteNode()
         {
-            return new Texture2DAssetNode { texture = value.textureArray };
+            return new Texture2DArrayAssetNode { texture = value.textureArray };
         }
 
         public override AbstractShaderProperty Copy()
--- a/com.unity.shadergraph/Editor/Data/Nodes/Input/Gradient/GradientNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Input/Gradient/GradientNode.cs
@@ -90,7 +90,7 @@ namespace UnityEditor.ShaderGraph
 
         public sealed override void UpdateNodeAfterDeserialization()
         {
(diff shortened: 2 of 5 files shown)


sebastienlagarde:
Katana is green, merging

PR #3498, opened 2019-05-01
Rework Sub Graph importer

pbbastian:
### Purpose of this PR
Remove SubGraphDatabase in favour of using `GatherDependenciesFromSourceFile` in importers for handling nested sub graphs. Currently, the sub graph importer still has to detect mutually recursive sub graphs, because of how they are handled by the import pipeline.

To make this faster, I've introduced `MinimalGraphData` which only deserialises exactly what it needs to determine dependencies.

`GetSourceAssetDependencies(List<string>)` has been removed from `AbstractMaterialNode`. Instead, nodes can provide a minimal type via `HasDependenciesAttribute`.

I tried to make Git track the renames nicely, but GitHub isn't following. Essentially I've merged `SubGraphAsset` and `SubGraphData`, as well as `ShaderSubGraphImporter` and `SubGraphDatabaseImporter`.

Fixes [FB1138262: Recursive Sub Graph check & error badge doesn't happen until file reload](https://fogbugz.unity3d.com/f/cases/1138262/). This issue turned out to be worse than assumed. What actually happened was that existing Shader Graph windows would get stuck on old sub graph data. We now reload sub graph data in all open Shader Graph windows after any sub graph has been imported.

Fixes [FB1076181: Having multiple tabs open and in "focus" causes multiple (maybe all) open assets to compile](https://fogbugz.unity3d.com/f/cases/1076181/). There is a separate issue with ADBv2 that will cause all Shader Graphs using Sub Graphs to re-compile their preview shader when any Sub Graph is saved. For ADBv1 users this is not an issue.

---
### Testing status
**Katana Tests**: Green except for unrelated VFX failure https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?unity_branch=trunk&ScriptableRenderLoop_branch=sg/fix-1138262&automation-tools_branch=master

- Ensure no repro without fix
- Ensure repro with fix

**Automated Tests**: nil

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Low


pbbastian:
Hey Zack, I slightly changed how this works to also fix a related issue, and to make it a bit better. Can you take another look? 😊

zadi0648:
As for the second case that this fixes; no problems detected however it should be noted that it's difficult to notice/reproduce the original issue in master with async comp.
ManyOpenGraphs

pbbastian:
@zadi0648 I think that might have come in with the previous merge with master. Alex landed a fix for that recently, so I merged master again. Can you try again?

zadi0648:
@pbbastian The "unexpected token '-'" error has been resolved. There is a problem now is that the error badge goes away when saving 2ndRecursionSubG (you can get the badge to show up again by removing the connect edge or removing the Sub Graph node and readding it. Save to make it go away again). Also it doesn't show in 1stRecursionSubG.

There is errors in console when saving.

Nevertheless this is an improvement over the old repro and we could proceed / make a new fogbugz case.

RecursionBadgeGoesAwayOnSave

pbbastian:
@zadi0648 Alright, let's try again. The work I've been doing to fix the importer to behave well with ADBv2 was based on these changes, so since I haven't managed to land this yet I've merged the importer fixes (killing SubGraphDatabase) into this one. This should take care of the issues you reported above :)

pbbastian:
Current status: Re-creating branch due to massively conflicting changes in `master`

PR #3515, opened 2019-05-02
HDRP half support

jessebarker:
### Purpose of this PR

For HDRP, on platforms that really support fp16 data and ALU ops, our abstraction of switching real between half and float strictly on shader API was not sufficient to cover all of the cases (might have been fine for LWRP, but there are also core library functions that are unused by LWRP).  Some cases really need to be float all the time and some cases needed holes plugged where we end up generating NaN or Inf values in shaders.  The code here preserves the previous default behavior (should be safe for LWRP).
---
### Release Notes

This should not impact the release notes.

---
### Testing status
**Katana Tests**: green

- [x] Built a player
- Other: 

---
### Overall Product Risks
Low
Medium (some changes to core to support the HDRP changes)
---
### Comments to reviewers

It would be good to understand if the approaches for eliminating NaNs and Infs are the optimal solution for these cases.


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -653,6 +653,7 @@ void EncodeIntoGBuffer( SurfaceData surfaceData
 
     // Pre-expose lighting buffer
     outGBuffer3 *= GetCurrentExposureMultiplier();
+    outGBuffer3 = max(outGBuffer3, 0); // NaN guard
```
sebastienlagarde: I personnaly don't like hidding NaN like this. This hide the true problem making it very difficult to localize when we have a NaN killer like this. We have Already a NaN killer that happen during postprocess pass (if enabled), so my suggestion is to remove this and use the NaN killer (that have a cost) if needed. We don't want to hide issue,we want to show them so we can solve it.

sebastienlagarde:
I am good with the PR except the case that hidde NaN

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
jessebarker: OK, I can pull this line.  Are you OK with the clamp on the specularReflectivity term in the next change down (lines 1055/1056)?

sebastienlagarde: yes, it is not a clamp but something to avoid divide by 0 :), should be fine.

sebastienlagarde:
Adding @EvgeniiG in case he have something to say about the specularReflectivity. I suppose you have review the PR with him already.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
victor-unity: for this one, once we fix the IntegrateGGXAndDisneyDiffuseFGD for Mac&IO, we can get rid of it. On Mac&iOS, the texture has zero in it.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -1052,7 +1053,7 @@ PreLightData GetPreLightData(float3 V, PositionInputs posInput, inout BSDFData b
     // We deem the intensity difference of a couple of percent for high values of roughness
     // to not be worth the cost of another precomputed table.
     // Note: this formulation bakes the BSDF non-symmetric!
-    preLightData.energyCompensation = 1.0 / specularReflectivity - 1.0;
+    preLightData.energyCompensation = 1.0 / max(specularReflectivity, 0.01) - 1.0;
```
victor-unity: for this one, we can get rid of the max once we fix the IntegrateGGXAndDisneyDiffuseFGD for Mac&IO. On Mac&iOS, the texture has zero in it.

jessebarker:
I've pushed up an additional change that removes the NaN clamp when encoding GBuffer3.  Let me know if there's anything else (katana still running).

com.unity.render-pipelines.core/ShaderLibrary/Packing.hlsl
```diff
@@ -494,7 +494,7 @@ real UnpackFloatFromR8G8(real2 f)
 }
 
 // Pack float2 (each of 12 bit) in 888
-real3 PackFloat2To888(real2 f)
```
EvgeniiG: Why was this a problem?

com.unity.render-pipelines.core/ShaderLibrary/Packing.hlsl
```diff
@@ -506,7 +506,7 @@ real3 PackFloat2To888(real2 f)
 }
 
 // Unpack 2 float of 12bit packed into a 888
-real2 Unpack888ToFloat2(real3 x)
```
EvgeniiG: And this.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
EvgeniiG: I think it's better to not touch this part and rather fix the texture or the texture generation code (make sure it does not write 0).

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
EvgeniiG: Yes, I don't like this one, either.

com.unity.render-pipelines.core/ShaderLibrary/Packing.hlsl
victor-unity: Because according to xcode, 1.0f by 4095.5 resulted 4096.

EvgeniiG: Yes, and we should use 4095 instead. That's the real bug.

victor-unity: Aren't 4095 & 4096 the same number in 16bits ?

EvgeniiG: Good point, it's 4094 and then 4096. :/ So I guess it can't work in half precision, anyway.

victor-unity: Unless we change back the format to 888 maybe, but I understood that we do not wan to diverge too much between platform.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
jessebarker: I've pushed a commit that reverts this.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
jessebarker: I've pushed up a commit that reverts this one as well.

PR #3517, opened 2019-05-03
2d/global light refactor

unity-cchu:
Work done on global light refactor


com.unity.render-pipelines.lightweight/Runtime/2D/Light2D.cs
```diff
@@ -37,16 +32,61 @@ internal static BoundingSphere[] boundingSpheres
...
+                {
+                    // If we found the first global light in our prefab stage
+                    if (PrefabStageUtility.GetPrefabStage(light.gameObject) == PrefabStageUtility.GetCurrentPrefabStage())
+                    {
+                        color = light.color;
+                        return true;
+                    }
+                    else
+                    {
+                        if (!foundGlobalColor)
```
rainsing: Shouldn't we return a black color in this case? It's kinda weird to let a global light that's not in a prefab to affect the global color when you're editing that prefab.

com.unity.render-pipelines.lightweight/Runtime/2D/Light2D.cs
```diff
@@ -37,16 +32,61 @@ internal static BoundingSphere[] boundingSpheres
...
-                    return s_Instance.m_GlobalClearColorsForPrefab;
-#endif
-                return s_Instance.m_GlobalClearColors;
+                Light2D light = lights[i];
+                if (light.IsLitLayer(sortingLayerIndex))
+                {
+                    // If we found the first global light in our prefab stage
+                    if (PrefabStageUtility.GetPrefabStage(light.gameObject) == PrefabStageUtility.GetCurrentPrefabStage())
+                    {
+                        color = light.color;
```
rainsing: I think previously it was light.m_Intensity * light.m_Color rather than just the color?

com.unity.render-pipelines.lightweight/Runtime/2D/Light2D.cs
```diff
@@ -352,47 +359,18 @@ internal bool IsLightVisible(Camera camera)
             return isVisible;
         }
 
-        static internal void AddGlobalLight(Light2D light2D, bool overwriteColor = false)
+        static internal void ErrorIfDuplicateGlobalLight(Light2D light2D)
```
rainsing: There're probably more occasions where we need to trigger this error. For example when a light's sorting layer changes.

com.unity.render-pipelines.lightweight/Runtime/2D/Light2D.cs
rainsing: Let's leave it this way.

com.unity.render-pipelines.lightweight/Runtime/2D/Light2D.cs
rainsing: Fixed.

PR #3518, opened 2019-05-03
Fix for refraction causing NaNs when IOR is set to 1 and refraction mode is Box

FrancescoC-unity:
When IOR is precisely 1, it can happen that the refracted ray end up being perpendicular to the normal, leading to a dot(-N, R) == 0 and therefore a divide by 0. 

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp%2Ffix-for-refraction-potentially-creating-nans&unity_branch=trunk&sort=1-asc [Just kicked with the new fix]


sebastienlagarde:
I have added a changelog

PR #3526, opened 2019-05-03
Shadergraph material properties features

alelievr:
### Purpose of this PR
Necessary features for the material render-states in shadergraph. 

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=shadergraph%2Ffeature%2Fmaterial_properties&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


sebastienlagarde:
Hey @marctem we need your feedback on the preview hack in this PR. 
We need access to the preview material PR and this is the hacky solutoin we come with to do our prototype. But now we need a proper access. What is your suggestion ?

 // HACK to access the preview material for HDRP, does not support more than one shader graph window at the same time
    public static class HackedPreview

Thanks

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -15,6 +15,16 @@ namespace UnityEditor.ShaderGraph.Drawing
 {
     delegate void OnPrimaryMasterChanged();
 
+    // HACK to access the preview material for HDRP, does not support more than one shader graph window at the same time
```
marctem: I don't think you actually need to add anything to PreviewManager! We already have PreviewManager.masterRenderData, from which you can get the Material for the current master preview. You can access that and just set the props you need when the user changes settings. You can also register for PreviewManager.onPrimaryMasterChanged to be notified if the user sets a different master node as active to know whether the current master node you're processing should set the material properties or not.

alelievr: Okay, that sounds good but how do we access the current PreviewManager for a given master node ?

marctem: Good point, looks like you'll need to make GraphEditorView.previewManager public. You can get the GraphEditorView anytime you're in UI code with GetFirstAncestorOfType<GraphEditorView>.

Or if you already have access to the MasterPreviewView, you could create a public accessor for its reference to the PreviewManager. Hopefully one of those routes will work for you, whichever is cleaner in your implementation. I'll approve either.

alelievr: I just pushed my changes using the previewManager in the MasterPreviewView. I didn't found better solution than sending the actual master preview view to the master so i can setup the keywords we need depending of the master node type, I hope this is okay.
Also, please note that this solution needs to be tested more to see if it doesn't break in some conditions.

pbbastian: Which HDRP class needs access to the data? I think it's a bit overkill to add 2 linked callbacks for 1 special case. Would it be possible for you to also add the HDRP-specific changes to this PR? That makes it a lot easier to consider the API :)

alelievr: If you want to see the whole diff, you can have it here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/HDRP/feature/shadergraph_blendstates_refactor?expand=1, it's not yet a PR but if you want I can create a draft PR for the diff.
The classes that needs access to the preview material are all the HDRP master nodes. I'm aware that sending GUI stuff to the master node which is technically just logic and code generation is not ideal, i would prefer to have some kind of per-node view access so each pipeline have an entry point in the shadergraph UI and then from this access the preview.
Anyway, if we can find a better solution than the current one that would be great :)

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -102,6 +102,20 @@ public override void OnImportAsset(AssetImportContext ctx)
             }
         }
 
+        public static AbstractMaterialNode GetOutputNode(string path)
```
alelievr: I've added this function here to access the output node from HDPR, i discovered that we need the master node type for our material upgrader so it is able to patch shader keywords based on their type.
What do you thnik about this ?
Also, it would be cool to include these functions somewhere:
```csharp
internal static bool IsShaderGraph(this Shader shader)
{
    AssetImporter shaderImporter = AssetImporter.GetAtPath(AssetDatabase.GetAssetPath(shader));
    return shaderImporter is ShaderGraphImporter;
}

internal static AbstractMaterialNode GetShaderGraphOutputNode(this Shader shader)
{
    return ShaderGraphImporter.GetOutputNode(AssetDatabase.GetAssetPath(shader));
}
```

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -102,6 +102,20 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+        {
+            try
+            {
+                var textGraph = File.ReadAllText(path, Encoding.UTF8);
+                var graph = JsonUtility.FromJson<GraphData>(textGraph);
+                return (AbstractMaterialNode)graph.GetNodes<IMasterNode>().FirstOrDefault() ?? graph.GetNodes<SubGraphOutputNode>().FirstOrDefault();
+            }
```
marctem: Here you should use `graph.outputNode` to get the active master node (or subgraph output node). Graphs can have multiple master nodes and you want the one currently set as active.

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
marctem: I think these helpers can stay in HDRP for now until we see a need for them in LW as well.

pbbastian:
I made some changes to `GetOutputNode`:
* It's now living in `GraphUtil` so that we can keep the importer class entirely internal.
* It now returns the type of the node rather than the full node, as that interface will be easier to maintain. I have another PR incoming that adds a class (internal) for peeking into SG files without having to fully deserialise everything. By returning just the type we can make use of this once it lands.
* It now uses `GraphData.outputNode` to get the output node.

com.unity.shadergraph/Editor/Data/Graphs/EnumShaderProperty.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public override string GetPropertyBlockString()
+        {
+            var result = new StringBuilder();
+            if (hidden)
+            {
+                result.Append("[HideInInspector] ");
+            }
+
+            string enumValuesString = ""; // TODO
```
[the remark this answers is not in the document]
alelievr: It's to support this kind of attribute with a C# enum: 
```[Enum(UnityEngine.Rendering.BlendMode)]```

File: com.unity.shadergraph/Editor/Data/Graphs/EnumShaderProperty.cs [hunk 5]
```diff
@@ -0,0 +1,126 @@
...
+            return new PreviewProperty(PropertyType.Vector1)
+            {
+                name = referenceName,
+                floatValue = value
+            };
+        }
+
+        public override AbstractMaterialNode ToConcreteNode()
+        {
+            throw new Exception("Enums as node are not yet supported");
```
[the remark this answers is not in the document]
alelievr: How could this be called, the enum property is not exposed in the blackboard ? I only implemented it for script generation. If we implement the node, we also need to do the blackboard UI with a reorderable list to edit the enum and an enum slot for the node default values. (As well as adding a new EnumControl field that support enums that are not C# (i.e a list of string and values))

com.unity.shadergraph/Editor/Data/Graphs/EnumShaderProperty.cs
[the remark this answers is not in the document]
alelievr: That was not necessary for this PR, I can add it if you want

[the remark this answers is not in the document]
alelievr: Finally I added it with a new EnumType: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3526/files#diff-7d614c8a46d4f96c83da38b113f48d5fR13

com.unity.shadergraph/Editor/Data/Graphs/Vector1ShaderProperty.cs
```diff
@@ -73,7 +83,47 @@ public Vector2 rangeValues
...
+            set
+            {
+                if (m_EnumType == value)
+                    return;
+                m_EnumType = value;
+            }
+        }
+    
+        [SerializeField]
+        Type m_CSharpEnumType;
```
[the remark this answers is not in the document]
Kink3d: Since we'll have to use reflection for this, I think thats a good idea. Less risk of changing our mind about the data layout.

sebastienlagarde:
@marctem @Kink3d Hey, if you can give your update today on this PR with the latest change (as Friday and Monday are day off for you in US :) ).
Ideally we would like to be able to move forward on Friday with it before we start the package release next week.
Thanks.

com.unity.shadergraph/Editor/Data/Graphs/Vector1ShaderProperty.cs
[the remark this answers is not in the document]
pbbastian: @sebastienlagarde All the other `[SerializeField]` added in this PR should also be removed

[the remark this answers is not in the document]
alelievr: done

sebastienlagarde:
PR have addressed the various feedback. Katana is green and we have get approval from Peter by Slack. We need to move forward with this PR as other PR are waiting for this change for next week package. We can still push fix during the beginning of next week, so feel free to add comment here if you want more change. Will be easier to push as bugfix.

PR #3541, opened 2019-05-06
Correcting various bugs due linked to ray tracing

anisunity:
- Adding a rayTracedAreaShadowIndex in light data and a padding size done with @FrancescoC-unity 
- Adding init for the contact shadows in ray tracing loop
- Changing the noise texture for ray tracing (and correcting a sampling bug)
- Correcting problems related to history buffers

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FCorrectingRayTracingProblems&automation-tools_branch=master&unity_branch=trunk&sort=1-asc


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightDefinition.cs
```diff
@@ -110,13 +110,11 @@ public struct LightData
...
-        // We store the ray traced area shadow index as a negative value inside the contactShadowMask.
-        // Contact shadows are disabled for area lights and setting the index as negative allows for still
-        // disabling contact shadows in the shader code (checks for => 0)
-        public int     rayTracedAreaShadowIndex { get => -contactShadowMask; set => contactShadowMask = -value; }
-#endif
-        public int     contactShadowMask;      // negative if unused (TODO: 16 bit)
+        public int     contactShadowMask;       // negative if unused (TODO: 16 bit)
+
+        // TODO: Instead of doing this, we should pack the ray traced shadow index into the tile cookie for instance
+        public Vector3 unusedData;
```
sebastienlagarde: you don't need this.  "public Vector3 unusedData;"
we don't need to pack as we use a structured buffer on hlsl side and on C# side there is no need of pading.

anisunity: @FrancescoC-unity

FrancescoC-unity: Gah of course, I was thinking of cbuffers

PR #3560, opened 2019-05-07
[ShaderGraph] Fix FB Case 1152247 - SubGraph Preview Generation Fails

alindmanUnity:
### Purpose of this PR
Sub Graph preview generation fails when we append the slotId to the variable name. Sub Graph properties used Hash Code which return a negative int and produce invalid syntax. This fixes that regression by ensuring the string is never build with a signed int. 
Previous behavior: 
image

Fixed (expected) Behavior: 
image

---
### Release Notes

---
### Testing status
Running

- Other: 

---
### Overall Product Risks
Low - unchecked supresses warnings but casting to uInt can inherently cause issues

Low - it does affect the AbstractMaterialNode but exclusive to preview generation

---
### Comments to reviewers
This is the smallest footprint fix to handle this edge case, and has the potential to catch any other negative int edge cases that could come  up in preview generation. 
Bug: https://fogbugz.unity3d.com/f/cases/1152247/


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - When you set a Master Node as active, the Main Preview now shows the correct result.
 - When you save a graph that contains a Sub Graph node, the Shader Graph window no longer freezes.
 - Fixed an error that occured when using multiple Sampler State nodes with different parameters.
+- Sub Graph input slots no longer produce errors on preview shader generation.
```
sharlenet: Please change to below (if correct).
Sub Graph input slots no longer produce errors when you generate preview shaders.

Kink3d:
This is tested green on Yamato and we really need this to land before package is cut, so its going in...

pbbastian:
I think this is the wrong approach. Negative `slotId` is invalid input, so an `ArgumentOutOfRangeException` should be thrown from `GetVariableNameForSlot`. The caller of `GetVariableNameForSlot` relies on their `slotId`s being unique, but by casting to `uint` you risk conflicts. The fix should instead be done in `SubGraphNode`, which shouldn't be generating negative `slotId`s.

PR #3562, opened 2019-05-07
Remove ShaderGraph tests from HD and LW Test Projects.

marctem:
### Purpose of this PR
Shader Graph tests are already run by HD and LW via the package dependency. Having the SG project marked as testable duplicated the test runs. I am adding a test to the SG Test Project which relies on assets in the project, and which fails when run as part of another project. We've been meaning to do this for a while but had no direct reason to. Now we do.

---
### Testing status
**Katana Tests**: Green! https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Fremove-tests-from-rps&unity_branch=trunk&automation-tools_branch=master&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: None - just removes redundant tests

**Halo Effect**: None


phi-lira:
We have a test in lwrp test project to test shadergraph lwrp integration / master nodes. The test is currently disabled due to some issues. I suppose this PR  should not prevent us from having such a test. Am I correct?

marctem:
> We have a test in lwrp test project to test shadergraph lwrp integration / master nodes. The test is currently disabled due to some issues. I suppose this PR should not prevent us from having such a test. Am I correct?

Yeah, that should be fine. This just stops the tests defined under com.unity.shadergraph/Tests from appearing and running in your LWTestProject.

PR #3563, opened 2019-05-08
Force preIntegratedFGD_GGXDisneyDiffuse and volumetricLighting shaders to use float when they generate data.

victor-unity:
### Purpose of this PR
Fix lighting flickering because of the lack of precision during the generation of the precomputed FGDGGX texture and during the generate of volumetric lighting data, by forcing the use of float int this shader. 

---
### Release Notes
Force us of float during the generation of the precomputed FGDGGX texture.

---
### Testing status
**Katana Tests**: running

- [x] Opened test project
- [x] Took a GPU capture to confirm the texture is correct.
- [x] Open SampleScene, switch platform to iOS, check there are no black blocs breaking the construction scene.

Launch Katana : [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=fp32-for-preintegratedFGD_GGXDisneyDiffuse-&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low, very local to the pre integration shader.

---

### Comments to reviewers
The #define PREFER_HALF 0 forces use of float even on mobile.


sebastienlagarde:
Please add a changelog
+ There is several preIntegratedFGD file in HDRP, we should add the change to all of them (AXF, CharlieFabric etc...) Look for all files with PreIntegratedFGD in it. Thanks!

victor-unity:
I will. It seems we have the same problem/solution with VolumetricLighting.compute. Do you prefer to have another PR for VolumetricLighting.compute or I add it to this PR (modified/renamed of course) ?

sebastienlagarde:
> I will. It seems we have the same problem/solution with VolumetricLighting.compute. Do you prefer to have another PR for VolumetricLighting.compute or I add it to this PR (modified/renamed of course) ?

you can do in same PR, thanks

sebastienlagarde:
Note: for branch naming, use HDRP/fp32-for-preIntegratedFGD_GGXDisneyDiffuse

PR #3569, opened 2019-05-09
Fixed crash on exception when saving, case 1136543

AscendingMan:
### Purpose of this PR
Case https://fogbugz.unity3d.com/f/cases/1136543/ lists an issue where attempting to save a read-only (not checked out) graph when VCS is disconnected results in a crash.
---
### Testing status
In progress

- [x] Built a player
- Other: 
---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

### DIFF
--- a/com.unity.shadergraph/Editor/Util/FileUtilities.cs
+++ b/com.unity.shadergraph/Editor/Util/FileUtilities.cs
@@ -5,9 +5,9 @@ using UnityEditor.VersionControl;
 
 namespace UnityEditor.ShaderGraph
 {
-    public class FileUtilities
+    static class FileUtilities
     {
-        public static bool WriteShaderGraphToDisk<T>(string path, T data, bool prettyPrint = false)
+        public static bool WriteShaderGraphToDisk<T>(string path, T data)
         {
             if (data == null)
             {
@@ -18,7 +18,7 @@ namespace UnityEditor.ShaderGraph
 
             try
             {
-                File.WriteAllText(path, EditorJsonUtility.ToJson(data, prettyPrint));
+                File.WriteAllText(path, EditorJsonUtility.ToJson(data, true));
             }
             catch (Exception e)
             {
@@ -27,7 +27,7 @@ namespace UnityEditor.ShaderGraph
                 {
                         FileInfo fileInfo = new FileInfo(path);
                         fileInfo.IsReadOnly = false;
-                        File.WriteAllText(path, EditorJsonUtility.ToJson(data, prettyPrint));
+                        File.WriteAllText(path, EditorJsonUtility.ToJson(data, true));
                         return true;
                 }
                 Debug.LogException(e);
--- a/com.unity.shadergraph/Editor/Util/FileUtilities.cs
+++ b/com.unity.shadergraph/Editor/Util/FileUtilities.cs
@@ -1,0 +1,28 @@
+using System;
+using System.IO;
+using Debug = UnityEngine.Debug;
+
+namespace UnityEditor.ShaderGraph
+{
+    public class FileUtilities
+    {
+        public static bool WriteShaderGraphToDisk<T>(string path, T data)
+        {
+            if (data == null)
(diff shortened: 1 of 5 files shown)


com.unity.shadergraph/Editor/Util/FileUtilities.cs
```diff
@@ -0,0 +1,28 @@
...
+            if (data == null)
+            {
+                throw new ArgumentNullException(nameof(data));
+            }
+
+            try
+            {
+                File.WriteAllText(path, EditorJsonUtility.ToJson(data, true));
+            }
+            catch (Exception e)
```
aras-p: Hmm so this only attempts to catch the exception, but not do anything that would have caused the file to not be writable in the first place?

e.g. if path was not "open for edit", attempt to check it out.

or if file exists and has a read-only attribute, then remove the read-only attribute first, etc.

com.unity.shadergraph/Editor/Util/FileUtilities.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static bool WriteShaderGraphToDisk<T>(string path, T data)
+        {
+            if (data == null)
+            {
+                throw new ArgumentNullException(nameof(data));
+            }
+
+            try
+            {
+                File.WriteAllText(path, EditorJsonUtility.ToJson(data, true));
```
aras-p: Only one calling place was setting `prettyPrint=true` in ToJson call previously, now all four callers would be doing that. Maybe add an argument to WriteShaderGraphToDisk to keep the previous behavior.

com.unity.shadergraph/Editor/Util/FileUtilities.cs
```diff
@@ -6,23 +6,52 @@ namespace UnityEditor.ShaderGraph
...
+                }
                 Debug.LogException(e);
                 return false;
             }
             return true;
         }
+
+        static void CheckoutIfValid(string path)
+        {
+            if (VersionControl.Provider.enabled && VersionControl.Provider.isActive)
```
aras-p: I know you're just moving a function (good!), but the way the function was written before is kinda weird and not how it's supposed to be done IMHO :)

1. replace CheckoutIsValid with IsOpenForEdit
2. all the writing code only writes the asset file, and not the .meta file, so replace CheckoutMode.Both with CheckoutMode.Asset

com.unity.shadergraph/Editor/Util/FileUtilities.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.IO;
+using Debug = UnityEngine.Debug;
+using UnityEditor.VersionControl;
+
+namespace UnityEditor.ShaderGraph
+{
+    public class FileUtilities
```
pbbastian: Remove `public`, make `static`

com.unity.shadergraph/Editor/Util/FileUtilities.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public static bool WriteShaderGraphToDisk<T>(string path, T data, bool prettyPrint = false)
```
pbbastian: When writing to disk we should always be pretty printing, so remove this parameter. I seems that some of the replaced code paths were not doing that for some reason.

AscendingMan:
> 
> 
> Good job 👍 Can we also add the prompt like described in the FB case? Then we can close the case entirely :)

@pbbastian It was decided not to add the prompt since MakeEditableWithPrompt in native would have to become a public API and that is not the case now. There would also be backwards-compatibility issues with it as it only recently landed in 2019.3.0a3. @aras-p might want to comment on this too.

marctem:
Looks good, thanks!

marctem:
WTF, github. Don't close the PR, just my review. 🙄

marctem:
@pbbastian could you merge please?

PR #3578, opened 2019-05-09
Lw/docs upd comparison table

Katstodian:
Update the Comparison Table in the Docs to reflect the current state of features.

### DIFF
--- a/com.unity.render-pipelines.lightweight/CHANGELOG.md
+++ b/com.unity.render-pipelines.lightweight/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-* Updated the [Feature Comparison Table](lwrp-builtin-feature-comparison.md) to reflect the current state of LWRP features.
+- Updated the [Feature Comparison Table](lwrp-builtin-feature-comparison.md) to reflect the current state of LWRP features.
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.lightweight/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added SpeedTree Shaders.
 
+### Changed
+
```
phi-lira: please remove empty line

com.unity.render-pipelines.lightweight/Documentation~/shading-model.md
```diff
@@ -1,62 +1,58 @@
-- - # Shading models in Lightweight Render Pipeline
```
phi-lira: Is the change to this file intenational? Seems only formatting is changed.

Katstodian: Yes, the formatting was changed, because it was causing major errors in the published site. It should be formatted correctly for the output now.

com.unity.render-pipelines.lightweight/CHANGELOG.md
Katstodian: Done.

PR #3596, opened 2019-05-10
Lw/set camera properties

egomeh:
### Purpose of this PR

Shader variables related to time was out of sync, and some times a frame behind. This PR adds, in lightweight, a manual set of time related variables.

---
### Release Notes

This is a fix to [bug 1142495](https://fogbugz.unity3d.com/f/cases/1142495/)

---
### Testing status
**Katana Tests**: Tests are started.

- [ ] Built a player
- Other: 

Bug case has reproduction steps

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lw/set-camera-properties&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Medium

---
### Comments to reviewers


com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -74,6 +101,25 @@ public ScriptableRenderer(ScriptableRendererData data)
...
         }
 
+        static PerFrameConstants GetFrameConstants(ref Camera camera)
+        {
+            // We make these parameters to mirror those described in `https://docs.unity3d.com/Manual/SL-UnityShaderVariables.html
+            float timeEights = Time.time / 8f;
+            float timeFourth = Time.time / 4f;
+            float timeHalf = Time.time / 2f;
+
+            PerFrameConstants perFrameConstants = new PerFrameConstants();
```
phi-lira: This is allocating memory every frame increasing GC pressure.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -74,6 +101,25 @@ public ScriptableRenderer(ScriptableRendererData data)
...
+
+            // Time values
+            perFrameConstants.time = Time.time * new Vector4(1f / 20f, 1f, 2f, 1f);
+            perFrameConstants.sinTime = new Vector4(Mathf.Sin(timeEights), Mathf.Sin(timeFourth), Mathf.Sin(timeHalf), Mathf.Sin(Time.time));
+            perFrameConstants.cosTime = new Vector4(Mathf.Cos(timeEights), Mathf.Cos(timeFourth), Mathf.Cos(timeHalf), Mathf.Cos(Time.time));
+            perFrameConstants.deltaTime = new Vector4(Time.deltaTime, 1f / Time.deltaTime, Time.smoothDeltaTime, 1f / Time.smoothDeltaTime);
```
phi-lira: We don't need to store these values. We should only cache Time.time. We could have a function that sets Time constants given the time instead.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -136,6 +182,9 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
 
             SortStable(m_ActiveRenderPassQueue);
 
+            PerFrameConstants perFrameConstants = GetFrameConstants(ref camera);
+            perFrameConstants.SetFrameValues(null);
```
phi-lira: Sadly the SetupCameraProperties will setup time constants again. In that case it might be that there's still a mismatch in time between the render passes that execute before and after SetupCameraProperties. 

We need to cache Time.time and set it before and after SetupCameraProperties to do this.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
egomeh: I have now changed to a free function that takes only the time and a commandbuffer, which may be null.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
egomeh: They are now not stored, but computed in a function that compute them as needed.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
egomeh: The struct is now gone. We only hold the time as a float in the updated version of the PR.

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -73,8 +80,15 @@ public LightweightRenderPipeline(LightweightRenderPipelineAsset asset)
             PerFrameBuffer._GlossyEnvironmentColor = Shader.PropertyToID("_GlossyEnvironmentColor");
             PerFrameBuffer._SubtractiveShadowColor = Shader.PropertyToID("_SubtractiveShadowColor");
 
+            PerFrameBuffer._Time = Shader.PropertyToID("_Time");
+            PerFrameBuffer._SinTime = Shader.PropertyToID("_SinTime");
+            PerFrameBuffer._CosTime = Shader.PropertyToID("_CosTime");
+            PerFrameBuffer.unity_DeltaTime = Shader.PropertyToID("unity_DeltaTime");
+
             PerCameraBuffer._InvCameraViewProj = Shader.PropertyToID("_InvCameraViewProj");
+            PerCameraBuffer._ScaledScreenParams = Shader.PropertyToID("_ScreenParams");
```
phi-lira: Typo here. It should be _ScreenParams.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -20,6 +20,35 @@ namespace UnityEngine.Rendering.LWRP
...
     {
+        void SetShaderTimeValues(float time, CommandBuffer cmd = null)
+        {
+            // We make these parameters to mirror those described in `https://docs.unity3d.com/Manual/SL-UnityShaderVariables.html
+            float timeEights = Time.time / 8f;
+            float timeFourth = Time.time / 4f;
+            float timeHalf = Time.time / 2f;
+
+            // Time values
+            Vector4 timeVector = Time.time * new Vector4(1f / 20f, 1f, 2f, 1f);
```
phi-lira: w component should be 3f.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -20,6 +20,35 @@ namespace UnityEngine.Rendering.LWRP
     /// </summary>
     public abstract class ScriptableRenderer
     {
+        void SetShaderTimeValues(float time, CommandBuffer cmd = null)
```
phi-lira: You should not call Time.time inside this function. This is why we pass time as parameter. You should consider to pass Time.smoothDeltaTime as parameter as well so no Time api is called within this function..

PR #3600, opened 2019-05-10
Precision

Kink3d:
### Purpose of this PR
This PR adds the ability to set the precision of generated code per graph and per node.

Image
_Green lines are an unrelated styling bug in trunk_

### High level changes

- Graph precision is set in the tool bar of each graph. This is the base precision for inheritance.
- Precision can be overridden per node in the nodes cog-wheel settings menu.
- Most nodes use `Inherit` as their default precision mode. Exceptions are:
   - Position (float)
- Precision can be overridden per property in the Blackboard.
- When resolving inherited precision:
   - Highest precision of all inputs is used.
   - If an input is disconnected that input uses graph precision.
   - If an input has a connection that input uses the precision of connected node
   - If the node has no inputs, the graph precision is used.
- Sub Graphs:
   - Precision is precalculated for each Sub Graph and the output code is detokenised in the database.
   - Sub Graph function input arguments use the "property" precisions from the Blackboard
   - Sub Graph function output arguments use the Sub Graph Output node precision
   - Sub Graph calls created by each reference node use the precision of that Sub Graph's input/output arguments when defining variables and adapting outputs for inlining code.
   - Sub Graph reference node cannot set precision (as precision is prebaked as stated above).

### Notes on this draft

- This PR targets `sg/body-code-ssb` as that PR is expected (and required) to land before this PR lands to master. A Pr for this branch is o n the way.
- This PR will require minor modifications once a PR has landed from `sg/node-color-by-category`. This change will only add the integration needed to display precision as node colors, and will not affect the way the code currently in this PR functions.
- This PR requires #2845 to land before merging. This is simply that this PR modifies the generated code for all nodes, and to merge it without automation is too risky. 

### Questions for draft reviewers

- I have exposed `real` precision. I believe this is the correct call but it is up for discussion.
- Currently all preview properties are `float`. Should we use the correct node precisions or is this ok?
- Should any other nodes be `float` precision by default? (see list above)
- Bumping scripted importers. Main importer version should be bumped (I have refactored some function code). Do I also need to bump the Sub Graph importer?

### Known issues
- Functions inside Sub Graphs do not change their precisions when they are edited and saved. I believe this is due to the function source being cached somehow. Investigating.
- There is a line in the settings menu UI (not in all nodes). I believe this to be an overlapping UI element. Investigating.

---
### Release Notes
Will update changelog when PR is out of draft.

---
### Testing status
**Katana Tests**: Will run when out of draft (see Notes of the draft)

- [ ] Built a player
- Other: 

Working on this.

N/A

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High? High - If the detokenising fails, the shader compilation fails. This could potentially happen anywhere.

**Halo Effect**: None, Low, Medium, High? High - deeply embedded through all code generation


pbbastian:
I'm not sure if it makes sense to have both `real` and `half`. @phi-lira @sebastienlagarde?

pbbastian:
> Currently all preview properties are float. Should we use the correct node precisions or is this ok?
I think that's fine for now.

> Bumping scripted importers. Main importer version should be bumped (I have refactored some function code). Do I also need to bump the Sub Graph importer?
Yep, sounds like a good idea.

pbbastian:
It seems weird to me that the precision setting, which is saved into the asset, lives in the toolbar with the editor viewing settings. How about putting it in the blackboard?

phi-lira:
Right, I also don't think it makes sense and it's actually confusing to have both real and half. I'll suggest saying that this is a precision hint. You expose float and half. Half is a hint to both platform and pipeline.

sebastienlagarde:
>Questions for draft reviewers
>I have exposed real precision. I believe this is the correct call but it is up for discussion.

I think it is fine to expose "real precision". At least on HDRP this should be the "default" as then HD Mobile will have half and other platform will have float. Only people working specifically on mobile only project will have a need for half type. So fine.

>Currently all preview properties are float. Should we use the correct node precisions or is this ok?
yes

>Should any other nodes be float precision by default? (see list above)
Be very careful:  anything related to uv coordinate should be float (otherwise it mean you can sample at texel precision in a texture that is higher than 1k by 1k). Fine for LW, not fine for HD Mobile...

Any world transformation (for animation etc...), i.e everything that is connect to world offset must be in float, otherwise you will get precision issue. In particular any intersection / geometric transform / position must be in float.

sebastienlagarde:
> I also don't think it makes sense and it's actually confusing to have both real and half.

See my comment above, for HDRP it could make sense (depends if you compile for HD Mobile or not). But yeah, you are maybe right, there shouldn't have issue here as we are on material side and not on lighting side. HOWEVER if "real" isn't available, it is mandatory to not convert uv to half in this case as explain above... As most HDRP project have 2K or even 4K texture...

Kink3d:
Ok I have removed real (we can always add it back later) and moved the dropdown as per earlier meeting feedback.

Note about UVs. By default every graph is set to float and all nodes inherit that, so UV will always be float until you switch the graph to half anyway. Also note: all nodes that use UVs directly from the active fields/requirements (texture samplers, procedurals, etc) will **always** use float UVs regardless of anything you do in the graph anyway (varyings precisions are not customisable at this stage) unless you connect another node to that port. I dont want to go down the rabbit hole of setting all UV nodes (theres 8 of them) to override float but we can continue this discussion anyway.

com.unity.shadergraph/Editor/Data/Nodes/Procedural/Noise/GradientNoiseNode.cs
```diff
@@ -22,34 +22,33 @@ protected override MethodInfo GetFunctionToConvert()
             [Slot(1, Binding.None, 10, 10, 10, 10)] Vector1 Scale,
             [Slot(2, Binding.None)] out Vector1 Out)
         {
-            return "{ Out = unity_gradientNoise(UV * Scale) + 0.5; }";
+            return
+                @"
+{ 
+    $precision p = UV * Scale;
```
Kink3d: This should be `$precision2`

Kink3d:
All tests green except known failure `VFXGraphicsTests.21_DepthModules`.
Manual testing done by @zadi0648 and @spaztronaut. More is required but no mjaor issues found. This PR is required for 19.2 release (obligations) so I'm going to merge it, any issues found after the fact will be fixed over the next week.

PR #3603, opened 2019-05-10
[2019.2] Reduce the group size of the deferred lighting pass from 16x16 to 8x8

EvgeniiG:
### Purpose of this PR
Nvidia claims there is a large perf loss from using large group sizes on their HW.
On AMD, it may or may not be faster, and should not be slower.
Therefore, this is a large optimization for Nvidia and a no-op for AMD.

---
### Release Notes
Reduce the group size of the deferred lighting pass from 16x16 to 8x8

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fdeferred-group-size&automation-tools_branch=master&unity_branch=2019.2%2Fstaging

**Manual Tests**: We need to test this on the PS4. I can test on Nvidia.

---
### Comments to reviewers
Need help perf testing this one.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
@@ -122,18 +122,20 @@ CBUFFER_END
 RW_TEXTURE2D_X(float3, diffuseLightingUAV);
 RW_TEXTURE2D_X(float4, specularLightingUAV);
 
-
+#define GROUP_SIZE (TILE_SIZE_FPTL / 2) // 4x 8x8 groups per a 16x16 tile
 
 #ifdef USE_INDIRECT
 
 StructuredBuffer<uint> g_TileList;
 // Indirect
-[numthreads(TILE_SIZE_FPTL, TILE_SIZE_FPTL, 1)]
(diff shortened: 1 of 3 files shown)


sebastienlagarde:
Target master ? and changelog :)

EvgeniiG:
Let's test perf on a stable version first.

FrancescoC-unity:
My gut feeling is that this could even be slightly faster on AMD, but as with everything profiling will tell the story. I will try to run a perf test today at some point on both consoles.

EvgeniiG:
These are huge gains on the PS4, much higher than I expected. Thanks for testing!

sebastienlagarde:
So we need to update PR to master and run the test

EvgeniiG:
I changed the target branch to master. Clearly, the 2019.2 branch diverged from the master branch.
My recommendation would be to change the target branch back to master, and cherry pick this commit for master.

sebastienlagarde:
Indeed, weird I don't see reasons for it. Can you create a new PR tagerting master and cherry-pick the commit? Thanks

EvgeniiG:
Switched this PR back to 2019.2, created another PR for master:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3654

PR #3605, opened 2019-05-10
Color Mode Feature - color nodes by category or custom user colors

marctem:
### Purpose of this PR
Adds support for Color Modes. User can select their mode from a dropdown. Initial modes are:
None - no color (default)
Category - Colors are defined by node category (see image)
User Defined - User-selected color saved per-node (or selection)

image

Note that final colors are with UX team for approval.

Currently the UI only supports a single user-defined color per node, but the data is saved as a dictionary of ProviderTitle -> Color, allowing us in the future to add additional user colors (or allow users to create their own custom providers) without breaking the node save format.

Since the specific category colors are defined in USS, users can easily override them (see docs).
---
### Release Notes
Adds Color Modes, allowing users to view their nodes colored by category or choose their own custom node colors to make graphs more readable and pretty.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Fnode-color-by-category&unity_branch=trunk&automation-tools_branch=master

- [X] Checked new UI names with UX convention
- Other: Loaded and saved lots of graphs in user-defined as well as category modes.

QA Test Doc:
https://docs.google.com/document/d/1fnT1_2U5E_YcKoz0nU9h-vhwWrZsVRLUbpmZVK82tyA/edit#

---
### Overall Product Risks
**Technical Risk**: Low - mostly a UI feature, also saves custom color data to nodes and the current provider to the graph.

**Halo Effect**: Low - UI code and a little plumbing

---
### Comments to reviewers
Please test out feature per the docs and see if anything breaks! Note that UX is still finalizing the colors.


martint-unity:
1st: Border line is not as visible as a full header. Also the text in the header is now not centered vertically. I would prefer it to be the header as we tried out before. Just my 2 cents.

2nd: Have the colors been globally agreed upon? Are the other graphview projects on board with the color scheme we have chosen for the categories? 
"Note that UX is still finalizing the colors." I guess that answer this question. Just wanted to bring it up. And it should be agreed upon before it goes out in a package.
PLUS have this design of using only a border line been agreed upon by the other graphview teams?

3rd: Code looks good IMO :)

4rth: This looks not so good. Should probably be fixed.
image
Looks like the drop down is belonging to things on the right side. Text is cut off.

5th:
User can only change color per node and not category. I would prefer if the user can also change the category colors.

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (string.IsNullOrEmpty(activeColors))
+                activeColors = DefaultProvider;
+
+            foreach (var colorType in TypeCache.GetTypesDerivedFrom<IColorProvider>())
+            {
+                var provider = (IColorProvider) Activator.CreateInstance(colorType);
+                m_Providers.Add(provider);
+            }
+            
+            m_Providers.Sort((p1, p2) =>  string.Compare(p1.Title, p2.Title, StringComparison.InvariantCulture));
```
martint-unity: nit picky here but there are 2 spaces here :)

marctem:
> 1st: Border line is not as visible as a full header. I would prefer it to be the header as we tried out before. Just my 2 cents.
True, it's not as visible, especially when zoomed out. This was UX's preference, though, to avoid contrast issues with the title text - it lets us much brighter colors which made the artists here happy. Also, you can easily override the CSS to change it back to full header (set background-color instead of border-color).

> 1.5th: Also the text in the header is now not centered vertically.
It is centered when the color is off - I prefer it not to move when turning color on and I think it looks good kind of sitting on the color line. It's a matter of preference, but I did that deliberately with input from Alex.

> 2nd: Have the colors been globally agreed upon? Are the other graphview projects on board with the color scheme we have chosen for the categories?
> "Note that UX is still finalizing the colors." I guess that answer this question. Just wanted to bring it up. And it should be agreed upon before it goes out in a package.
We're working on aligning on the colors and border vs background - Seb from GV is on the PR.

> 4th: This looks not so good. Should probably be fixed.
> image
> Looks like the drop down is belonging to things on the right side. Text is cut off.
I agree! Help! I tried messing with the IMGUI there for half an hour and couldn't get it to look decent. Any ideas? I'm going to seek out other help today.

> 5th:
> User can only change color per node and not category. I would prefer if the user can also change the category colors.
For MVP, they can change them in the USS file, (this is documented). I think that will be good enough, but if enough teams want a UI for that we could build one.

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
```diff
@@ -0,0 +1,31 @@
...
+            var cat = title.title[0];
+            
+            if (string.IsNullOrEmpty(cat))
+                return true;
+            
+            el.AddToClassList(cat);
+            return true;
+        }
+    }
+}
```
poussinfrais: Missing eol

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public string Title => "Category";
+        public bool AllowCustom => false;
+
+        public Color? GetColor(AbstractMaterialNode node)
+        {
+            return null;
+        }
+
+        public bool ApplyColorTo(AbstractMaterialNode node, VisualElement el)
```
poussinfrais: Intent is not clear from name

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
```diff
@@ -0,0 +1,31 @@
...
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ShaderGraph.Drawing.Colors
+{
+    class CategoryColors : IColorProvider
+    {
+        public string Title => "Category";
+        public bool AllowCustom => false;
+
+        public Color? GetColor(AbstractMaterialNode node)
```
poussinfrais: Personal note: I am not a megafan of nullables...but in any case, I think you should check for null values when GetColor() is called (I think I saw a few calls that use the return value blindly)

com.unity.shadergraph/Editor/Drawing/Colors/IColorProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        // If your color must be set programatically, return it here.
+        // If your colors are in USS and set via classes, return null here.
+        Color? GetColor(AbstractMaterialNode node);
+        
+        // If your color is defined in USS and set via classes, set them on the element here and return true.
+        // If your color must be set programatically, return false here.
+        bool ApplyColorTo(AbstractMaterialNode node, VisualElement el);
+    }
+}
```
poussinfrais: Missing eol

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
```diff
@@ -0,0 +1,31 @@
...
+            return null;
+        }
+
+        public bool ApplyColorTo(AbstractMaterialNode node, VisualElement el)
+        {
+            if (!(node.GetType().GetCustomAttributes(typeof(TitleAttribute), false).FirstOrDefault() is TitleAttribute title))
+                return true;
+
+            var cat = title.title[0];
+
```
poussinfrais: Extra line :)

com.unity.shadergraph/Editor/Drawing/Colors/IColorProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+    // Implement this to provide colors based on whatever factor you want
+    interface IColorProvider
+    {
+        string Title { get; }
+        
+        bool AllowCustom { get; }
+
+        // If your color must be set programatically, return it here.
+        // If your colors are in USS and set via classes, return null here.
+        Color? GetColor(AbstractMaterialNode node);
```
poussinfrais: Have to admit function names do not make it clear for me :(

com.unity.shadergraph/Editor/Drawing/Colors/NoColors.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            return null;
+        }
+
+        public bool ApplyColorTo(AbstractMaterialNode node, VisualElement el)
+        {
+            return true;
+        }
+    }
+}
```
poussinfrais: Missing eol

com.unity.shadergraph/Editor/Drawing/Colors/NoColors.cs
```diff
@@ -0,0 +1,22 @@
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ShaderGraph.Drawing.Colors
+{
+    class NoColors : IColorProvider
+    {
+        public static string NoColorTitle = "<None>";
```
poussinfrais: Maybe NoColorTitle is not needed here, Title can return directly

com.unity.shadergraph/Editor/Drawing/Colors/UserColors.cs
```diff
@@ -0,0 +1,35 @@
...
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ShaderGraph.Drawing.Colors
+{
+    class UserColors : IColorProvider 
+    {
+        string m_Title = "User Defined";
+        public bool AllowCustom => true;
+
+        public UserColors() {}
```
poussinfrais: Redundant ctor :)

com.unity.shadergraph/Editor/Drawing/Colors/UserColors.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            return node.GetColor(m_Title);
+        }
+
+        public bool ApplyColorTo(AbstractMaterialNode node, VisualElement el)
+        {
+            return false;
+        }
+    }
+}
```
poussinfrais: ;)

com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
```diff
@@ -139,6 +148,15 @@ public GraphEditorView(EditorWindow editorWindow, GraphData graph, MessageManage
 
                     GUILayout.FlexibleSpace();
 
+                    var newIdx = EditorGUILayout.Popup("Color Mode:", m_ColorManager.activeIndex, colorProviders);
```
poussinfrais: This should be UIElements.  As a reference, we use UnityEditor.UIElements.ToolbarMenu in GraphView -- I suspect you would use the same since you're using the control in a toolbar.

poussinfrais: Plus you would fix the width problem via uss ;)

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
poussinfrais: Could you return StyleColor noColor that you define later in that PR instead of null?

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
```diff
@@ -0,0 +1,76 @@
...
+            set
+            {
+                if (value < 0 || value >= m_Providers.Count)
+                    return;
+                
+                m_ActiveIndex = value;
+            }
+        }
+
+        public ColorManager(string activeColors)
```
poussinfrais: activeProvider?

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace UnityEditor.ShaderGraph.Drawing.Colors
+{
+    [Serializable]
+    class SerializableUserColor
+    {
+        public string Key = String.Empty;
+        public Color Value = Color.black;
+
+        public SerializableUserColor() {  }
```
poussinfrais: Redundant

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
```diff
@@ -0,0 +1,63 @@
...
+        public SerializableUserColor(KeyValuePair<string, Color?> pair) { Key = pair.Key; Value = pair.Value ?? Color.black; }
+    }
+
+    [Serializable]
+    public class CustomColorData : ISerializationCallbackReceiver
+    {
+        [NonSerialized]
```
poussinfrais: Not sure if [NonSerialized] is needed here

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                var provider = (IColorProvider) Activator.CreateInstance(colorType);
+                m_Providers.Add(provider);
+            }
+            
+            m_Providers.Sort((p1, p2) =>  string.Compare(p1.Title, p2.Title, StringComparison.InvariantCulture));
+            activeIndex = m_Providers.FindIndex(provider => provider.Title == activeColors);
+        }
+
+        public void SetColor(IShaderNodeView nodeView)
```
poussinfrais: From the name of the function I was expecting having to provide a color

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            var curProvider = m_Providers[m_ActiveIndex];
+            nodeView.colorElement.ClearClassList();
```
poussinfrais: Should we clear the whole class list?  Won't it remove layout styles too?

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
```diff
@@ -0,0 +1,63 @@
...
+        Dictionary<string, Color?> m_CustomColors = new Dictionary<string, Color?>();
+        [SerializeField]
+        List<SerializationHelper.JSONSerializedElement> m_SerializableColors = new List<SerializationHelper.JSONSerializedElement>();
```
poussinfrais: For the GraphView PR, I would like to see an abstraction layer to separate the data & let the implementation be the project's responsibility (some might prefer something other than JSON for example)

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -223,6 +224,22 @@ public void ClearMessage()
...
+
+        static readonly StyleColor noColor = new StyleColor(StyleKeyword.Null);
+        public void SetColor(Color? color)
+        {
+            m_TitleContainer.style.borderColor = color ?? noColor;
+        }
+
+        public Color GetColor()
+        {
+            return m_TitleContainer.style.borderColor.value;
```
poussinfrais: It may be safer to return m_TitleContainer.resolvedStyle.borderColor

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -18,6 +19,10 @@ sealed class MaterialNodeView : Node, IShaderNodeView
     {
         PreviewRenderData m_PreviewRenderData;
         Image m_PreviewImage;
+        // Remove this after updated to the correct API call has landed in trunk. ------------
```
poussinfrais: Is this comment still true?

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
poussinfrais: Especially if the colorElement is the nodeView itself...

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -32,10 +37,10 @@ sealed class MaterialNodeView : Node, IShaderNodeView
 
         GraphView m_GraphView;
 
-
         public void Initialize(AbstractMaterialNode inNode, PreviewManager previewManager, IEdgeConnectorListener connectorListener, GraphView graphView)
         {
             styleSheets.Add(Resources.Load<StyleSheet>("Styles/MaterialNodeView"));
+            styleSheets.Add(Resources.Load<StyleSheet>($"Styles/{ColorManager.StyleFile}"));
```
poussinfrais: Not sure StyleFile should be living in ColorManager if only used in MaterialNodeView (especially since style file in question contains MaterialNodeView-only selectors)

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
marctem: Required by serialization system unfortunately. I'll comment it.

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
marctem: Definitely. This whole class is SG-specific; I'm still thinking about what makes sense from an API standpoint for the data.

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
marctem: Yeah, that's nicer.

com.unity.shadergraph/Editor/Drawing/Colors/NoColors.cs
marctem: It's used explicitly in ColorManager since it's the default provider. I don't love it; do you have a better idea?

com.unity.shadergraph/Editor/Drawing/Colors/UserColors.cs
marctem: The reflection-based instantiation in ColorManager requires an empty constructor. The other ctor (and ChangeTitle) is actually unnecessary future-proofing against allowing users to rename their user color modes - I'll remove them for now.

com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
marctem: It should, but that's a different PR. =) UIE.ToolbarMenu didn't exist when we made this. Will fix the width issue though.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
marctem: I don't know, I just copy/pasted these 3 lines from elsewhere in the file. I didn't bother to figure that out, probably not necessary anymore.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
marctem: This is a definite weak spot in the design. I had it hard-coded here initially, but that seemed not quite right. Do you have a better suggestion?

com.unity.shadergraph/Editor/Drawing/Colors/NoColors.cs
marctem: I blame Rider for all of these. 😆 It says "Always put a newline at the end of file" in our .editorconfig, but when it auto-generated these files for me it obviously didn't do it!

com.unity.shadergraph/Editor/Drawing/Colors/IColorProvider.cs
marctem: I'll try to come up with better names. I don't love having two different functions but with two different approaches for applying color (class names vs values stored in code/data), we need to handle both.

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
marctem: Good catch. 😬  I'll move the clearing back to the individual providers and only clear the specific classes they set.

com.unity.shadergraph/Editor/Drawing/Colors/IColorProvider.cs
poussinfrais: Gotcha

com.unity.shadergraph/Editor/Drawing/Colors/CategoryColors.cs
marctem: Changed to "ApplyClassToElement".

com.unity.shadergraph/Editor/Drawing/Colors/NoColors.cs
poussinfrais: NoColors.Title would work, no?

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
marctem: Changed to `UpdateNodeView`

com.unity.shadergraph/Editor/Drawing/Colors/IColorProvider.cs
marctem: Changed to `ProvideColorForNode`

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            get => curProvider.GetTitle();
+        }
+
+        public bool activeSupportsCustom
+        {
+            get => curProvider.AllowCustom();
+        }
+
+        IColorProvider curProvider => m_Providers[activeIndex];
```
poussinfrais: Suggestion: currentProvider

com.unity.shadergraph/Editor/Drawing/Colors/CustomColorData.cs
marctem: True, will remove.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
marctem: How does that work? borderColor is not a Color, why would I return the StyleColor?

com.unity.shadergraph/Editor/Drawing/Colors/ColorManager.cs
marctem: Yep

com.unity.shadergraph/Editor/Drawing/Colors/UserColors.cs
```diff
@@ -0,0 +1,19 @@
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ShaderGraph.Drawing.Colors
+{
+    class UserColors : ColorProviderFromCode 
+    {
+        string m_Title = "User Defined";
```
poussinfrais: This can be const (const string k_Title = ...)

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
poussinfrais: I would be tempted to remove ColorManager.StyleFile, and instead add the ColorMode stylesheet here directly.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
poussinfrais: Actually it is (Note in my comment: m_TitleContainer.*resolved*Style.borderColor)

marctem: Ah, whoops, missed the "resolved" part. Done.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
marctem: Moved it back here as suggested.

com.unity.shadergraph/Editor/Drawing/Colors/UserColors.cs
marctem: The idea was that this would not be const, the user can create multiple UserColors providers with different names. But not doing that yet, so will remove.

marctem:
Katana is green except for 1 VFX test that's also broken in master. Node tests passed on Yamato on Windows, so this PR is good to go!

PR #3630, opened 2019-05-15
Remove post process condition on skipping pyramid since bloom runs its own pyramid

FrancescoC-unity:
All in the title really :-)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -3133,8 +3133,8 @@ namespace UnityEngine.Experimental.Rendering.HDPipeline
             }
             else
             {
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -3133,8 +3133,8 @@ void RenderColorPyramid(HDCamera hdCamera, CommandBuffer cmd, bool isPreRefracti
             }
             else
             {
-                // TODO: This final Gaussian pyramid can be reuse by Bloom and SSR in the future, so disable it only if there is no postprocess AND no distortion
-                if (!hdCamera.frameSettings.IsEnabled(FrameSettingsField.Distortion) && !hdCamera.frameSettings.IsEnabled(FrameSettingsField.Postprocess) && !hdCamera.frameSettings.IsEnabled(FrameSettingsField.SSR))
+                // TODO: This final Gaussian pyramid can be reused by SSR in the future, so disable it only if there is no distortion
```
FrancescoC-unity: Actually the TODO here is not needed

PR #3643, opened 2019-05-15
Fix volumetric rendering with single-pass instancing

fabien-unity:
### Purpose of this PR
- handle camera-relative
- avoid volume linear interpolation between XR views by adding manual clamp
- fix _VBufferCoordToViewDirWS for XR

---
### Testing status
**Katana Tests**: [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-fog-spi&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VBuffer.hlsl
```diff
@@ -40,8 +40,15 @@ float4 SampleVBuffer(TEXTURE3D_PARAM(VBuffer, clampSampler),
...
-        // XRTODO: figure out a better way to avoid leaks between eyes (double-wide might be safer, or guard bands, or 2 separate textures)
+        // With XR single-pass instancing, one 3D buffer is used to store all views (split along w)
         w = (w + unity_StereoEyeIndex) * _VBufferRcpInstancedViewCount;
+
+        // Manually clamp w with a safe limit to avoid linear interpolation from the others views
+        float limitSliceRange = 0.5f * _VBufferRcpSliceCount;
+        float lowerSliceRange = (unity_StereoEyeIndex + 0) * _VBufferRcpInstancedViewCount;
+        float upperSliceRange = (unity_StereoEyeIndex + 1) * _VBufferRcpInstancedViewCount;
+
+        w = clamp(w, lowerSliceRange + limitSliceRange, upperSliceRange - limitSliceRange);
```
EvgeniiG: Does this handle the half-texel offset?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -695,5 +691,7 @@ void VolumetricLighting(uint3 dispatchThreadId : SV_DispatchThreadID,
     context.shadowContext = InitShadowContext();
     uint featureFlags     = 0xFFFFFFFF;
 
+    ApplyCameraRelativeXR(ray.originWS);
```
EvgeniiG: How is this different than non-XR? You're not using the head position as the origin?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VBuffer.hlsl
fabien-unity: Yes, I was able to demonstrate the issue on a test scene and this code fixes it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -605,6 +604,21 @@ void UpdateVolumeParameters()
                 volumeAnchor = camera.transform;
         }
 
+        public void GetPixelCoordToViewDirWS(Vector4 resolution, ref Matrix4x4[] transforms)
+        {
+            if (xr.instancingEnabled)
+            {
+                for (int viewIndex = 0; viewIndex < viewCount; ++viewIndex)
```
EvgeniiG: Maybe it's more clean to set view count to 0 if XR is off?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
fabien-unity: The CPU does the lights setup and culling from the original camera position only once, but the shaders has to account for the offset between the original camera position and the actual position of the 'eye.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
fabien-unity: A lot of code expects viewCount to be 1 when XR is off. I think we should merge mainViewConstants and xrViewConstants into one array at some points to simplify some logic.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
EvgeniiG: My point is that perhaps the shaders should use the matrices computed using the head (rather than the eye) as the camera-relative origin.

EvgeniiG: If you think doing it this way is simpler then that's fine, too.

sebastienlagarde: We have discuss that for other topics as well (using one eye or the center), don't really matter as long as it is the same for all processing and choosing an eye is simpler (no need for specific calculation).

PR #3650, opened 2019-05-16
Fixing syntax in yamato yml files

sophiaaar:
### Purpose of this PR
Liquid template language needs 'and' instead of '&&'.
Previously, I had an if statement like `{% if testplatform.name == "editmode" && platform.name == "Windows64" %}` but because the `&&` was being ignored, OSX editmode pipelines were incorrect and using the wrong VM type.

---
### Testing status
Tested in master

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers
Backport of #3648


sophiaaar:
accidentally added docs - no changelogs have changed

PR #3652, opened 2019-05-16
Ray Tracing Errors and AO To volume system

anisunity:
Correcting Raytracing Errors due to depth beging a texture array
Moving Ambient occlusion parameters to the volume system
Adding the tier option in the HDRP Asset

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FRaytracingVolumeErrors&automation-tools_branch=master&unity_branch=trunk


com.unity.render-pipelines.core/Runtime/Volume/VolumeParameter.cs
```diff
@@ -169,6 +169,13 @@ public BoolParameter(bool value, bool overrideState = false)
             : base(value, overrideState) {}
     }
 
+    [Serializable, DebuggerDisplay(k_DebuggerDisplay)]
+    public class LayerMaskParameter : VolumeParameter<LayerMask>
```
Chman: This isn't used anywhere in the PR, is it intended or is the PR missing a file?

anisunity: It was initially using it and it will be used later, but yeah for the moment it is not. Should I remove it?

Chman: I'd keep it.

PR #3654, opened 2019-05-16
[master] Reduce the group size of the deferred lighting pass from 16x16 to 8x8

EvgeniiG:
### Purpose of this PR
Nvidia claims there is a large perf loss from using large group sizes on their HW.
On AMD, it may or may not be faster, and should not be slower.
Therefore, this is a large optimization for Nvidia and a no-op for AMD.

---
### Release Notes
Reduce the group size of the deferred lighting pass from 16x16 to 8x8

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fdeferred-group-size-master&automation-tools_branch=master&unity_branch=trunk
**Manual Tests**: We need to test this on the PS4. I can test on Nvidia.

---
### Comments to reviewers
Need help perf testing this one.


sebastienlagarde:
Missing changelog - I added it

PR #3656, opened 2019-05-17
Ray traced reflections to volume system

anisunity:
Indenting parameters of raytraced ambient occlusion
Moving reflection parameters from raytracing environement to volume system
Supporting the reflect sky parameter in raytraced reflections
Splitting raytraced reflections in tiers

This is based on an other PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3652


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
```diff
@@ -27,6 +27,12 @@ public enum SupportedLitShaderMode
             Both = ForwardOnly | DeferredOnly
         }
 
+        public enum RaytracingTier
+        {
+            Tier1 = 1 << 0,
```
sebastienlagarde: Let's try to find a better name in future PR :)

PR #3660, opened 2019-05-17
Fix volumetric reprojection with XR

fabien-unity:
### Purpose of this PR
- reprojection for volumetrics is now working with XR single-pass instancing
- update direct matrix references with macros so that it's fully compatible with XR
- clear of few XRTODOs

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-vol-reproj&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed an issue with history buffers causing effects like TAA or auto exposure to flicker when more than one camera was visible in the editor
 - The correct preview is displayed when selecting multiple `PlanarReflectionProbe`s
-- Fixed volumetric rendering with camera-relative code and XR stereo instancing
+- Fixed volumetric rendering and reprojection with camera-relative code and XR stereo instancing
```
sebastienlagarde: Please do a separate changelog as in 2019.2 it is separate version now

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
```diff
@@ -542,6 +542,11 @@ void FillVolumetricLightingBuffer(LightLoopContext context, uint featureFlags,
         float4 normalizedVoxelValue = linearizedVoxelValue * rcp(dt);
         float4 normalizedBlendValue = normalizedVoxelValue;
 
+    #if (SHADEROPTIONS_CAMERA_RELATIVE_RENDERING != 0) && defined(USING_STEREO_MATRICES)
+        // With XR single-pass instancing, remove the camera-relative offset for the reprojected sample
+        centerWS -= _WorldSpaceCameraPosViewOffset;
```
EvgeniiG: Why do this per sample? That's going to be slow.

EvgeniiG: You need to fix the ray origin rather than this.

fabien-unity: The ray origin is used for the actual sampling (current frame) but in this case, we need to only offset for the previous frame to get correct results.

EvgeniiG: Ah, right, I looked at the wrong place in code.

I still don't think this is the right fix. Look at how variables are used below:

centerWS should be cam relative w.r.t. the cam pos during the current frame.
_PrevCamPosRWS should be cam relative w.r.t. the cam pos during the current frame.
UNITY_MATRIX_PREV_VP should be cam relative w.r.t. the cam pos during the current frame.

So, first, you either don't support camera relative rendering for any of these, which is bad, or you only change one of the three variables, which is wrong.

I don't really understand how this can work?

sebastienlagarde:
@fabien-unity any update?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.compute
fabien-unity: I agree this is not very clear.
First observation: it looks good when I try it out. Without the offset, the reprojected coordinates are off and the result is not converging and create a "trail".

The explanation:
SampleVBuffer() has 2 versions: one takes positionNDC as input and the other (used by volumetric reproj) takes positionWS and viewProjMatrix to compute positionNDC.

With XR single-pass instancing, each view/eye use camera-relative so that in effect, camera position is float3(0) for *each* views. When using SampleVBuffer() to transform from positionWS to positionNDC, we have to make sure positionWS is in the same space.

At the start of kernel VolumetricLighting, we do : ApplyCameraRelativeXR(ray.originWS);
The goal is to convert the ray from "eye space" to "center/cyclops space" to match the density volume information that was computed on the CPU (camera-relative code).
Later in the code for reprojection, centerWS is derived from ray.originWS so it's still in the same space.
But in order to compute the correct positionNDC for reprojection, we need to convert centerWS  back to the initial view/eye space.

Another way to do this would be to keep track of both version of ray.originWS (one in eye space and the other in cyclops space) but it would be more expensive and intrusive.

I hope this makes more sense :)

PR #3661, opened 2019-05-18
fix for Distortion effect

macagu:
### Purpose of this PR
The distortion effect was not working with unlit shaders. Its very useful on action games to make a classic shockwave effect with a simple quad that must not be affected by light.

---
### Release Notes
Fixed distortion effect on particle unlit shader.

---
### Testing status

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


phi-lira:
Sounds good to me. Added @Verasl and we need to add changelog entry.

phi-lira:
Will merge to working branch so I run tests and add changelog.

PR #3662, opened 2019-05-20
Fix LWRP GC spikes

phi-lira:
### Purpose of this PR
 - Removes code causing heap allocations every frame in LWRP. 
 - There's still an issue with the TryGetComponent causing mem allocs when component is not present in game object. A fix is about to land in trunk. 

---
### Release Notes
  Fixed Garbage Collections spikes caused by heap allocations every frame.

---
### Testing status

- [x] Built a player
- Other: 

**Automated Tests**: Did not added.

Any test projects to go with this to help reviewers? You can use LW RP template to test.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lw/fix-mem-allocations&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: Low. Changes core render pass execution but it should be safe.

**Halo Effect**: Medium. Changes core bits of LWRP. 

---
### Comments to reviewers
 The execute block is not doing the smartest thing. I've started work to compute block range indices so when executing each block it will know the start/end indices to execute. 

 I need more testing to push this work and I'll have to switch to another task now. So I've stashed my changes and will push in another PR.


com.unity.render-pipelines.lightweight/Runtime/RenderingUtils.cs
```diff
@@ -146,5 +146,16 @@ internal static void RenderObjectsWithError(ScriptableRenderContext context, ref
...
             context.DrawRenderers(cullResults, ref errorSettings, ref filterSettings);
         }
+
+        // Caches render texture format support. SystemInfo.SupportsRenderTextureFormat allocates memory due to boxing.
+        static Dictionary<RenderTextureFormat, bool> m_RenderTextureFormatSupport = new Dictionary<RenderTextureFormat, bool>();
+
+        internal static bool SupportsRenderTextureFormat(RenderTextureFormat format)
+        {
+            if (!m_RenderTextureFormatSupport.ContainsKey(format))
+                m_RenderTextureFormatSupport.Add(format, SystemInfo.SupportsRenderTextureFormat(format));
```
pbbastian: Are we completely sure that `SystemInfo.SupportsRenderTextureFormat(RenderTextureFormat)` never changes?

phi-lira: Since graphics emulation is not supported in SRP, I guess the only case it could change is when switching graphics API in editor, which iirc it will restart the editor anyway. 

Just for safety, I'm clearing the cache everytime the render pipeline is rebuilt. This should be safe enough imo.

pbbastian: It is possible to change graphics device in the Editor as well. I wonder if that is possible in the Player.

phi-lira:
Cached render passes in block ranges. Also fixed some issues.

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -69,6 +69,22 @@ protected List<ScriptableRenderPass> activeRenderPassQueue
             get => m_ActiveRenderPassQueue;
         }
 
+        static class RenderPassBlock
+        {
+            // Executes render passes that are inputs to the main rendering
+            // but don't depend on camera state. They all render in monoscopic mode. f.ex, shadow maps.
+            public static readonly int BeforeRendering = 0;
```
pbbastian:
```suggestion
            public const int beforeRendering = 0;
```

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -69,6 +69,22 @@ protected List<ScriptableRenderPass> activeRenderPassQueue
...
+
+            // Main bulk of render pass execution. They required camera state to be properly set
+            // and when enabled they will render in stereo.
+            public static readonly int MainRendering = 1;
```
pbbastian:
```suggestion
            public const int mainRendering = 1;
```

com.unity.render-pipelines.lightweight/Runtime/ScriptableRenderer.cs
```diff
@@ -69,6 +69,22 @@ protected List<ScriptableRenderPass> activeRenderPassQueue
...
+
+            // Execute after Post-processing.
+            public static readonly int AfterRendering = 2;
```
pbbastian:
```suggestion
            public const int afterRendering = 2;
```

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -99,6 +99,8 @@ public LightweightRenderPipeline(LightweightRenderPipelineAsset asset)
             Lightmapping.SetDelegate(lightsDelegate);
 
             CameraCaptureBridge.enabled = true;
+
+            RenderingUtils.ClearSystemInfoCache();
```
pbbastian: Is the RP re-created when the active device changes? Otherwise, this doesn't fix the potential issue. I think it might be okay for now, given that the underlying issue with `SystemInfo` is fixed soon such that `RenderingUtils` can be removed.

com.unity.render-pipelines.lightweight/Runtime/RenderingUtils.cs
```diff
@@ -146,5 +146,21 @@ internal static void RenderObjectsWithError(ScriptableRenderContext context, ref
 
             context.DrawRenderers(cullResults, ref errorSettings, ref filterSettings);
         }
+
+        // Caches render texture format support. SystemInfo.SupportsRenderTextureFormat allocates memory due to boxing.
+        static Dictionary<RenderTextureFormat, bool> m_RenderTextureFormatSupport = new Dictionary<RenderTextureFormat, bool>();
```
pbbastian:
```suggestion
        static readonly Dictionary<RenderTextureFormat, bool> s_RenderTextureFormatSupport = new Dictionary<RenderTextureFormat, bool>();
```

com.unity.render-pipelines.lightweight/Runtime/RenderingUtils.cs
```diff
@@ -146,5 +146,21 @@ internal static void RenderObjectsWithError(ScriptableRenderContext context, ref
...
+        {
+            m_RenderTextureFormatSupport.Clear();
+        }
+
+        internal static bool SupportsRenderTextureFormat(RenderTextureFormat format)
+        {
+            if (!m_RenderTextureFormatSupport.ContainsKey(format))
+                m_RenderTextureFormatSupport.Add(format, SystemInfo.SupportsRenderTextureFormat(format));
+
+            return m_RenderTextureFormatSupport[format];
```
pbbastian: Suggestion that only does 1 dictionary lookup when the key exists:
```
if (!s_RenderTextureFormatSupport.TryGetValue(format, out var isSupported))
{
    isSupported = s_RenderTextureFormatSupport[isSupported] = SystemInfo.SupportsRenderTextureFormat(format);
}
return isSupported;
```

phi-lira: @pbbastian Nice. Can we do this fix in a following PR?

pbbastian: It's only a suggestion, so I'll leave that up to you :)

phi-lira: ah it's because github showed me you requested changes. Ok. I'll fix in a following PR.

PR #3665, opened 2019-05-20
Hdrp/removed unused shadow data

sebastienlagarde:
Removed unused shadow data code

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fremoved-unused-shadow-data&automation-tools_branch=master&unity_branch=trunk&sort=1-asc


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/AdditionalShadowData.cs
```diff
@@ -178,8 +98,6 @@ void OnEnable()
         {
             m_ShadowAlgorithm = serializedObject.FindProperty("shadowAlgorithm");
             m_ShadowVariant = serializedObject.FindProperty("shadowVariant");
-            m_ShadowData = serializedObject.FindProperty("shadowData");
```
alelievr: I think this:
```        
UnityEditor.SerializedProperty m_ShadowData;
UnityEditor.SerializedProperty m_ShadowDatas;	```
can be deleted too if we remove the property initialization

PR #3703, opened 2019-05-23
[WIP] Refactor the shading framework (bis)

sebastienlagarde:
### Purpose of this PR

This PR is a replacement for https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3250
The previous PR was hard to merge due to several change in both HDRP test and the code based + addition of DXR fail. Starting from fresh HDRP/Staging instead.

Refactor the shading framework to make it more easy to support arbitrary materials. Most of the non-material-specific code is now shared. The new CBSDF function which replaces the BSDF function is more flexible and allows to fully define the behavior of up to 4 types of lobes (more lobes can be grouped by type). The resulting code is efficient (automatic early-outs), simple and flexible.

Note: This PR is a draft and don't contain the change for StackLit 

Note2: I guess we have change the behavior of transmittance with directional is different, need to check that (i.e when there is transmission and no shadow with dir light)

Change in this PR:
- Introduce CBSDF, EvaluateBSDF, IsNonZeroBSDF to replace BSDF functions
- Update GetBSDFAngles parameters
- Update LightEvaluatoin and SurfaceShading functions, introduce define to enable/disable features of it
- Replace ComputeMicroShadowing by a function GetAmbientOcclusionForMicroShadowing() 
- Add shadow tint parameter on light
- Remove SunDisk and Sun Halo size
- Rename WorldToTangent to TangentToWorld as it was incorrectly named

---
### Release Notes
Refactor the shading framework to make it more easy to support arbitrary materials.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fshading-framework-refactor-bis&automation-tools_branch=master&unity_branch=trunk#
**Manual Tests**: WIP

What did you do?
- [ ] Built a player
- [X ] C# and shader warnings (supress shader cache to see them)
- Other: 

**Automated Tests**: None.

Any test projects to go with this to help reviewers? No.

---
### Overall Product Risks
**Technical Risk**: High.

**Halo Effect**: High.

---
### Comments to reviewers

Compare to PR  https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3250
- I use the name CBSDF instead of CxBSDF for consistency with all other naming (BSDFData, PostEvaluateBSDF, GetBSDFAngles etc...).
- I use IsNonZeroBSDF instead of IsNonZeroCBSDF
- I have replace ComputeMicroShadowing by a function GetAmbientOcclusionForMicroShadowing() that return ambient occlusion of BSDF data
- I fixed little mistake found here and there (like not the correct roughness clamp for hair, wrong name for diffuse variable for lambert in lit etc...)
- I added simple Lit support and stacklit
- I use define to avoid to have to add transmittance to AxF

@slunity I have done change for stacklit, you can update it if there is mistake. I think there is opportunity to clean some code here (code is quite complex currently), but I have try to keep same behavior.


com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
```diff
@@ -400,29 +411,40 @@ void BSDF(  float3 V, float3 L, float NdotL, float3 positionWS, PreLightData pre
...
+    // We don't multiply by 'bsdfData.diffuseColor' here. It's done only once in PostEvaluateBSDF().
+    return cbsdf;
 }
 
 //-----------------------------------------------------------------------------
 // Surface shading (all light types) below
 //-----------------------------------------------------------------------------
 
-// Here we need to make sure
-#define USE_DIFFUSE_LAMBERT_BRDF
```
EvgeniiG: ???

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.cs
```diff
@@ -106,8 +106,6 @@ public struct BSDFData
             // Anisotropic
             [SurfaceDataAttributes("", true)]
             public Vector3 hairStrandDirectionWS;
-            public float roughnessT;
-            public float roughnessB;
```
EvgeniiG: Expected?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -27,11 +27,6 @@
 // #define LIT_DISPLAY_REFERENCE_IBL
 #endif
 
-// In forward we can chose between reading the normal from the normalBufferTexture or computing it again
-// This is tradeoff between performance and quality. As we store the normal conpressed, recomputing again is higher quality.
-// Uncomment this to get speed (to measure), let it comment to get quality
-// #define FORWARD_MATERIAL_READ_FROM_WRITTEN_NORMAL_BUFFER
```
EvgeniiG: Why was this removed?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -427,13 +416,6 @@ BSDFData ConvertSurfaceDataToBSDFData(uint2 positionSS, SurfaceData surfaceData)
     bsdfData.geomNormalWS        = surfaceData.geomNormalWS;
     bsdfData.perceptualRoughness = PerceptualSmoothnessToPerceptualRoughness(surfaceData.perceptualSmoothness);
 
-    // Check if we read value of normal and roughness from buffer. This is a tradeoff
-#ifdef FORWARD_MATERIAL_READ_FROM_WRITTEN_NORMAL_BUFFER
-#if (SHADERPASS == SHADERPASS_FORWARD) && !defined(_SURFACE_TYPE_TRANSPARENT)
-    UpdateSurfaceDataFromNormalData(positionSS, bsdfData);
-#endif
-#endif
```
EvgeniiG: Why was this removed?

sebastienlagarde:
adding anis and emmanuel for the raytraceAreashadow.compute change. Note: This file need to be clean, we should use normalData as we have the GBuffer and BSDFdata available (in another PR)

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: We have never tested it and no users will be aware of it, so better to remove it and add it again in the future if someone want to try it or think it can be useful.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
sebastienlagarde: See comment above

com.unity.render-pipelines.high-definition/Runtime/Material/Hair/Hair.cs
sebastienlagarde: Yes, those variables aren't used with hair as we use perceptualRoughness and secondaryPerceptualRoughness instead

com.unity.render-pipelines.high-definition/Runtime/Material/Fabric/Fabric.hlsl
sebastienlagarde: With your change we define transmission inside the EvaluteBSDF function. there is no need anymore for a define USE_DIFFUSE_LAMBERT_BRDF when we include lightevaluation.hlsl and surfaceshading.hlsl . Before it was used to force lambert for transmission in those files.

EvgeniiG:
I started looking at the updated code...
Should we move transmittance to preLightData instead? This would remove the need for MATERIAL_INCLUDE_TRANSMISSION in most cases, I think.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,126 @@
...
+            #ifdef MATERIAL_INCLUDE_TRANSMISSION
+            // If we have transmission and transmittance is non zero here it mean we are thin transmission
+            bool isThinTransmission = Max3(bsdfData.transmittance.r, bsdfData.transmittance.g, bsdfData.transmittance.b) > 0.0;
+            #else
+            bool isThinTransmission = false;
+            #endif
+
+            float shadow = 1.0;
+            // For thin transmission we don't want micro shadow, contact shadow or baked shadows.
+            if (isThinTransmission)
```
EvgeniiG: Why only for thin? Do we need to do this at all?..

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,126 @@
...
+            {
+                // This will disable contact shadow and shadow mask
+                light.contactShadowMask = 0;
+                light.shadowMaskSelector.x = -1;
+            }
+            else
+            {
+                shadow = ComputeMicroShadowing(GetAmbientOcclusionForMicroShadowing(bsdfData), abs(dot(bsdfData.normalWS, L)), _MicroShadowOpacity);
```
EvgeniiG: Micro shadowing on transmission?

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,126 @@
...
+            }
+            // This code works for both surface reflection and thin object transmission.
+            shadow *= EvaluateShadow_Directional(lightLoopContext, posInput, light, builtinData, GetNormalForShadowBias(bsdfData));
+            
+            lightColor.rgb *= ComputeShadowColor(shadow, light.shadowTint);
```
EvgeniiG: All of this code should be in shadow evaluation...

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -138,147 +129,109 @@ DirectLighting ShadeSurface_Directional(LightLoopContext lightLoopContext,
...
 
 #ifdef DEBUG_DISPLAY
-    if (_DebugLightingMode == DEBUGLIGHTINGMODE_LUX_METER)
-    {
-        // Only lighting, not BSDF
-        lighting.diffuse = color * attenuation * saturate(NdotL);
-    }
+            // The step with the attenuation is required to avoid seeing the screen tiles at the end of lights because the attenuation always falls to 0 before the tile ends.
+            // Note: g_DebugShadowAttenuation have been setup in EvaluateShadow_Punctual
+            if (_DebugShadowMapMode == SHADOWMAPDEBUGMODE_SINGLE_SHADOW && light.shadowIndex == _DebugSingleShadowIndex)
```
EvgeniiG: All of this stuff must be in the shadow code.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -49,15 +49,22 @@ void ApplyDebug(LightLoopContext lightLoopContext, PositionInputs posInput, BSDF
...
                     DirectionalLightData light = _DirectionalLightDatas[_DirectionalShadowIndex];
-                    float3 shadowBiasNormal = GetNormalForShadowBias(bsdfData);
-                    shadow = EvaluateRuntimeSunShadow(lightLoopContext, posInput, light, shadowBiasNormal);
+
+                    float3 L = -light.forward;
+                    shadow = GetDirectionalShadowAttenuation(context.shadowContext,
+                                                             posInput.positionSS, posInput.positionWS, GetNormalForShadowBias(bsdfData),
+                                                             light.shadowIndex, L);
+#else
+                    shadow = GetScreenSpaceShadow(posInput);
```
EvgeniiG: Should merge these.

com.unity.render-pipelines.core/ShaderLibrary/CommonLighting.hlsl
```diff
@@ -303,6 +303,10 @@ real ComputeMicroShadowing(real AO, real NdotL, real opacity)
 	return lerp(1.0, microshadow, opacity);
 }
 
+real3 ComputeShadowColor(real shadow, real3 shadowTint)
+{
+    return float3(1.0, 1.0, 1.0) - ((1.0 - shadow) * (float3(1.0, 1.0, 1.0) - shadowTint));
```
EvgeniiG: ???

com.unity.render-pipelines.core/ShaderLibrary/BSDF.hlsl
```diff
@@ -372,7 +386,7 @@ real Lambert()
 real DisneyDiffuseNoPI(real NdotV, real NdotL, real LdotV, real perceptualRoughness)
 {
     // (2 * LdotH * LdotH) = 1 + LdotV
-    // real fd90 = 0.5 + 2 * LdotH * LdotH * perceptualRoughness;
+    // real fd90 = 0.5 + (2 * LdotH * LdotH) * perceptualRoughness;
```
anisunity: 😄

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -1,135 +1,126 @@
...
-            float intensity = attenuation * wrapNdotL;
-
-            // We use diffuse lighting for accumulation since it is going to be blurred during the SSS pass.
-            // Note: Disney's LdoV term in 'diffuseBsdf' does not hold a meaningful value
-            // in the context of transmission, but we keep it unaltered for performance reasons.
-            lighting.diffuse  = transmittance * (diffuseBsdf * (intensity * light.diffuseDimmer));
-            lighting.specular = 0; // No spec trans, the compiler should optimize
+            #ifdef MATERIAL_INCLUDE_TRANSMISSION
+            // If we have transmission and transmittance is non zero here it mean we are thin transmission
+            bool isThinTransmission = Max3(bsdfData.transmittance.r, bsdfData.transmittance.g, bsdfData.transmittance.b) > 0.0;
```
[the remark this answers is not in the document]
sebastienlagarde: Why is it wrong? 

Here is some extra info.
In the material feature we don't have the "generic" information that we are thin transmission, we only have MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THICK_OBJECT.
Otherwise it is MATERIAL_LIT_TRANSMISSION or MATERIAL_STACKLIT_TRANSMISSION or other that we aren't allowed to access. Before the refactor this was tested with HaveMaterialTransmission, but we have removed it in favor of cleaner code.

We can't use a "MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THIN_OBJECT" flags in addition to MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THICK_OBJECT because it mean that the neutral value isn't 0 anymore. We need to have neutral value when we don't go through the code of FillTransmission to be 0. This is a problem we have already met when we have done the change of THICK to THIN. 

So here is the logic:
If we don't have the flag MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THICK_OBJECT , then we have either thin transmission or no transmission.

transmittance is init to 0 by default unless you go through FillMaterialTransmission which is the only place where we put a value in transmittance.
If call to FillMaterialTransmission give a value of transmittance 0, then it doens't matter that we have isThinTransmission to false as it have no effect.

Thus why I do this test above. 
Thought?

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: thick transmission take the other path where none of the shadow code is call, and rely only on closest shadow.
And yes, we want to avoid contact shadow, it was a request from Torbjorn on book of the dead as contact shadow was significantly affecting foliage.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
sebastienlagarde: Screenspace shadow will change in another PR from Anis, so I don't touch it

com.unity.render-pipelines.core/ShaderLibrary/CommonLighting.hlsl
sebastienlagarde: In this PR I have integrated a request from Film verticals, they wanted to have tint shadow. This is what it is used for, color shadow. (Note: True that I should do a separate PR but I was a bit lazy to redo all the test).

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: Maybe. But we need to do the step here otherwise the display is tiled. Shadow code can't handle that. Maybe in future we need to revisit it.

sebastienlagarde:
Update about the status of this PR:
- One VFX test simpleLit is failing - seems that the test was already wrong before :)
- Hair test 1501 is failing 
- ShaderGraph transmission fail on 8201 opaque.

So 3 test left. Looking at it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: Well, you you don't want contact shadows on both thick AND thin transmission (read: any transmission).

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: OK.

com.unity.render-pipelines.core/ShaderLibrary/CommonLighting.hlsl
EvgeniiG: OK.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: OK, for this one, fair enough, I agree. For the one above, it should be in the shadow code.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
EvgeniiG: OK.

com.unity.render-pipelines.core/ShaderLibrary/CommonLighting.hlsl
EvgeniiG: OK.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: if (isThinTransmission) { disable } 
else { shadow = ComputeMicroShadowing }
???

EvgeniiG: Note that transmission is only active when NdotL is negative. Microshadowing could easily detect that.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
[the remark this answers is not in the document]
EvgeniiG: bsdfData.transmittance is non-zero whenever you have any transmittance, thin or thick.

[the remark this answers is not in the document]
sebastienlagarde: but we are in the else of if (ShouldEvaluateThickObjectTransmission(V, L, preLightData, bsdfData, light.shadowIndex))
or you assume that if (ShouldEvaluateThickObjectTransmission(V, L, preLightData, bsdfData, light.shadowIndex)) could return false even if MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THICK_OBJECT is setup ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: for thick case we are in another branch (see code: MATERIALFEATUREFLAGS_TRANSMISSION_MODE_THICK_OBJECT ), so here it is only about thin transmisison. But yes, we are agree :). I can move the comment earlier and make it more general

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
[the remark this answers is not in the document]
EvgeniiG: Maybe the variable name is bad.
If you need a transmission flag, just add one, call it MATERIALFEATUREFLAGS_TRANSMISSION, and make it material-agnostic.

sebastienlagarde:
>Should we move transmittance to preLightData instead? This would remove the need for >MATERIAL_INCLUDE_TRANSMISSION in most cases, I think.

Don't think have the same issue? it still mean we need to add transmittance to all prelightData that require it and then do #ifdef MATERIAL_INCLUDE_TRANSMISSION  around the code anyway in surface shading?

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: I don't like the fact that material is configuring the light at all.
It's probably a better idea for the lighting code to check the sign of the NdotL.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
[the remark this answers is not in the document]
EvgeniiG: Okay, I can see why it would work (couldn't see from the diff), but the code is now too complicated.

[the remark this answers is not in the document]
EvgeniiG: else
        {
            #ifdef MATERIAL_INCLUDE_TRANSMISSION
            // If we have transmission and transmittance is non zero here it mean we are thin transmission
            bool isThinTransmission = Max3(bsdfData.transmittance.r, bsdfData.transmittance.g, bsdfData.transmittance.b) > 0.0;
            #else
            bool isThinTransmission = false;
            #endif

            float shadow = 1.0;
            // For thin transmission we don't want micro shadow, contact shadow or baked shadows.
            if (isThinTransmission)
            {
                // This will disable contact shadow and shadow mask
                light.contactShadowMask = 0;
                light.shadowMaskSelector.x = -1;
            }
            else
            {
                shadow = ComputeMicroShadowing(GetAmbientOcclusionForMicroShadowing(bsdfData), abs(dot(bsdfData.normalWS, L)), _MicroShadowOpacity);
            }
            // This code works for both surface reflection and thin object transmission.
            shadow *= EvaluateShadow_Directional(lightLoopContext, posInput, light, builtinData, GetNormalForShadowBias(bsdfData));
            
            lightColor.rgb *= ComputeShadowColor(shadow, light.shadowTint);
        }

There's too much mess here, all of this code except for shadow eval should go away somewhere.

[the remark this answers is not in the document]
sebastienlagarde: hmm... so indeed ShouldEvaluateThickObjectTransmission isn't always true...

>if you need a transmission flag, just add one, call it MATERIALFEATUREFLAGS_TRANSMISSION

That's an issue as it mean that "neutral" isn't 0 by default anymore (remember that for tile classification we have added compile tile code to removed branch, so we need to have a struct initialize to 0 and no FillTransmission code call at all.). But yes, I am agree that this code isn't really lisible. Let's see if we can improve. I guess we have added a test last time to catch the issue with material classification and non neutral value, should help (I need to check that this test is still enabled)

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: ah I see what you mean. we can still be there with thick if we don't have shadow.
So it mean that if there is no shadow we fallback on thin transmission in a way?
So indeed in this case the variable is bably name and should be call "anyTransmision" instead of thin transmission.

So if we let the current code and just rename thinTransmisison to anyTransmisison with the test on transmittance > 0 + NdotL < 0), then we should be good. Thought ?

sebastienlagarde: and indeed microshadowing shouldn't have effect in this case, good point

EvgeniiG:
> > Should we move transmittance to preLightData instead? This would remove the need for >MATERIAL_INCLUDE_TRANSMISSION in most cases, I think.
> 
> Don't think have the same issue? it still mean we need to add transmittance to all prelightData that require it and then do #ifdef MATERIAL_INCLUDE_TRANSMISSION around the code anyway in surface shading?

Well, transmittance is fundamentally the property of the path between the surface we shade and the light. So it's not a property of the surface, but rather something we precompute before lighting (which is wrong but ...), so preLightData is probably a better fit.

I also don't see why declaring a variable which is zero is worse than all those ifdefs all over the place?

EvgeniiG:
Fundamentally, we have the same problem, though, of course.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: yes and no. Here shadow color apply to all shadow code (shadow, contact, shadow mask, micro shadow etc...) which is nice. Inside shadow code we only affect shadow. + it allow to keep single float for shadow for a longuer time (otherwise we need handle three color contact shadow)

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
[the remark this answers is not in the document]
EvgeniiG: So TLDR with all these comments is that I cleaned the code and now it's as confusing as it was before. It should be very simple.

IF thick transmission -> update transmission value
ELSE /* thin or no transmission */ ->
  use baked transmission value
  eval shadows /* which deal with the fact that there are no contact or baked shadows for back faces */
ENDIF

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: All shadow code should be within the eval shadow call.

EvgeniiG: Perf wise I think we already have a big mess. Did you profile recently?

sebastienlagarde: >All shadow code should be within the eval shadow call.

The code haven't change compare to what we have before regarding this. It is possible to have shadow contact without having shadow. Shadow mask are also separate. Shadow code should be only about the atlas shadow. Or we need a two level shadow, one for atlas shadow, then for all the other shadow type, but this is exactly what EvaluateShadow_Directional do, so don't see need for it. 
I think current code is fine regarding shadow, contact shadow, shadow mask. Let's avoid to refactor code too much time. 

Lats profiling was when Francesco have move code in lightloop to win VGPR with directional shadow. Shader code haven't really change since this, so should be ok to compare with. Haven't launch profiling on this one yet, focus on getting it passing the test

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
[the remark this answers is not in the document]
EvgeniiG: You can combine transmission and shadow together into an attenuation along the shadow ray (it should be RGB).

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
sebastienlagarde: >So it mean that if there is no shadow we fallback on thin transmission in a way

nope, bad idea. When we don't have shadow map we should be unshadowed, as Martin asked and was the previous behavior.

sebastienlagarde:
>So it's not a property of the surface, but rather something we precompute before lighting

I give it a try as it is ok to move in preLightData, but Hair use transmittance as an input of master node.
So I guess it is just simpler to let it in bsdfData.

btw, I can see that hair don't do anything for the bake lighting regarding transmittance. Is it on purpose?

void ModifyBakedDiffuseLighting(float3 V, PositionInputs posInput, SurfaceData surfaceData, inout BuiltinData builtinData)
{
    // To get the data we need to do the whole process - compiler should optimize everything
    BSDFData bsdfData = ConvertSurfaceDataToBSDFData(posInput.positionSS, surfaceData);
    PreLightData preLightData = GetPreLightData(V, posInput, bsdfData);

    // Add GI transmission contribution to bakeDiffuseLighting, we then drop backBakeDiffuseLighting (i.e it is not used anymore, this save VGPR)
    {
        // TODO: disabled until further notice (not clear how to handle occlusion).
        //builtinData.bakeDiffuseLighting += builtinData.backBakeDiffuseLighting * bsdfData.transmittance;
    }

    // Premultiply (back) bake diffuse lighting information with diffuse pre-integration
    builtinData.bakeDiffuseLighting *= preLightData.diffuseFGD * bsdfData.diffuseColor;
}

sebastienlagarde:
@EvgeniiG what about this proposal change for testing transmittance: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/d67a5f9944750bf7768edd040586a7af28009169 ?

EvgeniiG:
> btw, I can see that hair don't do anything for the bake lighting regarding transmittance. Is it on purpose?

Yes, hair transmission is non-PBR, so it doesn't work with baked lighting.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
```diff
@@ -137,15 +137,15 @@ float3x3 BuildWorldToTangent(float4 tangentWS, float3 normalWS)
...
+    float3x3 tangentToWorld = CreateTangentToWorld(unnormalizedNormalWS, tangentWS.xyz, tangentWS.w > 0.0 ? 1.0 : -1.0);
 
     // surface gradient based formulation requires a unit length initial normal. We can maintain compliance with mikkts
     // by uniformly scaling all 3 vectors since normalization of the perturbed normal will cancel it.
-    worldToTangent[0] = worldToTangent[0] * renormFactor;
-    worldToTangent[1] = worldToTangent[1] * renormFactor;
-    worldToTangent[2] = worldToTangent[2] * renormFactor;		// normalizes the interpolated vertex normal
+    tangentToWorld[0] = tangentToWorld[0] * renormFactor;
+    tangentToWorld[1] = tangentToWorld[1] * renormFactor;
+    tangentToWorld[2] = tangentToWorld[2] * renormFactor;		// normalizes the interpolated vertex normal
```
slunity: About normalization: Maybe not the appropriate PR to make those changes, and maybe moot since the compiler might treat the length() as subexpression and optimize them out anyway (and it is cheap), but it reminded me: we have many extra normalize() (if I'm not missing anything) in the code: In the shared ShaderGraph templates, ie HDLitRaytracingPass.template, SharedCode.template.hlsl and the equivalent FragInputs.hlsl, we normalize tangentToWorld[2] but it's clearly normalized as we see here. The mikkt TBN leaves the tangent and bitangent unormalized but still "to scale" with how shortened the normal was by interpolation.

Also, I had modified the comments for .core\ShaderLibrary\Common.hlsl : Orthonormalize( )
to underline the fact that the tangent need not be normalized before the call.

So in every shader code where we call Orthonormalize() for the surfaceData.tangentWS, we never need to normalize( ) the TBN tangent before. We only pay that cost in ShaderGraphs though if not "$Tangent" (ie no tangent input, and we use the interpolated tangent).

EvgeniiG:
> @EvgeniiG what about this proposal change for testing transmittance: [d67a5f9](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/d67a5f9944750bf7768edd040586a7af28009169) ?

Still too messy and complicated IMHO. I still prefer my version:

`IF thick_transmission THEN transmission = compute_transmission
ELSE /* thin or no transmission */
  transmission = baked_transmission /* 0 if no transmission */
  transmission *= eval_shadows /* deals with the fact that there are no contact or baked shadows for back faces */
ENDIF`

Ultimately, it is the fault of the shadow code that it does not work correctly when NdotL is negative. It has nothing to do with transmission. So I feel like the shadow code should handle this more gracefully.
Also, why are there baked or contact shadows for negative NdotL? We shouldn't even compute and store them if they have nonsensical values.

Another point is that thick transmission is a bad hack which I am certain will be removed at some point. We should make doing that less painful.

My 2c.

EvgeniiG:
> > @EvgeniiG what about this proposal change for testing transmittance: [d67a5f9](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/d67a5f9944750bf7768edd040586a7af28009169) ?
> 
> Still too messy and complicated IMHO. I still prefer my version:
> 
> ```
> IF thick_transmission THEN transmission = compute_transmission
> ELSE /* thin or no transmission */
>   transmission = precomputed_transmission /* 0 if no transmission */
>   transmission *= eval_shadows /* deals with the fact that there are no contact or baked shadows for back faces */
> ENDIF
> ```
> 
> Ultimately, it is the fault of the shadow code that it does not work correctly when NdotL is negative. It has nothing to do with transmission. So I feel like the shadow code should handle this more gracefully.
> Also, why are there baked or contact shadows for negative NdotL? We shouldn't even compute and store them if they have nonsensical values.
> 
> Another point is that thick transmission is a bad hack which I am certain will be removed at some point. We should make doing that less painful.
> 
> Finally, nothing prevents us from storing transmission in PreLightData. For hair, we can just copy the value from the BSDFData to PreLightData (recall that hair is a big hack as well).
> 
> My 2c.

Of course, if you disagree, we can go with whatever solution you have as long as it works.

sebastienlagarde:
Katana is finally green here:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fshading-framework-refactor-bis-shots&automation-tools_branch=master&unity_branch=trunk&sort=1-asc#

I have update SimpleLit vfx test as now we handle correctly the disabled of cookie.
However I have update the 1301_stacklit_SG test. Thin Transmission was the same but thin transmission with directional shadow was different. Haven't find the origin of the discrepancy, guess it is due to the bias of shadow in normal direction.
And lastly I have update the 8101 opaque. Again because of thin transmission - 3 of the sphere. But this time it was due to thin transmission shadow but to thin transmission itself. The new result looks better (we now see the wrapparound lighting compare to previously. So suppose there was something before that was not correct. Haven't found the origin of the discrepency.

Note: I have move test of contact and baked shadow for negative in evaluated shadow as suggested in feedback. But I let transmittance in bsdfdata this doesn't save any code or readability to me but require to update all generated structure.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl
sebastienlagarde: > About normalization: Maybe not the appropriate PR to make those changes, and maybe moot since the compiler might treat the length() as subexpression and optimize them out anyway (and it is cheap), but it reminded me: we have many extra normalize() (if I'm not missing anything) in the code: In the shared ShaderGraph templates, ie HDLitRaytracingPass.template, SharedCode.template.hlsl and the equivalent FragInputs.hlsl, we normalize tangentToWorld[2] but it's clearly normalized as we see here. The mikkt TBN leaves the tangent and bitangent unormalized but still "to scale" with how shortened the normal was by interpolation.
> 
> Also, I had modified the comments for .core\ShaderLibrary\Common.hlsl : Orthonormalize( )
> to underline the fact that the tangent need not be normalized before the call.
> 
> So in every shader code where we call Orthonormalize() for the surfaceData.tangentWS, we never need to normalize( ) the TBN tangent before. We only pay that cost in ShaderGraphs though if not "$Tangent" (ie no tangent input, and we use the interpolated tangent).

Let's see in another PR

EvgeniiG:
> Katana is finally green here:
> https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fshading-framework-refactor-bis-shots&automation-tools_branch=master&unity_branch=trunk&sort=1-asc#
> 
> I have update SimpleLit vfx test as now we handle correctly the disabled of cookie.
> However I have update the 1301_stacklit_SG test. Thin Transmission was the same but thin transmission with directional shadow was different. Haven't find the origin of the discrepancy, guess it is due to the bias of shadow in normal direction.
> And lastly I have update the 8101 opaque. Again because of thin transmission - 3 of the sphere. But this time it was due to thin transmission shadow but to thin transmission itself. The new result looks better (we now see the wrapparound lighting compare to previously. So suppose there was something before that was not correct. Haven't found the origin of the discrepency.

I think all of these are due to more correct handling of shadow bias.

> Note: I have move test of contact and baked shadow for negative in evaluated shadow as suggested in feedback. But I let transmittance in bsdfdata this doesn't save any code or readability to me but require to update all generated structure.

Excellent!

PR #3706, opened 2019-05-24
Refactor shadow atlas in preparation for Render Graph

JulienIgnace-Unity:
### Purpose of this PR
Refactor of Shadow atlas preparing for RenderGraph
- Shadow rendering and blur functions are now static function in order to make them compatible with RenderGraph
- Blur functions are now entirely handled by the atlas instead of the manager
- Slightly changed EVSM so that the final texture is always the first one of the two ping-pong-ed textures.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Frefactor-shadow-atlas&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAtlas.cs
```diff
@@ -340,46 +404,50 @@ public void AreaShadowBlurMoments(CommandBuffer cmd, HDCamera hdCamera)
...
 
                     }
                 }
 
                 // We patch up the atlas with the requests that, due to different count of blur passes, remained in the copy
-                for (int i = 0; i < m_ShadowRequests.Count; ++i)
+                for (int i = 0; i < parameters.shadowRequests.Count; ++i)
                 {
-                    if (m_CurrentAtlasMomentSurface != finalAtlasTexture[i])
+                    if (0 != finalAtlasTexture[i])
```
FrancescoC-unity: Is yoda notation allowed by our standard?

JulienIgnace-Unity: I have no idea to be honest

FrancescoC-unity: If I were to nitpick, since I never saw it used, I'd convert this back to standard notation

JulienIgnace-Unity: Will do

sebastienlagarde: Don't know if it is in the standard, but this is a good way to catch error.

i.e 
if (0 == finalAtlasTexture[i]) will fail if you do a typo like this: if (0 = finalAtlasTexture[i])
whereas it doesn't faild if you do a typo in the other case if (finalAtlasTexture[i] = 0)
but anyway, not  a big deal

FrancescoC-unity: Yup I know, I like the principle of yoda notation, but it is a thing that I feel should be a all or nothing throughout the code base. In mean jump between that and standard notation create a cognitive dissonance for a second

alelievr: I don't think this is relevant in C#, there is an error for assignations in if: 
```error CS0665: Assignment in conditional expression is always constant; did you mean to use == instead of = ?```

PR #3711, opened 2019-05-24
Hdrp/feature/shadow layers

iM0ve:
**Sumary:** Found an issue regarding Undo/Link Light Layer function, fixed a minor scene related issue, updated the 2501 scene with additional objects

**Undo/Link Light Layer issue:**
q8KWxPaIpW
When the Link Light Layer checkbox is set, the undo or unchecking does not restore previous Light Layer settings

**Minor scene related issue:**
Capture
The shadow layer light had Link Light Layer checked, so as I clicked on the light it updated(see above) and did not match the screenshot provided by alelievr. I separated light and shadows to make it match the image.

**Updated the 2501 scene:**
image
Added spheres that only receive light(left sphere) and only cast shadows(right sphere). Test was successful, all sources worked as intended.

Katana Tests: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP/feature/shadow_layers&automation-tools_branch=master&unity_branch=trunk


sebastienlagarde:
I have redonne a proper PR from your commit here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3732

PR #3712, opened 2019-05-24
XR single-pass instancing for VFX Graph

fabien-unity:
### Purpose of this PR
- add shader macros to VFX templates
- handle VFX indirect dispatch by adding an instanceCountMultiplier
---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=xr-vfx-spi-wip&unity_branch=trunk)

- [ ] Built a player

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Medium

---
### Comments to reviewers
- this PR is self-contained: no need for C++ changes
- each SRP will need to set a constant **srpInstanceCountMultiplier** (or it'll default to 1 if not set)
- @julienf-unity : with the changes to the template files, you need to open each effect to have it recompiled ... is there a better way to handle this like a version bump ?
- @julienf-unity : the VFXFillIndirectArgs compute shader will need to be run for every camera, or we need to output multiple argument list to multiply the instanceCount depending on the render mode (single-pass instancing requires 2 in most cases)
- @peeweek : I've run most scenes from the test projects but could you make sure nothing's broken? Or can you recommend a QA to handle this?


julienf-unity:
VFXFillIndirectArgs is camera independant, it is run during the simulation pass. But yes the issue I see here is that when an effect is needed to be rendered with multiple cameras, some with single pass instanced and some without, it will always be rendered with the amount of particles doubled. Actually it should fail with shadows too. Maybe additional particles are culled in the VS (I need to check the code) but still it is twice the amount of vertices issued for non single pass instanced renders.

One solution would be to fill 2 slots per indirect draw needed in the indirect buffer (one normal and one with the number of instances doubled) and offset per camera/pass depending on whether we need single pass instancing or not.

julienf-unity:
We have a version number in C++ and plan to add one in C# too. We also need to add an auto recompile in case fx are outdated. At the moment, you can trigger manually a recompilation of all effects via the menu.

julienf-unity:
But as a side note, this is an issue (the need for two arg list) that will exist with any DrawInstancedIndirect call and single pass intancing VR. Is drawInstancedIndirect used anywhere else out of VFX?

fabien-unity:
I think the only other system relying on dispatch indirect is the deferred lighting with classification in HDRP and we also handle it manually for single-pass instancing. I don't see a way to make it generic for VR.
Let's chat more about VFXFillIndirectArgs.

fabien-unity:
Follow-up after discussing with @julienf-unity:
- we need to handle the indirect draw in C++
- I'm going to cancel this PR in favor of 2 others : one for just the macros and one later for the change to VFXFillIndirectArgs which will required the C++ change

PR #3715, opened 2019-05-27
Initial support of lightweight pipeline on Stadia platform

Groovounet:
### Purpose of this PR
Stadia platform doesn't support XR which resulted in XR being excluded on the platform. However, Some XR code is systematically so we need to exclude it from lightweight pipeline.

---
### Release Notes
Initial support of lightweight pipeline on Stadia platform.

---
### Testing status

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=stadia&automation-tools_branch=master&unity_branch=trunk&sort=1-asc
The test is green

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: low

---
### Comments to reviewers


Groovounet:
Apparently Yamato test are not ready yet for production. Running katana instead based on Filipe advice.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=stadia&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

It's green!

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -222,6 +222,13 @@ static void InitializeCameraData(LightweightRenderPipelineAsset settings, Camera
 
             if (Camera.main == camera && camera.cameraType == CameraType.Game && camera.targetTexture == null)
             {
+                // There's no exposed API to control how a backbuffer is created with MSAA
+                // By settings antiAliasing we match what the amount of samples in camera data with backbuffer
+                // We only do this for the main camera and this only takes effect in the beginning of next frame.
+                // This settings should not be changed on a frame basis so that's fine.
+                QualitySettings.antiAliasing = msaaSamples;
```
brandon-unity3d: If msaaSamples is 2 and QualitySettings.antiAliasing was 4, QualitySettings.antiAliasing will now be 2 as well.  A change in sample count has happened however when running in XR, the MSAA eye texture will not be updating due to line 234.  
currentQualitySettingsSampleCount, which will be the value of QualitySettings.antiAliasing, will be 2 and msaaSamples will be 2 so msaaSampleCountHasChnaged will not get set to true.  I think the QualitySettings.antiAliasing = msaaSamples should remain where it was which is after the msaaSampleCountHasChanged logic.

Groovounet: I handled this feedback in the following commit

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -237,8 +231,16 @@ static void InitializeCameraData(LightweightRenderPipelineAsset settings, Camera
                     msaaSampleCountHasChanged = true;
                 }
 
+                // There's no exposed API to control how a backbuffer is created with MSAA
+                // By settings antiAliasing we match what the amount of samples in camera data with backbuffer
+                // We only do this for the main camera and this only takes effect in the beginning of next frame.
+                // This settings should not be changed on a frame basis so that's fine.
+                QualitySettings.antiAliasing = msaaSamples;
+
```
brandon-unity3d: Minor suggestion.  Perhaps it would be better to move the comment, starting at line 234, to line 222 or above the 
"if (Camera.main == camera && camera.cameraType == CameraType.Game && camera.targetTexture == null)" line of code.  The reason is that the comment applies both to XR and monoscopic rendering cases.  I think doing so may lead to less possible confusion in the future as some may think it only applies to XR.

phi-lira:
LGTM. Added gfx docs for review then will merged it. Sorry about the delay on this one.

com.unity.render-pipelines.lightweight/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [7.0.0] - 2019-XX-XX
 ### Added
+- Initial Stadia platform support.
```
JordanL8: Added initial support for the Google Stadia platform.

PR #3718, opened 2019-05-27
Deferred Lighing refactoring (lightloop side)

JulienIgnace-Unity:
### Purpose of this PR
Refactored Deferred Lighting in preparation for Render Graph implementation.
- Instead of one big function we now have 3 static functions for each case (compute, compute as pixel, debug pixel full screen pass)
- Compute as Pixel shader version is now available on all platform (by disabling Compute Lighting in the frame settings)
- Unfified Compute and ComputeAsPixel in term of variants. Now both of them always use deferred tile.
- Fixed an issue with lighting quad generation in ComputeAsPixel version.

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Frefactor-deferredlighting&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)

- [ ] Built a player


sebastienlagarde:
There is several failure on both vfx graph and HDRP graphic test

PR #3721, opened 2019-05-28
Fix case 1152439 - 

phi-lira:
### Purpose of this PR
 - Fix bandwidth consumption regression. case 1152439 https://issuetracker.unity3d.com/issues/lwrp-mobile-increased-memory-usage-and-extra-rendering-steps
 - Disables MSAA on screen space shadow resolve pass
 - Fix store and load of screen space shadow resolve pass due to Blit using src and dest as same texture.
 - Backbuffer is not created with MSAA samples when generating intermediate render texture
 - Optimized RenderingUtils.SupportsTextureFormat to do a single look-up. 

---
### Release Notes
 Added

---
### Testing status

- [ ] Built a player
- Other: 

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lw/optimization&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: Medium. MSAA related changes are moderately risky.

**Halo Effect**: Medium. Can affect multiple platforms.

---
### Comments to reviewers

When rendering to a intermediate render texture we don't need to create a backbuffer with MSAA. We resolve MSAA before post of final blit and that's it. 
We need a better way to handle changing MSAA backbuffer format. I'm thinking about exposing a setting that will allow users to select between auto or the number of samples explicitly. 

Bandwidth before changes

Bandwidth after changes

Screen space shadow resolve issue


com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
```diff
@@ -163,8 +163,13 @@ public static void RenderSingleCamera(ScriptableRenderContext context, Camera ca
                 return;
             }
 
-            CommandBuffer cmd = CommandBufferPool.Get(k_RenderCameraTag);
-            using (new ProfilingSample(cmd, k_RenderCameraTag))
+#if UNITY_EDITOR
+            string tag = camera.name;
```
aleks01010101: Wouldn't it be useful to know, which camera is being rendered in the player as well?

com.unity.render-pipelines.lightweight/Runtime/ForwardRenderer.cs
```diff
@@ -291,6 +300,26 @@ void CreateCameraRenderTarget(ScriptableRenderContext context, ref CameraData ca
             CommandBufferPool.Release(cmd);
         }
 
+        void SetupBackbufferFormat(int msaaSamples, bool stereo)
+        {
+            bool msaaSampleCountHasChanged = false;
+            int currentQualitySettingsSampleCount = QualitySettings.antiAliasing;
+            if (currentQualitySettingsSampleCount != msaaSamples &&
```
aleks01010101: This can be a one-liner

com.unity.render-pipelines.lightweight/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -64,6 +64,10 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             ref CameraData cameraData = ref renderingData.cameraData;
             if (cameraData.isStereoEnabled || cameraData.isSceneViewCamera || cameraData.isDefaultViewport)
             {
+                // This set render target is necessary so we change the LOAD state to DontCare.
+                cmd.SetRenderTarget(BuiltinRenderTextureType.CameraTarget, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store);
+                if (Application.isMobilePlatform || Application.platform == RuntimePlatform.Switch)
```
aleks01010101: This can be setup once and stored, no need to query each time

com.unity.render-pipelines.lightweight/Runtime/LightweightRenderPipeline.cs
brandon-unity3d: I agree

phi-lira: camera.name allocates memory. :(

com.unity.render-pipelines.lightweight/Runtime/ForwardRenderer.cs
phi-lira: Just moved the code. @brandon-unity3d you wrote that one. Why does it need the second condition?

com.unity.render-pipelines.lightweight/Runtime/Passes/FinalBlitPass.cs
phi-lira: Right! will fix it.

PR #3744, opened 2019-05-29
Vfx/feature/output refactor

julienf-unity:
### Purpose of this PR
- Refactor outputs so there can be specific data per SRP
- Move HDRP specific outputs to HDRP package (HDRP has now a dependency on VFX Graph)
- Automatic detection of SRP
- Add render queue selection for HDRP
- A few minor features from staging

---
### Release Notes

---
### Testing status
All tests green locally
Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=vfx%2Ffeature%2Foutput-refactor&unity_branch=trunk&automation-tools_branch=master&sort=1-asc

---
### Comments to reviewers
This is the first pass of the refactor for 19.2
- HDRP templates are still in VFX Graph package because we need to handle multi directory for template includes
- Lots of specific HDRP features still need to be added


com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubOutput.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace UnityEditor.VFX
+{
+    class VFXHDRPSubOutput : VFXSRPSubOutput
+    {
+        [VFXSetting(VFXSettingAttribute.VisibleFlags.InInspector), Header("HDRP")]
+        public OpaqueRenderQueue opaqueRenderQueue = OpaqueRenderQueue.Default;
+
+        [VFXSetting(VFXSettingAttribute.VisibleFlags.InInspector), Header("HDRP")]
+        public TransparentRenderQueue transparentRenderQueue = TransparentRenderQueue.Default;
+
```
PaulDemeulenaere: Nice interface, in the end, I will be able to add "generate motion vector" option here, is that right ?

com.unity.visualeffectgraph/Editor/Models/Contexts/Implementations/VFXAbstractRenderedOutput.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        protected virtual void WriteBlendMode(VFXShaderWriter writer)
+        {
+            var blendModeStr = subOutput.GetBlendModeStr();
+            if (!String.IsNullOrEmpty(blendModeStr))
+                writer.WriteLine(blendModeStr);
+        }
+
+        [SerializeField]
+        private List<VFXSRPSubOutput> m_SubOutputs;
```
PaulDemeulenaere: Did you check the behavior of copy past of this ? I'm not sure these sub-object are correctly grabbed while doing this kind of operation.

julienf-unity: Right copy/paste will probably not take that into account. Can be fixed later though

com.unity.render-pipelines.high-definition/Editor/VFXGraph/VFXHDRPSubOutput.cs
julienf-unity: Yes if we want motion vectors to be HDRP only. But I heard LWRP was going to have some at some point. Anyway we can still sanitize when it is the case

PR #3750, opened 2019-05-30
Adding HDRP SpeedTree shaders [2019.3]

parashar-unity:
### Purpose of this PR
Enable support SpeedTree 7 and SpeedTree 8 shaders for HDRP.  Both are dependent on (and by extension, are variants of) the standard Lit model.

---
### Release Notes
Dependent, at least, on the core trunk changes that made the LWRP SpeedTree shaders possible, due to the fact that these changes added default SpeedTree shaders as members of the RenderPipelineAsset. 

---
### Testing status
**Katana Tests**: All green.  Newly added tests are now passing.

**Manual Tests**: 
- Created test projects and scenes, testing both SpeedTree7 and SpeedTree8 sample assets.
- Tested custom shader UIs (particularly for SpeedTree8 shader)
- C# and shader warnings cleared for now, but there are some to check on multiple platforms after the recent merge with master.
- Tested UI multi-edition + Undo/Redo

**Automated Tests**:
Added a test scene to HDRP_Tests under Materials.  Image result shown below
1801_SpeedTreeLit

From the left, the first two trees are both SpeedTree7 assets.  The tree to the far right is a SpeedTree8 asset.

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FSpeedTree&unity_branch=trunk&automation-tools_branch=master&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: Medium
Brand new shader and test added.  New material is subject to be affected as the Lit material and standard shaders in HDRP change over time.

**Halo Effect**: Medium
Needed to add some #ifdefs to some parts of the shader library because of the fact that SpeedTree packs and interprets data a little differently than we do for our own geometry, which means I had to add some custom resolve routines -- this meant adding some #ifdefs that should not affect other shaders, but is worth watching out for.

---


sebastienlagarde:
Hi ,
It would have been better to discuss about this PR before bring it like this. Particularly as we are in the process of doing package for 19.2 and not adding feature. Also we haven't schedule it on our roadmap.
We will take the time to review it but after we release the 19.2 package and template for the beta. 
This PR will be part of 19.3.
But my main question (sorry have check the whole PR, have a very quick look). Why isn't this a shader graph node? Adding new Lit inherited shader is bad for maintenance.

parashar-unity:
> Hi ,
> It would have been better to discuss about this PR before bring it like this. Particularly as we are in the process of doing package for 19.2 and not adding feature. Also we haven't schedule it on our roadmap.
> We will take the time to review it but after we release the 19.2 package and template for the beta.
Now I'm getting mixed signals.  On one side, I'm hearing clamor to get this in as quickly as possible (and it was technically done some time back, and simply needed tests to be added, but was put on the back burner as I shifted over), so I tried to move this forward quickly....  and now I'm hearing that the very notion that it was coming was seemingly not even expected.

> But my main question (sorry have check the whole PR, have a very quick look). Why isn't this a shader graph node - From my understanding Speedtree only add vertex animation control, right ? Adding new Lit inherited shader is bad for maintenance.

Vertex animation control and a custom pack/unpack of the interpolators.  Also, as I mentioned, much work was done some weeks back ... during a time when the ShaderGraph API was experiencing quite a bit of flux.  I was basically thinking of having something usable at all.  If this anyway has to wait out for the 19.3 version of the package, then that is a time window to switch over to a graph node if the maintenance cost is really that much worse.

russoh:
The SpeedTreeImporter writes attributes into the mesh vertex streams, that then get selectively copied into interpolators to the fragment shader. That usage isn't very well defined/fixed, there ends up being a lot of (manual) packing logic based on which features are being used etc. ShaderGraph would need some changes to support the extra data needed if it can't just be packed via uv channels and handled generically? The shared wind shader code could be a custom node.
Has any of this been tested on Mac yet?

zadi0648:
Removing myself. Readd me if Wyatt's review cannot happen.

parashar-unity:
Will create a new one with the ShaderGraph version when that is ready.

PR #3751, opened 2019-05-31
Experimental Version of the PBR Sky System

EvgeniiG:
### Release Notes
Adds an experimental version of the PBR Sky System.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fpbr-sky&automation-tools_branch=master&unity_branch=trunk

---

### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
Still need to clean the code, but everything mostly works.

Known limitations:
- reflection probe is computed at the position of the camera, which does not respect parallax
- ambient probe is computed at the position of the camera, which does not take the height above the sea level and NdotL into account
- planet rendering does not write any depth (it should)
- transparent alpha blending is wrong
- no sun disk support - users are expected to use an emissive mesh for sun/moon, as we do not want the sky in the reflection probe or SH to avoid double lighting
- no lighting information well below the horizon (currently limited to 120 deg) - this will present an issue for large atmospheres
- undersampling/banding
- MS contribution seems rather weak, could be a bug - need a path-traced implementation to investigate


com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.cs
```diff
@@ -33,7 +33,17 @@ public void PushShaderParametersCommon(HDCamera hdCamera, CommandBuffer cmd, Fog
         {
             Debug.Assert(hdCamera.frameSettings.IsEnabled(FrameSettingsField.AtmosphericScattering));
 
-            cmd.SetGlobalInt(HDShaderIDs._AtmosphericScatteringType, (int)type);
+            int pbrSkyAtmosphereFlag = 0;
+
+            var visualEnvironment = VolumeManager.instance.stack.GetComponent<VisualEnvironment>();
+
+            if (visualEnvironment != null)
```
JulienIgnace-Unity: This should not be needed. There is always a component of any type in the volume system. If none is in the scene, it will just be default values.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -48,30 +50,38 @@ Shader "Hidden/HDRP/OpaqueAtmosphericScattering"
...
+            float2 positionSS  = input.positionCS.xy;
+            float3 V           = GetSkyViewDirWS(positionSS);
+            float  depth       = LoadCameraDepth(positionSS);
+            float3 surfColor   = LOAD_TEXTURE2D_X(_ColorTexture, (int2)positionSS).rgb;
+
+            float3 volColor, volOpacity;
+            AtmosphericScatteringCompute(input, V, depth, volColor, volOpacity);
 
-            return AtmosphericScatteringCompute(input, V, depth);
+            return float4(volColor + (1 - volOpacity) * surfColor, 1); // Premultiplied alpha (over operator)
```
JulienIgnace-Unity: Why manual blending here? What's the advantage over hw blending?
Besides, we read and write into the same texture right? Might be some devices where it's problematic.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1912,6 +1914,15 @@ HDProbe SelectProbe(VisibleReflectionProbe probe, PlanarReflectionProbe planarPr
                     // And if needed rescale the whole atlas
                     m_ShadowManager.LayoutShadowMaps(debugDisplaySettings.data.lightingDebugSettings);
 
+                    bool isPbrSkyActive = false;
+
+                    var visualEnvironment = VolumeManager.instance.stack.GetComponent<VisualEnvironment>();
+
+                    if (visualEnvironment != null)
```
JulienIgnace-Unity: Same as above.

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/PbrSkyRenderer.cs
```diff
@@ -0,0 +1,440 @@
...
+                        break;
+                    }
+
+                    // Re-illuminate the ground with each bounce.
+                    cmd.DispatchCompute(s_GroundIrradiancePrecomputationCS, firstPass, (int)PbrSkyConfig.GroundIrradianceTableSize / 64, 1, 1);
+                }
+            }
+        }
+
+        // 'renderSunDisk' parameter is meaningless and is thus ignored.
```
JulienIgnace-Unity: re: comment about sun disk in the PR description
renderSunDisk is not meaningless. Its very purpose is to handle the case where reflection probes are baked.
See procedural sky implementation for an example but basically, the "includeSunInBaking" parameter in SkySettings will be used by the sky manager to determine whether or not sun disk should be rendered (visual sky versus baking based on this parameter).
So it's totally possible to implement the sun disk in pbr sky and still avoid double lighting.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
JulienIgnace-Unity: nvm, did not see that volOpacity was float3

EvgeniiG: Yes, Unity does not support dual source blending. :(

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.cs
EvgeniiG: OK, fixed.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
EvgeniiG: OK, fixed.

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/PbrSkyRenderer.cs
EvgeniiG: Ah, I see. Either way, users should use an emissive mesh since I don't want to deal with having multiple stars, sun/moon, situations where you can fly towards the sun in space, etc.

EvgeniiG: I will update the comment.

EvgeniiG:
> See the few comments I left.

Addressed.

EvgeniiG:
I have some failing tests, trying to fix them.

EvgeniiG:
1206 Distortion test appears to be unstable.

EvgeniiG:
2307 Shadows is also unstable.

EvgeniiG:
8101 Opaque is also unstable.

EvgeniiG:
So the only test which is really broken is 9401 MSAA. Will try to fix...

EvgeniiG:
OK, all tests pass locally (at least, after running them several times) except for 1206, 2307 and 8101, which fail due to tiny differences. Will re-run Katana, and if they fail there, update the images.

sebastienlagarde:
>except for 1206, 2307 and 8101, 
hmm... I don't see any reasons that those test fail on katana. These test are stable since a while (as can be seen here: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-Test%20PlayMode%20-%20Windows?ScriptableRenderLoop_branch=master&automation-tools_branch=master&unity_branch=trunk (note the red failure aren't about those test).

I will avoid to update the image as your PR isn't suppose to affect them (if I get it correctly - but as I haven't look at the PR I don't know). Or it is because the extra pass (read / write) ?

sebastienlagarde:
Note that your PR must target HDRP/staging. 
(We have finally move the development to HDRP/Staging to avoid failure on master)

There is various update in HDRP/staging on the test (see https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3743) and you will require to update the test anyway.

Note: we also need a test for the physically based sky itself. Thanks

sebastienlagarde:
@laurenth-unity For awareness so you can experiment with the sky :)

sebastienlagarde:
Adding Remy for QA. This is still experimental but good to have early review by QA.

EvgeniiG:
OK, I switched the target to HDRP/staging.

EvgeniiG:
Merged staging, will run Katana again.

sebastienlagarde:
FYI, on HDRP/Staging, all the test pass, but katana is red due to the new HDRP_Test_DXR that we have added (so don't be surprise if it is red), we will fix that tomorrow.

EvgeniiG:
So, looks like it now passes?
https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/

EvgeniiG:
> So, looks like it now "passes" all tests?
> https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/

Confirm / Deny / **Ignore** ? :P

sebastienlagarde:
> So, looks like it now "passes" all tests?
> https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/

This is weird, Don't know why there is 85 test total but only 79 pass without any failing... LEt's say it is Mac and LW so ignore.
However you have a lot of test failing for VFX, suppose you miss VFX function update:

<![CDATA[
Unhandled log message: '[Error] Shader error in 'Hidden/VFX/Distortion/System 2/Lit Cube Output': 'EvaluateAtmosphericScattering': no matching 2 parameter function at Packages/com.unity.visualeffectgraph/Shaders/RenderPipeline/HDRP/VFXCommon.cginc(82) (on d3d11)
 
 Compiling Vertex program
 Platform defines: UNITY_ENABLE_REFLECTION_BUFFERS UNITY_USE_DITHER_MASK_FOR_ALPHABLENDED_SHADOWS UNITY_PBS_USE_BRDF1 UNITY_SPECCUBE_BOX_PROJECTION UNITY_SPECCUBE_BLENDING UNITY_ENABLE_DETAIL_NORMALMAP SHADER_API_DESKTOP UNITY_HARDWARE_TIER3 UNITY_LIGHTMAP_RGBM_ENCODING'. Use UnityEngine.TestTools.LogAssert.Expect
]]>
</message>
<stack-trace>
<![CDATA[ 
 ]]>
</stack-trace>
</failure>
<output>
<![CDATA[

 ##utp:{"type":"TestStatus","time":1559687846526,"version":2,"phase":"Begin","processId":4576,"name":"UnityEngine.VFX.Test.VFXGraphicsTests.FlipbookMotionBlend","state":0}
 Shader error in 'Hidden/VFX/Distortion/System 2/Lit Cube Output': 'EvaluateAtmosphericScattering': no matching 2 parameter fun
]]>

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -756,6 +757,19 @@ float DecodeLogarithmicDepth(float d, float4 encodingParams)
     return encodingParams.x * exp2(d * encodingParams.y);
 }
 
+real4 CompositeOver(real4 front, real4 back)
```
sebastienlagarde: Sanity check: just wonderingif the calculation on the alpha component is what you want.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -62,6 +62,7 @@ enum Expandable
             ShadowQuality = 1 << 8
         }
 
+        // That's not a real English word...
```
sebastienlagarde: What do you try to mean by Advanceable here? :)

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
EvgeniiG: Yes.

image

com.unity.render-pipelines.high-definition/Editor/Lighting/SerializedHDLight.cs
```diff
@@ -124,7 +125,7 @@ public SerializedHDLight(HDAdditionalLightData[] lightDatas, AdditionalShadowDat
                     affectSpecular = o.Find(x => x.affectSpecular),
                     nonLightmappedOnly = o.Find(x => x.nonLightmappedOnly),
                     lightTypeExtent = o.Find(x => x.lightTypeExtent),
-                    spotLightShape = o.Find("m_SpotLightShape"),
+                    spotLightShape = o.Find("m_SpotLightShape"), // WTF?
```
sebastienlagarde: it some legacy inheritance iirc

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.hlsl
```diff
@@ -105,13 +110,229 @@ float4 EvaluateAtmosphericScattering(PositionInputs posInput, float3 V)
...
-    return float4(fogColor * GetCurrentExposureMultiplier(), fogFactor);
+    fogColor *= GetCurrentExposureMultiplier();
+
+    color   = fogColor;
+    opacity = fogOpacity;
+
+    int atmosphereType = _AtmosphericScatteringType & ~FOGTYPE_MAX_VALUE;
+
+    // Below, we assume there is NO sky occlusion AT ALL.
+    if (atmosphereType == ATMOSPHERETYPE_PBR)
```
sebastienlagarde: Do you expect other type than ATMOSPHERETYPE_PBR?

EvgeniiG: Well, there is "None"... I dunno? I could make it a bool hasAtmosphere.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.hlsl
```diff
@@ -105,13 +110,229 @@ float4 EvaluateAtmosphericScattering(PositionInputs posInput, float3 V)
...
+            if (!_DirectionalLightDatas[i].interactsWithSky) continue;
+
+            float3 L             = -_DirectionalLightDatas[i].forward.xyz;
+            float3 lightRadiance =  _DirectionalLightDatas[i].color.rgb;
+
+            // TODO: solve in spherical coords?
+            float  NdotL = dot(N, L);
+            float3 projL = L - N * NdotL;
+            float3 projV = V - N * NdotV;
+            float  phiL  = acos(clamp(dot(projL, projV) * rsqrt(max(dot(projL, projL) * dot(projV, projV), FLT_EPS)), -1, 1));
```
sebastienlagarde: acos => Can this be replace by a FastACos() ? or precision matter here?

EvgeniiG: No optimizations as a part of this PR.

EvgeniiG:
> > So, looks like it now "passes" all tests?
> > https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/
> 
> This is weird, Don't know why there is 85 test total but only 79 pass without any failing... LEt's say it is Mac and LW so ignore.
> However you have a lot of test failing for VFX, suppose you miss VFX function update:
> 
> ##utp:{"type":"TestStatus","time":1559687846526,"version":2,"phase":"Begin","processId":4576,"name":"UnityEngine.VFX.Test.VFXGraphicsTests.FlipbookMotionBlend","state":0}
> Shader error in 'Hidden/VFX/Distortion/System 2/Lit Cube Output': 'EvaluateAtmosphericScattering': no matching 2 parameter fun
> ]]>

My bad, will fix.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/OpaqueAtmosphericScattering.shader
```diff
@@ -91,11 +101,12 @@ Shader "Hidden/HDRP/OpaqueAtmosphericScattering"
         // 1: MSAA
         Pass
         {
-            Cull Off ZTest  Always ZWrite Off Blend One OneMinusSrcAlpha // Premultiplied alpha
+            Cull Off    ZTest Always    ZWrite Off    Blend Off // Manual blending
 
             HLSLPROGRAM
                 #pragma vertex Vert
                 #pragma fragment FragMSAA
+                #define MSAA 1
```
sebastienlagarde: is this define use somewhere?

EvgeniiG: Not anymore, will remove.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs
```diff
@@ -117,6 +117,14 @@ public sealed class ShaderResources
             public Shader gradientSkyPS;
             [Reload("Runtime/Sky/AmbientProbeConvolution.compute")]
             public ComputeShader ambientProbeConvolutionCS;
+            [Reload("Runtime/Sky/PbrSky/OpticalDepthPrecomputation.compute")]
+            public ComputeShader opticalDepthPrecomputationCS;
+            [Reload("Runtime/Sky/PbrSky/GroundIrradiancePrecomputation.compute")]
+            public ComputeShader groundIrradiancePrecomputationCS;
+            [Reload("Runtime/Sky/PbrSky/InScatteredRadiancePrecomputation.compute")]
+            public ComputeShader inScatteredRadiancePrecomputationCS;
+            [Reload("Runtime/Sky/brSky/PbrSky.shader")]
```
sebastienlagarde: there is a typo here: brSky

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/GroundIrradiancePrecomputation.compute
```diff
@@ -0,0 +1,90 @@
+#pragma enable_d3d11_debug_symbols
```
sebastienlagarde: remove enable_d3d11_debug_symbols

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/GroundIrradiancePrecomputation.compute
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    _GroundIrradianceTableOrder[tableCoord] = tableEntry; // One order
+    _GroundIrradianceTable[tableCoord]      = tableEntry; // All orders
+
+#else  // MULTIPLE_SCATTERING
+
+    // Gather the volume contribution.
+    // Arbitrary number of samples...
+    const int numVolumeSamples = 89;
```
sebastienlagarde: is it random or is it value you tested to have good result and lower number is not good?

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/InScatteredRadiancePrecomputation.compute
```diff
@@ -0,0 +1,384 @@
+#pragma enable_d3d11_debug_symbols
```
sebastienlagarde: enable_d3d11_debug_symbols

EvgeniiG:
> > > So, looks like it now "passes" all tests?
> > > https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/
> > 
> > 
> > This is weird, Don't know why there is 85 test total but only 79 pass without any failing... LEt's say it is Mac and LW so ignore.
> > However you have a lot of test failing for VFX, suppose you miss VFX function update:
> > ##utp:{"type":"TestStatus","time":1559687846526,"version":2,"phase":"Begin","processId":4576,"name":"UnityEngine.VFX.Test.VFXGraphicsTests.FlipbookMotionBlend","state":0}
> > Shader error in 'Hidden/VFX/Distortion/System 2/Lit Cube Output': 'EvaluateAtmosphericScattering': no matching 2 parameter fun
> > ]]>
> 
> My bad, will fix.

Bad idea to have several versions of the same function, now I have to play the game of figuring out the difference...

image

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/OpticalDepthPrecomputation.compute
```diff
@@ -0,0 +1,144 @@
+#pragma enable_d3d11_debug_symbols
```
sebastienlagarde: enable_d3d11_debug_symbols

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/GroundIrradiancePrecomputation.compute
EvgeniiG: Haven't extensively tested.

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/PbrSkyCommon.hlsl
```diff
@@ -0,0 +1,360 @@
...
+}
+
+float3 AtmospherePhaseScatter(float LdotV, float height)
+{
+    return AirPhase(LdotV) * AirScatter(height) + AerosolPhase(LdotV) * AerosolScatter(height);
+}
+
+// Returns the closest hit in X and the farthest hit in Y.
+// Returns a negative number if there's no intersection.
+float2 IntersectSphere(float sphereRadius, float cosChi, float radialDistance)
```
sebastienlagarde: more it to gemetry.hlsl?

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/OpticalDepthPrecomputation.compute
EvgeniiG: Removed.

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
```diff
@@ -401,15 +414,26 @@ public void RenderSky(HDCamera camera, Light sunLight, RTHandleSystem.RTHandle c
...
-                HDUtils.DrawFullScreen(cmd, m_OpaqueAtmScatteringMaterial, colorBuffer, depthBuffer, propertyBlock, isMSAA? 1 : 0);
+                if (isMSAA)
+                    propertyBlock.SetTexture(HDShaderIDs._ColorTextureMS, colorBuffer);
+                else
+                    propertyBlock.SetTexture(HDShaderIDs._ColorTexture,   colorBuffer);
+                // Color -> Intermediate.
+                HDUtils.DrawFullScreen(cmd, m_OpaqueAtmScatteringMaterial, intermediateBuffer, depthBuffer, propertyBlock, isMSAA? 1 : 0);
+                // Intermediate -> Color.
+                // Note: Blit does not support MSAA (and is probably slower).
+                cmd.CopyTexture(intermediateBuffer, colorBuffer);
```
sebastienlagarde: arf... hmm... Maybe at least we should do this as an option. i.e pbr sky we do this, but when non pbr we use hardware blending (until at some point we implement the dual hardware blend).

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyRenderer.cs
```diff
@@ -20,5 +21,18 @@ protected float GetExposure(SkySettings skySettings, DebugDisplaySettings debugS
             }
             return ColorUtils.ConvertEV100ToExposure(-(skySettings.exposure.value + debugExposure));
         }
+
+        public static void SetGlobalNeutralSkyData(CommandBuffer cmd)
```
sebastienlagarde: why is this here? This is specific to PBR sky

sebastienlagarde:
> > > > So, looks like it now "passes" all tests?
> > > > https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Mac__Intel_/30651926_04_06_2019_20_01_35_%2B0000/logs/
> > > 
> > > 
> > > This is weird, Don't know why there is 85 test total but only 79 pass without any failing... LEt's say it is Mac and LW so ignore.
> > > However you have a lot of test failing for VFX, suppose you miss VFX function update:
> > > ##utp:{"type":"TestStatus","time":1559687846526,"version":2,"phase":"Begin","processId":4576,"name":"UnityEngine.VFX.Test.VFXGraphicsTests.FlipbookMotionBlend","state":0}
> > > Shader error in 'Hidden/VFX/Distortion/System 2/Lit Cube Output': 'EvaluateAtmosphericScattering': no matching 2 parameter fun
> > > ]]>
> > 
> > 
> > My bad, will fix.
> 
> Bad idea to have several versions of the same function, now I have to play the game of figuring out the difference...
> 
> image

@julienf-unity Why is there a redefinition of fog function here ?

sebastienlagarde:
Great work btw, so much stuff here, well done :)

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyManager.cs
EvgeniiG: Can optimize in another PR.

com.unity.render-pipelines.high-definition/Runtime/Sky/SkyRenderer.cs
EvgeniiG: Because we have the "sky system" and every sky is treated the same way. How do you propose I do it instead?

EvgeniiG: Ah, wait, I see what you mean. Let me think.

EvgeniiG: Well, one of the problems is that since there are no #defines for the PBR sky (I just use dynamic branching), we have these textures bound (but maybe not accessed) to all lighting shaders (including forward for transparents). If I don't bind anything, Unity acts stupid and produces an error.

com.unity.render-pipelines.high-definition/Runtime/Sky/PbrSky/PbrSkyCommon.hlsl
EvgeniiG: It's pretty sky-specific, I don't think it's generally useful...

julienf-unity:
@sebastienlagarde I guess by the time we did it it was not easy to use the function as is (either because of the HDRP specific defines or some other stuff I dont remember).

Anyway now I have refactored some stuff to handle several SRPs, it will be very easy to get rid of this and just call your EvaluateAtmosphericScattenring function directly.

Let's talk about this tomorrow about what's the best time frame to do it (tomorrow or in a follow up PR).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs
EvgeniiG: Fixed.

EvgeniiG:
I updated the VFX function, will re-run tests.

EvgeniiG:
Opaque test still fails with microscopic differences...

julienf-unity:
> Opaque test still fails with microscopic differences...

Well as this has nothing to do with the VFXApplyFog and we use the HDRP path for fog on opaque, I guess you can just replace the ref image. We tuned the error threshold for tests so that it is the lower possible. Are the differences deterministic?

sebastienlagarde:
>Opaque test still fails with microscopic differences...

ok, my guess is that it is due to the copy of the buffer. Hardware blending and in shader blending may have a different precision. thought?

EvgeniiG:
> > Opaque test still fails with microscopic differences...
> 
> Well as this has nothing to do with the VFXApplyFog and we use the HDRP path for fog on opaque
Indeed, this one has nothing to do with VFX. :-)

EvgeniiG:
> > Opaque test still fails with microscopic differences...
> 
> ok, my guess is that it is due to the copy of the buffer. Hardware blending and in shader blending may have a different precision. thought?
> 
> I haven't look at the screenshots, is it everywhere or just the sky?

OK, we have exponential fog in the scene for some reason, so that's why... Should disable it, but meanwhile I will just update the screenshot.

EvgeniiG:
Also had to update the shadow test ref img. The test has both the sky and the fog even though it's only supposed to test shadows.

EvgeniiG:
Also, I think I still have some failures in the VFX tests. Will try to investigate...

EvgeniiG:
I updated the test img for vfx fog. Seems somewhat different, so I suppose the behavior changed... I guess HDRP premultiplies something, and you don't?

EvgeniiG:
VFX tests now pass on my machine.
But since I switched to HDRP/staging, HDRP added new non-fog-related tests and those fail now.
What am I supposed to do with these?

image

EvgeniiG:
Also, I still don't have a way to apply (opaque) atmospheric scattering in reflection probes.

sebastienlagarde:
Yes. The new failing test on reflection probe failing are on staging. I just discovered that. It is related to a PR of fred that fix point of capture byt missed to update screenshots. There is 5 test failing. You have the list on the hdrp/staging katana if needed.

EvgeniiG:
So I should wait for him to fix it?

sebastienlagarde:
> So I should wait for him to fix it?

Test are fixed now

EvgeniiG:
Merged 'HDRP/staging' again, old failures are fixed, and I have two new ones.

image

And 9005 after post process is messed up (broken sorting order):

image

sebastienlagarde:
yes, I discover the two news failure, they are on HDRP staging, consider your katana to be green.
sorting order is due to a PR from Antoine, the other one is a mess in a the fix of Fred. Sadly as katana take sooooo long to build, we don't detect issue faster.

EvgeniiG:
Testing locally takes only 10 minutes. Everyone should test locally first.

EvgeniiG:
So, what's the plan regarding my PR?

TomasKiniulis:
PR looks to be in a pretty stable shape. Some issues occur with planar reflection probes, fog, planet atmosphere produces strong ring artifacts with high planetary radius, atmosphere part drawn in front of objects or planet.
Sharing doc to some findings and issues: https://confluence.hq.unity3d.com/display/HDRP/Experimental+Version+of+the+PBR+Sky+System

EvgeniiG:
Merged the staging branch again.
SSAO tests are broken (2601, 2602).
8101 opaque has microscopic differences again (will update).
9001 LOD transition looks pretty broken:

image

sebastienlagarde:
> Merged the staging branch again.
> 
> SSAO tests are broken (2601, 2602).
> 8101 opaque has microscopic differences again (will update).
> 9001 LOD transition looks pretty broken:
> 
> image

hmm... Weird for the SSAO test, it isn't expected.
8101 have the same micro difference than before, it is because I have update it, so you need to update again.

For 9001 this is really weird. I have get it two times already and on katana. I guess it could have been introduced by the refactor PR, but don't know why. The image is failing only the first time the test is executed, if you load the scene all is correct. And worse of all. It seems that if you don't have other error, the test success... So seems to be an instability related to what was the previous failure, I need to discuss that with Remy. So only AO test here are weird as they are suppose to work. A conflict in a merge ?

sebastienlagarde:
update, the test 9001 is also failing on master :( 
https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Windows/30804961_12_06_2019_08_48_46_%2B0000/logs/

EvgeniiG:
Might be an issue with HDRenderPipelineResources.asset, will double check locally.

EvgeniiG:
Still failing the SSAO test locally.

EvgeniiG:
SSAO tests pass on Katana, but fail locally.

EvgeniiG:
Okay, seems green for HDRP and VFX tests.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/AtmosphericScattering.hlsl
sebastienlagarde: yes, it will be better (can be done in next PR)

zero-radiance:
Nice UI changes, thank you. :-)

EvgeniiG:
Nice UI changes, thank you. :-)

EvgeniiG:
As promised, I renamed the sky class, and got rid of the atmospheric scattering type.
Will run Katana again.

EvgeniiG:
Katana is green!

PR #3170, opened 2019-03-06
ShaderGraph Tests in LWRP

spaztronaut:
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?
Adds some ShaderGraph tests to the LWRP Test Project
Modifies LWGraphicsTest.cs to check for LWRP Test Settings or SGLW Test Settings

---
### Release Notes
Please add any useful notes about the feature/fix that might be helpful for others.
3 test scenes and reference images added for Windows + DX11:
1. Unlit (includes alpha clip)
2. Lit (includes alpha clip)
3. Vertex offsets with ShaderGraph in LWRP

---
### Testing status
**Katana Tests**: First off we need to make sure the Katana SRP tests are green?

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [ ] C# and shader warnings (supress shader cache to see them)
- Other: 

**Automated Tests**: What did you setup?

Any test projects to go with this to help reviewers?
TestProjects/LWGraphicsTest

---
### Overall Product Risks
**Technical Risk**: None, Low, Medium, High?
Low

**Halo Effect**: None, Low, Medium, High?

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #3752, opened 2019-05-31
Optional Shadow Pass

zadi0648:
### Purpose of this PR
The start of optional passes: optional shadow pass.

Completes this favro card: https://favro.com/organization/c564ede4ed3337f7b17986b6/d06b8de3e40da6d340b5b063?card=Uni-80150

Fixes this ancient "bug": https://fogbugz.unity3d.com/f/cases/1046827/

Implemented for:
- LWRP Unlit
- LWRP PBR
- HDRP Unlit
- HDRP PBR

OptionalShadowPass

**Bonus Feature**: PBR & Unlit Master Nodes now show & hide setting menu options depending on if they are applicable based on Surface type.

(HDRP in use below)
OptionalShadowPass2

---
### Implementation Notes
- Did not implement for any HDRP specific Master Nodes, that's up to them.
- Only possible for opaque. The option becomes hidden and then disabled for transparent. (Decided on this after talk with Chris) Could easily change that.
- Created an Interface for Master Nodes so that potentially templates wouldn't have to check each Master Node type.
- Started to split up LW templates to make passes modular.
- Default for PBR = true, default for Unlit = false.

---
### Testing status
**Katana Tests**: 

**Manual Tests**:

- Opened test project + Run graphics tests locally.
- Checked new UI names with UX convention.
- Tested UI multi-edition + Undo/Redo.
- Various material and setting combinations for the 4 combinations, vertex animation, ect.
- Switching to transparent with Cast Shadow still set results in now shadow for the 4 combinations.
- Messing with Shadow Settings in the LWRP & HDRP assets.

**Automated Tests**:

**HDRP:**

Added a test to test for Shader Graph Shadow Cast:
8206_Shadows

**LWRP:**

Looks like #3170 is not in yet, will either add a test directly to that PR or after.

**Status:**

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Foptional-shadow-pass&automation-tools_branch=master&unity_branch=trunk
? Yamato: ...

---
### Overall Product Risks

**Technical Risk**: Low
**Halo Effect**: Medium

---
### Questions

What other passes do we care to make optional?


com.unity.shadergraph/Editor/Data/MasterNodes/PBRMasterNode.cs
```diff
@@ -91,7 +92,23 @@ public AlphaMode alphaMode
...
+        {
+            get { return new ToggleData(m_shadowCast); }
+            set
+            {
+                if (m_shadowCast == value.isOn)
+                    return;
+
+                m_shadowCast = value.isOn;
+                Dirty(ModificationScope.Graph);
+            }
```
pbbastian: Why is `ToggleData` used instead of `bool` if `isEnabled` is discarded?

com.unity.shadergraph/Editor/Drawing/Views/PBRSettingsView.cs
```diff
@@ -36,14 +37,28 @@ public PBRSettingsView(PBRMasterNode node)
...
+                        {
+                            field.value = m_Node.alphaMode;
+                            field.RegisterValueChangedCallback(ChangeAlphaMode);
+                        });
                     });
-                });
+            }
+            else
+            {
+                ps.Add(new PropertyRow(new Label("Cast Shadow")), (row) =>
```
pbbastian:
```suggestion
                ps.Add(new PropertyRow(new Label("Is Shadow Caster")), (row) =>
```

com.unity.shadergraph/Editor/Drawing/Views/PBRSettingsView.cs
```diff
@@ -84,6 +99,14 @@ void ChangeAlphaMode(ChangeEvent<Enum> evt)
             m_Node.alphaMode = (AlphaMode)evt.newValue;
         }
 
+        void ChangeShadowCast(ChangeEvent<bool> evt)
+        {
+            m_Node.owner.owner.RegisterCompleteObjectUndo("Shadow Cast Change");
```
pbbastian:
```suggestion
            m_Node.owner.owner.RegisterCompleteObjectUndo("Change Is Shadow Caster");
```

com.unity.shadergraph/Editor/Drawing/Views/UnlitSettingsView.cs
```diff
@@ -15,9 +15,10 @@ class UnlitSettingsView : VisualElement
         public UnlitSettingsView(UnlitMasterNode node)
         {
             m_Node = node;
-
             PropertySheet ps = new PropertySheet();
 
+            bool opaqueOptions = (m_Node.surfaceType == SurfaceType.Opaque);
```
pbbastian:
```suggestion
            bool showOpaqueOptions = (m_Node.surfaceType == SurfaceType.Opaque);
```

com.unity.shadergraph/Editor/Data/Interfaces/IOptionalShadowPass.cs
```diff
@@ -0,0 +1,9 @@
+using UnityEditor.Graphing;
+
+namespace UnityEditor.ShaderGraph
+{
+    interface IOptionalShadowPass
```
pbbastian: I don't think we should introduce this interface. It's only used internally in methods that already deal with a specific master node type, so it doesn't actually need the indirection. IMO it makes the code harder to follow. I suggest adding a property `isShadowCaster` to `PBRMasterNode` and `UnlitMasterNode`

sebastienlagarde:
Hi, 

I am curious about the motivation for this PR. We have already an option on the mesh itself to not cast shadow https://docs.unity3d.com/Manual/class-MeshRenderer.html why is there a need to do it in Material?

Any stuff like text mesh pro etc... already disable cast shadow.
I could see the usage for Unlit in general but I don't see the motivation for PBR version).

>Only possible for opaque. The option becomes hidden and then disabled for transparent. (Decided on >this after talk with Chris) Could easily change that.

The question of supporting this optional pass in HDRP master node or not could be problematic for the statement above. In HDRP we now (in HDRP/staging) have move renderstate control into the Material (i.e in the Material you can select if you are transparent or opaque). Mean that if you don't generate the pass you will have an issue if a user change it on Material.

As the move of render state control to material is really something that I will recommend for LW (in particular because it will become more general pipeline and will require to have optimization for shader variant), I think not generating the pass will just be like shooting ourselves in the foot (a bit like when we have created multi-scene with lighting settings) :).

My advice: Always generate the shadow pass in the shader, but in the Material you can disable the pass or not based on the settings (we do this in HDRP for Motion vector pass for example to handle motion vector, and LW will need to support that as well in the future to support correct motion blur and TAA for vertex animation). i.e think about the future evolution of LW.

Any thought?

Adding @alelievr for awareness of this PR.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRSubShader.cs
```diff
@@ -442,8 +442,15 @@ public string GetSubshader(IMasterNode iMasterNode, GenerationMode mode, List<st
...
+                bool includeShadowPass = true;
+                IOptionalShadowPass optionalShadow = masterNode as IOptionalShadowPass;
+                if (optionalShadow != null)
+                {
+                    includeShadowPass = optionalShadow.ShadowPassActive();
+                }
+
                 GenerateShaderPassLit(masterNode, m_PassMETA, mode, materialOptions, subShader, sourceAssetDependencyPaths);
-                GenerateShaderPassLit(masterNode, m_PassShadowCaster, mode, materialOptions, subShader, sourceAssetDependencyPaths);
+                if (includeShadowPass) GenerateShaderPassLit(masterNode, m_PassShadowCaster, mode, materialOptions, subShader, sourceAssetDependencyPaths);
```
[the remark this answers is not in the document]
zadi0648: Done.

com.unity.render-pipelines.high-definition/Editor/Material/Unlit/ShaderGraph/UnlitSubShader.cs
```diff
@@ -304,7 +304,14 @@ public string GetSubshader(IMasterNode iMasterNode, GenerationMode mode, List<st
...
 
-                GenerateShaderPassUnlit(masterNode, m_PassShadowCaster, mode, materialOptions, subShader, sourceAssetDependencyPaths);
+                bool includeShadowPass = true;
+                IOptionalShadowPass optionalShadow = masterNode as IOptionalShadowPass;
+                if (optionalShadow != null)
+                {
+                    includeShadowPass = optionalShadow.ShadowPassActive();
+                }
+
+                if (includeShadowPass) GenerateShaderPassUnlit(masterNode, m_PassShadowCaster, mode, materialOptions, subShader, sourceAssetDependencyPaths);
```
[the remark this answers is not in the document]
zadi0648: Done.

com.unity.render-pipelines.lightweight/Editor/ShaderGraph/LightWeightPBRSubShader.cs
```diff
@@ -120,17 +141,33 @@ public string GetSubshader(IMasterNode masterNode, GenerationMode mode, List<str
...
 
+                bool includeShadowPass = true;
+                IOptionalShadowPass optionalShadow = masterNode as IOptionalShadowPass;
+                if (optionalShadow != null)
+                {
+                    includeShadowPass = optionalShadow.ShadowPassActive();
+                }
+
+                if (includeShadowPass)
+                    subShader.AppendLines(GetShaderPassFromTemplate(
```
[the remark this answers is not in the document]
zadi0648: Done.

com.unity.render-pipelines.lightweight/Editor/ShaderGraph/LightWeightUnlitSubShader.cs
```diff
@@ -62,14 +76,20 @@ public string GetSubshader(IMasterNode masterNode, GenerationMode mode, List<str
...
 
             var templatePath = GetTemplatePath("lightweightUnlitPass.template");
-            var extraPassesTemplatePath = GetTemplatePath("lightweightUnlitExtraPasses.template");
-            if (!File.Exists(templatePath) || !File.Exists(extraPassesTemplatePath))
+            var shadowPassTemplatePath = GetTemplatePath("lightweightUnlitShadowPass.template");
+            var depthPassTemplatePath = GetTemplatePath("lightweightUnlitDepthPass.template");
+
+            if (!File.Exists(templatePath) || !File.Exists(shadowPassTemplatePath) || !File.Exists(depthPassTemplatePath))
+            {
+                Debug.LogError("One or more LW Unlit template files not found.");
```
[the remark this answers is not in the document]
zadi0648: No longer applicable

com.unity.render-pipelines.lightweight/Editor/ShaderGraph/LightWeightUnlitSubShader.cs
```diff
@@ -15,16 +15,16 @@ class LightWeightUnlitSubShader : IUnlitSubShader
     {
         static readonly NeededCoordinateSpace k_PixelCoordinateSpace = NeededCoordinateSpace.World;
 
-        struct Pass
+        private struct Pass
```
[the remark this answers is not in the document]
zadi0648: Fixed.

com.unity.shadergraph/Editor/Drawing/Views/PBRSettingsView.cs
zadi0648: So here I wanted to ask you about the Is stuff. If I do this, it becomes inconsistent with the others, such as Two Sided. For example should I change Two Sided to be "Is Two Sided"? I could do that.

com.unity.shadergraph/Editor/Data/Interfaces/IOptionalShadowPass.cs
zadi0648: I did this because I figured it would be good practice if other (or new) master nodes want to make their shadow pass option. I can still remove this, but I wanted to run the reason past you. Thoughts?

com.unity.render-pipelines.universal/Editor/ShaderGraph/UniversalPBRSubShader.cs
```diff
@@ -258,10 +300,18 @@ public string GetSubshader(IMasterNode masterNode, GenerationMode mode, List<str
...
+            if (optionalShadow != null)
+            {
+                includeShadowPass = optionalShadow.ShadowPassActive();
+            }
+
             // Passes
             var forwardPass = pbrMasterNode.model == PBRMasterNode.Model.Metallic ? m_ForwardPassMetallic : m_ForwardPassSpecular;
             var forward2DPass = pbrMasterNode.model == PBRMasterNode.Model.Metallic ? m_ForwardPassMetallic2D : m_ForwardPassSpecular2D;
-            var passes = new Pass[] { forwardPass, m_DepthShadowPass, forward2DPass };
+            var passes = includeShadowPass ? new Pass[] { forwardPass, m_ShadowPass, m_ExtraPasses, forward2DPass } : new Pass[] { forwardPass, m_ExtraPasses, forward2DPass };
```
zadi0648: I just wanted to note that I don't think this is ideal, because it's not scalable. My initial plan was to pass in a null if there was no shadow pass, and modify GetSubShader to skip nulls. However this is a Struct array. So I was then going to implement something else such as making it a list, however I thought it ought to be discussed first as doing something like that changes code in many places. I don't intend this to stay like this, however maybe it should until a refactor.

com.unity.shadergraph/Editor/Drawing/Views/PBRSettingsView.cs
zadi0648: Same as above.

zadi0648:
@sebastienlagarde Hey Seb,

Yeah I didn't meant to bypass your comment, I ran out of time to reply as I was pushing the fixes. Indeed we should discuss the future of how we handle this - is this approach right or should control be entirely in the materials?

This would just be the start of making more passes optional, the idea behind it that was discussed here is that some user will want to cut down the number of passes / get rid of passes they know they won't use to make for smaller shaders. For example users who create all shaders for their game, as opposed to using any built in materials, that is likely something they want.

To be clear, the moving of the renderstate control to the material is for built-in materials? Or do you envision it being something any SG generated shader material would support in the future?

Adding @Kink3d & @pbbastian to the discussion.

sebastienlagarde:
@zadi0648 We simply need to define what we are expecting here.
If the goal is to NOT generate a pass, then this is fine as long as there is no exception (i.e all Material from this shader will not use this pass as it doesn't exist).

The issue I see is with this statement:
>Only possible for opaque. The option becomes hidden and then disabled for transparent. (Decided on >this after talk with Chris) Could easily change that.

Then having the pass or not is an option base not only on a single boolean but also on other feature (I am transparent or not). And this is the problem as such thing could be control by the Material.

So there is two level of disabling from my point of view:
The shader don't support a pass at all, no exception. **the pass isn't generated**. No material instance can enable it. Fine.
The shader optionally support a pass. In this case **the pass must always be generated** and the material decide if it is enabled or not.


>This would just be the start of making more passes optional, the idea behind it that was discussed here >is that some user will want to cut down the number of passes / get rid of passes they know they won't >use to make for smaller shaders. For example users who create all shaders for their game, as opposed >to using any built in materials, that is likely something they want.

Note that remove shadow pass don't make shaders smallers. Each pass in shader labs generate a shaders. 
So what it mean is that you have one variant of the shader (or several if there is a multi-compile) that isn't generated. Shader aren't smaller. 
In addition according to what Aras explain me, if a compiled shader is the same than another, it is discard. As Shadow Pass is always the same (except when using alpha test), then shader generated by shadow pass should be discarded and in memory we don't store several variant.

So what people really save here is compile time (which is already great!), but no performance, nor memory. I am not against the PR itself, just we have already an option to achieve the same goal
 https://docs.unity3d.com/Manual/class-MeshRenderer.html but fair enough.

So to sum up. I am fine with the PR IF the shadow pass option is always present and don't depends on surface type, so the shadow pass is permanently removed or not for this shader.

Kink3d:
As this PR is now so far behind master, and the entire pipeline for generation is completely different i am going to close this PR. If we want to implement optional passes the method to do so is now much simpler (And completely different) so we should discuss a proper path for it.

PR #3755, opened 2019-05-31
Fix for decals showing blurry low res mip when using texture mip map streaming

lyndon-unity:
### Purpose of this PR
Fix for decals showing blurry low res mip when using texture mip map streaming
With this fix decals pick the highest mip for streaming rather than defaulting to lowest permitted mip
(due to nothing calculating the required mip level).

Bug report. 
https://fogbugz.unity3d.com/f/cases/1140772/
Note this had a PR to fix it with Unity code but that needs to be reverted due to side effects.

The following PR will undo the above bug fix, so an alternative fix is needed in the Decal system to prevent a regression
https://ono.unity3d.com/unity/unity/pull-request/88212/_/core/graphics/texture-streaming-preload-fix

---
### Release Notes
Fixed decal rendering when texture mip map streaming is enabled.

---
### Testing status
**Katana Tests**: 
Launched Katana
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=decals-texture-streaming-fix&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

- Created a decal with texture containing mips and set to streaming.
- Enabled texture streaming and set max mip level reduction high (to 7)
- Prior to this fix the texture was limited to a low mip so looks blurry.
- With this fix all the mips are loaded and is no longer blurry.
There is also a test in the bug report listed above.

**Automated Tests**: 
- Nothing added

---
### Overall Product Risks
**Technical Risk**: Low, 
**Halo Effect**: Low

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added custom BaseShaderPreprocessor in HDEditorUtils.GetBaseShaderPreprocessorList()
 - Fixed compile issue when USE_XR_SDK is not defined
 - Fixed procedural sky sun disk intensity for high directional light intensities
+- Fixed Decal mip level when using texture mip map streaming to avoid dropping to lowest permitted mip (now loading all mips)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


sebastienlagarde:
Hey, thanks for the PR. Note that this only work for inspector decal. For shader graph decal this code will not have effect (unless texture are name the same way).

PR #3761, opened 2019-06-03
Don't GC alloc in motion blur code

FrancescoC-unity:
A dumb mistake on one of the parameter allocated a new parameter every frame. 

I am using this occasion also to remove that parameter as it is really not a valid one to expose to the user.


sebastienlagarde:
Please add a changelog,

sebastienlagarde:
(and target HDRP/staging)

PR #3781, opened 2019-06-06
Lwrp rename to universal

stramit:
### Purpose of this PR
Rename LWRP -> Universal RP.

---
### Release Notes
See changelogs
---

### Testing status
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=LwrpToUniversal&unity_branch=trunk

- [x] Built a player
- [x] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
-[] Ran projects to validate update works

---
### Overall Product Risks
**Technical Risk**:  HIGH

**Halo Effect**:  HIGH
---

### Comments to reviewers
Forgive me for an unreviewable PR


hippocoder:
A small suggestion if I may?  Name it as Default Pipeline instead of Universal Pipeline. They end up having a similar rationale (covers all bases by default or is universal) but one is easier to type and understand for a wider range of English competency.

stramit:
> A small suggestion if I may? Name it as Default Pipeline instead of Universal Pipeline. They end up having a similar rationale (covers all bases by default or is universal) but one is easier to type and understand for a wider range of English competency.

This led to a 4 hour internal meeting with marketing and a lot of the graphics people at unity. In the end we will continue with Universal. Big concerns with default is that is what some people are calling the 'built-in' pipeline currently and we really want to avoid that confusion.

PR #3785, opened 2019-06-07
Time input & motion vectors

pbbastian:
### Purpose of this PR
Extracts time as a graph input, such that HDRP Shader Graphs can calculate motion vectors for time-based vertex animation.

A vertex animated Shader Graph in HDRP with TAA and motion blur enabled:
loop

Motion vector debug view in HDRP:
SGMotionVectors

---
### Release Notes
See updated SG and HDRP change log.

---
### Testing status
**Katana Tests**: On the way

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=sg%2Ftime-input&automation-tools_branch=master&unity_branch=trunk#

**Manual Tests**: Manually verified with HDRP motion vector debug view that motion vectors get generated and look correct. Also verified that the shaders work fine with TAA and motion blur in HDRP. See GIFs above.

**Automated Tests**: I tried to set up a test scene, but since this feature deliberately only works with global time, I couldn't get it to produce consistent results for the test.

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
If you want to test this out, remember to enable "Motion Vector For Vertex Animation" on the material.


com.unity.render-pipelines.high-definition/Editor/ShaderGraph/VertexAnimation.template.hlsl
```diff
@@ -35,10 +35,13 @@ VertexDescriptionInputs AttributesMeshToVertexDescriptionInputs(AttributesMesh i
     return output;
 }
 
-AttributesMesh ApplyMeshModification(AttributesMesh input)
+AttributesMesh ApplyMeshModification(AttributesMesh input, float time, float sinTime, float cosTime)
```
sebastienlagarde: Arf, it is really too bad and awfult that we need to do such prototype for this function... I know it isn't related to this PR but due to legacy Unity format. 

In HDRP we can change thing, and I really would like to move to 
_Time containing float4 (Time, Sin(Time), Cos(Time), ?) instead. Current version with the various scale factor useless, is it used in LW 5(?)
Same for _LastTime.
We can do the change in HDRP easily. Or if this is really a problem we can create new variables
i.e _TimeParameter, _LastTimeParameter instead, and then do 

AttributesMesh ApplyMeshModification(AttributesMesh input, float4 timeParameter)
and use float4 all allong the code.

Thought? Is it doable for LW ? in HDRP it is in HDCamera.cs, we just need to add this.

My wory is that here we define a prototype like this:
AttributesMesh ApplyMeshModification(AttributesMesh input, float time, float sinTime, float cosTime)
for the next severals years...

Still I suggest that we merge this PR as it is (as it is a feature and we need it for fixing motion vector on our foliage) and do the change as a "fix"

sebastienlagarde:
Thanks for this PR! See the comment above about merging all _Time and _CosTime/Sin_Time parameter in a single variable. If it is doable for LW we should really do it, this will avoid to pollute the code everywhere and just have _TimeParameters / _LastTimeParameters  instead of three of them that also don't exist for _LastTime (and LW in the future will require it too and you don't want to do sin / cos in shader for LW...)

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/VertexAnimation.template.hlsl
pbbastian: Yeah, I was actually considering doing that, as this prototype is not so nice :( If there's not anything sensible to put in the 4th component, what do you think about making it a float3 (Time, Sin(Time), Cos(Time))?

sebastienlagarde: I am fine with float3. 
Still if we need something else in the future (like delta time :) ), at least we have a slot available if we used float4 without breaking the API. And for the uniform we should use float4 anyway, so I guess we can keep it float4. But as you wish, not a strong opinion here.

sebastienlagarde:
Launched Katana: I get 15 fail. Most are motion blur test, so it is expected as now motion vector works with shader graph. However 3 are about Terrain because shader don't compile

Unhandled log message: '[Error] Shader error in 'HDRP/TerrainLit': 'ApplyMeshModification': no matching 4 parameter function at Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/VertMesh.hlsl(111) (on d3d11)

pbbastian:
@zadi0648 Like I said in the description, I tried to add a test and did not succeed due to the nature of the feature. It makes sense to add a test when we eventually support motion vectors for user-defined properties, and not just global time. Or maybe if there's a way to override global time.

sebastienlagarde:
I have update the PR with my suggestion on _TimeParameters for both LW and HD. I have also removed the wind code from HDRP that isn't used anymore and was relying on the time.

sebastienlagarde:
Katana is green, as explain in the comment I will merge this PR for 19.2 purpose

PR #3797, opened 2019-06-11
Refactored forward pass in preparation for Render Graph

JulienIgnace-Unity:
### Purpose of this PR
- Refactored forward pass in preparation for Render Graph
- Removed FrameSettings member from light loop as it wasn't necessarily always up to date and caused errors (FrameSettings should be passed with HDCamera when needed)

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Frendergraph-refacto-forward&automation-tools_branch=master&sort=1-asc)

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1900,7 +1875,7 @@ void Callback()
                 RenderSubsurfaceScattering(hdCamera, cmd, hdCamera.frameSettings.IsEnabled(FrameSettingsField.MSAA) ? m_CameraColorMSAABuffer : m_CameraColorBuffer,
                     m_CameraSssDiffuseLightingBuffer, m_SharedRTManager.GetDepthStencilBuffer(hdCamera.frameSettings.IsEnabled(FrameSettingsField.MSAA)), m_SharedRTManager.GetDepthTexture());
 
-                RenderForward(cullingResults, hdCamera, renderContext, cmd, ForwardPass.OpaqueEmissive);
+                RenderDecalsForwardEmissive(hdCamera, cmd, renderContext, cullingResults);
```
sebastienlagarde: Call it RenderForwardEmissive as it is possible that in the future we do emissive object here too

JulienIgnace-Unity: Fixed in latest commit

PR #3814, opened 2019-06-13
Grant internal access to ShaderGraph for future SpeedTree package

parashar-unity:
### Purpose of this PR
We are planning to move a lot of SpeedTree support into its own package, and since this will include some ShaderGraph nodes, the idea is to give that package access to ShaderGraph internals.

---
### Release Notes
This is basically just future-proofing for when the SpeedTree package actually exists.  There is an intention to backport to 2019.2 (and package v6.8?) as well, so that when the SpeedTree package is created, it will "just plain work"(tm) all the way back to 2019.2.

---
### Testing status
Other than verifying that nothing gets broken (which has been verified in editor, and also being checked in Katana), there's no need for further testing beyond that at this stage as the package in question does not yet exist.

---
### Overall Product Risks
**Technical Risk**: None... yet.
**Halo Effect**: None... again, yet.

---
### Comments to reviewers
Yes, I know I am putting this in a branch labeled terrain, but since it's such a minuscule change, I didn't feel the need to create a whole other branch just for this.  I'm just using a branch I'm anyway going to use for another separate PR.

### DIFF
--- a/com.unity.shadergraph/Editor/AssemblyInfo.cs
+++ b/com.unity.shadergraph/Editor/AssemblyInfo.cs
@@ -3,3 +3,4 @@ using System.Runtime.CompilerServices;
 [assembly: InternalsVisibleTo("Unity.ShaderGraph.Editor.Tests")]
 [assembly: InternalsVisibleTo("Unity.RenderPipelines.Lightweight.Editor")]
 [assembly: InternalsVisibleTo("Unity.RenderPipelines.HighDefinition.Editor")]
+[assembly: InternalsVisibleTo("Unity.SpeedTree.Editor")]


pbbastian:
I'm not entirely comfortable with this. Where will the package live? Our rule so far has been that only packages that are part of the SRP release cycle can have access to internals. This is because we aren't properly following semantic versioning yet, and so breaking changes to API is not tracked properly.

So if the package is going to live in the SRP repository and be part of the SRP release cycle, then there is no need to add internal access until then. This is because we only support SRP packages at the exact same version.

If the package is not going to live in the SRP repository, then I can't approve this until the SRP packages are versioned independently, and follow semantic versioning.

What functionality that you need for this is missing from CustomFunctionNode? Because that can be expected to not break.

cdxntchou:
What breaks in a separate repo?  Is it just that the SG version gets bumped unnecessarily, causing separate packages (i.e. speedtree package) to suddenly become incompatible?

cdxntchou:
I'm hoping it doesn't need internal access, and we can get by with CustomFunctionNode.
But if not, would be nice if we have the option to release a 2019.2 compatible version of the package.

parashar-unity:
In theory, the most likely case for needing internal access would be the
custom UI, and while you could avoid all that by just having everything in
SRP itself, that effectively defeats the purpose of the SpeedTree package.

cdxntchou:
Also, semantic versioning does't really apply to internal access, does it?  Semantic versions don't get bumped for internal only changes, just for public API.  I'm thinking we need to do very explicit version checks in the speedtree package to verify it is compatible with the SG version.

pbbastian:
> What breaks in a separate repo? Is it just that the SG version gets bumped unnecessarily, causing separate packages (i.e. speedtree package) to suddenly become incompatible according to semantic versioning rules? (even though they are still actually compatible?)

With SpeedTree in a separate repository, we can no longer require that users run the exact same version like we currently do across SRP packages. Since the repository isn't tracking breaking changes via semantic versioning this leaves users with absolutely no indicator as to which versions will work together. As in they will just get a spew of compilation errors, or maybe just bugs. This won't be an issue when packages get independently versioned and use semantic versioning.

> In theory, the most likely case for needing internal access would be the
custom UI, and while you could avoid all that by just having everything in
SRP itself, that effectively defeats the purpose of the SpeedTree package.

We should have a talk about what kind of UI you need, and when you need it :)

> Also, semantic versioning does't really apply to internal access, does it? Semantic versions don't get bumped for internal only changes, just for public API.

It does. When internal APIs get exposed to other packages they effectively become public. From the user-side of things, they'll try to install 2 packages that looks like they should work together, and then get a bunch of weird errors.

> I'm thinking we need to do very explicit version checks in the speedtree package to verify it is compatible with the SG version.

This could be quite a bit of work on your end to keep track of. You would have to verify that each new Shader Graph version didn't change any of the APIs you're using, and then release a new SpeedTree package version that whitelists the new package. Otherwise, you run the risk that a new Shader Graph version _does_ change APIs and then users will just see a bunch of compilation errors if they select the wrong pair of versions. As things are now, those changes could happen even in a patch release.

parashar-unity:
Based on some discussion, I'm thinking I'll suspend this PR and we'll just roll the SpeedTree ShaderGraph implementation into SRP for the short term and work out some way to move that into the package where all the other SpeedTree stuff will one day live and there will be peace and harmony throughout all the lands...  at least until this misery resurfaces anyway.

Also, as the branch name implies, I set up this branch for HDRP terrain shader work, and that's a little higher priority at the moment.

PR #3821, opened 2019-06-13
Lookdev/second draft

RSlysz:
### Purpose of this PR
Second batch of devlopment of the LookDev

It includes inspectors, camera manipulation, user feedbacks and some preparatory work for light composition and debug modes.
Debug modes are included but are still slightly buggy when script reload (without crash) and they affect both view in comparison mode (will be fixed later)

Everything is in an Experimental namespace.

---
### Release Notes
First Experimental version of the LookDev. Works with all SRP. Only branched on HDRP at the moment.

---
### Testing status
**Katana Tests**: 

**Manual Tests**: 
- [x] Opened test project + Run graphic tests locally (all green but 1206)
- [x] Built a player
- Other: 

**Automated Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=lookdev%2Fsecond-draft&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: High

**Halo Effect**: Low

---
### Comments to reviewers
@Chman for the Post-process isolation but we already review it together


com.unity.render-pipelines.core/Runtime/Volume/VolumeManager.cs
```diff
@@ -343,5 +352,30 @@ static void SortByPriority(List<Volume> volumes)
                 volumes[j + 1] = temp;
             }
         }
+
+        static bool IsVolumeRenderedByCamera(Volume volume, Camera camera)
+        {
+#if UNITY_2018_3_OR_NEWER && UNITY_EDITOR
```
Chman: The `UNITY_2018_3_OR_NEWER` define is not needed as we target 19.3 at this point anyway. Not really important but it should be removed at some point for the same of keeping the code clean.

PR #3829, opened 2019-06-13
Hdrp/update shadow maps at custom rates

FrancescoC-unity:
### Purpose of this PR

This PR introduces a first pass at a rudimentary caching mechanism for shadow map. What it does is simply not clearing the atlas region with the cached shadow and forces the usage of the old data when rendering with the shadow. 

This at the moment allows for the following update policies: 

- Every frame: same as now. 
- On Enable: renders the shadow map only when the light is enabled. 
- On Demand: renders the shadow map every time the users invoke the method RequestShadowMapRendering() 

---
### Release Notes
Adding option to update shadow maps only On Enable and On Demand

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fupdate-shadow-maps-at-custom-rates&automation-tools_branch=master&unity_branch=trunk [Just kicked]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**:  Medium


FrancescoC-unity:
Ouch conflicts, will try to resolve asap.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -34,6 +34,9 @@ public struct HDShadowData
 
         public Vector4      shadowFilterParams0;
 
+        public Vector3      cacheTranslationDelta;
+        public float        _padding1;
```
sebastienlagarde: There is no need for padding as we use structure buffer in hlsl

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAtlas.cs
```diff
@@ -332,6 +332,9 @@ RenderShadowsParameters PrepareRenderShadowsParameters()
 
             foreach (var shadowRequest in parameters.shadowRequests)
             {
+                if (shadowRequest.shouldUseCachedShadow)
```
alelievr: You need to check if debugClearAtlas is enabled, if so the atlas is always entirely cleared.

FrancescoC-unity:
> Looks good for me (just the issue with debug clear atlas).
> One question though: How does it behaves with the global shadow scale factor in the debug window ?
> image

It does not change the current state of the shadow map, which I think is the correct thing anyway as changing resolution would require re-rendering. If it is in "on demand" mode, then when retriggering the rendering will take that scale into account.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAtlas.cs
FrancescoC-unity: Good point! I guess in that case the best course of action is to disable caching altogether, otherwise we either don't respect what the debug says (i.e. we don't clear the cached areas) or we need to move the cached shadow to another target with all the associated costs that come with that.

JMargevics:
[Link to Confluence report](https://confluence.hq.unity3d.com/display/HDRP/Update+shadow+maps+at+custom+rates) Since the modes are meant for static lights I now think it would be nicer to move the object around and leave the light. But otherwise everything looks good.

FrancescoC-unity:
Just found a potential issue while playing a bit more with the rescaling. I might change how that works in the cached case (I might have to reforce the update).

FrancescoC-unity:
> Just found a potential issue while playing a bit more with the rescaling. I might change how that works in the cached case (I might have to reforce the update).

Ok this is done! Now writing the docs .

PR #3866, opened 2019-06-17
Adding a standard Lit model that is our go to for indirect lighting for tier1

anisunity:
All in the title

This includes previous PRs


com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.shader
```diff
@@ -461,8 +461,10 @@ Shader "HDRP/Unlit"
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Material.hlsl"
 
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/Deferred/RaytracingIntersectonGBuffer.hlsl"
-
             #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.hlsl"
+            #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/NormalBuffer.hlsl"
```
sebastienlagarde: do we need normal buffer for unlit?

anisunity: You are right, i do not rememeber why i did add it. I think it was a bad dependency that i did not investigate

PR #3696, opened 2019-05-23
Merge Hdrp/staging to master

sebastienlagarde:
### Purpose of this PR
Why is this PR needed, what hard problem is it solving/fixing?

- Disable sky when in matcap view (#3667)
- Update AdditionalShadowData.cs (#3665)
- Remove async compil selection (#3675) 
- Better handling of fully metallic objects in matcaps (#3677)
- Fix default contact shadow texture type mismatch (#3676)
- Moving Indirect Diffuse, Recursive Rendering and Light cluster data to the volume system (#3664)
- Small pass of refactoring for Render Graph preparation (#3674)
- Hdrp/fix generate shader includes (#3684)
- Hdrp/fix cleaned texture in cascade inspector (#3685)
- Update CSharpToHLSL.cs (#3686)
- Fix crash in FrameSettingsHistory when double disabling a camera (#3691)
---
### Release Notes
Part of each PR

---
### Testing status
**Katana Tests**: First off we need to make sure the Katana SRP tests are green?

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fstaging&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---

PR #3870, opened 2019-06-17
Hdrp/terrain layer blending and custom UI improvements

parashar-unity:
### Purpose of this PR
This brings some of the changes that the corresponding changes going into LWRP terrain shader over to the HDRP version so that they are both basically identical in functionality.  The main addition is that it allows backwards compatibility support for legacy-style blending operations to be available on a per-layer basis.  This means that mixed setups with some masked and some un-masked layers will still blend properly, even with height-based or density-based blending.

As a result, this change also fixes these issues -
https://fogbugz.unity3d.com/f/cases/1150934/
https://fogbugz.unity3d.com/f/cases/1161116/

UI changes affect which values are used in the non-mask mode layers help by maintaining the already existing separation between the remap controls and the conventional sliders (at least those that are already there).  This means that changes made in one mode do not "stomp" on one another when you add or remove a mask map.

---
### Release Notes
Core changes that are needed for support went into 2019.3.0a5 with 2019.2 and 2019.1 backports in flight.  Since most of the advanced blending operations are already present in HDRP, the total set of changes are relatively small compared to LW, which actually needed all that functionality to be added to the shader in the first place.

---
### Testing status
**Katana Tests**: Tests are green in build #3696

**Manual Tests**:
- Created test project + Run graphic tests locally
- Built a player
- Cleared C# and shader warnings

Katana tests link --
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2FterrainLayer&unity_branch=trunk&automation-tools_branch=master&sort=1-asc

---
### Comments to reviewers
Parametrization changes in the UI do not affect serialized data or compatibility thereof.  They are merely there to allow the UI controls to make slightly more sense.

---
### Overall Product Risks
**Technical Risk**: Medium.  Causes some changes to HDRP terrain shader, but the main effect is backwards compatibility between masked and "legacy" mode for blending layers can operate on a per-layer level.  Katana tests also indicate that existing content should not be affected unless it was already running in a buggy state.
**Halo Effect**: Low.  Only affects terrain shader and terrain layer UI.


com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit_Splatmap.hlsl
```diff
@@ -1,4 +1,8 @@
 TEXTURE2D(_Control0);
+#ifndef _TERRAIN_BASEMAP_GEN
+// To avoid duplicate definitions, we put this inside the #ifdef
```
pbbastian: How about putting include guards on the file, and then only have the field defined one time in one file?

parashar-unity: That's actually what I tried initially.  I felt like this was cleaner because it made the fewest changes, and didn't require reordering the includes or making includes that were otherwise unnecessary actually necessary.  Moving the definition of the field to that file means moving it out of the CB_DEBUG_VARS. All fine if TerrainLit_Splatmap.hlsl is included, which unfortunately, it isn't except for BasemapGen, so either way, you're going to have multiple definitions unless you want to clutter all the files with a full include of that TerrainLit_Splatmap.hlsl

There are still a few ways to do it without as much include hell, but to me, any of them would constitute more files modified, which I kind of felt was uncalled for for something this minor.

pbbastian: Ah okay, makes sense :)

com.unity.render-pipelines.high-definition/Editor/Material/TerrainLit/TerrainLitGUI.cs
```diff
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.Rendering;
+using UnityEngine.Experimental.Rendering;
```
sebastienlagarde: We should remove experimental now (HDRP is out of experimental)

sebastienlagarde:
@parashar-unity . Sorry for the delay regarding this PR. 
LW have been rename universal - so the file don't exist anymore, could you merge the PR with master and report change on the correct file? Then relaunch katana.

also @zeroyao can you validate the PR? thanks

sebastienlagarde:
@parashar-unity I get this error in katana:
Packages\com.unity.render-pipelines.high-definition\Editor\Material\TerrainLit\TerrainLitGUI.cs(155,24): error CS0103: The name 'GraphicsFormatUtility' does not exist in the current context
Packages\com.unity.render-pipelines.high-definition\Editor\Material\TerrainLit\TerrainLitGUI.cs(155,62): error CS0103: The name 'GraphicsFormatUtility' does not exist in the current context
-----CompilerOutput:-stderr----------
-----EndCompilerOutput---------------

Could you fix it and merge master again and trigger katana. Sorry about all this sucessive failure, but seems to be due to C++ change out of our control... THanks!

parashar-unity:
Ah...  yeah, IIRC, that was the only reason I needed to include
Rendering.Experimental in the first place -- just to have access to that.
I had a feeling that would happen so I ran that test last night.

sebastienlagarde:
After lot of back and forth with Katana I was finally able to get final result of this PR for graphic test.
There is two test failing: 

https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Windows/31297670_12_07_2019_12_32_41_%2B0000/logs/TestResults_playmode_HDRP_Tests.xml

TerrainLit and another one related to Terrain . I can get the same issue when launching test locally, there is a shit of position of terrain element. Is it expected ? I was assuming the PR shouldn't change anything.

Was wondering if it was the change in the UI but as it is in the UI, it only apply if you touch something form the test. Which maybe should be done. Any thought about what is happening?

parashar-unity:
The UI change is nothing more than a display change so that fields related
to height-based blending are all using the same units (the existing UI uses
different units for different fields presumably so that numbers in one
field aren't an order of magnitude larger than the other) -- it doesn't
actually change anything in the actual stored TerrainLayer data, nor does
it change anything in the values moved downstream to the shader.

There's nothing in the PR that should change positions per se.  There is,
however, a change in the splatmap UV when sampling, though.  That is
actually a bugfix that fixes alignment when going across terrain tiles, but
if the tests have canonicals with the bug in place, it makes sense that a
test should fail.  I will take a look to see if the issue is actually
related to that, which if it is, the canonical image itself needs a change.

sebastienlagarde:
@parashar-unity .

I have update the two screenshots to move forward with the PR and have it in 7.0.0 package. But please, look at the diff of the screenshots and provide a fix for next pacakge (7.1.0) of HDRP if it appears that the change was not expected. Thanks

sebastienlagarde:
@parashar-unity . on request of @phi-lira I have revert all change from LW and Universal. So it isn't merged. You will required to do a separate PR for LW / Universal. Check with Felipe for it. Thanks!

PR #3878, opened 2019-06-18
Simplify the shadow biases we have

FrancescoC-unity:
This simplifies a bit the biases in the shadow options. While the previous options when combined can definitively be superior, they were harder to tune (due to the amount of knobs to turn) and more expensive to run. 

Current solution is a standard dirty constant bias + normal bias  (cleaned up a tiny bit the code to prune dead code).  It might have some downfalls compared to the other method, but it doesn't have other and it is cleaner+faster. 

This whole thing was discussed (and greenlit) by Laurent. 

_The heuristics to determine the bias are possibly going to change once we can get more data point with usage._ 

A good side effect of this is that is fairly easy to explain how to tune. The constant bias if pushed too high will cause peter panning (solved by contact shadows), the normal bias gives better acne fixing tools but it distorts shadows (can create problems with contact shadows). So we can simply explain:

> If you use contact shadows: push the constant bias higher more freely and normal bias less. 
> If you _don't_ use contact shadows: rely more on the normal bias and keep the constant bias as low as possible. 
> 

I might change the default values soon though. 

As far as perf gain go: 

**VGPR improvements:** 

Variant3 (the lit standard + punctual): -12 VGPR
Variant13: -8 VGPR
Variant23: -12 VGPR
Variant26: -4 VGPR

ShadowMask Variant3 : -12 VGPR
ShadowMask Variant6: -4 VGPR
ShadowMask Variant13: -4 VGPR
ShadowMask Variant18: -8 VGPR
ShadowMask Variant23: -12 VGPR
ShadowMask Variant26: -4 VGPR

Note that this change will also require probably some pass over reshuffling code that might even give us a bit more. 

**Timing**

*Very very* simple scene and materials. 

Forced Forward:  **-9%** 
Deferred: **-14%**


FrancescoC-unity:
I found few problems at high qualities, I'll update content of review asap.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -498,13 +527,8 @@ void SetCommonShadowRequestSettings(HDShadowRequest shadowRequest, Vector3 camer
...
             float n = shadowNearPlane;
             shadowRequest.zBufferParam = new Vector4((f-n)/n, 1.0f, (f-n)/n*f, 1.0f/f);
-            shadowRequest.viewBias = new Vector4(m_ShadowData.viewBiasMin, m_ShadowData.viewBiasMax, m_ShadowData.viewBiasScale, 2.0f / shadowRequest.deviceProjectionYFlip.m00 / viewportSize.x * 1.4142135623730950488016887242097f);
-            shadowRequest.normalBias = new Vector3(m_ShadowData.normalBiasMin, m_ShadowData.normalBiasMax, m_ShadowData.normalBiasScale);
-            shadowRequest.flags = 0;
-            shadowRequest.flags |= m_ShadowData.sampleBiasScale     ? (int)HDShadowFlag.SampleBiasScale : 0;
-            shadowRequest.flags |= m_ShadowData.edgeLeakFixup       ? (int)HDShadowFlag.EdgeLeakFixup : 0;
-            shadowRequest.flags |= m_ShadowData.edgeToleranceNormal ? (int)HDShadowFlag.EdgeToleranceNormal : 0;
-            shadowRequest.edgeTolerance = m_ShadowData.edgeTolerance;
+            shadowRequest.worldTexelSize = 2.0f / shadowRequest.deviceProjectionYFlip.m00 / viewportSize.x * 1.4142135623730950488016887242097f;
```
sebastienlagarde: can you replace by sqrt(2) here instead of 1.4142135623730950488016887242097f?

FrancescoC-unity: Sure, this is a copy paste of what was there :)

sebastienlagarde:
Relaunched katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fslim-down-biases&automation-tools_branch=master&unity_branch=trunk#

sebastienlagarde:
katana is green, merging

PR #3879, opened 2019-06-18
Refactored Decal Rendering in preparation for RenderGraph

JulienIgnace-Unity:
### Purpose of this PR
- Refactored Decal Rendering in preparation for RenderGraph
- Main functions are now static and moved logic from DBufferManager to HDRenderPipeline

---
### Testing status
**Katana Tests**: [Running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Frendergraph-refaco-decals&automation-tools_branch=master&unity_branch=trunk)

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -123,6 +123,10 @@ public partial class HDRenderPipeline : UnityEngine.Rendering.RenderPipeline
         ShaderTagId[] m_ForwardErrorPassNames = { HDShaderPassNames.s_AlwaysName, HDShaderPassNames.s_ForwardBaseName, HDShaderPassNames.s_DeferredName, HDShaderPassNames.s_PrepassBaseName, HDShaderPassNames.s_VertexName, HDShaderPassNames.s_VertexLMRGBMName, HDShaderPassNames.s_VertexLMName };
         ShaderTagId[] m_DecalsEmissivePassNames = { HDShaderPassNames.s_MeshDecalsForwardEmissiveName, HDShaderPassNames.s_ShaderGraphMeshDecalsForwardEmissiveName };
         ShaderTagId[] m_SinglePassName = new ShaderTagId[1];
+        ShaderTagId[] m_Decals4RTPassNames = { HDShaderPassNames.s_MeshDecalsMName , HDShaderPassNames.s_MeshDecalsAOName , HDShaderPassNames.s_MeshDecalsMAOName, HDShaderPassNames.s_MeshDecalsSName ,
```
sebastienlagarde: Caution: I guess there is a mistake here, HDShaderPassNames.s_MeshDecalsSName should also be in the list of m_Decals3RTPassNames (and we must set keyword somewhere in the code)
// All decal pass name can be see in Decalsystem.s_MaterialDecalPassNames and Decalsystem.s_MaterialSGDecalPassNames
                // All pass that have 3RT in named are use when perChannelMask is false. All 4RT are used when perChannelMask is true.
                // There is one exception, it is DBufferProjector_S that is used for both 4RT and 3RT as mention in Decal.shader
                // there is a multi-compile to handle this pass, so it will be correctly removed by testing m_Decals3RT or m_Decals4RT

sebastienlagarde: you can put the comment above here (it come from the stripper)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2531,8 +2531,9 @@ DepthPrepassParameters PrepareDepthPrepass(CullingResults cull, HDCamera hdCamer
 
             var result = new DepthPrepassParameters();
 
+            bool decalsEnabled = hdCamera.frameSettings.IsEnabled(FrameSettingsField.Decals);
```
sebastienlagarde: just for more explanation. the use of m_DbufferManager.enableDecals was initially to only trigger renderdecal if there is decal on screen and framesettings was enabled. But of course we have never been able to implement it as it rely on C++. So here you have change the semantic of the test. But not a big deals we can re-introduce the correct term later if this work on day.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2675,94 +2676,146 @@ void RenderGBuffer(CullingResults cull, HDCamera hdCamera, ScriptableRenderConte
...
+            {
+                // We still bind black textures to make sure that something is bound (can be a problem on some platforms)
+                m_DbufferManager.BindBlackTextures(cmd);
                 return;
+            }
+
+            // We need to copy depth buffer texture if we want to bind it at this stage
+            CopyDepthBufferIfNeeded(cmd);
 
             using (new ProfilingSample(cmd, "DBufferRender", CustomSamplerId.DBufferRender.GetSampler()))
```
sebastienlagarde: Maybe we should add a global "Render decals" marker ?

PR #3891, opened 2019-06-19
Rename PBR sky to physically based sky

sebastienlagarde:
Rename PBR sky to physically based sky


EvgeniiG:
Can we port it to master too, please. Thanks.

PR #3900, opened 2019-06-20
Remove IMS setting from shadows 

FrancescoC-unity:
Due to artist request :-) 

Code is left there to allow advanced users of HDRP to bring it back. I left comments to help in the process HDShadowSampling.hlsl  file.

I however removed the tier from the multi compile stuff to reduce variants since we don't use it.


sebastienlagarde:
please add changelog :)

sebastienlagarde:
katana is green, merging

PR #3902, opened 2019-06-20
Use medium as default for shadows in deferred

FrancescoC-unity:
As in the title... 

... Although I just realized this will require to take visual tests all over again... Maybe worth merging after the bias one as that will change the same tests.


FrancescoC-unity:
Closing as merged in https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3878

PR #3913, opened 2019-06-25
Change deprecated UIElement to use new accessor.

marctem:
### Purpose of this PR
UIE team removed `borderColor`, replaced it with top, bottom, left, and right. We needed to change to get rid of a deprecation warning that breaks tests.
---
### Testing status
**Yamato**: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Ffix-border-color

### Overall Product Risks
**Technical Risk**: None - tiny visual change to fix Color Mode

**Halo Effect**: None - one file only


sebastienlagarde:
Merging as katana is failing on master. Note that yamato is currently expected to fail on HDRP playmode test, so we can't rely on it (out of memory)

marctem:
> Merging as katana is failing on master. Note that yamato is currently expected to fail on HDRP playmode test, so we can't rely on it (out of memory)

Was just digging into that to see, everything else passed so thanks for merging!

PR #3927, opened 2019-07-02
Change shadow resolution settings so that by default use shadow tiers

FrancescoC-unity:
### Purpose of this PR
Introduce shadow resolution tiers, then changes the resolution settings so that by default resolution is set with pre sets that are defined on the asset. 

User can still select custom resolutions, but they will be clamped by the max value set on the assets

---
### Release Notes
Introduced shadow resolution tiers and changed how the resolutions are decided on shadow maps. 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fshadow-tier-settings&automation-tools_branch=master&unity_branch=trunk [Running]

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -75,7 +75,10 @@ protected static class HDStyles
             public static readonly GUIContent ColorTemperature = EditorGUIUtility.TrTextContent("Color Temperature");
             public static readonly GUIContent Shadows = EditorGUIUtility.TrTextContent("Shadows");
             public static readonly GUIContent ContactShadows = EditorGUIUtility.TrTextContent("Contact Shadows");
-            public static readonly GUIContent ShadowResolution = EditorGUIUtility.TrTextContent("Shadows Resolution");
+            public static readonly GUIContent ShadowResolutionTier = EditorGUIUtility.TrTextContent("Shadows Resolution Tier");
+            public static readonly GUIContent UseCustomShadowResolution = EditorGUIUtility.TrTextContent("Use Custom Shadow Resolution");
```
sebastienlagarde: you miss this update about quality settings

sebastienlagarde:
katana is green - but now we conflict we antoine PR that in between have gone in....

PR #3953, opened 2019-07-03
Occlusion mesh support with XR SDK

fabien-unity:
### Purpose of this PR
Render occlusion mesh from XR SDK into depth buffer to exclude invisible pixels

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=hdrp-xr-sdk-occlusion-mesh&unity_branch=trunk&sort=1-asc)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/XROcclusionMesh.shader
```diff
@@ -0,0 +1,47 @@
...
+
+        struct Varyings
+        {
+            float4 vertex : SV_POSITION;
+        };
+
+        Varyings Vert(Attributes input)
+        {
+            Varyings output;
+            output.vertex = mul(GetRawUnityObjectToWorld(), input.vertex);
```
sebastienlagarde: Wont you loose precision here if with VR you are far away from the center ? shouldn't we still use camera relative?

fabien-unity: in this case, it's only used for screenspace occlusion meshes

PR #3320, opened 2019-03-29
Exclude objects that tag the stencil with a specific bit from TAA [2019.2]

FrancescoC-unity:
### Purpose of this PR
This PR allows part of the frame to not execute TAA, but go straight through with just the unjittered sample of the scene color. 

The client for this is particles, so the tagging of the stencil will need to happen on VFX side. 

I know reviewers are not in this week, but I thought of opening the PR anyway for book keeping 🙂 

---
### Testing status
**Katana Tests**:  https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fexclude-from-taa&unity_branch=trunk [ Green ] 

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [ ] C# and shader warnings (supress shader cache to see them)
- Other: 

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low

PR #3963, opened 2019-07-04
Exclude objects that tag the stencil with a specific bit from TAA (bis)

FrancescoC-unity:
This is essentially exactly the same PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3320  but manually ported to current state (so much as changed) 

@julienf-unity for visibility on what needs to be tagged from VFX (feature is not explicitly exposed in HDRP) 

New Katana: 
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fexclude-objects-from-taa-bis&unity_branch=trunk&sort=1-asc


sebastienlagarde:
@FrancescoC-unity you need to update the HDRP resources asset too, thanks

PR #3973, opened 2019-07-05
Fix compilation issues on player in HDAdditionalLightData.cs 

FrancescoC-unity:
Fix compilation errors when building player.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -2530,8 +2530,9 @@ public void SetShadowDimmer(float shadowDimmer = 1, float volumetricShadowDimmer
         /// </summary>
         /// <param name="lightmapBakeType"></param>
         /// <returns></returns>
+#if UNITY_EDITOR
         public LightmapBakeType SetLightmapBakeType(LightmapBakeType lightmapBakeType) => legacyLight.lightmapBakeType = lightmapBakeType;
```
FrancescoC-unity: lightmapBakeType is defined only in editor apparently.

alelievr: Since this is public part of the public API, we must move it to this file: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.high-definition/Editor/Lighting/Light/HDAdditionalLightDataEditorExtension.cs because it's editor only.
If we let it here, and users start to use it them wont be able to compile. By putting the function in our editor namespace they will be aware that this function is editor only

FrancescoC-unity: Wasn't aware of that file ;)

PR #3975, opened 2019-07-05
DXR Indirect Diffuse Deferred

anisunity:
Implement a deferred mode for Indirect diffuse
Refactor Deferred Lighting to be compatible with render graph


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.shader
```diff
@@ -865,6 +865,7 @@ Shader "HDRP/Lit"
             #pragma multi_compile _ LIGHTMAP_ON
             #pragma multi_compile _ DIRLIGHTMAP_COMBINED
             #pragma multi_compile _ DYNAMICLIGHTMAP_ON
+            #pragma multi_compile _ DIFFUSE_LIGHTING_ONLY
```
sebastienlagarde: Arf!

PR #3981, opened 2019-07-06
Hdrp/additional velocity changes support for alembic

jeanblouin:
### Purpose of this PR
This PR is adding support for additional velocity changes in HDRP materials for alembic models.

Alembic models can be imported with additional velocity information per vertex.
With this PR we will enable to use this additional velocity information in the motion vector pass.

The alembic package will set the Texcoord5 with the velocity information. This information
is then used to modify the mesh vertex previous position.

Added  new shader property and a #define that will create variants if this feature is enabled.
By default this feature is not enabled.

The same changes are applied to the hd shaders as well as for the master node for the shader graph versions.
---
### Release Notes
This PR adds a new advanced option to the HDRP materials Additional Velocity Change.
This option is also added in the ShaderGraph master node for the hdrp materials.

add_velocity_option
advanced_option_litmat
alembic_velocity_support_in_hdrp

---
### Testing status

- [no] Built a player  (Error in \Lighting\Light\HDAdditionalLightData.cs(2613,42): error CS0246: The type or namespace name 'SerializedProperty' could not be found (are you missing a using directive or an assembly reference?) Not related to this branch.

- [X] Checked new UI names with UX convention
- [X] Tested UI multi-edition + Undo/Redo
- Other: 

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Fadditional_velocity_change_for_alembic&automation-tools_branch=master&unity_branch=trunk#

---
### Overall Product Risks
**Technical Risk**: Medium
This PR adds a new options (additional velocity for alembic) to most HDRP shaders.
**Halo Effect**: Low.
If you don't enable this opt-in feature this should have minimal impact.
---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.shader
```diff
@@ -92,6 +95,7 @@ Shader "HDRP/Unlit"
 
     #pragma target 4.5
     #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
+    #pragma enable_d3d11_debug_symbols
```
sebastienlagarde: Caution, this must not be push

sebastienlagarde:
Nice! I launched katana 
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp%2Fadditional_velocity_change_for_alembic&automation-tools_branch=master&unity_branch=trunk#

One request. Could you test that it work currently with a HDLit shader graph? 

Then we are good

com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.shader
jeanblouin: Oops, forgot to remove this, thanks for making the fix in 5f385db.

sebastienlagarde:
katana is green for windows, which is sufficient for these test. Merging. Note, give feedback about shader graph test. thanks

johnpars:
Tested and working on all shader graphs, barring Unlit. I can push a patch.

One thing of note however is that while normal HD material UI no longer requires the "Motion Vector for Vertex Animation" box, it looks like it still exists on shadergraph material UI block. Would it make more sense to move the additional velocity option under that? Otherwise, motion vector settings would be spread out, which might be confusing.
image

sebastienlagarde:
> looks like it still exists on shadergraph material UI block.
yes, it exist in case you are doing vertex animation in shader graph, then you need to enable the motion vector pass.

sebastienlagarde:
@johnpars So for your other comment, what we should do is that we should automatically enable the motoin vector pass if the option additonalVelocityChange is enabled.

This control is in ShaderGraphUIBlock in function

 void DrawMotionVectorToggle()

currently we test this:
enabled = EditorGUILayout.Toggle("Motion Vector For Vertex Animation", enabled);
but we should add a test on additonalVelocityChange  == true IF the property exist. this will avoid to have to add a button.

johnpars:
Yes- I'm on it

PR #3988, opened 2019-07-08
Introducing shadow resolution quality tiers

FrancescoC-unity:
Same as https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3927 , just manually ported after big antoine PR landed (solving merge conflict was more of a pain than port manually) 

Ran the test locally and they all pass, the original katana was green too.  I can re-launch if needed, but didn't yet to not put pressure on katana.


sebastienlagarde:
ok, please close the other PR. Thanks

PR #3996, opened 2019-07-09
Hdrp/alembic motionvectors shadergraph patch

johnpars:
### Purpose of this PR
This is an update to the recently added `_ADDITIONAL_VELOCITY_CHANGE` keyword added to the motion vector pass, for alembic velocity support.

This is a shadergraph UI update that will automatically enable the motion vector pass if the additional velocity change is requested in the graph.

**The effected shader graphs are: Lit, Unlit, Stacklit, Fabric, Hair**

**The PR also includes a minor bug fix in `BaseUnlitGUI.cs`**

---
### Release Notes

Motion vector pass will be automatically enabled when a user compiles a shader graph with additional velocity changes.

AlembicShadergraphMotionVectors

---
### Testing status

- [x] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Medium: This PR adds a new shader property generation to multiple graphs.

**Halo Effect**:  Low


sebastienlagarde:
Hey, yes, the way you have done it is correct (at least we I don't think about other way currently :) ). 
thanks for this PR

sebastienlagarde:
Note: We have real issue with shader graph and material dependency that aren't solve currently. Here your code will work only if users select their material one by one after changing the option in shader graph... we have other similar issue and will try to find a global solution and update this code later.

johnpars:
> Note: We have real issue with shader graph and material dependency that aren't solve currently.

Indeed, I almost felt forced to use a shader_feature, but am glad this way works for now (until the global solution).

PR #4003, opened 2019-07-09
Fix remaining ray tracing shader warnings

anisunity:
All in the title


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/RaytracingLightCluster.compute
```diff
@@ -34,150 +34,151 @@ static const float3 CornerSubDirections[8] = {
...
+        // If this light should be skipped, skip it
+        // if(_RaytracingLightCullResult[lightIdx] == 1) continue;
+        // if(_LightVolumes[lightIdx].active == 0) continue;
+
+        bool intersects = false;
+
+        if (currentLight.shape == 0)
+        {
+            // When the shape we are processing is a sphere, the radius is stored in the three channels
+            float squareRange = currentLight.range.x * currentLight.range.x;
```
anisunity: Warning was due to this line

PR #4016, opened 2019-07-09
Improve parametrization of the PBR sky and the volumetric fog

EvgeniiG:
### Purpose of this PR
Make parametrization more unified and user-friendly.

---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fimprove-vol-param&automation-tools_branch=master&unity_branch=trunk

image

So now we uniformly use Albedo, Attenuation Distance (mean free path) at the sea/base level and Max Height (w.r.t. y = 0) or Max Altitude (w.r.t. the sea level given by the position and the radius of the planet).

The major remaining difference is the fact the volumetric fog is defined on a plane, whereas the PBR sky is defined on a sphere.

Max Altitude/Height is of course a hack (we simply determine the point at which density falls to 0.1% of the sea level value), but seems to work OK.


sebastienlagarde:
@JordanL8 I will merge this Pr to be in the next staging to master batch, but feel free to create another Pr for the documentation fix. Thanks!

PR #4041, opened 2019-07-11
Added Version.hlsl file to Core ShaderLibrary

JulienIgnace-Unity:
### Purpose of this PR
Added a generic render pipeline version define (major and minor) and helper macros to check them in client code.


com.unity.render-pipelines.core/ShaderLibrary/Version.hlsl
```diff
@@ -0,0 +1,7 @@
+#define RENDER_PIPELINE_VERSION_MAJOR 7
+#define RENDER_PIPELINE_VERSION_MINOR 0
```
phi-lira: Maybe SHADER_LIBRARY_VERSION?

PR #4052, opened 2019-07-11
Shadergraph properties sync

alelievr:
### Purpose of this PR
+ Added a function to synchronize the properties of the master node with the material in the inspector.
+ When a shader graph property is changed, the keyword reset is now called.

---
### Testing status

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Ffix%2Fshadergraph_properties_sync&automation-tools_branch=master&unity_branch=trunk#

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


pbbastian:
This is only triggered from the material GUI, and not from within Shader Graph, right?

alelievr:
Yes, it's a temporary fix waiting for us to find a better solution about properties sync between shaders and the material. (it doesn't touch the ShaderGraph source code)

sebastienlagarde:
>This is only triggered from the material GUI, and not from within Shader Graph, right?

We have discover issue due to lack of dependency between shader graph and corresponding material. i.E Unity don't have the concept of Material Instance :( . In the mean time this is the best we can do and we must do it. We are starting discussion about how to be able to have better dependency tracking between a material and its shader graph. Goal will be that changing something in shader graph trigger update of all depenendent Material. We will have this issue with Keyword switch PR too.

PR #4063, opened 2019-07-11
 Fix for case 1168215

parashar-unity:
Basically made sure all normalization needs to happen after clipping when in the add pass.

### Purpose of this PR
Issue that appears with normalization in multi-pass in terrain shader on Universal.  This shows up when there are more than 4 layers in the terrain splatmaps.

### Testing status

Manually tested on the sample test case.  Manipulated the test case so that the layers are color-coded in order to have a little more clarity to verify that indeed things were showing up correctly.

Result before the fix with 6 layers  --
MultiPassBefore

Corrected end result after fix --
MultiPassTest


com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -118,22 +118,18 @@ void SplatmapMix(float4 uvMainAndLM, float4 uvSplat01, float4 uvSplat23, inout h
...
+#if !defined(SHADER_API_MOBILE) && !defined(SHADER_API_SWITCH) && defined(TERRAIN_SPLAT_ADDPASS)
     clip(weight == 0.0h ? -1.0h : 1.0h);
 #endif
 
-#ifndef TERRAIN_SPLAT_ADDPASS
     // Normalize weights before lighting and restore weights in final modifier functions so that the overal
-    // lighting result can be correctly weighted.  In the add pass, we can assume the weights
-    // are already properly normalized in the layer below, so we don't want to renormalize again.
+    // lighting result can be correctly weighted.
     splatControl /= (weight + HALF_MIN);
```
phi-lira: Suggestion for next PR: you could multiply by rcp(weight + HALF_MIN) instead.

phi-lira:
Yamato tests passed.

PR #4069, opened 2019-07-12
Reset debug views on teardown

FrancescoC-unity:
As in the title, this also fixes the issue with the sky and matcap left on when closing

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
+++ b/com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
@@ -714,7 +714,6 @@ namespace UnityEngine.Rendering
 
         public static void ForceSceneLightingEnable()
         {
-            // Determine whether the "No Scene Lighting" checkbox is checked for the current view.
             foreach (UnityEditor.SceneView sv in UnityEditor.SceneView.sceneViews)
             {
                 sv.sceneLighting = true;
--- a/com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
+++ b/com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
@@ -712,6 +712,15 @@ namespace UnityEngine.Rendering
             return disabled;
         }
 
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.core/Runtime/Utilities/CoreUtils.cs
```diff
@@ -712,6 +712,15 @@ public static bool IsSceneLightingDisabled(Camera camera)
             return disabled;
         }
 
+        public static void ForceSceneLightingEnable()
+        {
+            // Determine whether the "No Scene Lighting" checkbox is checked for the current view.
```
FrancescoC-unity: Copy-pasted comment, let me remove it

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -826,6 +826,11 @@ void DisposeProbeCameraPool()
             }
 
             CameraCaptureBridge.enabled = false;
+
+            // Disable all debug views
+            DebugManager.instance.Reset();
```
sebastienlagarde: I am wondering if this is ok? I guess we create and delete render pipeline various time already. Have you try for example to enable the matcap view, then change a parameter on HDRP asset? Do it diasble the matcap view?

FrancescoC-unity: Uhm you are right, haven't thought of it;  it actually leaves stuff in a weird state, investigating.

FrancescoC-unity: So, it disables the matcap on our side, however, the engine reverts for some reason to the old no scene lighting debug view...

FrancescoC-unity:
Yeah this is going to give some grievances when swapping asset due to something on C++ side. Closing this, I am going to fix the matcap sky issue in another way.

PR #4092, opened 2019-07-12
Don't apply atmospheric scattering to sky during the atmospheric scattering pass

EvgeniiG:
### Purpose of this PR
Request from @JulienIgnace-Unity and @sebastienlagarde.
Additionally expose intensity and exposure hacks.

### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fno-atmospheric-scattering-during-atmospheric-scattering-pass&automation-tools_branch=master&unity_branch=trunk&sort=1-asc


EvgeniiG:
> * If you look at the sky reflection, there are many directions where it's just black. From what I checked, this is due to the code going through the test `if (tFrag < tEntry)` where black is returned. This was due to GetViewForwardDir() being wrong for cubemap faces. This needs to be passed as a parameter to solve this.

That's bad. Are we planning to set up the constant buffers correctly per face?

JulienIgnace-Unity:
> > * If you look at the sky reflection, there are many directions where it's just black. From what I checked, this is due to the code going through the test `if (tFrag < tEntry)` where black is returned. This was due to GetViewForwardDir() being wrong for cubemap faces. This needs to be passed as a parameter to solve this.
> 
> That's bad. Are we planning to set up the constant buffers correctly per face?

It's actually set correctly. When we render a reflection probe in a cube map, we actually have 6 cameras, with each their own hdCamera that pushes the right parameters so that already works.
The issue here is that it's not a full render in a cube map, it's only a tiny part of the Render Loop that does a very specific render to a cubemap as part of the frame so we can't expect our global variables to work here.
Here we need to use a specific uniform value. That should be easily doable by adding the face index (-1 for regular rendering) in BuiltinSkyParameters and passing it properly in the loop in SkyRenderingContext.RenderSkyToCubemap then each sky renderer can do whatever they please with that information.

EvgeniiG:
> > That's bad. Are we planning to set up the constant buffers correctly per face?
> 
> It's actually set correctly. When we render a reflection probe in a cube map, we actually have 6 cameras, with each their own hdCamera that pushes the right parameters so that already works.
> The issue here is that it's not a full render in a cube map, it's only a tiny part of the Render Loop that does a very specific render to a cubemap as part of the frame so we can't expect our global variables to work here.
> Here we need to use a specific uniform value. That should be easily doable by adding the face index (-1 for regular rendering) in BuiltinSkyParameters and passing it properly in the loop in SkyRenderingContext.RenderSkyToCubemap then each sky renderer can do whatever they please with that information.

Passing parameters one by one in a bunch of places and hunting for bugs like this is not a very efficient use of dev time. Nevertheless, I will try to work around the system in place.

EvgeniiG:
I fixed the issue with cube map rendering.
Trying to fix the issue with atmospheric scattering pass at the horizon.

EvgeniiG:
I fixed the artifact, still need to test.

EvgeniiG:
Sky and fog works locally, other stuff fails. Katana fails. Will try to merge staging and re-test.

EvgeniiG:
So, I cleaned/fixed Julien's code, so now the pre-computation properly supports all sky update modes.
One big downside with the current approach is that before, we always pre-computed 1 bounce/frame.
Now, unless the sky is set as real-time, we pre-compute ALL bounces every frame, so editing the sky parameters is very sluggish.

The only remaining issue is that during start up, the sky probe is black.
It happens because we use the directional lights in the scene to light the sky, and there are no lights during the first frame. It's easy to verify this by adding the following line to PhysicallyBasedSky.shader:

if (_DirectionalLightCount == 0) skyColor = float3(1,0,0);

I will see what I can do.

EvgeniiG:
Building the light list already runs before the cube map rendering.
I thought this must be it...

image

EvgeniiG:
OK, so this is the issue. We cull lights at the right time, but upload them way too late.

image

EvgeniiG:
PushLightLoopGlobalParams happens during BuildGPULightLists, which is too late.

image

EvgeniiG:
I fixed it. See the commit as to how.

PR #4110, opened 2019-07-15
Try to make writeable cubemaps when baking reflection probes

fredpointzero:
### Purpose of this PR
When you setup a project from a perforce server but with Unity not configured to the perforce server,
The reflection probes' cubemaps are read-only, and Unity can't checkout them.

So we try to make the file writeable before writing it. There will still be a warning about missing Perforce credential, but at least, this won't break the workflow with an `UnauthorizedException`.

---
### Testing status
**Katana Tests**: -

Tested manually on A2

**Automated Tests**: -
---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---
### Comments to reviewers


sebastienlagarde:
please. add changelog and target hdrp staging. thanks

PR #4112, opened 2019-07-15
Removed shadow mask debug and fixed an issue where selecting various material debug modes would not reset the others.

JulienIgnace-Unity:
### Purpose of this PR
Removed shadow mask debug and fixed an issue where selecting various material debug modes would not reset the others.


sebastienlagarde:
Please add changelog :)

PR #4124, opened 2019-07-16
Fix the Basemap generation pass issue within case #1168215

parashar-unity:
### Purpose of this PR
Earlier fix for case #1168215 solved the issues for general render passes of the splatmaps both for single-pass and multi-pass.  The normalization steps are actually causing blowout in the basemap generation pass because it occurs after the weights and weight sums have already been corrected.

---
### Testing status
**Katana Tests**:

Tested with sample scenes in the original ticket as well as created several others.
RenderDoc captures as well to verify that the results are being written into the BaseMap Render Textures correctly.

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: Low

---
### Comments to reviewers
This should fix the lingering issues seen within the ticket.  There is the other issue of the fact that there was a corrupted layer in the terrain data within the sample scene, and that does get sidestepped by this fix (such that it doesn't result in any visual flaw, though it would with more real-world content in the other layers as opposed to just checkerboards).  That is ultimately not related to this particular bug, however, and should be considered a separate issue as far as how to deal with those types of layers.


phi-lira:
Yamato failing due to trunk issue.

PR #4145, opened 2019-07-17
 fixed shader compiler errors in gles2.

phi-lira:
### Purpose of this PR
 Fix PPv3 shader compiler errors on gles2 https://fogbugz.unity3d.com/f/cases/1169687/


com.unity.render-pipelines.universal/Editor/Overrides/BloomEditor.cs
```diff
@@ -39,6 +41,9 @@ public override void OnInspectorGUI()
             PropertyField(m_Clamp);
             PropertyField(m_HighQualityFiltering);
 
+            if (m_HighQualityFiltering.overrideState.boolValue && m_HighQualityFiltering.value.boolValue && SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES2)
```
Verasl: This will never work in editor, we should check `PlayerSettings.GetGraphicsAPIs` and see if GLes2 is in there

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -392,6 +392,8 @@ void DrawAdditionalData()
                 {
                     EditorGUI.indentLevel++;
                     hasChanged |= DrawIntPopup(m_AdditionalCameraDataAntialiasingQuality, ref selectedAntialiasingQuality, Styles.antialiasingQuality, Styles.antialiasingQualityOptions, Styles.antialiasingQualityValues);
+                    if (SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES2)
```
Verasl: This will never work in editor, we should check `PlayerSettings.GetGraphicsAPIs` and see if GLes2 is in there

PR #4156, opened 2019-07-17
Use UAV API properly for TAA

FrancescoC-unity:
This should fix the issue we had on Metal and on PC, while still using UAV writes on PS.

FWIW, ClearRandomWriteTargets does not clear the content, but clears the binding slots. This is somewhat confusing name, but anyway, whatever works. 

@alelievr  could you check whether this fixes the issues on mac, whenever you have a second :-)


FrancescoC-unity:
I checked myself as I had my mac with me today, can confirm it fixes the issue we had on Metal, so this is good to go

PR #4165, opened 2019-07-18
Different handling of atlas placement for cached shadows.  [DO NOT MERGE]

FrancescoC-unity:
This changes how the cached shadow are handled with regards to atlas placement. 

To put it simply, even when are removed from view, they hold their spot in the atlas unless the space runs out in the atlas. In that case, the cached data is invalidated and empty slots are pruned. 
If the atlas is full, cached shadows are instead rendered in real time. 

There is lots of extra handling for the above cases, but the layouting policy has been left the same. This change doesn't prevent us from changing the layouting policy in the future as long as the placement is deterministic (same order of light leads to same placing). 

Katana: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Fdifferent-shadow-caching-handling-for-atlas-placement&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

This will need some testing from QA, I tested in sample scene and fontainebleau and it all looked correct there.


alelievr:
Since i'm currently checking the GC calls and i saw that there are few in HDShadowMap.cs (from the deep copy of the list) that i fixed on my branch it would be nice it you do a pass to check if there is no garbage allocated each frame in your code.
(Be careful to Linq queries too, they tend to allocate memory)

FrancescoC-unity:
> Since i'm currently checking the GC calls and i saw that there are few in HDShadowMap.cs (from the deep copy of the list) that i fixed on my branch it would be nice it you do a pass to check if there is no garbage allocated each frame in your code.
> (Be careful to Linq queries too, they tend to allocate memory)

There were indeed a couple due to Linq, I have removed them!

sebastienlagarde:
Miss changelog :)

sebastienlagarde:
I have updated to master and  relaunched Katana

sebastienlagarde:
Katana is green

FrancescoC-unity:
Please ignore the changes for now, I will ping you guys when ready to merge.

FrancescoC-unity:
The merge was non trivial and for problems with my git repro I had to create a new branch here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4303

PR #4178, opened 2019-07-19
[VFX] Better integration of VFXManager in UniversalRP

PaulDemeulenaere:
### Purpose of this PR
Replace `VFX.VFXManager.ProcessCamera(camera);` but this couple of function : 
- `VFX.VFXManager.PrepareCamera` : Should be called before Culling
- `VFX.VFXManager.ProcessCameraCommand` : Should be called once all global variable setup (typically view & projection matrices)

`VFX.VFXManager.ProcessCamera(camera);` is actually already doing both but it will be deprecated at some point.

Additionally, `VISUAL_EFFECT_GRAPH_1_0_0_OR_NEWER` has been added to filter out VFXManager call : even if it does nothing when there isn't any visual effect component within the scene, it's better to avoid a useless native call.

---
### Testing status
**Katana Tests**:  🔶 

**Manual Tests**:
- [x] Built a player
_all_test_standalone

Launch Katana :
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=vfx%2Ffix%2Fintegration-in-universal&unity_branch=graphics%2Fupdate-srp-package-7.0.x&automation-tools_branch=master

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**: Low


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -533,6 +536,14 @@ static void SetupPerCameraShaderConstants(CameraData cameraData)
             Shader.SetGlobalMatrix(PerCameraBuffer._InvCameraViewProj, invViewProjMatrix);
         }
 
+        static void ProcessPerCameraAfterCulling(CameraData cameraData, CommandBuffer cmd)
+        {
+#if VISUAL_EFFECT_GRAPH_0_0_1_OR_NEWER
+            //Triggers dispatch per camera, all global parameters should have been setup at this stage.
+            VFX.VFXManager.ProcessCameraCommand(cameraData.camera, cmd);
```
phi-lira: What is this doing? Asking to figure out if it should be called before the Setup or after.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -140,8 +140,10 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
             foreach (Camera camera in cameras)
             {
                 BeginCameraRendering(renderContext, camera);
-
-                VFX.VFXManager.ProcessCamera(camera); //Visual Effect Graph is not yet a required package but calling this method when there isn't any VisualEffect component has no effect (but needed for Camera sorting in Visual Effect Graph context)
+#if VISUAL_EFFECT_GRAPH_0_0_1_OR_NEWER
```
phi-lira: I'm not sure I understand how this works. This define is currently set in UniversalRP, which means it will always be defined. Shouldn't it be defined in VFX manifest so, when it defines this only when VFX Graph is present?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
phi-lira: Ah I see. Sorry read the description now. Sadly UniversalRP still depends on a context.SetupCameraProperties, this happens in the execute step. Maybe call ProcessCameraCommand here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L221-L222

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
PaulDemeulenaere: It relies on user manifest.json file, it is only better for code stripping.
- https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/TestProjects/VisualEffectGraph_LWRP/Packages/manifest.json <= This one has `com.unity.visualeffectgraph`, thus, `VISUAL_EFFECT_GRAPH_0_0_1_OR_NEWER` will be defined while importing UniversalRP.
- https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/TestProjects/UniversalGraphicsTest/Packages/manifest.json <= This one didn't include VisualEffectGraph, `VISUAL_EFFECT_GRAPH_0_0_1_OR_NEWER` will not be defined and this code automatically stripped (even if it does nothing when there isn't any component within the scene).

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
PaulDemeulenaere: I moved it right before "ExecuteBlock" after all setup : https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4178/commits/3e5d6e50da01f85abf59203779731640a019d677
What do yo think of this ?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -178,7 +178,7 @@ public virtual void FinishRendering(CommandBuffer cmd)
         /// </summary>
         /// <param name="context">Use this render context to issue any draw commands during execution.</param>
         /// <param name="renderingData">Current render state information.</param>
-        public void Execute(ScriptableRenderContext context, ref RenderingData renderingData)
+        public void Execute(ScriptableRenderContext context, ref RenderingData renderingData, CommandBuffer cmd)
```
phi-lira: You can't change public API. This will break user projects. You have access to the context, create a command buffer inside execute method instead. We might reuse it for other things.

phi-lira:
Should we add a changelog entry as well?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
PaulDemeulenaere: I used a local commandBuffer in `renderer.Execute`

PaulDemeulenaere:
I have run VFX graphic tests locally in editor & standalone, ready to be merged @sebastienlagarde 
Katana has been also launched but I get the known issue : `'[Assert] Assertion failed on expression: 'SUCCEEDED(hr)''` with HDRP test.
https://artifacts-ui.bf.unity3d.com/proj57-Test_PlayMode_-_Windows/31652512_06_08_2019_13_14_51_%2B0000/logs/
Otherwise, everything is 🍏

julienf-unity:
Should this be merged for 2019.3 ?

PaulDemeulenaere:
```Should this be merged for 2019.3 ?``` It's not necessary to merge it in 2019.3.
I will remerge this PR and rerun test on yamato, I will let you know when it's ready to be merged.

**Comment by PaulDemeulenaere** [comment]
image
it can be merged now !

File: com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs [hunk 4]
```diff
@@ -234,6 +234,14 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
             if (stereoEnabled)
                 BeginXRRendering(context, camera);
 
+#if VISUAL_EFFECT_GRAPH_0_0_1_OR_NEWER
+            var localCmd = CommandBufferPool.Get(string.Empty);
```
julienf-unity: Shouldnt there be a string here?

PaulDemeulenaere: It was a suggestion from @phi-lira because he has a PR that will add a command buffer to execute some postprocessing stuff there and will patch this code.
So, he advised me to remove the tag var and create cmd buffer with empty string.

julienf-unity: ok

PR #4185, opened 2019-07-22
Config entry for area light and shadow filtering

FrancescoC-unity:
### Purpose of this PR
Moved the shader config to the config package.

Added options to disable area lights and to control shadow filtering for deferred. 

For QA: I tested it myself as it is quite simple so not notifying QA explicitly. If anyone wants to check further, the workflow is modify ShaderConfig.cs, regenerate shader include and verify that the options are acting as modified in the cs file (e.g. if AreaLight is set to 0, Area Lights are not active anymore) 

---
### Release Notes
Moved ShaderConfig handling from the main package to the config package.
Added option to globally disable Area Lights.
Added option to change the shadow filtering quality for deferred. 
 
---
### Testing status
**Katana Tests**: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fhdrp-config-for-area-light-and-shadow-filtering&unity_branch=trunk&sort=1-asc [Just kicked]

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Medium


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1185,9 +1185,12 @@ static Vector3 EvaluateAtmosphericAttenuation(Vector3 L, Vector3 positionWS)
             VisibleLight light, Light lightComponent, HDAdditionalLightData additionalLightData,
             int lightIndex, int shadowIndex, ref Vector3 lightDimensions, DebugDisplaySettings debugDisplaySettings, ref int screenSpaceShadowIndex)
         {
+            if (ShaderConfig.s_AreaLights == 0 && gpuLightType.IsAreaLight())
```
sebastienlagarde: I think we should do this test earlier, here we are already too low in the loop. Mean you don't need to do the test in GetLightVolumeDataAndBound either

sebastienlagarde:
There is a conflict, can you resolve it ?

sebastienlagarde:
I made the change + merged master, if you can check

sebastienlagarde:
@iM0ve could you validate that the PR is fine? 
Goal is to control area lights from config file. So we setup area light to 0 in config.cs and in config.cs.hlsl and it should not render area light anymore. Thanks

sebastienlagarde:
Note: I will merge this PR in HDRP/staging as I need it. But please comment here. THanks

iM0ve:
image Found one concern and no issues. If it is by design, the PR is ready to be merged. More details:
https://confluence.hq.unity3d.com/display/HDRP/Area+light+and+shadow+filtering

sebastienlagarde:
Good point about the concern. We will fix it in another PR. Indeed Emissive Mesh shouldn't be display anymore. But yes by design currently...

FrancescoC-unity:
I have a fix ready now, will make that on staging directly.

PR #4201, opened 2019-07-23
Added test to detect gc allocs when rendering what is on main camera on Universal RP

roberto-renzetti-unity:
### Purpose of this PR

This PR adds a check for GC_Alloc in a generic way so that it is run against all our existing UniversalRP tests, similar to what was done on the HDRP tests. 

Right now all the tests fail as there are GC Allocations, so I surrounded the tests with the following define:

      CHECK_ALLOCATIONS_WHEN_RENDERING

Add this define to the Player settings to activate the tests for GC Allocs. The define is the same as the one for HDRP so its easy to remember it. This way developers can add this define locally, run the tests, see which ones failed, fix the gc allocs. Note: I did NOT add this define to this PR so that the tests won't fail in Katana.

Once all the gc allocs are identified/fixed, we can take this define out from the code and make the test permanent.

---
### Release Notes

---
### Testing status
**Katana Tests**: Since the define CHECK_ALLOCATIONS_WHEN_RENDERING is not set on this PR, these changes won't affect Katana.

- [ ] Built a player
- Other: 

**Automated Tests**: Please see description. This is a change that affects all UniversalRP automated tests, but not fully enabled yet.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers

### DIFF
--- a/TestProjects/UniversalGraphicsTest/Assets/CommonAssets/Scripts/LWGraphicsTests.cs
+++ b/TestProjects/UniversalGraphicsTest/Assets/CommonAssets/Scripts/LWGraphicsTests.cs
@@ -15,7 +15,7 @@ public class LWGraphicsTests
     [UnityTest, Category("LightWeightRP")]
     [PrebuildSetup("SetupGraphicsTestCases")]
     [UseGraphicsTestCases(lwPackagePath)]
-    
+
 
     public IEnumerator Run(GraphicsTestCase testCase)
     {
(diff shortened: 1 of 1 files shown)


phi-lira:
I'm wondering if we should merge this PR. Then once fixing the GC allocs we remove the define?

roberto-renzetti-unity:
@phi-lira That was my idea. We could merge it now so the developers can use it to test locally while fixing the problems, once all the problems are all gone we take the defines away so that new code won't re-introduce the problems.

phi-lira:
Adding @sophiaaar for review.

PR #4212, opened 2019-07-24
Replace Z by Depth in the material UI

alelievr:
Replace Z by Depth in the material UI


sebastienlagarde:
@alelievr can you do a backport for 2019.2. thanks

PR #4234, opened 2019-07-25
Universal RP fix for issue #1161116

parashar-unity:
### Purpose of this PR
Fixing issue that was affecting splatmap UV alignment across multiple tiles.  This was identified initially in HDRP, but the same issue was spotted in LW/Universal.  Those changes are now in this PR.

Also, added a new test scene for the recently added terrain blending features (these are old hat for HD, but relatively new to LW/Universal)

### Release Notes

---
### Testing status
Manual tests to verify that the bug in question was fixed.
Newly added tests have also been run in Yamato, and show green at least on Windows.  Mac only fails because of not finding reference images, which it does for nearly all tests in spite of those references actually existing.

**Automated Tests**: 
038_Shader_TerrainBlend

Test scene validates the following :
    * Per-pixel variation over normal, smoothness, metallic, etc.
    * Mix of mask map and non-mask map layers maintaining per-layer backwards compatibility
    * Use of modulation and min-max remap functions
    * Use of multipass/add-pass by having more than 4 terrain layers
    * Use of diffuse alpha-channel as density

---
### Overall Product Risks
**Technical Risk**: Low.  The only real change to code is a bugfix.  This fix is shown to have some slight shift in UVs, but is not perceptible without either zooming in close (e.g. in the instance of the bug), or in extremely low resolution.
**Halo Effect**: Low (mostly from adding the automated test).


parashar-unity:
> QA did a pass and found some issues https://docs.google.com/document/d/12YG3xx295sEyIKVXEpm_v-94S4-GAocXUXmZ4_5rQzc
> Android does not play well with the test scene

From the looks of things, it looks like the add pass is just plain overwriting the previous base pass instead of "adding" as it should.  Good way to verify this is if you still get the same issue on Android after deleting two of the layers from the terrain (i.e. reducing the layer count to 4, thus making it single-pass).  If you don't, then the issue is definitely with the add pass and/or some settings/state when it is executed.

parashar-unity:
I tried making a small change that may help.  Unfortunately, I can't be 100% from my testing alone because trying things out in an emulator did not give me any failure in the first place.  However, the apparent visuals in the document very clearly looked like the add pass was the problem and it looked like it was not clipping anything.  So I tried a bit of a change to ensure that it would.

Verasl:
Recent changes fix the black terrain on Android devices, @LucasCZ tested and updated the test doc.

PR #4236, opened 2019-07-26
Automated tests for SRP batcher

TomasKiniulis:
### Purpose of this PR
Check functionality of SRP batcher in automated tests for SkinnedMeshRenderer and MeshRenderer - compare scenes when SRP batcher is on and when SRP batcher is off.

Before we had only manual test project: https://qatestrail.hq.unity3d.com/index.php?/cases/view/401299&group_by=cases:section_id&group_order=asc&group_id=73404

---
Added 4 automated scenes to test if SRP batcher is working fine.
9601_SkinnedMeshBatching-Off
9602_SkinnedMeshBatching-On
C21HH7WbTc

9603_MeshRendererBatching-Off
9604_MeshRendererBatching-On
image (14)

---
### Testing status
Test PlayMode - Windows: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders/proj57-Test%20PlayMode%20-%20Windows?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP/tests/srpBatcher&unity_branch=trunk

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low


sebastienlagarde:
Hey, if you test SRP batcher, we should also add a test on shadow as we get recent failure on it (it is fix now but have been undetected for a while, the shadow where flickering). Could eb done in a next PR. Thanks

sebastienlagarde:
+ note: the head is green because there is no diffusion profile. You should assign one, otherwise we could be in trouble.

sebastienlagarde:
Note: I launched graphic test katana as this is graphic test: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=HDRP%2Ftests%2FsrpBatcher&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

TestProjects/HDRP_Tests/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -14,7 +14,7 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/GraphicTests/Scenes/1x_Materials/1103_Unlit_Distortion_DepthTest.unity
     guid: 27d04b0368c7fef4eb423d76cc0e48dc
-  - enabled: 0
+  - enabled: 1
```
sebastienlagarde: CAUTION: you are re-enabling several test here that were disable because instable or not working. This is something we need to work on, but it is high likely that katana will fail due to this.

TomasKiniulis:
@sebastienlagarde thanks for checking! Added missing diffusion profile. Reverted EditorBuildSettings.asset to one from staging and enabled the new test scenes. I'm rechecking if scenes pass tests locally on mac and win and will rerun the katana tests. Shadow tests are on the way!

TomasKiniulis:
@sebastienlagarde katana tests finally passed (also checked if they pass on two local machines), one thing preventing for the animated skinned mesh scenes passing is that SinRotate.cs was using Time.frameCount. I've changed this to counting frames incrementing an integer variable in the update and it works correctly. SinMove.cs script already uses this approach - that's why MotionVector scenes do not fail, so the change shouldn't cause an issue.

PR #4238, opened 2019-07-26
Universal/fix case 1163320

aleks01010101:
### Purpose of this PR

- Fix case 1163320 (https://issuetracker.unity3d.com/issues/lwrp-mobile-secondary-camera-background-alpha-value-is-lost-when-hdr-and-opaque-texture-are-enabled-in-lwrp-asset).
- Fix linear rendering on Adreno 3xx when postprocessing is active.

---
### Release Notes
- Fix case 1163320 (https://issuetracker.unity3d.com/issues/lwrp-mobile-secondary-camera-background-alpha-value-is-lost-when-hdr-and-opaque-texture-are-enabled-in-lwrp-asset).
- Fix linear rendering on Adreno 3xx when postprocessing is active.

---
### Testing status
In progress

- [x] Built a player
- Other: 
Manual testing on two devices (Mali G71 and Adreno 330).

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=universal%2Ffix_case_1163320&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
Low

Low (mobile only)

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -165,9 +165,9 @@ void SortCameras(Camera[] cameras)
                 desc.height = (int)((float)desc.height * renderScale);
             }
 
-            // TODO: when preserve framebuffer alpha is enabled we can't use RGB111110Float format.
-            bool useRGB111110 = Application.isMobilePlatform && RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.RGB111110Float);
-            RenderTextureFormat hdrFormat = (useRGB111110) ? RenderTextureFormat.RGB111110Float : RenderTextureFormat.DefaultHDR;
+            RenderTextureFormat rt32BitHDR = needsAlpha ? RenderTextureFormat.ARGB2101010 : RenderTextureFormat.RGB111110Float;
+            bool use32BitHDR = Application.isMobilePlatform && RenderingUtils.SupportsRenderTextureFormat(rt32BitHDR);
```
Chman: I would argue that we should use 32bit HDR on all platforms by default given the performance gain we get from it (unless we need alpha obviously).
@phi-lira thoughts? On console it makes quite a big difference, even more so on Switch.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -165,9 +165,9 @@ void SortCameras(Camera[] cameras)
                 desc.height = (int)((float)desc.height * renderScale);
             }
 
-            // TODO: when preserve framebuffer alpha is enabled we can't use RGB111110Float format.
-            bool useRGB111110 = Application.isMobilePlatform && RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.RGB111110Float);
-            RenderTextureFormat hdrFormat = (useRGB111110) ? RenderTextureFormat.RGB111110Float : RenderTextureFormat.DefaultHDR;
+            RenderTextureFormat rt32BitHDR = needsAlpha ? RenderTextureFormat.ARGB2101010 : RenderTextureFormat.RGB111110Float;
```
phi-lira: This is not going to work. ARGB32 is UNORM format. If preserve alpha is not the default option I'd say it's fine to use FP16 as long as we warn users about the option to preserve alpha and HDR. Also always check if format is supported.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
phi-lira: I think adding clause for 32bit in Switch is safe. I'm concerned about other platforms. Should we use cameraHDrformat from Tier settings? This allows users to choose.

Chman: Yes I guess we should use the one from Tier settings instead of hard-coding it. Where is the "keep alpha" setting exposed on the UniversalRP side though?

Chman: One thing I really want to stress out is that the only reason to use a full fp16 buffer is to have HDR + good precision alpha. For everything else, really, we shouldn't, it's a bandwidth killer. And not just on Switch, on everything.

aleks01010101: Keep alpha is not on the SRP side. Player settings :)

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
aleks01010101: Perhaps not from tier settings, but expose it on the pipeline then?

aleks01010101: Preserve alpha is not the default option. Thinking about this: I have not changed it so that offscreen cameras would always need alpha with HDR, which is probably not desired...
Should we make a per-camera setting then regarding "does this camera need alpha channel"?

phi-lira: Yes. A per-camera setting is ideal. Could you add a toggle on the additional camera data component? @unity-cchu for the 2D  renderer you'll also want this, but until you do it you can hide this from the inspector.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
aleks01010101: Fixed.
Now if a camera has a target RT, it's format is used. For cameras targeting backbuffer, RG11B10F is used when supported and alpha needs to be killed, DefaultHDR otherwise.

com.unity.testframework.graphics/Runtime/ImageAssert.cs
```diff
@@ -60,7 +60,10 @@ public static void AreEqual(Texture2D expected, IEnumerable<Camera> cameras, Ima
             // This PR adds a dummy rendered frame before doing the real rendering and compare images ( test already has frame delay, but there is no rendering )
```
sophiaaar: does this comment need updating?

aleks01010101: No, this didn't change.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
aleks01010101: Fixed

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -165,14 +165,24 @@ void SortCameras(Camera[] cameras)
                 desc.height = (int)((float)desc.height * renderScale);
             }
 
-            // TODO: when preserve framebuffer alpha is enabled we can't use RGB111110Float format.
-            bool useRGB111110 = Application.isMobilePlatform && RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.RGB111110Float);
-            RenderTextureFormat hdrFormat = (useRGB111110) ? RenderTextureFormat.RGB111110Float : RenderTextureFormat.DefaultHDR;
-            desc.colorFormat = isHdrEnabled ? hdrFormat : renderTextureFormatDefault;
-            desc.depthBufferBits = 32;
+            bool use32BitHDR = !needsAlpha && RenderingUtils.SupportsRenderTextureFormat(RenderTextureFormat.RGB111110Float);
```
phi-lira: This is a little bit confusing. 
!needsAlpha = camera.targetTexture != null || !Graphics.preserveFramebufferAlpha || !PlatformNeedsToKillAlpha();

camera.targetTexture check is irrelevant here since hdrFormat is only used when targetTexture is null. however it still can cause side effects and break things if someone changes the loop below. I recommend instead of passing this needsAlpha as argument to this function, already compute a !needsAlpha only inside the else case below.

phi-lira:
Katana is red but failures are not introduced by this PR.

PR #4240, opened 2019-07-26
Disable Emissive Mesh if area lights are turned off in config

FrancescoC-unity:
As in title


sebastienlagarde:
Hey. We shouldn't do it this way. It will not work.
I can explain you the details at the office, there is better to do.

FrancescoC-unity:
> Hey. We shouldn't do it this way. It will not work.
> I can explain you the details at the office, there is better to do.

Sure! For what is worth though, this was working just fine locally here

sebastienlagarde:
It work because you use the light :). Try to create several emissive mesh. then disable the option and don't touch anything else. Mesh will not disappeard.

FrancescoC-unity:
> It work because you use the light :). Try to create several emissive mesh. then disable the option and don't touch anything else. Mesh will not disappeard.

They actually do disappear! I had about 6/7 of them and as soon as the option is turned off, they all disappear. 

There is indeed a problem though, they do not reappear once I enable the option back, unless I touch the light.

PR #4265, opened 2019-07-31
Added Check Out buttons to ShaderGraph/VFXGraph asset editors, fix case 1136620

AscendingMan:
### Purpose of this PR
This PR fixes [case 1136620](https://fogbugz.unity3d.com/f/cases/1136620/) where the VFX Asset Inspector would allow editing assets that are not checked out in version control.

Two Check Out buttons were also added to VFXGraph Editor and ShaderGraph Editor windows.

image
image
---
### Testing status

Manual UI testing

---
### Overall Product Risks

**Technical Risk**:  Low

**Halo Effect**: None


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -222,6 +225,27 @@ public void PingAsset()
             }
         }
 
+        public bool IsGraphAssetCheckedOut()
+        {
+            var path = AssetDatabase.GUIDToAssetPath(selectedGuid);
```
aras-p: Other places in code check for selectedGuid being null before using it. Why not in this function?

com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
```diff
@@ -159,6 +163,15 @@ public GraphEditorView(EditorWindow editorWindow, GraphData graph, MessageManage
                             node.Dirty(ModificationScope.Graph);
                         }
                     }
+                    
+                    if (!isCheckedOut())
```
aras-p: maybe this needs to check for isCheckedOut being null before calling it

AscendingMan:
> could the Shader Graph checkout button be drawn between `Save Asset` and `Show in Project`, as it's a one-time action related the Saving, it's a nice UX.
> 
> we're also double checking some behaviors for the check out action, but we have to get set up with a perforce server to test it with, please hold tight!

Currently, the button is made to disappear if checking out is not required so I don't think it would be a good idea to surround it with other buttons that would shift positions once the asset is checked out. It was made that way to behave the same as the Check Out button behaves in the Prefab Editor.

sebastienlagarde:
@PaulDemeulenaere please review

aras-p:
Looks like superseded by #4492

PR #4271, opened 2019-08-02
[Yamato] Adding Mac Metal builders

sophiaaar:
### Purpose of this PR
* Adding Mac with Metal to our Yamato testing
* Keeping OpenGLCore, but disabling it from running in CI
* Testing all projects on all platforms will now only happen daily on master, rather than on every commit

---
### Release Notes

---
### Testing status
**Yamato**: https://yamato.stg.cds.internal.unity3d.com/jobs/45-ScriptableRenderPipeline/tree/macmini-setup

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None

---
### Comments to reviewers


.yamato/upm-ci-universal.yml
```diff
@@ -4,8 +4,12 @@ platforms:
   - name: Windows64
     type: Unity::VM::GPU
     image: sdet/gamecode_win10:stable
-    flavor: b1.large
+    flavor: m1.large
```
phi-lira: Is this intentional? Why are we going back to m1?

.yamato/upm-ci-universal.yml
```diff
@@ -4,8 +4,12 @@ platforms:
...
   - name: OSX
+    type: Unity::metal::macmini
+    image: slough-ops/macos-10.14-xcode
+    flavor: m1.mac
```
phi-lira: Same here

.yamato/upm-ci-hdrp.yml
```diff
@@ -50,18 +50,29 @@ projects:
       
     {% else %}
       
+    - git clone git@github.cds.internal.unity3d.com:unity/utr.git TestProjects/{{ project.folder }}/utr
```
phi-lira: Seems like this git clone is the same for all if/else/if paths. could be out of the loop reducing code duplication/improving maintanability.

.yamato/upm-ci-universal.yml
sophiaaar: not intentional, good spot! thanks!

.yamato/upm-ci-universal.yml
sophiaaar: yes, m1 is all there is for mac (as far as i know) but i will double check

sophiaaar: confirmed that they need to be m1.mac

.yamato/upm-ci-universal.yml
sophiaaar: fixed

.yamato/upm-ci-hdrp.yml
sophiaaar: good shout

sophiaaar: fixed

PR #4302, opened 2019-08-08
Manully backported #4283

rainsing:
### Purpose of this PR
Manully backported #4283.

---
### Testing status

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=2d%2Frelease%2F6.9.x%2Fbug-fixes&automation-tools_branch=master&unity_branch=2019.2%2Fstaging&sort=1-asc


phi-lira:
I'm not sure if we are release another 6.9.x package. Need to double check, if we are releasing a 6.10, then this is not required.

rainsing:
> I'm not sure if we are release another 6.9.x package. Need to double check, if we are releasing a 6.10, then this is not required.

As long as we have these fixes in any version targeting 2019.2, then it's OK.

PR #4310, opened 2019-08-08
Single-pass instancing support for XR SDK

fabien-unity:
### Purpose of this PR
- Enabled single-pass instancing support for XR SDK with new API cmd.SetInstanceMultiplier()
- Replaced usage of ENABLE_VR in XRGraphics.cs by a version define (ENABLE_VR_MODULE) based on the presence of the built-in VR module
- Replaced usage of ENABLE_VR in XRSystem.cs by version defines based on the presence of the built-in VR and XR modules

---
### Testing status
**Katana Tests**: [running](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=hdrp-xr-pure-instancing&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low

---
### Comments to reviewers
ENABLE_VR is set from the C++ engine per platform.
The VR and XR modules are built-in packages and can be disabled either per platform from the C++ engine or by project with the package manager or by editing the manifest.
The presence of the module will expose the API to scripts:
- VR module will expose UnityEngine.XR.XRSettings
- XR module will expose the XR SDK API

By using package version define, we can rely on the presence of the package to gate the inclusion of code and use the API only if it's actually present in the project to avoid compilation errors.
By doing so, we can exclude VR/XR code from HDRP if the modules are not enabled.


sebastienlagarde:
Hey, can you target master for this PR (as it require people from other team to valid. Thanks)

fabien-unity:
Closing this one.
new version with master as target: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4313

PR #4313, opened 2019-08-09
Single-pass instancing support for XR SDK

fabien-unity:
### Purpose of this PR
- Enabled single-pass instancing support for XR SDK with new API cmd.SetInstanceMultiplier()
- bump min version to 2019.3.0a12
- Replaced usage of ENABLE_VR in XRGraphics.cs by a version define (ENABLE_VR_MODULE) based on the presence of the built-in VR module
- Replaced usage of ENABLE_VR in XRSystem.cs by version defines based on the presence of the built-in VR and XR modules

---
### Testing status
**Katana Tests**: [green](https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=xr-sdk-pure-instancing&automation-tools_branch=master&unity_branch=trunk&sort=1-asc)

- [x] Built a player

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low

---
### Comments to reviewers
ENABLE_VR is set from the C++ engine per platform.
The VR and XR modules are built-in packages and can be disabled either per platform from the C++ engine or by project with the package manager or by editing the manifest.
The presence of the module will expose the API to scripts:
- VR module will expose UnityEngine.XR.XRSettings
- XR module will expose the XR SDK API

By using package version define, we can rely on the presence of the package to gate the inclusion of code and use the API only if it's actually present in the project to avoid compilation errors.
By doing so, we can exclude VR/XR code from HDRP if the modules are not enabled.


sebastienlagarde:
Katana is green. Please @phi-lira , @sophiaaar could you approve the change to min 2019.3a12 version, so we can move forward with this PR. @fredpointzero please merge this PR in master when approved. Thanks

mikesfbay:
QA @DennisDeRykeUnity please test XRSDK Oculus with 19.3a12 and also legacy XR for regressions.

com.unity.testframework.graphics/package.json
```diff
@@ -3,7 +3,7 @@
     "displayName":"Graphics Tests Framework",
     "version": "7.1.1",
     "unity": "2019.3",
-    "unityRelease": "0a10",
+    "unityRelease": "0a12",
```
sophiaaar: @fabien-unity did you intend for these to change?

fabien-unity: I was told to bump all the SRP sub-project at once. If this is not required, we can only bump HDRP related ones.

sophiaaar: no thats fine, i just wanted to check it wasnt an accidental change :)

PR #4317, opened 2019-08-09
HDRP Custom pass

alelievr:
### Purpose of this PR
Implement custom passes in HDRP

---
### Release Notes
This PR adds a new component called `Custom Pass Volume` that allows you to inject custom passes in the HD graphic pipeline. Currently there are 3 injection points supported:
+ Before Rendering
+ Before Transparent
+ Before PostProcess

In each `Custom Pass Volume` you have a list of passes that will be executed from top to bottom, here is an example of a DrawRenderers pass:
image
And a fullscreen shader pass:
image

### Features list
+ Separate volume component that allows you to have multiple volumes overlapping with different injection points.
+ DrawRenderers pass, allow you to choose a specific subset (from the material renderqueue) of objects and render them.
+ FullScreen pass, render a fullscreen quad using the material in the pass setting. ShaderGraphs are not supported currently. Alpha blending is supported by this pass and enabled by default in the custom fullscreen pass template.
+ Custom target: you have access to a builtin custom buffer (color + depth) in addition to the camera buffer, when a pass is executed you can choose which is bound. Additionally you have access to the ClearFlag parameter which allow you to do a clear operation on either color or depth when the buffer is bound.
+ Material Override (DrawRenderers): allow you to replace the material pass used to render objects that have passed the filters (layer + renderqueue + shader passes) shaderGraphs are supported here, you just have to choose the right pass in the material (i.e Forward or ForwardOnly in most cases).
+ Shader templates: there are two templates available (see picture below):
+ + one for fullscreen pass which gives you access to most of the inputs you need, here is the generated code:
```hlsl

    float4 FullScreenPass(Varyings varyings) : SV_Target
    {
        float depth = LoadCameraDepth(varyings.positionCS.xy);
        PositionInputs posInput = GetPositionInput(varyings.positionCS.xy, _ScreenSize.zw, depth, UNITY_MATRIX_I_VP, UNITY_MATRIX_V);
        float3 viewDirection = GetWorldSpaceNormalizeViewDir(posInput.positionWS);

        // You can add your custom pass code here
        float4 color = float4(posInput.positionNDC, 0.0, 1);

        return color;
    }
```

+ + The other one is for the material override for the DrawRenderers, it provides control over the vertex and fragment shader:
```hlsl
            #define HAVE_MESH_MODIFICATION
            AttributesMesh ApplyMeshModification(AttributesMesh input, float3 timeParameters)
            {
                input.positionOS *= 1.0001; // inflate a bit the mesh to avoid z-fight
                return input;
            }

            // Put the code to render the objects in your custom pass in this function
            void GetSurfaceAndBuiltinData(FragInputs fragInputs, float3 viewDirection, inout PositionInputs posInput, out SurfaceData surfaceData, out BuiltinData builtinData)
            {
                float2 colorMapUv = TRANSFORM_TEX(fragInputs.texCoord0.xy, _ColorMap);
                float4 result = SAMPLE_TEXTURE2D(_ColorMap, s_trilinear_clamp_sampler, colorMapUv) * _Color;
                float opacity = result.a;
                float3 color = result.rgb;

#ifdef _ALPHATEST_ON
                DoAlphaTest(opacity, _AlphaCutoff);
#endif

                // Write back the data to the output structures
                ZERO_INITIALIZE(BuiltinData, builtinData); // No call to InitBuiltinData as we don't have any lighting
                builtinData.opacity = opacity;
                builtinData.emissiveColor = float3(0, 0, 0);
                surfaceData.color = color;
            }
```
+ + This template is based on the unlit shader, so currently you can't access light data from here.
+ + There are also some new shader functions you can access to sample the custom buffer:
```hlsl
float3 SampleCustomColor(float2 uv);
float3 LoadCustomColor(uint2 pixelCoords);
float LoadCustomDepth(uint2 pixelCoords);
float SampleCustomDepth(float2 uv);
```

image

### Things to test
+ Injection points and volume stacking (note that there is no blending between the list of different volumes, only one list can be used at the same time for a certain injection point)
+ Custom buffer access in Shaders
+ Custom FullScreen pass / Renderers shader templates + code
+ Settings exposed in the UI for the DrawRenderers pass
+ Override camera depth
+ Scripting API

### Cool effects to make (for testing)
+ Edge detection
+ Glitch for only certain objects
+ Objects visible behind walls

### Scripting
You should be able to access the `Custom Pass Volume` by code, most of the functions have been documented (but the API is not yet finished). Here is an example of how to create a fullscreen pass and a custom pass by script:

```csharp
public class VolumeScripting : MonoBehaviour
{
    CustomPassVolume    volume;

    void Start()
    {
        volume = GetComponent< CustomPassVolume >();   

        volume.injectionPoint = CustomPassInjectionPoint.BeforeTransparent;

        var pass = ScriptableObject.CreateInstance< MyCustomPass >();

        CustomPass.CreateFullScreenPass(new Material(Shader.Find("Fullscreen/Test")));

        volume.customPasses.Add(pass);
    }
}

class MyCustomPass : CustomPass
{
    // It can be used to configure render targets and their clear state. Also to create temporary render target textures.
    // When empty this render pass will render to the active camera render target.
    // You should never call CommandBuffer.SetRenderTarget. Instead call <c>ConfigureTarget</c> and <c>ConfigureClear</c>.
    // The render pipeline will ensure target setup and clearing happens in an performance manner.
    protected override void Setup(ScriptableRenderContext renderContext, CommandBuffer cmd)
    {
        // Setup code here
    }

    protected override void Execute(ScriptableRenderContext renderContext, CommandBuffer cmd, HDCamera camera, CullingResults cullingResult)
    {
        // Executed every frame for all the camera inside the pass volume
    }

    protected override void Cleanup()
    {
        // Cleanup code
    }

```

Custom Passes created via script will also be available in the custom pass list when adding a new one:
image

For shader scripting you have the option to create new shaders from the Assets/Create menu (see above). This allows you to do cool stuff like this: (adding splat of emissive material on top of some objects)
CustomPass

### Known issues
+ When an object is moved via a handle, the DrawRenderers skips this object

---
### Testing status
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffeature%252Fcustom_pass/.yamato%252Fupm-ci-hdrp.yml%2523HDRP_Windows64_playmode_trunk/412819/job

- [ ] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Medium

**Halo Effect**:Low


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/DrawRenderersCustomPassDrawer.cs
```diff
@@ -0,0 +1,256 @@
...
+		SerializedProperty      m_OverrideMaterial;
+		SerializedProperty      m_OverrideMaterialPass;
+		SerializedProperty      m_SortingCriteria;
+
+	    ReorderableList         m_ShaderPassesList;
+
+	    protected override void Initialize(SerializedProperty customPass)
+	    {
+		    // Header bools
+			m_FilterFoldout = customPass.FindPropertyRelative("filterFoldout");
```
sebastienlagarde: Indent

sebastienlagarde: all indent in this files seems fucked ?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineMenuItems.cs
```diff
@@ -122,6 +122,30 @@ static void MenuCreateDiffusionProfile()
             var icon = EditorGUIUtility.FindTexture("ScriptableObject Icon");
             ProjectWindowUtil.StartNameEditingIfProjectWindowExists(0, ScriptableObject.CreateInstance<DoCreateNewAssetDiffusionProfileSettings>(), "New Diffusion Profile.asset", icon, null);
         }
+        
+        [MenuItem("Assets/Create/Shader/HDRP/Custom FullScreen Pass")]
+        static void MenuCreateCustomFullScreenPassShader()
+        {
+            const string customPassTemplateGUID = "2e9bd7db24646bc409f60aa1d91e8883"; // CustomPassFullScreenShader.template
```
sebastienlagarde: use path instead of GUID

sebastienlagarde:
Remy, please check on hdrp/staging

PR #4319, opened 2019-08-11
Fix Mac test in HDRP/staging introduce with PR #4236

sebastienlagarde:
HDRP/Staging katana was 100% crashing on Mac Editor test. This PR fix it.

@TomasKiniulis  Looks like something bad happen with the change of custom render piple in the PR #4236. After a lot of try on katana, this seems the root cause of failure. Katana is green with it.


TomasKiniulis:
@sebastienlagarde so it was the changed GraphicsSettings.asset that was causing a crash right? Strange that it is causing it and only on mac. Tried running tests locally now and at least here and now it's not crashing anymore. I'll be more careful staging files to commit.

PR #4325, opened 2019-08-12
Universal RP Asset and Renderer workflow changes

Verasl:
### Purpose of this PR
The workflow for using custom Renderers was a bit unintuitive and also lead to it being hard to keep track of what was assigned to cameras in scene, this made it also hard at build time to know exactly what renderer and what settings were needed as we would have had to traverse every scene and prefab to check if they had custom renderers assign.

The Pipeline Asset now contains a list of all the renderers you wish to use in your project, this currently will be upgraded from previous assets but creating a forward renderer if type was ForwardRenderer, if the type was Custom, it will just insert the custom data into the first entry on the list.
Screenshot 2019-08-09 at 09 40 44
Here you can see an Object slot for the ScriptableRendererData, there is also a Default button to set which renderer is classed as the one any new camera will use. The Cog button is a shortcut to the renderer asset.

The Camera now uses a dropdown to select which renderer to use, this is stored as a int and has safety fallbacks if the requested renderer no longer appears.
Screenshot 2019-08-09 at 09 42 03
Screenshot 2019-08-09 at 09 42 31

---
### Release Notes
TODO - Need to voice this out well as there could potentially be some minor manual upgrading required.

---
### Testing status

- [ ] Built a player
- Other: 

**Automated Tests**: TODO

https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=universal%2Fasset-renderer-refactor&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fasset-renderer-refactor

---
### Overall Product Risks
**Technical Risk**: High - Changes to how renderers are loaded and stored, this can break rendering in both editor and builds.

**Halo Effect**: Low - Affects Universal RP Renderer logic.

---
### Comments to reviewers


TestProjects/UniversalGraphicsTest/Assets/Test/Runtime/CameraCallbackTests.cs
```diff
@@ -29,7 +29,9 @@ public CameraCallbackTests()
 
 	public override void Create()
 	{
-		ForwardRendererData data = UniversalRenderPipeline.asset.m_RendererData as ForwardRendererData;
+		ForwardRendererData data = UniversalRenderPipeline.asset.m_RendererDataList[0] as ForwardRendererData;
```
phi-lira: This will throw null reference exception if asset is null. 
Fix: Change to UniversalRenderPipeline?.asset.m_RendererDataList[0]

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -353,5 +376,92 @@ void DrawAdvancedSettings()
...
+                    if (GUI.Button(selectRect, Styles.rendererSettingsText))
+                    {
+                        Selection.SetActiveObjectWithContext(prop.GetArrayElementAtIndex(index).objectReferenceValue,
+                            null);
+                    }
+                }
+                else if(index == defaultRenderer)
+                {
+                    // missing defaut
+                    if (GUI.Button(selectRect, Styles.rendererDefaultMissingText))
```
phi-lira: In this case you don't need the if(). Just call GUI.Button.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -353,5 +376,92 @@ void DrawAdvancedSettings()
...
+                    {
+                        //
+                    }
+                }
+                else
+                {
+                    // missing renderer
+                    if (GUI.Button(selectRect, Styles.rendererMissingText))
```
phi-lira: Same here.

com.unity.render-pipelines.universal/Runtime/Data/PostProcessData.cs
```diff
@@ -22,7 +22,7 @@ public override void Action(int instanceId, string pathName, string resourceFile
             }
         }
 
-        [MenuItem("Assets/Create/Rendering/Universal Render Pipeline/Post-process Data", priority = CoreUtils.assetCreateMenuPriority1)]
+        [MenuItem("Assets/Create/Rendering/Universal Render Pipeline/Post-process Data", priority = 500)]
```
phi-lira: Instead of calling a single number here create another assetCreateMenuPriority in CoreUtils.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -486,6 +478,20 @@ bool DrawIntPopup<T>(SerializedProperty prop, ref T value, GUIContent style, GUI
             EndProperty();
             return hasChanged;
         }
+        
+        bool DrawBasicIntPopup(SerializedProperty prop, ref int value, GUIContent style, GUIContent[] optionNames, int[] optionValues)
```
phi-lira: Consider if this is generic enough to move to EditorUtils.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -93,13 +95,19 @@ public enum ColorGradingMode
...
 
         // Default values set when a new UniversalRenderPipeline asset is created
-        [SerializeField] int k_AssetVersion = 4;
+        [SerializeField] int k_AssetVersion = 5;
+        bool m_NeedsUpgrade;
 
+        // Deprecated settings for upgrading sakes
         [SerializeField] RendererType m_RendererType = RendererType.ForwardRenderer;
-        [SerializeField] internal ScriptableRendererData m_RendererData = null;
+        [SerializeField]internal ScriptableRendererData m_RendererData;
```
phi-lira: Why removing initialization to null?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -162,13 +170,15 @@ public class UniversalRenderPipelineAsset : RenderPipelineAsset, ISerializationC
 
         public static readonly string packagePath = "Packages/com.unity.render-pipelines.universal";
 
-        public static UniversalRenderPipelineAsset Create()
+        public static UniversalRenderPipelineAsset Create(ScriptableRendererData rendererData)
```
phi-lira: Public API change.  Maybe make default argument for rendererData = null and then add a case if that's null to create a RendererAsset in Assets folder.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -177,17 +187,46 @@ internal class CreateUniversalPipelineAsset : EndNameEditAction
...
             public override void Action(int instanceId, string pathName, string resourceFile)
             {
-                AssetDatabase.CreateAsset(Create(), pathName);
+                //Create asset
+                AssetDatabase.CreateAsset(Create(CreateRendererAsset(pathName, RendererType.ForwardRenderer)), pathName);
             }
         }
 
-        [MenuItem("Assets/Create/Rendering/Universal Render Pipeline/Pipeline Asset", priority = CoreUtils.assetCreateMenuPriority1)]
+        [MenuItem("Assets/Create/Rendering/Universal Render Pipeline/Pipeline Asset/Forward Renderer", priority = CoreUtils.assetCreateMenuPriority1)]
```
phi-lira: Consider if we really need to do this. It's adding one more depth to an already long menu, plus we have to update docs.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -177,17 +187,46 @@ internal class CreateUniversalPipelineAsset : EndNameEditAction
...
+            if (relativePath)
+                dataPath =
+                    $"{Path.Combine(Path.GetDirectoryName(path), Path.GetFileNameWithoutExtension(path))}_Renderer{Path.GetExtension(path)}";
+            else
+                dataPath = path;
+            AssetDatabase.CreateAsset(data, dataPath);
+            return data;
+        }
+        
+        static ScriptableRendererData CreateRenderer(RendererType type)
```
phi-lira: CreateRendererData?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -233,33 +272,44 @@ UniversalRenderPipelineEditorResources editorResources
         }
 #endif
 
-        public ScriptableRendererData LoadBuiltinRendererData()
+        public ScriptableRendererData LoadBuiltinRendererData(RendererType type)
```
phi-lira: You don't seem to need RendererType here. Also this is a public API change.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -297,21 +347,86 @@ public ScriptableRenderer scriptableRenderer
         {
             get
             {
-                if (scriptableRendererData.isInvalidated || m_Renderer == null)
-                    m_Renderer = scriptableRendererData.InternalCreateRenderer();
+                if (m_RendererDataList[m_DefaultRenderer] == null)
```
phi-lira: Should also check m_DefaultRenderer is not out of bounds here.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -297,21 +347,86 @@ public ScriptableRenderer scriptableRenderer
...
+                {
+                    Debug.LogError("Default renderer is missing from the current Pipeline Asset.", this);
+                    return null;
+                }
+                if (scriptableRendererData.isInvalidated || m_Renderers[m_DefaultRenderer] == null)
```
phi-lira: Add an asset to verify m_Renderers size is same as m_RendererDataList size. This help us catch issues and also avoid doing bounds check again for m_DefaultRenderer when indexing m_Renderers.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -93,13 +95,19 @@ public enum ColorGradingMode
...
 
+        // Deprecated settings for upgrading sakes
         [SerializeField] RendererType m_RendererType = RendererType.ForwardRenderer;
-        [SerializeField] internal ScriptableRendererData m_RendererData = null;
+        [SerializeField]internal ScriptableRendererData m_RendererData;
+        
+        // Renderer settings
+        [SerializeField] internal ScriptableRendererData[] m_RendererDataList = new ScriptableRendererData[1];
+        internal ScriptableRenderer[] m_Renderers;
+        [SerializeField] int m_DefaultRenderer = 0;
```
phi-lira: m_DefaultRendererIndex?

com.unity.render-pipelines.universal/Tests/Editor/EditorTests.cs
```diff
@@ -18,9 +18,12 @@ public void CreatePipelineAssetWithoutErrors()
 
         try
         {
-            UniversalRenderPipelineAsset asset = UniversalRenderPipelineAsset.Create();
+            ForwardRendererData data = ScriptableObject.CreateInstance<ForwardRendererData>();
+            UniversalRenderPipelineAsset asset =
+                UniversalRenderPipelineAsset.Create(data);
```
phi-lira: Minor. Fix identation.

com.unity.render-pipelines.universal/Tests/Editor/EditorTests.cs
```diff
@@ -63,7 +66,9 @@ public void ValidateBuiltinResourceFiles()
     [Test]
     public void ValidateNewAssetResources()
     {
-        UniversalRenderPipelineAsset asset = UniversalRenderPipelineAsset.Create();
+        ForwardRendererData data = ScriptableObject.CreateInstance<ForwardRendererData>();
+        UniversalRenderPipelineAsset asset =
+            UniversalRenderPipelineAsset.Create(data);
```
phi-lira: same identation issue here.

com.unity.render-pipelines.universal/Tests/Editor/EditorTests.cs
```diff
@@ -77,16 +82,19 @@ public void ValidateNewAssetResources()
...
 
     // When changing LWRP settings, all settings should be valid.
     [Test]
     public void ValidateAssetSettings()
     {
         // Create a new asset and validate invalid settings
-        UniversalRenderPipelineAsset asset = UniversalRenderPipelineAsset.Create();
+        ForwardRendererData data = ScriptableObject.CreateInstance<ForwardRendererData>();
+        UniversalRenderPipelineAsset asset =
+            UniversalRenderPipelineAsset.Create(data);
```
phi-lira: and here.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -587,8 +702,44 @@ public void OnAfterDeserialize()
...
+                }
+            }
+        }
+
+
+        static void UpgradeAsset(UniversalRenderPipelineAsset asset)
+        {
+#if UNITY_EDITOR
+            // This upgrade block moved the renderers from a single reference to the list.
+            if(asset.k_AssetVersion == 5)
```
phi-lira: should it be < 5 as we just bumped to 5?

Verasl: I'm not sure how to handle this actually, because our main upgrader is run when deserializing the asset, at this point we cannot call a lot of asset functions and one of them is being able to create a new renderer asset, so this has to come after the initial upgrade, meaning we are already at version 5, this part of the code will break down quickly as we go forwards. We should have a better discussion offline.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Verasl: I left this blank while looking for API to open the asset browser, this button now opens a object selector with a ScriptableRendererData filter.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Verasl: This has been merged with the above check, now the icon is switched inline rather than an additional `if`

com.unity.render-pipelines.universal/Runtime/Data/PostProcessData.cs
Verasl: Wanted to add it into Universal EditorUtils but access to that class is unavailable in the PostProcessData. Added 3rd priority in Core.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
Verasl: There is a few GUI commands in here, and they all return bools to allow us to check if we need to add AdditionalCameraData to the camera game object, this might be useful for when we add AdditionalLightData so potentially all could be moved?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: no I repurposed the same name, but should this be suffixed with `Index`?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -353,5 +377,89 @@ void DrawAdvancedSettings()
...
+            list.onRemoveCallback = li =>
+            {
+                if (li.serializedProperty.arraySize - 1 != m_DefaultRendererProp.intValue)
+                {
+                    if(li.serializedProperty.GetArrayElementAtIndex(li.serializedProperty.arraySize - 1).objectReferenceValue != null)
+                        li.serializedProperty.DeleteArrayElementAtIndex(li.serializedProperty.arraySize - 1);
+                    li.serializedProperty.arraySize--;
+                    li.index = li.count - 1;
+                }
+                else if (li.serializedProperty.arraySize == 1)
```
martint-unity: Check if this is needed

Verasl: Removed

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: Added back in

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: will change to brackets versions(also add 2D)

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: changed

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: added default type of forward to not change public api

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
phi-lira: We can leave here and move later when we need it.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
phi-lira: from API wise it makes more sense to be `Index`. Is this like this for upgrading reasons?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
phi-lira: Discussed this offline.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: Fixed

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: Fixed

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: I do this on create, which 'should' always happen before this.
```
if(m_Renderers == null || m_Renderers.Length < m_RendererDataList.Length)
                m_Renderers = new ScriptableRenderer[m_RendererDataList.Length];
```

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: Fixed with suggestion

com.unity.render-pipelines.universal/Runtime/UniversalAdditionalCameraData.cs
```diff
@@ -136,16 +135,16 @@ public bool requiresColorTexture
...
-
-                if (m_RendererData.isInvalidated || m_Renderer == null)
-                    m_Renderer = m_RendererData.InternalCreateRenderer();
+            get => UniversalRenderPipeline.asset.GetRenderer(m_RendererIndex);
+        }
 
-                return m_Renderer;
-            }
+        /// <summary>
+        /// Allows you to set this cameras current ScriptableRenderer to one listed on the render pipeline asset. Takes an index that maps to the list on the pipeline asset.
```
Katstodian: Please replace with:

Use this to set this Camera's current ScriptableRenderer to one listed on the Render Pipeline Asset. Takes an index that maps to the list on the Render Pipeline Asset.

com.unity.render-pipelines.universal/Runtime/UniversalAdditionalCameraData.cs
```diff
@@ -136,16 +135,16 @@ public bool requiresColorTexture
...
+        /// </summary>
+        /// <param name="index">The index that maps to the ReendererData list on the currently assigned pipeline asset</param>
```
Katstodian: Please change "pipeline asset" to "Render Pipeline Asset"

PR #4326, opened 2019-08-12
Post process quality settings

FrancescoC-unity:
This PR is the first instance of volume quality settings, for now it includes all post processing effects.  Once this go in and is approved, a similar PR will be done for lighting volume quality settings. 

Users can select tiers of quality in the HDRP Asset (Example for DoF)

image

Then on the volume, quality settings can be used for performance sensitive info, as follow: 

Before selecting "Use Quality Setting": 
image

After: 

image

This also will allow to set different tiers for different platforms as they're defined on the asset. 

Katana just kicked: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?automation-tools_branch=master&ScriptableRenderLoop_branch=HDRP%2Fvolumes-quality-settings&unity_branch=trunk&sort=1-asc


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineUI.cs
```diff
@@ -544,6 +555,145 @@ static void Drawer_SectionPostProcessSettings(SerializedHDRenderPipelineAsset se
...
+            }
+
+            m_ShowDoFHighQualitySection = EditorGUILayout.Foldout(m_ShowDoFHighQualitySection, k_HighQualityContent);
+            if (m_ShowDoFHighQualitySection)
+            {
+                int quality = (int)VolumeQualitySettingsLevels.High;
+                DrawDepthOfFieldQualitySetting(serialized, quality);
+            }
+        }
+
```
FrancescoC-unity: Let me remove the double space :P

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/Bloom.cs
```diff
@@ -26,18 +27,57 @@ public sealed class Bloom : VolumeComponent, IPostProcessComponent
...
+        {
+            get
+            {
+                if (!UsesQualitySettings())
+                {
+                    return m_Resolution.value;
+                }
+                else
+                {
+                    HDRenderPipeline pipeline = (HDRenderPipeline)RenderPipelineManager.currentPipeline;
```
Chman: Seeing how this is used everywhere we need quality settings, we might as well make it a protected getter in the VolumeComponent base class?
Also beware that this will trigger a nullref if an HDPipeline isn't set and the user tries to access it via script, but that's a fairly uncommon (and probably useless) use case.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
```diff
@@ -0,0 +1,88 @@
...
+            BloomRes[(int)VolumeQualitySettingsLevels.Low] = BloomResolution.Quarter;
+            BloomRes[(int)VolumeQualitySettingsLevels.Medium] = BloomResolution.Half;
+            BloomRes[(int)VolumeQualitySettingsLevels.High] = BloomResolution.Half;
+
+            BloomHighQualityFiltering[(int)VolumeQualitySettingsLevels.Low] = false;
+            BloomHighQualityFiltering[(int)VolumeQualitySettingsLevels.Medium] = true;
+            BloomHighQualityFiltering[(int)VolumeQualitySettingsLevels.High] = true;
+
+            /* Chromatic Aberration */
+            ChromaticAberrationMaxSamples[(int)VolumeQualitySettingsLevels.Low] = 4;
```
Chman: I would have set low to 3 (minimum possible value) as it matches the 3 color channel split. And maybe medium = 6 and high = 12. Higher values are really overkill except for film use IMO.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
```diff
@@ -0,0 +1,88 @@
...
+            MotionBlurSampleCount[(int)VolumeQualitySettingsLevels.Medium] = 8;
+            MotionBlurSampleCount[(int)VolumeQualitySettingsLevels.High] = 12;
+
+            /* Bloom */
+            BloomRes[(int)VolumeQualitySettingsLevels.Low] = BloomResolution.Quarter;
+            BloomRes[(int)VolumeQualitySettingsLevels.Medium] = BloomResolution.Half;
+            BloomRes[(int)VolumeQualitySettingsLevels.High] = BloomResolution.Half;
+
+            BloomHighQualityFiltering[(int)VolumeQualitySettingsLevels.Low] = false;
+            BloomHighQualityFiltering[(int)VolumeQualitySettingsLevels.Medium] = true;
```
Chman: Medium and High are the same quality here. Is this intended? Else I'd put hqfiltering to `false` for medium.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
FrancescoC-unity: 👍 Change done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
FrancescoC-unity: Yup, it is to mantain the default == Medium (we discussed this offline)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Components/Bloom.cs
FrancescoC-unity: 👍  Change done

PR #4329, opened 2019-08-12
Make PBR sky compile again on xbox

FrancescoC-unity:
The compiler gets confused at a fairly deep level with the current code path, to make it deviate, changing the number of samples work. This makes it compile again on xbox.

The new path is slower, but the shader is used in pre computation.


EvgeniiG:
That compiles, but visually speaking it will outright fail because 90 is not a Fibonacci number. We probably need a better fix.

FrancescoC-unity:
Gah, of course, I even thought of that, but forgot. 
Is 55 an acceptable value or is too low?

EvgeniiG:
Probably good enough, we just need to get it running to start with.

FrancescoC-unity:
Bumped to 144 and it compiles. It is probably overkill, but being precomputation as long as it compiles I'd say it's good.

PR #4338, opened 2019-08-13
Added HDRP test 2007_Light_Volumetrics

iM0ve:
Test is temporarily disabled. There is an issue when trying to disable volumetrics in light properties.
Fogbugz: https://fogbugz.unity3d.com/f/cases/1175970/
After the issue is fixed, I will make a ref image and enable the test.

Test result:
image


RemyUnity:
You can use the "Broken Test" prefab to highlight that the "volumetrics off" part is broken : https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/TestProjects/HDRP_Tests/Assets/GraphicTests/Common/ScenePrefabs/BrokenTestText.prefab
And still have the test enabled with the prefab included in the reference image, so it will effectively test for the other features of the scene.

sebastienlagarde:
hey, what is the status of this test? should we add it? note that this test number conflict with this one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4328 which I have merge. So please update current number of this test. thanks

iM0ve:
I did this test for Volumetric fog component, now there is no such thing. Its just Fog. The scene needs an overhaul to restore lost parameters, maybe add new ones to fully represent fog component.

PR #4340, opened 2019-08-13
Hybrid Renderer Custom Instancing

arnaudcarre:
### Purpose of this PR
this PR makes shadergraph compatible with custom material instanced data used in dots hybrid renderer.

### Testing status

We use our own project test ( git@github.cds.internal.unity3d.com:unity/TestHybridHDRP7.git )

 git@github.cds.internal.unity3d.com:unity/TestHybridHDRP7.git

---
### Overall Product Risks
low. Only affect people using dots hybrid renderer ( very few)

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaderUtilities.cs
```diff
@@ -759,10 +771,24 @@ public static bool GenerateShaderPass(AbstractMaterialNode masterNode, Pass pass
             ShaderGenerator instancingOptions = new ShaderGenerator();
             {
                 instancingOptions.AddShaderChunk("#pragma multi_compile_instancing", true);
-                if (pass.ExtraInstancingOptions != null)
+                if (instancedCount > 0)
                 {
-                    foreach (var instancingOption in pass.ExtraInstancingOptions)
-                        instancingOptions.AddShaderChunk(instancingOption);
+                    instancingOptions.AddShaderChunk("#if SHADER_TARGET >= 35 && (defined(SHADER_API_D3D11) || defined(SHADER_API_GLES3) || defined(SHADER_API_GLCORE) || defined(SHADER_API_XBOXONE) || defined(SHADER_API_PSSL) || defined(SHADER_API_VULKAN) || defined(SHADER_API_METAL))");
```
sebastienlagarde: This is not a good idea to generate all the define here as you don't have control in case of change. You should have a single define like

#define SUPPORT_DOTS_INSTANCING SHADER_TARGET >= 35 && (defined(SHADER_API_D3D11) || defined(SHADER_API_GLES3) || defined(SHADER_API_GLCORE) || defined(SHADER_API_XBOXONE) || defined(SHADER_API_PSSL) || defined(SHADER_API_VULKAN) || defined(SHADER_API_METAL))"

then in code generation you do 

instancingOptions.AddShaderChunk("#if SUPPORT_DOTS_INSTANCING ")

then if we add new platform (which we will do!), we don't need to regenerate all the shaders, just recompiling them

PR #4363, opened 2019-08-20
Fixed broken render scale when the final post-processing pass is active (1177660)

Chman:
### Purpose of this PR
Render scale would break rendering when the final post-processing pass is active due to bad viewport handling.

Fogbugz [#1177660](https://fogbugz.unity3d.com/f/cases/1177660/)

---
### Testing status
Local tests are green. Yamato running, see CI.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Medium, might have an impact on viewport handling in post but all automated tests and manual testing seem fine.


Chman:
Closing this, the fix is part of #4369 among other things.

PR #4390, opened 2019-08-23
Raytracing Node for HDRP shaders

fredpointzero:
### Purpose of this PR
Add a new Raytracing node to handle different code path for Raytracing passes.

---
### Release Notes
Add a new Raytracing node to handle different code path for Raytracing passes.

---
### Testing status
**Katana Tests**: -

**Automated Tests**: -

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---


anisunity:
@fredpointzero would be great to add a test scene in the dxr project

JulienIgnace-Unity:
@zadi0648 
> So we had a short discussion, and we aren't sure this should go in if it's only for a GDC demo, because then we have to support it indefinitely. And it could just be a keyword? (This isn't a hard PR rejection, just that the discussion needs to happen)

Who was this discussed with? This new node has nothing to do with GDC and is made to allow people to do "LoDs" for raytracing specifically. It's also not a keyword as it's not driven by the material but only by the way the shader graph is setup.

fredpointzero:
> 
> 
> @fredpointzero would be great to add a test scene in the dxr project

I added a test scene in the DXR project.

This is the expected result:
image

PR #4394, opened 2019-08-23
DXR warnings, tooltips and UX

anisunity:
Fixing a warning in the ray tracing sampling shader
Correcting issues in the UX for ray tracing volument components
Changing tool tips for ray tracing


iM0ve:
When DXR is enabled now all non DXR related settings are hidden, improving user experience. Tooltips adjusted.

ISSUES found:
1)**Screen Space Reflection->Bounce count** tooltip incorrect "Controls the length of reflection rays"
2) **Filter** is now renamed to **Denoiser**, however not in all instances. Shadows and GI override still have **Filter**. Same with Filter/Denoiser Radius

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/ShaderVariablesRaytracing.hlsl
```diff
@@ -20,3 +20,4 @@ int                                     _RayCountEnabled;
 float                                   _RaytracingCameraNearPlane;
 uint                                    _RaytracingDiffuseRay;
 RWTexture2D<uint4>                      _RayCountTexture;
+int 									_RaytracingPreExposition;
```
JulienIgnace-Unity: Indentation

PR #4397, opened 2019-08-23
Fix issue where XR RenderTargets were not being properly 

caryhaynie:
### Purpose of this PR
Currently, UniversalRP has an optimization to minimize the number of texture clears being done. Unfortunately, as implemented, this behavior fails to properly account for the fact that the default render targets in the BeforeRendering stage are then swapped out with the XR render targets when switching to XR rendering later in the frame.

The solution is to reset m_FirstCameraRenderPassExecuted when switching to XR to make sure the XR render targets are properly cleared.

---
### Release Notes
Fixed issue where XR render targets were not being correctly cleared.

---
### Testing status

- [ ] Built a player
- Other: 

---
### Overall Product Risks

---
### Comments to reviewers


caryhaynie:
Katana in progress: https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=xr%2Ffix-urp-on-legacy-xr&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

thomas-zeng:
Looks good!

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -232,7 +232,14 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
...
 
             if (stereoEnabled)
+            {
                 BeginXRRendering(context, camera);
+                // Enabling XR will modify the camera render targets,
+                // (the actual switch is performed in `context.SetupCameraProperties`)
+                // so we need to reset m_FirstCameraRenderPassExecuted
+                // to ensure the XR render targets are properly cleared.
+                m_FirstCameraRenderPassExecuted = false;
+            }
```
mikesfbay: @caryhaynie This seems like a roundabout way to get URP to do a clear. Are you debugging something that's rendering directly to device tex? The logic to check camera target is done in ExecuteRenderPass: 
            if (passColorAttachment == m_CameraColorTarget && !m_FirstCameraRenderPassExecuted)
            {
                m_FirstCameraRenderPassExecuted = true;
...
The better way would be to add additional XR target checks here instead.

thomas-zeng: @mikesfbay Hey mike, this bool m_FirstCameraRenderPassExecuted basically represents a state for the current camera target(whether it is cleared or not). In this case, the underlying camera target gets swapped to XR target, so we have to flush this cached state since it is no longer valid.

thomas-zeng: I am not sure adding additional XR target checks would help if we don't flush the invalid m_FirstCameraRenderPassExecuted state. Maybe I am understanding this wrong, could you elaborate a bit on what you are proposing?

caryhaynie: relatedly:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/804c09e684dba8a7dc4d926002faef314e230b18/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L391

that only gets called when we do the clear. Should we be tying the clear to the occlusion mesh like this?

mikesfbay: We shouldn't be overloading/overriding the meaning of variable m_FirstCameraRenderPassExecuted just so XR can workaround the clear. It'll be harder to debug issues later. If XR changes the target, we can add another variable (i.e. m_IsUsingXRRenderTarget = true) and do the clear based on a check on that variable.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -368,7 +371,10 @@ void ExecuteRenderPass(ScriptableRenderContext context, ScriptableRenderPass ren
                 passDepthAttachment = m_CameraDepthTarget;
             }
 
-            if (passColorAttachment == m_CameraColorTarget && !m_FirstCameraRenderPassExecuted)
+            bool needsClear = !m_FirstCameraRenderPassExecuted;
```
phi-lira: Please add a comment here explaining reasonings.

caryhaynie: Add comment for extra XR clear.

mikesfbay:
@caryhaynie Can you include the fogbugz case link? Also adding QA to verify on Rift/GO.

caryhaynie:
@mikesfbay https://fogbugz.unity3d.com/f/cases/1188600/

caryhaynie:
ping @DennisDeRykeUnity

DennisDeRykeUnity:
The Render Pipelines project on Katana has been decommissioned but I ran Universal and Universal Stereo tests on Yamato.  We can ignore the Windows standalone failures as well as the single Windows playmode test failure (044_Lighting_ReflectionProbe) due to known bugs.  The OSX standalone and playmode failures also occur in recent SRP master branch, so they are not the fault of this PR.

DennisDeRykeUnity:
I saw no new problems when I tested players built from the URP Template scene with Legacy XR enabled on Rift using: 
- Unity Editor built from 2010.1 (changeset 19a26876029b) 
- Stereo Instancing (other modes have known bugs) 
- SRP from dev branch xr/fix-urp-on-legacy-xr commit 13a6c23
- Both HDR and MSAA on, and both HDR and MSAA off 

Tomorrow I'll run similar tests on Oculus Go.

caryhaynie:
@phi-lira Should be ready to merge

phi-lira:
I couldn't see tests. Was yamato passing? If so it's fine to merge it trunk.

caryhaynie:
@phi-lira Yamato tests are now passing: **https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/xr%252Ffix-urp-on-legacy-xr/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/783308/job**

ellioman:
What's the status on this? Is this still needed and should be merged? @caryhaynie

ellioman:
After speaking with @caryhaynie it was decided to close this PR as its out of date and no longer relevant.

PR #4410, opened 2019-08-27
Inner cone angle is now working.

martint-unity:
### Purpose of this PR
This PR is exposing the inner cone angle handles and the User can now use those to drive the inner angle of the light.

inner_cone_angle

---

- [X] Built a player
- Other: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/urp%252Flight-inner-cone-angle/.yamato%252Fupm-ci-universal.yml%2523All_Universal/400465/job

---
### Overall Product Risks
Medium

Low

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -145,8 +145,8 @@ namespace UnityEngine.Rendering.Universal
 
 #if UNITY_EDITOR
             SceneViewDrawMode.ResetDrawMode();
-            Lightmapping.ResetDelegate();
 #endif
+            Lightmapping.ResetDelegate();
             CameraCaptureBridge.enabled = false;
         }
 
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -4,7 +4,6 @@ using Unity.Collections;
 using UnityEditor;
 using UnityEditor.Rendering.Universal;
 #endif
-using UnityEngine.Experimental.GlobalIllumination;
 using UnityEngine.Scripting.APIUpdating;
 using Lightmapping = UnityEngine.Experimental.GlobalIllumination.Lightmapping;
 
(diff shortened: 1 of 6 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -148,9 +145,8 @@ protected override void Dispose(bool disposing)
 
 #if UNITY_EDITOR
             SceneViewDrawMode.ResetDrawMode();
-#endif
-
             Lightmapping.ResetDelegate();
```
uygar3d: As you're always setting the delegate now, resetting it shouldn't be editor only.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -181,11 +184,71 @@ void SortCameras(Camera[] cameras)
...
+                        SpotLight spotLight = new SpotLight();
+                        LightmapperUtils.Extract(light, ref spotLight);
+                        spotLight.innerConeAngle = light.innerSpotAngle * Mathf.Deg2Rad;
+                        lightData.Init(ref spotLight);
+                        break;
+                    case LightType.Area:
+                        RectangleLight rectangleLight = new RectangleLight();
+                        LightmapperUtils.Extract(light, ref rectangleLight);
+                        lightData.Init(ref rectangleLight);
+                        light.lightmapBakeType = LightmapBakeType.Baked;
```
uygar3d: You shouldn't change light properties here. Rather set the baked type on the rectangle light before you call lightData.Init.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -181,11 +184,71 @@ void SortCameras(Camera[] cameras)
...
+                        break;
+                    case LightType.Disc:
+                        DiscLight discLight = new DiscLight();
+                        LightmapperUtils.Extract(light, ref discLight);
+                        lightData.Init(ref discLight);
+                        light.lightmapBakeType = LightmapBakeType.Baked;
```
uygar3d: Same here as above.

PR #4419, opened 2019-08-27
Refactor stencil management in the HDRP

EvgeniiG:
### Purpose of this PR
Unify stencil management across all materials and simplify the entire pipeline.
Rough outline here: https://docs.google.com/spreadsheets/d/1-85pJ1w7elwIMKR7HQmQGaekoZuPwHvlkusFrkRTstw/edit#gid=0.
Also use a single coarse stencil representation to remove all the HTILE crap.
Remove all redundant stencil copies.
Free G-Buffer space from the material ID. Classify materials with the stencil wherever possible.

---
### Release Notes
Improve stencil management in the HDRP.
---
### Testing status
**Katana Tests**: No.

**Manual Tests**: Will fail horribly unless you manually upgrade all scenes, materials and screenshots.

---
### Overall Product Risks
**Technical Risk**: HIGH.

**Halo Effect**: Medium.

---
### Comments to Reviewers
Start with BaseUnlitGUI.cs.
Lit.shader is the reference implementation.
Still need to clean the code of all the legacy crap.
Currently the focus is on making it pass tests.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1782,31 +1826,63 @@ AOVRequestData aovRequest
...
+            }
+
             // In both forward and deferred, everything opaque should have been rendered at this point so we can safely copy the depth buffer for later processing.
             GenerateDepthPyramid(hdCamera, cmd, FullScreenDebugMode.DepthPyramid);
             // Depth texture is now ready, bind it (Depth buffer could have been bind before if DBuffer is enable)
             cmd.SetGlobalTexture(HDShaderIDs._CameraDepthTexture, m_SharedRTManager.GetDepthTexture());
 
             if (shouldRenderMotionVectorAfterGBuffer)
             {
+                // THIS WILL NOT WORK WITH THE STENCIL CHANGES. It will overwrite the correct stencil value with an incorrect one.
```
sebastienlagarde: Tagging for future reference.
Motion vector pass need a full depth buffer. If there is no prepass, we must wait GBuffer to fill the Gbuffer. (Alternative would be to render during GBuffer pass, but not an option with current legacy management of motion vector.

sebastienlagarde:
Could you make a list of what is missing currently in the PR? Like, are all master node handled?

EvgeniiG:
> Could you make a list of what is missing currently in the PR? Like, are all master node handled?

Every material is handled (unless I forgot something...) except for the PBR master with a hardcoded stencil.
HDRP tests need to be upgraded.
Need to do something about motion vectors after G-Buffer.
The stencil setup code needs to know the material type, so it relies on MaterialExtension.GetMaterialId, which doesn't work with the shader graph.

EvgeniiG:
Oh, and I want to stress it again, we need to test, our test coverage is far from complete.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: Is this even true? Isn't our motion vector pass writing depth?

EvgeniiG:
The first step would be to see whether what I am doing makes sense and if it's what we want.

sebastienlagarde:
Repalce by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5581

PR #4352, opened 2019-08-16
Universal/fix zero length computebuffer

phi-lira:
### Purpose of this PR
- Fixed scene view and shader graph preview turning grey: https://fogbugz.unity3d.com/f/cases/1175094/
- Fixed errors about trying to create zero length compute buffers.
- Fixed VFX tests.
---
### Release Notes
 Added missing note for a Shader Graph preview fix. 
 The other fixes don't require changelog entries as they are regressions introduced only in master.

---
### Testing status
**Katana Tests**: First off we need to make sure the Katana SRP tests are green?

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

**Automated Tests**: Extended Reflection Probes tests to contain per-object and sky probes.

Launch Katana (Link on master here - update for your branch):
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=universal/fix-zero-length-computebuffer&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

Alternative, launch Yamato (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?

---
### Overall Product Risks
**Technical Risk**: Medium. Contains core lighting changes.

**Halo Effect**: Medium. Affects Vulkan, Metal, PS4 and Xbox One.

---
### Comments to reviewers
 For now disabling the SSBO path. It's not fully tested and it contains some Vulkan binding issues that need more investigation.

PR #4423, opened 2019-08-28
Added code to ensure we always create a buffer of at least one element

bobvodka:
### Purpose of this PR
Fixes an issue where an exception about a zero sized compute buffer would be thrown when using Vulkan or Metal as the graphics API.

Light culling it seems can sometimes return zero lights during the first frame, which causes the ShaderData to request a zero sized compute buffer which isn't valid. This change ensures we always request a buffer of at least 1 element.

---
### Release Notes
N/A

---
### Testing status

**Manual Tests**: Local tested just to make sure things worked as intended.
**Automated Tests**:  Nothing new added.

---
### Overall Product Risks
**Technical Risk**: None
**Halo Effect**: Low

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ShaderData.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ShaderData.cs
@@ -57,6 +57,7 @@ namespace UnityEngine.Rendering.Universal
 
         ComputeBuffer GetOrUpdateBuffer<T>(ref ComputeBuffer buffer, int size) where T : struct
         {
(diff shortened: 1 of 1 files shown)


phi-lira:
Hi, the zero compute buffer issue was fixed in this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4352/

However there was another issue found in Vulkan. For now we disabled the SSBO path until we can do wider testing with SSBO. Light limits are unnafected (still 256)

PR #4433, opened 2019-08-28
Restore camera alpha compositing to HDRP

JulienIgnace-Unity:
### Purpose of this PR
Alpha channel is now properly exported to camera render textures when using FP16 color buffer format.

---
### Testing status
Added a new graphics test (9007)

**Katana Tests**: [Running](https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Falpha-compositing)

- [ ] Built a player


FrancescoC-unity:
I think we should at least produce a warning for when it is not supported or when something silently fails (e.g. low res transparent)

JulienIgnace-Unity:
The issue is that we can't really know if this is used as it's not a "feature" per se. There is no checkbox for us to check and compare to the current settings.

FrancescoC-unity:
> The issue is that we can't really know if this is used as it's not a "feature" per se. There is no checkbox for us to check and compare to the current settings.

Ha, yes. 

Oh well, shame.

PR #4454, opened 2019-09-02
Hdrp/custom render resource sets

tlaedre:
### Purpose of this PR

A draft suggestion for allowing custom render resource sets to be attached to HDRP alongside the builtin renderpipeline resources.

See https://unity.slack.com/archives/C20E70CAC/p1566813874362300 for some context


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/SerializedHDRenderPipelineAsset.cs
```diff
@@ -7,43 +7,22 @@ class SerializedHDRenderPipelineAsset
     {
         public SerializedObject serializedObject;
 
-        public SerializedProperty renderPipelineResources;
         public SerializedProperty currentMaterialQualityLevel;
         public SerializedProperty materialQualityLevels;
+        public SerializedProperty renderPipelineResources;
```
sebastienlagarde: are we missing the raytracing resource here?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineResources.Migration.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
-    public partial class RenderPipelineResources : ScriptableObject, IVersionable<RenderPipelineResources.Version>
+    public partial class HDRenderPipelineResources : IVersionable<HDRenderPipelineResources.Version>
```
sebastienlagarde: what is the goal or renaming RenderPipelineResources  => HDRenderPipelineResources  ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineResources.cs
```diff
@@ -5,8 +5,10 @@
 namespace UnityEngine.Rendering.HighDefinition
 {
     [HelpURL(Documentation.baseURL + Documentation.version + Documentation.subURL + "HDRP-Asset" + Documentation.endURL)]
-    public partial class RenderPipelineResources : ScriptableObject
+    public partial class HDRenderPipelineResources : BaseRenderResources
```
sebastienlagarde: what is the goal or renaming RenderPipelineResources => HDRenderPipelineResources ?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/Settings/SerializedHDRenderPipelineAsset.cs
RSlysz: It seams to not be exposed in UI.
@anisunity Is it on purpose?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipelineResources.cs
tlaedre: It was mainly for naming consistency since I was changing these related files and names anyway, but I can rip out that change.

Chman: I actually welcome this change, having `RenderPipelineResources` as a name when everything else is `HDWhatever` (even the `HDRenderPipelineResource.cs` file itself!) always confused me. Assuming it won't break anything, I would keep that change.

sebastienlagarde: ok, we are fine with the change then

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/BaseRenderResources.cs
```diff
@@ -0,0 +1,52 @@
...
+        public override void OnInspectorGUI()
+        {
+            DrawDefaultInspector();
+
+            // Add a "Reload All" button in inspector when we are in developer's mode
+            // .. currently only for 'builtin' types to mimic existing behaviour.
+            //
+            // TODO: Why not just always show this reload button? For assets in read-only packages (like most cases
+            //       of SRP usage) this button will be disabled anyway.
+            //
```
tlaedre: thoughts on this TODO @sebastienlagarde @RSlysz? I don't quite see the value in going to the extra trouble of hiding this button for certain types unless in developer mode (which end-users will never be, but they might still be customizing HD)

RSlysz: If you display a disabled button, user will try to find a way why it is disabled. 
It was at start only a tool for us developer to force reloading in order to verify we do not make wrong path in the reloader.
Now that this is somewhat exposed in user land. I think it is preferable to let the user use this debug button too.
So show it up :)

FrancescoC-unity:
Tried to build a player (on PS4, but not sure it matters) with this on and the editor resources become unset when building, thus failing.

tlaedre:
@FrancescoC-unity I don't quite understand this statement:
> and the editor resources become unset when building, thus failing
they are supposed to become unset for the build as they are editor-only resources, are you saying they become permanently unset in the Editor source data after a build?

@FrancescoC-unity @Chman any hint about what failed? I built standalone windows for both HDRPTests and TheHeretic with this change without any failures.

Chman:
Right, it is supposed to be unset for the build, but there might be a check in place somewhere because when this gets unset it throws an error in the console and fails the build as a result. I'll have to look deeper into it tomorrow to see exactly why and how it fails.

Chman:
On 2020.1a7 it builds & run fine now.

sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp can one of you test this PR? need to build players on various platform

TomasKiniulis:
@tlaedre still getting failings builds. While building template project, Render Pipeline Editor Resources does get permanently unset after building, I also still get missing editor resources errors while building, but the build succeeds if only "Build" is pressed
image
if "Build & Run" is pressed it fails. Checked on 2020.1.0a7 and 2020.1.0a11

PR #4456, opened 2019-09-02
Hdrp/hd shader stripping changes

tlaedre:
### Purpose of this PR

Various optimizations to HD stripper as discussed here: https://unity.slack.com/archives/C20E70CAC/p1567065601011400

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
@@ -42,23 +42,22 @@ namespace UnityEditor.Rendering.HighDefinition
             // Here we can't strip based on opaque or transparent but we will strip based on HDRP Asset configuration.
 
             bool isMotionPass = snippet.passName == "MotionVectors";
-            bool isTransparentPrepass = snippet.passName == "TransparentDepthPrepass";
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -247,33 +247,38 @@ public void OnProcessShader(Shader shader, ShaderSnippetData snippet, IList<Shad
...
                             break;
                         }
                     }
 
                     if (removeInput)
-                    {
-                        inputData.RemoveAt(i);
-                        i--;
-                    }
+                        inputData[i] = inputData[--inputShaderVariantCount];
```
joravainen: I've no idea on how C# containers work internally. However, I'm wondering how the perf would compare to instead of modifying the input list, just building a new list out of the passing variants and then in the end swap the whole list?

tlaedre: Yeah, I did consider that approach as well @joravainen, and it's possible that it would be even faster. The reality is that the awful iteration time on this didn't really inspire a lot of exploration towards the optimal solution given the current limitations in data flow - I just tried to make it a little bit more bearable.

PR #4458, opened 2019-09-03
Add reflection fade distance

RSlysz:
### Purpose of this PR
Add falloff distance on Reflection Probe

Note that even if weight is 0 or distance is too high, the update is still done. This should be done in an other PR due to package release approaching. ( @laurenth-unity )

Unity_2019-09-03_09-54-02

---
### Release Notes
Added falloff distance on Reflection Probe and Planar Reflection Probe

---
### Testing status

- [x] Opened test project + Run graphic tests locally (only 5004 failing locally as on other computers)
- [ ] Built a player

**Automated Tests**: 

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: Low

---


laurenth-unity:
This is great but since the idea was to use it for optimization purposes and that part is not done yet I wouldn't prioritize merging this for the package. It's confusing to offer the option and not do the optimization that people will assume we do.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.Migration.cs
```diff
@@ -65,7 +66,9 @@ protected enum Version
             MigrationStep.New(Version.AddReflectionFrameSetting, (HDProbe data) =>
             {
                 FrameSettings.MigrateToNoReflectionSettings(ref data.m_ProbeSettings.camera.renderingPathCustomFrameSettings);
-            })
+            }),
+            MigrationStep.New(Version.AddFadeDistance, (HDProbe data)
```
fredpointzero: I don't think this is required, only a default assignment with the default value for an added property will do the job.

RSlysz:
^ rebased on tip of HDRP/staging as it was getting old (also removed the unnecessary migration as it was conflicting)

RSlysz:
@fredericv-unity3d This seams to work but we can see the non updated texture in the preview of planar which is strange. It should be computed by the preview camera while displaying it but I cannot find how at the moment.

RSlysz:
@sebastienlagarde : cannot remove previous fred id (fredpointzero) nor ask to review again.

PR #4461, opened 2019-09-03
Universal Asset upgrade improvements

Verasl:
### Purpose of this PR
The upgrade path was prone to not being fully called when upgrading a project with multiple pipeline assets, this PR now causes all upgrading to happen with the deserialisation step.

Also an extra null check was added in the event that you had a camera inspector visible while removing Universal RP form the graphics settings.

---
### Release Notes
N/A
---
### Testing status

- [ ] Built a player
- Other: 

Katana:
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=universal%2Fasset-upgrade-improvements&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fasset-upgrade-improvements

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers


phi-lira:
Adding Martin for awareness on the CameraEditor. Why are we changing test image files?

Verasl:
> Adding Martin for awareness on the CameraEditor. Why are we changing test image files?

I'm only adding missing ones, this is dues to getting failures because of them missing. The other test image files are just meta files, these were causing issues as the comparison was with compressed images, meaning we had very slight variation.

PR #4465, opened 2019-09-03
Fixing Ray Tracing bugs (XR, Lights)

anisunity:
Correcting a bug related to shadow maps when a screen space shadow is enabled
Removing exception when deleting a light
Correcting visual artifacts related to xr matrices
Ignoring disabled lights with disabled ligth components


enqui:
Thank you for making sure DXR works properly with XR too.  Please provide working DXR support in 19.3 Beta with a new package 7.xx HDRP build. Thank you.

PR #4472, opened 2019-09-04
Fix sample scene for shadergraph & vfx

PaulDemeulenaere:
WIP


PaulDemeulenaere:
First pass of clean fix here : https://github.cds.internal.unity3d.com/unity/vfx-graph/pull/179 (screenUV is now correct)

PR #4473, opened 2019-09-04
Hdrp/scalable settings ui update

fredpointzero:
### Purpose of this PR
[REQUIRES ScalableSettings PR](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4468)

Updates the UI of scalable settings in Volumes:
- Merge Use Quality Setting boolean and Quality Level enum in a single field
- Always show all the fields that previously depended on "use quality setting"

image
image

---
### Release Notes
-

---
### Testing status
**Katana Tests**: -
**Automated Tests**: -

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---


FrancescoC-unity:
This should sync up with  https://github.com/Unity-Technologies/ScriptableRenderPipeline/pulls?page=2&q=is%3Apr+is%3Aopen as well. The lighting stuff is already existing in that PR along with many other things.

FrancescoC-unity:
I have some issues in having only 3 levels for shadow resolutions, feels like the granularity here is too coarse.

While I appreciate the need of having only one set of levels, we really need to discuss this further. I can see in a game quite easily to have: 

- Crappy lights that need shadows (low)
- Average light (medium)
- High quality lights (hero lights in gameplay)
- Cinematic stuff (very high). 

Limiting to three sounds bad to me, or well, cuts the leg of the feature a bit.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
```diff
@@ -2,66 +2,76 @@
...
 {
-    public enum VolumeQualitySettingsLevels
+    [Serializable]
+    public sealed class ScalableSettingLevelParameter : IntParameter
     {
-        Low = 0,
-        Medium = 1,
-        High = 2
-    }
+        public ScalableSettingLevelParameter(ScalableSetting.Level level, bool useOverride, bool overrideState = false)
```
FrancescoC-unity: Shouldn't really live here anymore. See also the Volume Lighting Quality PR

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
```diff
@@ -2,66 +2,76 @@
...
+            : base(useOverride ? ScalableSetting.LevelCount : (int)level, overrideState)
+        {
 
-    [Serializable]
-    public sealed class QualitySettingParameter : VolumeParameter<VolumeQualitySettingsLevels> { public QualitySettingParameter(VolumeQualitySettingsLevels value, bool overrideState = false) : base(value, overrideState) { } }
+        }
+
+        public (ScalableSetting.Level level, bool useOverride) level
```
FrancescoC-unity: I think the naming here is confusing. 

the outer level should be called something else, the level.level pattern is confusing

FrancescoC-unity:
I am not a big fun of having Low (full spelling), Med (compressed form), High (full spelling). 

Not pointing a single file since is spread-ish

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
```diff
@@ -180,12 +145,12 @@ public static HDShadowAtlasInitParams GetDefault()
             directionalShadowsDepthBits         = k_DefaultShadowMapDepthBits,
             punctualLightShadowAtlas            = HDShadowAtlasInitParams.GetDefault(),
             areaLightShadowAtlas                = HDShadowAtlasInitParams.GetDefault(),
-            directionalLightsResolutionTiers    = HDShadowTierParams.GetDefault(),
-            punctualLightsResolutionTiers       = HDShadowTierParams.GetDefault(),
-            areaLightsResolutionTiers           = HDShadowTierParams.GetDefault(),
+            shadowResolutionDirectional         = new IntScalableSetting { low = 256, med = 512, high = 1024 },
```
FrancescoC-unity: I'd say for directional light 1024 is a far cry for high quality :-)

FrancescoC-unity: See my comment in the PR to discuss this.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
```diff
@@ -2,66 +2,76 @@
...
-        static int s_QualitySettingCount = Enum.GetNames(typeof(VolumeQualitySettingsLevels)).Length;
+        static int s_QualitySettingCount = ScalableSetting.LevelCount;
 
         public GlobalPostProcessingQualitySettings()
         {
             /* Depth of Field */
-            NearBlurSampleCount[(int)VolumeQualitySettingsLevels.Low] = 3;
-            NearBlurSampleCount[(int)VolumeQualitySettingsLevels.Medium] = 5;
-            NearBlurSampleCount[(int)VolumeQualitySettingsLevels.High] = 8;
+            NearBlurSampleCount[(int)ScalableSetting.Level.Low] = 3;
```
FrancescoC-unity: If I understood correctly the point of it, why aren't this arrays *ScalableSetting now?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/ScalableSetting.cs
```diff
@@ -0,0 +1,117 @@
...
+{
+    public static class ScalableSetting
+    {
+        public static readonly int LevelCount = Enum.GetValues(typeof(Level)).Length;
+
+        public enum Level
+        {
+            Low,
+            Medium,
+            High
```
FrancescoC-unity: This creates a very strong tie between the enum and the list of properties below. 

If we ever change this to include an extra quality level, lots of code needs changing. 

Why not having the fields as arrays of size LevelCount ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
fredpointzero: Then what would be appropriate default values?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
fredpointzero: It will be done in further PR, this one is intended for next release and to unify the UI on volumes.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/ScalableSetting.cs
fredpointzero: I don't think the enum is an issue here, it is always a good practice in C# to have this when you enumerate something.

This is more about the implementation of the `ScalableSetting<T>` itself.
I usually prefer to embed the properties when the umber of properties will not likely change over the time. (Once settled down on the number of level, this should not change much).
Also, arrays makes an allocation on the heap and insert write guards for the garbage collector, not ideal for small types like int.
And there is not much location you can write simpler with an array (only the `this` implementation).

Well that's a tradeoff because there is no inline generic array nor codegen/macro in C# (that would solve perfectly this)

fredpointzero:
>     * )

The number of quality level per object needs to be settled down. Currently 3 levels are chosen.
Also, we should use the same number of level for all scalable settings, for API and UX reasons.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/GlobalPostProcessingQualitySettings.cs
fredpointzero: Will move in another PR

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowManager.cs
FrancescoC-unity: As discussed in the other comment, I am not a big fan of having 3 values only, that's the core issue.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/ScalableSetting.cs
FrancescoC-unity: I was not referring to the enum.

 I was referring to the manual enumeration of fields below the enum. We have the enum to tell us how many fields there are, but we don't use that to define the fields. 
We have a dissonance where we use enum in some places to determine the fields and in some other places (fields in this class) we manually put the fields ignoring the enum. 

The array shouldn't be allocated every frame, so don't see the problem of allocating 12 bytes on the heap, unless I am missing something here.

FrancescoC-unity:
> > I have some issues in having only 3 levels for shadow resolutions, feels like the granularity here is  too coarse.
> > While I appreciate the need of having only one set of levels, we really need to discuss this further. I > can see in a game quite easily to have:
> > Crappy lights that need shadows (low)
> > Average light (medium)
> > High quality lights (hero lights in gameplay)
> > Cinematic stuff (very high).
> > Limiting to three sounds bad to me, or well, cuts the leg of the feature a bit.
> 
> The number of quality level per object needs to be settled down. Currently 3 levels are chosen.
> Also, we should use the same number of level for all scalable settings, for API and UX reasons.

And what I am saying is that we should further discuss this. 

I am fine on settling on a number, but here we need to be aware we are mixing the quality settings of oranges and potatoes and I think it is going to severely limit the usefulness of it. 
At least the defaults need changing, will speak with Laurent today to find good ones.

FrancescoC-unity:
> This should sync up with https://github.com/Unity-Technologies/ScriptableRenderPipeline/pulls?page=2&q=is%3Apr+is%3Aopen as well. The lighting stuff is already existing in that PR along with many other things.

I was meaning to link to this https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4411

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/ScalableSetting.cs
Chman: I think you missed the point Francesco was making. The problem isn't about using an enum, that's fine (and safe). The problem is about tying the enum values to what's hardcoded in that `ScalableSetting<T>` class.

> Also, arrays makes an allocation on the heap and insert write guards for the garbage collector, not ideal for small types like int.

The allocation isn't really an issue here, it'd be done once per setting. Beside, replacing the three hardcoded quality values (which are also allocated on the heap anyway) + that switch in the getter/setter by an array access isn't going to make a noticeable difference (we're talking pulling from an array vs executing jumps in a switch case), at least not compared to all the indirections there are with the rest of the code (CPU wise).

What Francesco is worried about mostly has to do with code maintenance in the future. And this pattern looks like a code smell to me too.

> I usually prefer to embed the properties when the umber of properties will not likely change over the time. (Once settled down on the number of level, this should not change much).

It shouldn't, but it could. I get what you're saying, but then why did we need this PR? If we take that logic further, the previous code was simple enough and didn't need to change because it wasn't likely to change over time. As far as I'm aware, this PR was done to make it more flexible so I don't understand why it doesn't apply to this specific bit too, despite everything else around it being engineered to be flexible & reusable.

sebastienlagarde:
@FrancescoC-unity @fredpointzero  Let's do a call for the 3 values

iM0ve:
Found one bug with Undo. Skipping some states. For example in Depth Of Field override you set quality from Medium to Custom and hit undo it will produce incorrect results.
image

Nice to have improvement would be to update the values in override when switching quality. I understand this might not be a priority right now though.
image

UI in some cases gets confusing. Tested on a few people in the office, they either misinterpreted or didnt understand what the dropdown does. UX should be addressed at some point.
image

PR #4476, opened 2019-09-04
Fog Refactoring

JulienIgnace-Unity:
### Purpose of this PR
Refactoring of fog components:
- Deprecated all old Fog volume components (Linear, Exponential and Volumetrics)
- Replaced them with a new "Fog" component. This fog is an exponential fog that uses the former Volumetrics Fog component distant fallback parametrization.
- Volumetric is now an option of this new Fog component.
- Temporarily removed "PBR Fog" generated by the PBR sky.
- Added a script to update Volume Profile with the new fog when possible (Edit->RenderPipeline -> Upgrade Fog Volume Component)
- Updated all tests scenes using fog.

---
### Testing status
**Katana Tests**: [Running](https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffog-refactoring)

- [ ] Built a player


com.unity.render-pipelines.core/Runtime/Volume/VolumeComponent.cs
```diff
@@ -18,6 +18,14 @@ public VolumeComponentMenu(string menu)
         }
     }
 
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public sealed class VolumeComponentDeprecated : Attribute
+    {
+        public VolumeComponentDeprecated()
```
Chman: Note: you don't even need the constructor here if you want to reduce the code size, C# will automatically create one for you if you don't provide one.

com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/Fog.cs
```diff
@@ -0,0 +1,154 @@
...
+        public static bool IsFogEnabled(HDCamera hdCamera)
+        {
+            var fogComponent = VolumeManager.instance.stack.GetComponent<Fog>();
+            return hdCamera.frameSettings.IsEnabled(FrameSettingsField.AtmosphericScattering) && fogComponent.enabled.value;
+        }
+
+        public static bool IsVolumetricLightingEnabled(HDCamera hdCamera)
+        {
+            var fogComponent = VolumeManager.instance.stack.GetComponent<Fog>();
+            return hdCamera.frameSettings.IsEnabled(FrameSettingsField.Volumetrics) && fogComponent.enableVolumetricFog.value;
```
Chman: Nitpicking here but as this is called at runtime every little cycle helps... You can save a potential lookup (`GetComponent`) by inlining it in the test. (also valid for `IsFogEnabled()`)

PR #4485, opened 2019-09-04
Add refresh after a frame on history element of debugmenu to handle sanitization

RSlysz:
### Purpose of this PR
On Runtime DebugMenu, when changing framesettings debug element to something that is not supported by sanitation, it display impossible value even if not applied by the engine.
This fix force a refresh one frame later to ensure sanitized value are showed for debug too.

Concrete example: If you put DeferredOnly as supported render mode in the HDRPAsset, on runtime debug menu, you can change it to forward and it remain blocked on it until you close and reopen it. 

---
### Release Notes

---
### Testing status

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: Low

---
### Comments to reviewers
@FrancescoC-unity  for notification


sebastienlagarde:
Please add a changelog in another PR.

PR #4504, opened 2019-09-05
Expose Camera Capture

dmm-john:
### Purpose of this PR
CameraCaptureBridge was created for Unity Recorder but since it uses reflection to pull actions from Recorder there was no way for anyone else to access this functionality.
Since all render pipelines support these actions, it makes sense to simply have the list of actions in render-core and for Unity Recorder and others to set the actions there.

---
### Release Notes
A PR with Unity Recorder is also being submitted.  That project is the only one directly affected by this change.

---
### Testing status

- Tested changes with Unity Recorder to make sure they are still compatible
- Created my own runtime capture that will make use of these exposed actions
---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**:  Low

---
### Comments to reviewers
This PR was the result of an open discussion between the Unity Recorder team and members of all render pipelines.
Initially, we were going to create a new package.  It was much simpler to just make a change to CameraCaptureBridge.
Apologies for whitespace changes.  The only change necessary is CameraCaptureBridge.cs


dmm-john:
Sorry, there's a good chance I'm requesting this merge into the wrong branch, master is probably wrong.   Let me know where this should be directed if it's wrong.

com.unity.render-pipelines.core/Runtime/Utilities/CameraCaptureBridge.cs
```diff
@@ -53,44 +61,86 @@ static CameraCaptureBridge()
...
             }
         }
 
-        static public IEnumerator<Action<RenderTargetIdentifier, CommandBuffer> > GetCaptureActions(Camera camera)
+        /// <summary>
+        /// Provides the set actions to the renderer to be triggered at the end of the render loop for camera capture
+        /// </summary>
+        /// <param name="camera">The camera to get actions for</param>
+        /// <returns>Enumeration of actions</returns>
+        public static IEnumerator<Action<RenderTargetIdentifier, CommandBuffer>> GetCaptureActions(Camera camera)
```
dmm-john: The GetCaptureActions is the important part here.  If reflection is used and actions are found, use them, otherwise fall through and try and find actions stored in the class.

dmm-john:
I've modified the code to maintain the existing code path.  This should allow no changes to the Recorder package until they're ready.

sebastienlagarde:
Yamato is green except a test due to VFX unrelated to the PR (it is on master). Let's move forward with this PR.

PR #4510, opened 2019-09-05
[DO NOT MERGE]Asset Upgrader fixes

Verasl:
### Purpose of this PR
On certain test projects and test runs, some Universal RP assets were not upgraded with the secondary upgrade path(one outside of deserialisation) due to the event not being called.

---
### Release Notes
Cleaned up upgrade code, much less prone to failure due to compilation issues and editor event order.

---
### Testing status
Launch Katana (running):
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=universal%2Ffix-asset-upgrader&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

Yamato (running Universal RP pipelines):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Ffix-asset-upgrader

---
### Overall Product Risks
**Technical Risk**: Medium - changes to how the upgrade system works

**Halo Effect**: None - does not affect any other code aside from the asset itself

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -711,13 +706,11 @@ public void OnBeforeSerialize()
 
         public void OnAfterDeserialize()
         {
-            k_AssetPreviousVersion = k_AssetVersion;
-
             if (k_AssetVersion < 3)
             {
                 m_SoftShadowsSupported = (m_ShadowType == ShadowQuality.SoftShadows);
+                k_AssetPreviousVersion = k_AssetVersion;
```
phi-lira: Why can't we update the previous version outside these ifs? I'm concerned about in some cases it will be initialized or when we add more if clauses we forget to assign the previous version.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -307,18 +306,14 @@ protected override RenderPipeline CreatePipeline()
...
-                    UpgradeAsset(this);
-                }
-                else
-                {
-                    Debug.LogError(
-                        $"Default Renderer is missing, make sure there is a Renderer assigned as the default on the current Universal RP asset:{UniversalRenderPipeline.asset.name}",
-                        this);
+                // If previous version and current version are miss-matched then we are waiting for the upgrader to kick in
+                if(k_AssetPreviousVersion != k_AssetVersion)
                     return null;
```
phi-lira: Why can't we call upgrade asset here?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: This is because deserialisation can happen multiple times before the delayed editor call is called, meaning if we are outside, we come back here with the version being 5, previous being set to 5 and then when the upgrade gets called from the delayed call it thinks it is all up to date and does not upgrading.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Verasl: Potentially, I will give it a go.

Verasl: Doing the upgrade here causes there renderer to run without upgrading happening, I think this is due to the asset database needing to create thumbnails for all assets, and the upgrade needing to run once the asset database has finish the initial loading of assets

PR #4551, opened 2019-09-09
Fix lit transparent refraction with XR single-pass rendering

fabien-unity:
### Purpose of this PR
Refraction was broken with single-pass

---
### Testing status
**Katana Tests**: not started

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low


FrancescoC-unity:
Uhm I am not entirely convinced by this change, we indeed end up with much coarser information now to deal with. The depth here is based on the proxy volume which is either a sphere or a box, so really really coarse, while before, we used actual scene information. 
My guess is that we'll have many more false negatives. I can double check on Monday as on a laptop setting up a full repro scene might be hard. 

What is that makes it break in SPI? I am sure we have a similar pattern in SSR?

fabien-unity:
Oh indeed, I thought the depth value would be identical but you're right, it's not.
I can restore the code and debug the SPI version.

PR #4552, opened 2019-09-09
Fix for flickering gizmos in matcap mode

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1178543/ . 

The simplest way to fix this is to just have the probe gizmos to render black when in matcap view. The problem is that gizmos are not rendered via HDRP so this is the simplest way to go around the problem of having excessively high values caused by the exposure of the main scene.


FrancescoC-unity:
@sebastienlagarde  only lit ones are going to become black, that is just the probe spheres AFAIK, all the other gizmos render as usual

sebastienlagarde:
Ok. Good for me

PR #4559, opened 2019-09-10
HDRP shader stripping optimization/refactor

joravainen:
### Purpose of this PR
HDRP shader stripping scripts were structured in a way that made it easy to make the mistake of doing per-shader work in the innermost variant stripping loop. When the shader variant amounts approach gazillions it can make a big difference.
 
I restructured the BaseShaderPreprocessor interface into three stages:
1) PrepareShaderStripping: Do all the global preparation here, pass name checks etc.
2) ShouldStripShader: Check if this whole shader pass could be stripped (e.g. SceneSelectionPass).
3) ShouldStripVariant: Check if a specific variant should be stripped (keyword based checks)

Torbjörn's recent optimization pass already removed the biggest slowness in the stripping. However, this PR shaves some ~10% further off a warm project build time. A synthetic stripping test was sped up by ~50%

---
### Release Notes
Refactored/optimized HDRP stripping scripts

---
### Testing status
Yamato tests look ok. HDRP_Standalone has the same failure as the master has atm (some lighting difference)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: None

---

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using UnityEditor.Build;
 using UnityEditor.Build.Reporting;
+using UnityEditor.Rendering.Utilities;
 using UnityEngine;
 using UnityEngine.Rendering;
 using UnityEngine.Rendering.HighDefinition;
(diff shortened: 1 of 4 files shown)


joravainen:
I think I was exaggerating a bit about the gazillions based on some assumptions :P
Now, to look at actual data from the builds I've been running:
Total number of OnProcessShader invocations: 1052
Remaining/total variants: 5132/191896

It seems that indeed majority of the invocations handle only a handful of variants (around 2-16). However there are some that hundreds. The biggest one I saw was 1296 (HDRP/Lit - Forward pass).

The ratio of script invocations vs number of variants per invocation might explain why my change caused slight perf drop on some cases as I introduced some overhead per invocation with the multi-step stripping approach.

I think I'll leave further investigation/optimization of the stripping script side to you guys as I've at least made my point here already. I'll be looking further on things on the native side as there seems to be some duplicate processing going on and indeed jobifying the stripping invocations could help quite a bit here.

PR #4561, opened 2019-09-10
Workaround a graphics driver bug in the GTAO shader

JussiKnuuttila:
### Purpose of this PR
The GTAO shader produces an incorrect result on Vulkan using Nvidia hardware. This is probably caused by a driver bug. By modifying the shader slightly, the issue disappears. This change should be reverted once the driver bug is fixed.

---
### Release Notes
Fix a driver issue in the GTAO shader on Vulkan.

---
### Testing status

- [x] Built a player
- Other: 

Manually compared 2602_SSAO_FullRes and 2601_SSAO_HalfRes. Attached screenshots from 2602_SSAO_FullRes before and after this PR.
Before:
2602-vulkan-before
After:
2602-vulkan-after

Ran Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/vulkan%252Fhdrp-test-fixes2/.yamato%252Fupm-ci-projects.yml%2523all_project_ci/431842/job/pipeline

2601_SSAO_HalfRes
2602_SSAO_FullRes

---
### Overall Product Risks
**Technical Risk**: Low
**Halo Effect**: None

---
### Comments to reviewers
The original shader had just "-dir", which was being disregarded by the driver shader compiler. The workaround replaces it with "(-dir + 1e-30)", which is practically the same thing (epsilon is extremely tiny), but compiles correctly.


JussiKnuuttila:
Looks like this should be going into HDRP/staging instead, attempting to change...

PR #4564, opened 2019-09-10
DXR UI and Fixes

anisunity:
Renaming ray tracing menu items and ray tracing environement labels
Changing the behavior for non denoised area light shadow (ray traced)
Fixing a exception occuring when LOD have different submesh count


iM0ve:
Can confirm that:
- In RaytracingEnvironment, Indirect diffuse now is called Global Illumination to match the override name
- Denoiser can now be enabled and disabled on Area lights

PR #4568, opened 2019-09-10
[VFX] Improve Motion Vector Test

PaulDemeulenaere:
### Purpose of this PR
- Modify render debug render view of Motion Vectors (see `DebugFullScreen.shader`)
-- Use a nine tap filter to evaluate the arrow direction (abitrary but it helps small particles detection)
-- Color value relies on motion vector speed per pixel
- Use this debug view to compare graphic test `24_MotionVector.unity` : compare motion vector instead of compare result of motion blur.

### Motion Vectors Debug
*Before*
image

*After*
image

The change is subtle but intensity now relies on length of motion vector :
image
Intensity was a constant of 0.15f now it linearly interpolates between 0.03f and 0.50f.

### Image reference for 24_MotionVector
**Before** : 512x512, motion blur is compared
image

**After**: 1024x1024, motion vector debug view is compared
image

N.B.: This change **doesn't** require any backport.


PaulDemeulenaere:
There was a change to commit in HDRP test play mode.
Before :
image
After :
image

PR #4577, opened 2019-09-11
Fixed linear to sRGB conversion happening twice on some Android devices.

aleks01010101:
### Purpose of this PR
Fixed linear to sRGB conversion happening twice on some Android devices.

---
### Release Notes
Fixed linear to sRGB conversion happening twice on some Android devices.

---
### Testing status

- [ ] Built a player
- Other: 
- [x] Built and ran manually on an Android device

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low (Android only)

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -243,7 +244,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
 
                     if (requiresFinalPostProcessPass)
                     {
-                        m_FinalPostProcessPass.SetupFinalPass(m_ActiveCameraColorAttachment);
+                        m_FinalPostProcessPass.SetupFinalPass(m_ActiveCameraColorAttachment, true);
```
phi-lira: Do we have a scenario this would be false. If not we should remove this boolean from API.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -116,14 +120,16 @@ public void Setup(in RenderTextureDescriptor baseDescriptor, in RenderTargetHand
             m_InternalLut = internalLut;
             m_IsFinalPass = false;
             m_HasFinalPass = hasFinalPass;
+            m_EnableSRGBConversionIfNeeded = enableSRGBConversion;
         }
 
-        public void SetupFinalPass(in RenderTargetHandle source)
+        public void SetupFinalPass(in RenderTargetHandle source, bool enableSRGBConversion)
```
phi-lira: Looking at the code I'm not sure we need this bool. We could initialize m_EnableSRGBConvertionIfNeeded to true.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -107,7 +111,7 @@ public PostProcessPass(RenderPassEvent evt, PostProcessData data)
             m_ResetHistory = true;
         }
 
-        public void Setup(in RenderTextureDescriptor baseDescriptor, in RenderTargetHandle source, in RenderTargetHandle destination, in RenderTargetHandle depth, in RenderTargetHandle internalLut, bool hasFinalPass)
+        public void Setup(in RenderTextureDescriptor baseDescriptor, in RenderTargetHandle source, in RenderTargetHandle destination, in RenderTargetHandle depth, in RenderTargetHandle internalLut, bool hasFinalPass, bool enableSRGBConversion)
```
phi-lira: From looking at the code it seems we can infer the srgb convertion if needed by looking at source and dest + if it's final target. I'd rather abstract this parameter if possible.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -225,10 +225,11 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             // we need to stay in a RT
             if (afterRenderExists)
             {
+                bool willRenderFinalPass = (m_ActiveCameraColorAttachment != RenderTargetHandle.CameraTarget);
```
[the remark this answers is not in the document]
aleks01010101: Well, it's used everywhere in this function...

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
aleks01010101: Removed this bool. Final pass indeed always does color conversion if needed.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
aleks01010101: I'm not sure if we can. Right now it doesn't looks like we use dest at all for figuring that out.
I'd prefer to leave this bool in here.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
aleks01010101: Fixed.

PR #4582, opened 2019-09-11
Fix for decal not affecting specular occlusion

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1178272/


FrancescoC-unity:
Updated this PR with also the cases for shadergraph

PR #4593, opened 2019-09-12
Fix for decal mask map not considered for spec occlusion

FrancescoC-unity:
IMPORTANT! Note this is dependent on this being merged first: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4582  (this branched off that branch)

Fix for https://fogbugz.unity3d.com/f/cases/1178380/

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed emissive area light mesh non-uniform scale issues (1171739)
 - Fixed Hair and PBR shader graphs double sided modes
 - Fixed an issue where updating an HDRP asset in the Quality setting panel would not recreate the pipeline.
+- Fixed issue that prevented decals from modifying specular occlusion (case 1178272).
 
 ## [7.1.1] - 2019-09-05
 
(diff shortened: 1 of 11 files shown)


FrancescoC-unity:
> In the case where Spec AO is based on AO, it make sense that the decals affect it, but in the other case it doesn't make sense.

At the moment there is nothing on the decal that tells us how to compute spec AO and this PR changes the behaviour only when the source doesn't have spec AO specified. What this PR is doing is saying that if Spec AO is determined only by decal, it will be determined via the maskmap of the decal. 

> So from my point of view, the fix is just to move AO applicaiton by decal before evaluation of spec AO, but that's all. If we setup specAO to be 1.0, we shouldn't change it. It mean it is a users choice to not have spec AO.

But here is different, the user decided that the receiver has no spec AO, but it could be that the decal material needs to receive spec AO (e.g. a metallic decal on a dielectric surface might want spec AO since we can modify also the normals for example). 

> (In current state the PR is non consistent anyway as it perform a check only for lit shader but not for shader graph version of fabric, eye etc...)


This is because the shader graph passes seems to always the spec ao from AO so there is no need to do anything extra, decal will affect SpecAO as is. 
In fact, the inconsistencies are here before this PR and this PR makes it a bit more consistent :)

sebastienlagarde:
We will discuss Monday. There is a misunderstanding.. also the sahder graph allow to do what you want with spec AO. In particular looks at stacklit spec AO (there is 3 different method)

FrancescoC-unity:
Closing this as it will be now part of a slightly bigger change (change of UI in Lit).

PR #4598, opened 2019-09-13
Clearing the directional screen space shadow (fixes a bug with ray traced shadows)

anisunity:
in a scene with a point light with a ray traced shadow, change the light type to directional and back to point. The directional shadow is still computed and outputed to the same buffer. This PR fixes that behavior


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1988,6 +1988,10 @@ HDProbe SelectProbe(VisibleReflectionProbe probe, PlanarReflectionProbe planarPr
                     m_CurrentScreenSpaceShadowData[i].valid = false;
                 }
 
+                m_CurrentSunLight = null;
```
JulienIgnace-Unity: This is done above for all but m_CurrentSunLightAdditionalLightData  parameter

PR #4603, opened 2019-09-13
Reworked how matcap view is triggered to avoid having it in game view when scene lighting toggle is off

FrancescoC-unity:
Tried quite a few things, but this is the least intrusive way. 

Fix for https://fogbugz.unity3d.com/f/cases/1156102/


FrancescoC-unity:
A problem with exposure handling has been found, will investigate monday

FrancescoC-unity:
This should now work, could you try it again @JMargevics? Thanks

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -3299,7 +3301,7 @@ static void RenderComputeAsPixelDeferredLighting(in DeferredLightingParameters p
             }
             else
             {
-                CoreUtils.SetKeyword(cmd, "DEBUG_DISPLAY", parameters.debugDisplaySettings.IsDebugDisplayEnabled());
+                CoreUtils.SetKeyword(cmd, "DEBUG_DISPLAY", true/* parameters.debugDisplaySettings.IsDebugDisplayEnabled()*/);
```
JulienIgnace-Unity: Is this intended?

FrancescoC-unity: Nope, debug code, good catch

PR #4625, opened 2019-09-17
Group node fixes

martint-unity:
### Purpose of this PR
Fixes for these fog bugzzzz

Sticky note not being grouped properly (FB-1181447)
Before:
group_sticky_note_broken

After:
group_sticky_note

Coy nodes from a group causes errors (FB-1182137)
Before:
copy_node_from_group_broken

After:
copy_node_from_group

---
### Testing status

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fui-ux-group-node-fixes/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph/468503/job

---
### Overall Product Risks
Low

Low

---


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - You can now use the right-click context menu to delete Sticky Notes.
 - You can now save your graph as a new Asset.
 
+### Fixed
+- Sticky Notes can now be grouped properly.
```
phi-lira: Please link the public bug from issue tracker

Kink3d: This is not a convention we follow (at this time). Why is this a valid reason to deny a PR?

martint-unity:
@zadi0648 @pbbastian 
Anything you guys wants to add? :)

martint-unity:
@zadi0648 Updated with your feedback

zadi0648:
Why make ungroup the last contextual menu item? Menu items of the same kind should be together ya?
GroupNodeFixes0

Instead I'd recommend just giving the sticky note the Group Option, for parity/consistency, and keeping Ungroup right under that
GroupNodeFixes1

martint-unity:
I totally agree! Done :)

zadi0648:
Improved!

Minor parity thing, but the option to group a single node/sticky note is greyed out for a sticky note, while not so for main nodes:

Grouping0
Grouping1

Also, as for the `Delete Group and Contents` item, could we perhaps improve that some more in this PR? Suggesting:

- Add a separator between it and the other stuff (Cut/Copy)
- Add another option with it above `Delete Group` (that only deletes the group, not the contents - follows the rule that every hotkey should come with a menu option somewhere)
- It doesn't show up when the right click happens to be on Nodes or Sticky Notes (see 2nd image). Perhaps it should remain that way? I'll leave that to your discretion.
- Perhaps the grouping stuff should remain together (despite it being greyed out at the bottom)

Grouping2
Grouping3

zadi0648:
Actually I find that Nodes are inconsistent of when you can and can't group them (when they are alone), do I make this a new fogbugz?

Repro:
1) Left Click GV Background
2) Right Click on Node
> Option to Group Node is available
3) Left Click GV Background
4) Right Click GV Background
5) Right Click on Node
> Option to Group Node is unavailable

CantGroup0

martint-unity:
Fixed the above mentioned issue.
Also added nicer menus.
Groups are also a bit transparent so that you can see if another group is under it.
Also added so that you can select a group "behind" another group and move it.

martint-unity:
I also added shortcuts to the menu. Should be OS agnostic.

martint-unity:
This fails due to Adds or something @sophiaaar Knows more about it. But Master has the same issues for the last 6 days so Nothing this branch introduced so this should be safe to push @pbbastian

PR #4626, opened 2019-09-17
Fix leak in box spotlight shadows 

FrancescoC-unity:
Fix for https://fogbugz.unity3d.com/f/cases/1184475/ 

For QA: The bug does not have a repro scene, but you can easily create one with something like this:
InkedUnity_Avyha94pJU_LI

(the above scene shows the issue, not the fix :) Please let me know if you need the repro scene I guess I can upload it, although is fairly easy to recreate)

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
@@ -26,15 +26,15 @@
 #endif
 
 #ifdef SHADOW_LOW
-#define PUNCTUAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias, maxCoord) SampleShadow_PCF_Tent_3x3(_ShadowAtlasSize.zwxy, posTC, tex, samp, bias, maxCoord)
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
```diff
@@ -26,15 +26,15 @@
 #endif
 
 #ifdef SHADOW_LOW
-#define PUNCTUAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias) SampleShadow_PCF_Tent_3x3(_ShadowAtlasSize.zwxy, posTC, tex, samp, bias)
-#define DIRECTIONAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias) SampleShadow_PCF_Tent_5x5(_CascadeShadowAtlasSize.zwxy, posTC, tex, samp, bias)
+#define PUNCTUAL_FILTER_ALGORITHM(sd, posSS, posTC, tex, samp, bias, bboxCoords) SampleShadow_PCF_Tent_3x3(_ShadowAtlasSize.zwxy, posTC, tex, samp, bias, bboxCoords)
```
FrancescoC-unity: I assume the name bboxCoords is a bad name, I can split it in minCoords and maxCoords or shadowMapMinMaxCoords

FrancescoC-unity:
Found an alternative, cheaper way to deal with the issue, will open a PR in few minutes.

PR #4640, opened 2019-09-17
Initial Targets implementation

Kink3d:
### Purpose of this PR
This PR adds the very beginnings of the Targets framework for code generation and integrations.

- Moves validity checks from `ISubShader` and `IMasterNode` to the per-pipeline Target variants for `MeshTarget`
- Passes `ITarget` instance through to `ISubShader.GetSubShader` for use in future PRs. This is temporary as eventually this will be passed directly to the generator instance and subshaders wont be involved, but right now it will allow us to move code from SubShaders and Resources files to the Target definitions.
- Moves `ISubShader` and `IMasterNode` to internal. This is a temporary change as these interfaces will be removed by stack node work, but there is a worry here as `ISubShader.GetSubShader` takes argument types from Shader Graphs private assembly. This method can be removed entirely once HDRP generators are in line with Universal and can be moved inside of the Shader Graph package.

NOTE: This PR is based on the current PR for preview refactor work. This is because it relies on a single definition for `ISubShader` and the removal of unneeded methods from that interface.


Kink3d:
Replaced by #4940

PR #4647, opened 2019-09-18
Failsafe for XR/PostFX when not in single-pass instanced mode

Chman:
### Purpose of this PR
Post-processing isn't currently supported in single-pass (non instanced). This PR disables postfx when the selected XR mode isn't single-pass instanced.

Note: there should also be a global warning in the editor when "multi-pass" is selected as the whole pipeline fails with that mode.


phi-lira:
Discussed with @daves-unity , we will not take this, a proper fix will be done.

PR #4661, opened 2019-09-18
Fix zBufferParams with XR single-pass rendering

fabien-unity:
### Purpose of this PR
- Restore original code from https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4551
- Fix zBufferParams and LinearEyeDepth(..) with XR single-pass rendering

---
### Testing status
**Katana Tests**: not run

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -465,7 +465,7 @@ void UpdateAllViewConstants(bool jitterProjectionMatrix, bool updatePreviousFram
                         combinedViewMatrix.SetColumn(3, combinedOrigin);
                     }
 
-                    projMatrix = combinedProjMatrix;
+                    projMatrix = GL.GetGPUProjectionMatrix(combinedProjMatrix, true);
```
sebastienlagarde: is it correct to do this?
just sanity check but GetGPUProjectionMatrix should only be use when we transfer the matrix to the GPU, not if the matrix is used on C# side.

(But agree that the code from before was like this, just wondering if we have side effects...)

fabien-unity: Note this is only for the VR path (and it was not correct). The standard path is untouched

fabien-unity: See line 490 in HDCamera.cs:
```
var gpuProj = GL.GetGPUProjectionMatrix(cameraProj, true); // Had to change this from 'false'
```

TomasKiniulis:
It feels like PR introduces undesired artifacts. Refracted objects get flipped and they become refractors themselves, which is an opposite result compared to Staging. Adding comparison gifs:

hdrp-xr-fix-zbufferparams
ZcHkTFZXnU

staging
hVp0lDiXl2

It also seems to fail to refract at certain angles
hdrp-xr-fix-zbufferparams
tmWgMXWsmu
staging
ChGmZAKoUk

fabien-unity:
@TomasKiniulis : could you compare before https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4551 ? This is where behavior was changed and I want to restore previous behavior.
To clarify: this current PR should be identical to behavior before https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4551

TomasKiniulis:
@fabien-unity compared now with how it was before #4551. Did not observe any differences whatsoever.
Before #4551
prev
Current PR
zbufferparams

So just to be clear, the current PR is the expected result to restore back? The refracted spheres appear to be flipped, is that a desired effect?

fabien-unity:
Yes, the flip is expected with refraction. cc @sebastienlagarde , @FrancescoC-unity

PR #4664, opened 2019-09-18
Fix font shaders in test projects for VR by using a Shader Graph version

fabien-unity:
### Purpose of this PR
- one step closer to be able to run XR tests on Yamato

---
### Testing status
**Katana Tests**: not run

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
```diff
@@ -9,20 +9,23 @@
 namespace UnityEngine.Rendering.HighDefinition
 {
     // XRTODO: custom user defined views + culling
-    internal enum XRLayoutOverride
+    public enum XRLayoutOverride
```
sebastienlagarde: caution: if you make this public you must document it, here it miss API documentation

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
```diff
@@ -9,20 +9,23 @@
...
     {
         None,                       // default layout
         TestComposite,              // split the  into tiles to simulate multi-pass
         TestSinglePassOneEye,       // render only eye with single-pass path
     }
 
-    internal class XRSystem
+    public class XRSystem
```
sebastienlagarde: same here

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
sebastienlagarde: is this what you want? otherwise we can make the HDRP test project seen internal of HDRP package (wondering if it is already the case?

fabien-unity: Oh yes I'd like know more to keep this internal but visible to com.unity.testing.hdrp
How to do this ?

sebastienlagarde: You need to edit the asm file of hdrp package (take a look at asm file of shader graph, they add other package to see their internal definition), here you can add com.unity.testing.hdrp in asmdef of hdrp

fabien-unity: So far no success. I get cyclic dependencies.
Do you know which file exactly is used for this trick with Shader Graph ?

sebastienlagarde:
Fixed the internal accessibility issue

PR #4670, opened 2019-09-19
Fix leak in box spotlight shadows

FrancescoC-unity:
Same problem as in  https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4626 (so fix bug https://fogbugz.unity3d.com/f/cases/1184475/) but alternative way that should be cheaper, even though less precise.

The heuristic is a bit empirically decided, but it worked well for lots of cases I tried. For QA, pls refer to the other PR for an example of broken case.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1282,6 +1282,18 @@ static Vector3 EvaluateAtmosphericAttenuation(Vector3 L, Vector3 positionWS)
...
                 lightData.right *= 2.0f / Mathf.Max(additionalLightData.shapeWidth, 0.001f);
                 lightData.up    *= 2.0f / Mathf.Max(additionalLightData.shapeHeight, 0.001f);
+
+                // If we have shadows, we need to shrink the valid range so that we don't leak light due to filtering going out of bounds.
+                lightData.boxLightSafeExtent = 1.0f;
+                if (shadowIndex >= 0)
+                {
+                    // We subtract a bit from the safe extent depending on shadow resolution
+                    float shadowRes = additionalLightData.shadowResolution.Value(m_ShadowInitParameters.shadowResolutionPunctual);
+                    shadowRes = Mathf.Clamp(shadowRes, 128.0f, 2048.0f); // Clamp in a somewhat plausible range.
```
sebastienlagarde: Plausible range would be at least 4K

FrancescoC-unity: The factor is used just as a scale for the safe extent, putting the max value to 4k will result in a lerp below going a bit too slow.  Conversely, IIRC, a case with 4K shadows work fine.

PR #4672, opened 2019-09-19
Improved color buffer history allocations

JulienIgnace-Unity:
### Purpose of this PR
- Color buffer history is no longer allocated if not needed (ie: no SSR/TAA/refract/disto)
- Preview cameras disable refraction now (allows to not allocate any color buffer pyramid for them)

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -3570,6 +3573,8 @@ RenderSSRParameters PrepareSSRParameters(HDCamera hdCamera)
 
         void RenderSSR(HDCamera hdCamera, CommandBuffer cmd, ScriptableRenderContext renderContext)
         {
+            return;
```
kubacupisz: Uhm...? ;)

JulienIgnace-Unity: Yeah ahem, nothing to see here!
(fixed in HDRP/staging :p )

PR #4675, opened 2019-09-19
Added test and reference images for working ShaderGraph in XR

soraryu:
### Purpose of this PR

XR rendering was broken in several past releases and is currently still broken on some stereo rendering modes in releases/7.1.x. (namely, Multi Pass and Single Pass Instanced).

### Release Notes

Added a new test that has a PBR shadergraph shader and a Unlit shadergraph shader. Also, added reference image.

### Testing status

- [X] Built a player
- Other: 
Note that the new test succeeds on Single Pass but fails on Multi Pass and Single Pass Instanced, which is kind of the point.

---
### Overall Product Risks

None

Slightly longer test time, I guess, but prevents breaking ShaderGraph in XR again.


soraryu:
I did this pull request on release/7.1.x because that's the place where I think it's needed most right now; please let me know if I should rebase this somewhere else.

alindmanUnity:
lgtm but do we need any mac coverage in addition to windows?

soraryu:
OK, I rebased on master.

@alindmanUnity I can't provide Mac XR reference images. Also it seems that none of the XR tests currently have any reference images for Mac.

sophiaaar:
thinking about this again, perhaps it would be better to put xr tests into a separate stereo project (like what we have for Universal) cc @zadi0648 @DennisDeRykeUnity

soraryu:
This pull request is actually adding tests to the separate stereo project for Universal, do you mean yet another one (a new one)?

sophiaaar:
Yes, they should be separate

soraryu:
I think that discussion shouldn't be part of the scope of this pull request - it can be split out later if deemed useful.

(Besides that, I think this would lead to an explosion on the amount of test projects if package combination tests would each sit in totally separate projects).

sophiaaar:
> I think that discussion shouldn’t be part of the scope of this pull request - it can be split out later if deemed useful.

I appreciate your contribution, and I’m glad you want to help improve our test coverage, but we want things to be done correctly the first time, rather than have to go back and correct them later.

> (Besides that, I think this would lead to an explosion on the amount of test projects if package combination tests would each sit in totally separate projects).

Creating more test projects with smaller scopes is the test strategy that we in the Graphics team have decided on going forward. Our internal testing infrastructure lends itself to testing multiple projects in parallel, and we want to make use of that. It’s more scalable, and it’s easier to see right off the bat which area has failed and allows for faster turnaround.

If you want to change your PR, that's great, and we'll merge it. If you don't want to make this change, that's okay too, but it means we won't be able to land this right away and it will take us a few weeks .

soraryu:
Oh, I'm totally fine to change it if you specify what to. I'm just not sure I can even create a new test project that's properly integrated with the internal CI (can I? - is that yamato?).

So, what are the actual changes you are requesting?

sophiaaar:
If you create the project, then I can hook it up to the CI.

I'd name the project ShaderGraphTestsStereo. The PostFX XR tests are fine to live in the URP Stereo project, as postfx is a feature of URP.

soraryu:
Do you think it makes sense to test ShaderGraph in a single project with both URP and HDRP (that's what the name suggests)?
Right now there is
UniversalGraphicsTest
UniversalGfxTestStereo

So
ShaderGraphTestStereo (with the question being whether that is for HDRP + URP then?)
or
UniversalShaderGraphTestStereo?

soraryu:
@sophiaaar any notes on this?
(I took a look at how HDRP handles that, and there ShaderGraph tests are directly embedded in the main testing project - so that is kinda "legacy"?)

sophiaaar:
@soraryu Perhaps go for ShaderGraphUniversalStereo?

soraryu:
@sophiaaar new project added + removed all the changes from the existing one.

sophiaaar:
Great, thanks so much!

marctem:
This should be backported if possible but can wait for 7.3.

PR #4679, opened 2019-09-19
Testing XR PostFX working on URP

soraryu:
### Purpose of this PR
PostFX are broken on current 7.1.x and master. This PR adds tests that show that and prevent it from happening again.

### Release Notes
Adds four different XR tests.
- toggling "Use Postprocessing" on on a camera
- using FXAA
- using SMAA
- adding a Volume to the scene with Bloom / Depth of Field that don't actually do something

### Testing status

- [X] Built a player
- Other: 

Reference image for all tests:
013_xr_PostFX_Simple
_Note: the tests for FXAA and SMAA should look different but since they totally fail currently I included this as baseline result. These reference images should be updated once the basics work and have a really small tolerance to detect failing FXAA/MSAA._

Actual current results:

**013**
_Note that 013 currently succeeds despite the actual & result images looking very different. This is a bug in the Graphics Test Framework. Workaround is reducing the tolerance on the camera. Additionally, the result during running the test differs from the result seen when pressing Play._
013_xr_PostFX_Simple

**014**
014_xr_PostFX_Complex_NoAA

**015**
015_xr_PostFX_FXAA

**016**
_Note that 016 doesn't fail because of image content but because of the depth buffer not being the same size as the color buffer._
016_xr_PostFX_SMAA

---
### Overall Product Risks
None

Low, increases test run time

### Comments to reviewers
Note that the numbering starts with 013 because of my previous pull request, #4675.
Also please note that you should run these tests with all of the available Stereo modes (Multi Pass, Single Pass, Single Pass Instanced) since they all fail in different ways.


sebastienlagarde:
Please target master branch, we can help for backport for release/2019.3 and then release/7.1.x

soraryu:
OK, rebased on master.

@seblagarde please note that there is something _really_ weird going on with 013. Pressing "Play" in the Editor shows that it doesn't work (same double-image issues as the others), but the test succeeds and it seems the test-generated image is actually correct. I think this might be a bug in the Graphics Test Framework, but I'm not sure.

phi-lira:
Adding @DennisDeRykeUnity as owner of XR Tests.

PR #4685, opened 2019-09-20
Fix for SSR reflections being clamped out with emissive values

FrancescoC-unity:
This a first part of the fix for https://fogbugz.unity3d.com/f/cases/1138739/ 

Also fixes an issue in GetInversePreviousExposureMultiplier

Second part will be for probes, but it is effectively a different problem.


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/ScreenSpaceReflections.compute
```diff
@@ -349,10 +349,6 @@ void ScreenSpaceReflectionsReprojection(uint3 dispatchThreadId : SV_DispatchThre
     float  opacity  = EdgeOfScreenFade(prevFrameNDC, _SsrEdgeFadeRcpLength)
                     * PerceptualRoughnessFade(perceptualRoughness, _SsrRoughnessFadeRcpLength, _SsrRoughnessFadeEndTimesRcpLength);
 
-    // Inverse pre-exposure
```
FrancescoC-unity: Inverting the exposure here causes saturation issues. Indeed moving it makes it slightly more expensive, but is broken otherwise.

FrancescoC-unity:
Discussing a sec this with Anis, the SSR texture is also used by ray traced reflections, I need to add a uniform branch to skip the undoing of the prev exposure for that case.

sebastienlagarde:
Please add changelog

FrancescoC-unity:
> Please add changelog

Done!

PR #4688, opened 2019-09-20
Removing ray tracing environement and consequent modifications

anisunity:
Removing the Ray Tracing Environement Component
Adding a Layer Mask to the ray Tracing volume components (Ao, Reflection, GI, Recursive tracing and path tracing) and using those layer mask to define which objects are included in the process
Making the ray count function public and writing documentation for it and the enum
Changing the return type of the ray count function from float to uint
Changing the ray tracing manager from class to HDRP partial class
Adding couple profiling id to make profiling easier
Creating a generic ray tracing setting volume component that contains global ray tracing parameters (ray bias for the moment)
Rebuilding completely the acceleration strucre every frame
Fetching all the renderers and lights every frame to build the RAS and Light Cluster
Re-enabling the light cluster (was commented by mistake)
Correcting a serialization bug for the light cluster


anisunity:
@eturquin

PR #4691, opened 2019-09-20
Add explicit exposure for probe rendering 

FrancescoC-unity:
Second part of the fix https://fogbugz.unity3d.com/f/cases/1138739/ even though not directly. 

This adds an explicit exposure _per probe_ that can be used when rendering probe content, this allow to have emissive materials look like this 

image

instead of this 

image


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1695,6 +1695,7 @@ internal bool GetEnvLightData(CommandBuffer cmd, HDCamera hdCamera, HDProbe prob
             envLightData.influenceShapeType = influence.envShape;
             envLightData.weight = weight;
             envLightData.multiplier = probe.multiplier * m_indirectLightingController.indirectSpecularIntensity.value;
+            envLightData.invRangeCompressionFactor = Mathf.Max(probe.rangeCompressionFactor, 1e-6f);
```
sebastienlagarde: I don't see inversion here but the property is name invert? "invRangeCompressionFactor "

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -392,7 +394,7 @@ float GetInverseCurrentExposureMultiplier()
 
 float GetInversePreviousExposureMultiplier()
 {
-    float exposure = GetCurrentExposureMultiplier();
+    float exposure = GetPreviousExposureMultiplier();
```
sebastienlagarde: Arf, good find about the typo. Note for @JulienIgnace-Unity I see that GetInversePreviousExposureMultiplier was use in the volumetric exposure PR that you have done... So it wasn't working correctly...

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -369,18 +371,18 @@ void ApplyCameraRelativeXR(inout float3 positionWS)
...
-    return 1.0;
+    return _ProbeExposureScale;
 #endif
 }
 
 float GetPreviousExposureMultiplier()
 {
 #if SHADEROPTIONS_PRE_EXPOSITION
-    return LOAD_TEXTURE2D(_PrevExposureTexture, int2(0, 0)).x;
+    return LOAD_TEXTURE2D(_PrevExposureTexture, int2(0, 0)).x * _ProbeExposureScale;
```
sebastienlagarde: I am wondering if we need this here, I guess we shouldn't. This is use only in SSR

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -369,18 +371,18 @@ void ApplyCameraRelativeXR(inout float3 positionWS)
 float GetCurrentExposureMultiplier()
 {
 #if SHADEROPTIONS_PRE_EXPOSITION
-    return LOAD_TEXTURE2D(_ExposureTexture, int2(0, 0)).x;
+    return LOAD_TEXTURE2D(_ExposureTexture, int2(0, 0)).x * _ProbeExposureScale;
```
sebastienlagarde: Let's add a comment here about what is _ProbeExposureScale and why we apply it

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
FrancescoC-unity: While we might not need it now, I don't like the idea of creating discrepancies. This come for free anyway (scalar mul)

PR #4693, opened 2019-09-20
Use real types for compute shadow sampling

jessebarker:
### Purpose of this PR
Allow shadow sampling using "real" typed data.  Previous attempts either left shadow sampling incompatible (mixing half and float interfaces in the plumbing between core and HDRP) - this is the current state of master, or broken for mobile platforms.

---
### Testing status

- [ x] Built a player
- [ x] C# and shader warnings (supress shader cache to see them)
- Other: 

---
### Overall Product Risks
Low
Low - Desktop should be unaffected, and Universal does not use the compute sampling routines.
---
### Comments to reviewers


sebastienlagarde:
I am fine with the change, but Kaychang is right. Regarding UV with fp16 you can only sample 1kx1K texture correctly, if you have higher resolution( like our shadow atlas), you should have precision issue.

Thought? Maybe you could make some test with a 4kx4k atlas, for the cascade shadow map just to check impact of the change at the edge of the shadow?

sebastienlagarde:
this could also be a valid constraint, on mobile we max the size at 1K

jessebarker:
I think I'm confused.  Are you worried about the theoretical limit (minimum distinguishable value) of fp16 (1/2048), or GPUs that don't support denormalized numbers (and would have an enormous zero hole in fp16)?  I'm not suggesting we push our luck, but 2048x2048 (and not a pixel more :-)) should be doable on a good GPU (assuming sane RNE and denorm behavior).

jessebarker:
(I'm not opposed to a 1Kx1K limit if it will keep reasonable quality)

kaychang-unity:
I agree that in theory, there shouldn't be an issue with sampling individual texels in a 2K texture with uv coordinates at half precision.
Empirically, I did saw sampling artefacts though, but I never dug deeper on the exact root cause (this could be some issue with the interpolation of the uv maybe, or other related math at half precision?).

sebastienlagarde:
ok let's merge and we will see with time :)

PR #4694, opened 2019-09-20
Fp16 fix error calculation

victor-unity:
### Purpose of this PR
On iOS and MacOs, when using fp16, the specular lighting values can go wrong because of precision issue. Added a saturate to a value that should be between 0 and 1 anyway, but if it goes beyond one, then some results can be wrong in the rest of the calculation.
---
---
### Testing status

- [x] Built a player
- Other: 

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/CHANGELOG.md
+++ b/com.unity.render-pipelines.core/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Restored usage of ENABLE_VR to fix compilation errors on some platforms.
 - Only call SetDirty on an object when actually modifying it in SRP updater utility 
+- Added a saturate to avoid fp16 calculation go above 1
 
 ## [7.1.1] - 2019-09-05
 
--- a/com.unity.render-pipelines.core/ShaderLibrary/BSDF.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/BSDF.hlsl
@@ -231,7 +231,7 @@ real V_SmithJointGGX(real NdotL, real NdotV, real roughness)
 // Inline D_GGX() * V_SmithJointGGX() together for better code generation.
 real DV_SmithJointGGX(real NdotH, real NdotL, real NdotV, real roughness, real partLambdaV)
 {
-    real a2 = Sq(roughness);
+    real a2 = saturate(Sq(roughness));
     real s = (NdotH * a2 - NdotH) * NdotH + 1.0;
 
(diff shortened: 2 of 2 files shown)


kaychang-unity:
FYI I found this old blog post about saturate(x) vs max(0, x): https://nyahoon.com/blog-en/590
I wonder if there is still a perf diff on mobile GPUs?

sebastienlagarde:
would suggets to also add a comment (i will do it), otherwise futher people could remove it

sebastienlagarde:
> FYI I found this old blog post about saturate(x) vs max(0, x): https://nyahoon.com/blog-en/590
> But you want min(1, x), probably saturate() is faster then?

saturate is free, min and max are one instruction.

sebastienlagarde:
>but if it goes beyond one,

 don't understand in which case roughness could be beyond 1. With current set of shader it shouldn't be possible, or people do bad thing in shader graph and input wrong value for roughness? 
It is a bit like NaN, yes we can provide NaN killer but ultimately it doesn't solve the issue of the wrong input from material that cause it and that certainly don't produce the result you want. Here same thing, I have the feeling that we hide the problem to the users instead of helping him to find that he have a problem with a Material.

Thought?

victor-unity:
Holà, 
I have tried to narrow down the cause of the bug, it is weird because if I replace a2 by 1.0001 for instance I do not see any problem. I have tried to write a specific shader with only this function, but I could not reproduce the bug. Moving the saturate later in the code works too : 
`real2 D = real2(saturate(a2), (s * s)); `
It seems that Filament has a similar issue (https://google.github.io/filament/Filament.md.html cf D_GGX function) ; so basically there is a lack of precision around 1 in fp16, and doing a computation like  :
 `1 - a_number_close_to_one `
 can lead to zero instead of a small number. 
But in our case, even if a2==0 I do not see the bug.
I do not understand how roughness can be higher than 1, I will  re-check the unpack functions. 
and continue to investigate...

sebastienlagarde:
>Or as KayChang suggested we modify codegen in order to store real as float.
>What do you think ?

What does this imply? I mean what will be the generated code in this case?

>For iOS, one solution would be to use only floats in the hlsl LightData structs. That solution fixes the >bug we saw here.
That's of course work, Kaychang changed it to make it faster for switch, so it mean we just revert the optimization. So Let's see what's the above solution imply, otherwise I fine with changing to float.

sebastienlagarde:
I am closing this PR for now, as it require a bit more work. Can re-open the PR or do another one in another form later (just to clean pull request list)

PR #4696, opened 2019-09-23
Fixed an issue where SSR would use an incorrect roughness value on Fo…

slunity:
### Purpose of this PR
Fixed an issue where SSR would use an incorrect roughness value on ForwardOnly (StackLit, AxF, Fabric, etc.) materials when the pipeline is configured to also allow deferred Lit.

The gbuffer2 alpha channel (carrying material features and coat mask) was incorrectly initialized to 1 and since ForwardOnly pixels never touched that coat mask, they took the "deferred coat mask" path in the SSR compute shader.

---
### Release Notes

SSR was reported to "disappear or not work" for ForwardOnly materials (eg StackLit, but also AxF, Fabric, etc.) when the pipeline is configured (via camera's frameSettings' LitShaderMode) in Forward only mode.

The opposite actually happened:

-When deferred Lit was allowed to render, it incorrectly allowed SSR to appear on objects where it shouldn't, eg on objects with a perceptual roughness value higher than what the volume settings "min smoothness" value for SSR (which is 0.9 by default) allowed.
Moreover, roughness was fixed and not taken into consideration on these materials, giving SSR an incorrect appearance at various roughnesses.
-When all Lit materials are forced in forward, SSR always works correctly.

---
### Testing status
Not run.
- [ ] Built a player
- Other: 

---
### Overall Product Risks
Low
Low
---
### Comments to reviewers


sebastienlagarde:
Good find! THanks!

PR #4708, opened 2019-09-24
Add local config package for dxr test

FrancescoC-unity:
Make sure the local dxr test works


LocalTestProjects/HDRP_DXR_Tests/Packages/manifest.json
```diff
@@ -1,12 +1,27 @@
 {
   "dependencies": {
+    "com.unity.2d.sprite": "1.0.0",
+    "com.unity.2d.tilemap": "1.0.0",
+    "com.unity.ads": "2.0.8",
+    "com.unity.analytics": "3.3.2",
```
sebastienlagarde: manifest.json must only contain restricted set of package (like before). You should only update config path and test framework version.

FrancescoC-unity: While I agree and will make the change, this also means that this will appear in every changeset anytime the project is open locally, since the packages are added automatically. 
If it isn't untracked, I guess we should make it so then post the config update

sebastienlagarde: There is a misunderstanding here :).
This is how our test framework work currently. We expect this manifest to be updated when we open the project locally and when trigger on yamato, it allow to always use correct package. i.e we want to automatically add all others package. We have get many problem in the past with other test projects by forcing a given version of package (i.e fail after trunk PR update etc...).

FrancescoC-unity: Ha yes indeed, first I didn't realise that it was that the reason (yamato) and most importantly, I thought these tests were meant for local execution and not online one, but I guess the plan is to run these too on yamato agents?

PR #4709, opened 2019-09-24
Refactoring the code and removing UnlitraytracingData litraytracingdata and making the layeredlit work

anisunity:
All in the title @eturquin FYI


com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLitData.hlsl
```diff
@@ -628,10 +628,10 @@ float3 ComputeMainBaseColorInfluence(float influenceMask, float3 baseColor0, flo
 
     // We want to calculate the mean color of the texture. For this we will sample a low mipmap
     float textureBias = 15.0; // Use maximum bias
-    float3 baseMeanColor0 = SAMPLE_UVMAPPING_TEXTURE2D_BIAS(_BaseColorMap0, sampler_BaseColorMap0, layerTexCoord.base0, textureBias).rgb *_BaseColor0.rgb;
-    float3 baseMeanColor1 = SAMPLE_UVMAPPING_TEXTURE2D_BIAS(_BaseColorMap1, sampler_BaseColorMap0, layerTexCoord.base1, textureBias).rgb *_BaseColor1.rgb;
-    float3 baseMeanColor2 = SAMPLE_UVMAPPING_TEXTURE2D_BIAS(_BaseColorMap2, sampler_BaseColorMap0, layerTexCoord.base2, textureBias).rgb *_BaseColor2.rgb;
-    float3 baseMeanColor3 = SAMPLE_UVMAPPING_TEXTURE2D_BIAS(_BaseColorMap3, sampler_BaseColorMap0, layerTexCoord.base3, textureBias).rgb *_BaseColor3.rgb;
+    float3 baseMeanColor0 = SAMPLE_UVMAPPING_TEXTURE2D_LOD(_BaseColorMap0, sampler_BaseColorMap0, layerTexCoord.base0, textureBias).rgb *_BaseColor0.rgb;
```
JulienIgnace-Unity: Seems like it's been mistakenly changed to _LOD

anisunity: The line before sets the bias to 15.0 which is equivalent to fetching lod 15 (the latest one). The problem being bias does not exist in ray tracing, this allows us to share the code without having a branch

com.unity.render-pipelines.core/ShaderLibrary/Sampling/SampleUVMapping.hlsl
```diff
@@ -42,14 +42,14 @@ struct UVMapping
 #undef SAMPLE_TEXTURE_FUNC
 
 // Lod sampling functions
-#define ADD_FUNC_SUFFIX(Name) Name##Lod
+#define ADD_FUNC_SUFFIX(Name) MERGE_NAME(Name, Lod)
```
sebastienlagarde: I see a lot of change on MERGE_NAME macro (in particular in Sampling .hslsl. 
I guess it was like this because otherwise it wasn't compiling on PS4. So we will need to check before accepting the change. Note that in 2020.1 the preprocessor have change, so maybe those change workon 2020.1 but not on 19.3

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLitData.hlsl
sebastienlagarde: Make sense. Initially the bias was control by the artists, but then we have change it to have less parameters.

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitData.hlsl
```diff
@@ -200,6 +215,20 @@ void GetSurfaceAndBuiltinData(FragInputs input, float3 V, inout PositionInputs p
...
     float alpha = GetSurfaceData(input, layerTexCoord, surfaceData, normalTS, bentNormalTS);
+    
+#ifdef _ALPHATEST_ON
+    #if defined(RAYTRACING_SURFACE_SHADER)
+        #if SHADERPASS == SHADERPASS_RAYTRACING_VISIBILITY 
+        if (_UseShadowThreshold && alpha < _AlphaCutoffShadow)
+            return false;
+        else
+        #else
+        if(alpha < _AlphaCutoff)
```
sebastienlagarde: We should invert the order
#if defined(RAYTRACING_SURFACE_SHADER)
    #ifdef _ALPHATEST_ON

sebastienlagarde:
Alternate proposition here (draft): https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5404

PR #4710, opened 2019-09-24
Olento this is a message for you. 

phi-lira:
Olento cares and we also care. Meet me at Unite in the lounge tomorrow the 25th at 16:15.


soraryu:
Can I join? :)

0lento:
Will do :D

phi-lira:
This event quest is available for anyone that can find this message, by any means. I'll be in the lounge at 16:15, carrying Unity on my back and a "Dev" piece of armour.

Chman:
I might drop by as well...

PR #4715, opened 2019-09-26
Fix linear depth computation in the orthographic mode

EvgeniiG:
### Purpose of this PR
Fix linear depth computation in the orthographic mode
Fix https://fogbugz.unity3d.com/f/cases/edit/1167427/

---
### Release Notes
Fix linear depth computation in the orthographic mode

---
### Testing status
**Katana Tests**: No

**Manual Tests**: Yes

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassDecal.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassDecal.hlsl
@@ -53,8 +53,10 @@ void Frag(  PackedVaryingsToPS packedInput,
         clip(-1);
 #else // Decal mesh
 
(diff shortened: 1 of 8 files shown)


FrancescoC-unity:
I was looking at a bug with the same root cause, made me realise we get linear depth in other places too that will break in ortho, I think we need to discuss whether we want to sprinkle branches in lots of places or just ignore orthographic issues

EvgeniiG:
We obviously cannot ignore the bugs.
My suggestions are:
1. Don't support orthographic in the HDRP - most likely not an option.
2. Emulate orthographic with a narrow FoV perspective under the hood - worse precision and will not look exactly the same, but is at least fast.
3. Add an ortho variant to all the relevant shaders.

FrancescoC-unity:
Of course, by ignore I meant essentially deprecating ortho views. I think yes, the options you put forward are the only one we have for now and we need a wider discussion to chose the bst option. 

I am personally inclined toward either 1 or 2. We cannot afford extra variants at the moment.

PR #4699, opened 2019-09-23
Added multi selection support for scalable settings in lights

fredericv-unity3d:
### Purpose of this PR
Fixes [1182968](https://fogbugz.unity3d.com/f/cases/1182968/)

Refactored ScalableSettings to support different number of level per setting.
Added support for multi selection

---
### Release Notes
Fixes [1182968](https://fogbugz.unity3d.com/f/cases/1182968/)

---
### Testing status
**Katana Tests**: -

**Automated Tests**: -

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---

PR #4718, opened 2019-09-26
Ray Traced Contact Shadows

anisunity:
Implementing ray traced contact shadows
Changing the lightdata.screenspaceshadow content to store the ray traced contact shadow flag

Has potentially a dependency on https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4699


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
```diff
@@ -806,6 +806,16 @@ static void DrawContactShadowsContent(SerializedHDLight serialized, Editor owner
                     hdrp
                 )
             );
+
+#if ENABLE_RAYTRACING
+            // TODO: At the moment, it is not possible to request this information, so we always display the option.
+            //if (serialized.serializedLightData.contactShadows.GetValue())
```
sebastienlagarde: expected?

anisunity: The comment explains it. It requires an API from this PR #4699

com.unity.render-pipelines.high-definition/Editor/Lighting/SerializedHDLight.cs
```diff
@@ -76,6 +76,9 @@ public sealed class SerializedLightData
             public SerializedProperty volumetricShadowDimmer;
             public SerializedProperty shadowFadeDistance;
             public SerializedScalableSettingValue contactShadows;
+#if ENABLE_RAYTRACING
+            public SerializedProperty rayTracedContactShadow;
```
sebastienlagarde: iirc we shouldn't have serialized data that depends on preprocessor. @JulienIgnace-Unity any comment? (a bit like when we have editor only szeiralized value. Not a problem to have it in all build. thought? or maybe  my memroy is ffuzzy. It is related to the fact you can make a build with raytrace or not depends on config asset and then we can be in ill define state.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1283,6 +1283,25 @@ public float shadowFadeDistance
         BoolScalableSettingValue m_UseContactShadow = new BoolScalableSettingValue { useOverride = true };
         public BoolScalableSettingValue useContactShadow => m_UseContactShadow;
 
+#if ENABLE_RAYTRACING
+        [SerializeField]
+        bool m_RayTracedContactShadow = false;
```
sebastienlagarde: same remark than before. Are we doing this for other values? does it work?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -1353,7 +1354,7 @@ static Vector3 EvaluateAtmosphericAttenuation(Vector3 L, Vector3 positionWS)
 
             lightData.cookieIndex = -1;
             lightData.shadowIndex = -1;
-            lightData.screenSpaceShadowIndex = -1;
+            lightData.screenSpaceShadowIndex = 0xffff;
```
sebastienlagarde: better to use a define here than 0xFFFF, we will certainly make mistake in future :)

com.unity.render-pipelines.high-definition/Editor/Lighting/SerializedHDLight.cs
JulienIgnace-Unity: Yes, this will be a problem if you build with raytracing and run a player without for example (or the other way around)

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightUI.cs
sebastienlagarde: PR have been merge, please update

PR #4731, opened 2019-09-30
Fix shader graph samples package

laurenth-unity:
### Purpose of this PR
Change procedural sky to PBR sky in the sample scene
Update shader graphs and materials to new version in order to fix issues seen here :
https://confluence.hq.unity3d.com/display/HDRP/Package+samples

---
### Comments to reviewers
Only touches samples package and changelog.


iM0ve:
Out of the box result:
box

Suggestions applied:
tweaked

Found one issue:
1) Volume is missing an override, hence you can see greenish diffusion on the sides of the balls
image

And a few cosmetic suggestions(optional):
2) The puddle looks a bit sad compared to materials balls. In the second image, I added a planar reflection, reduced the planes smoothness for better contrast and gave the puddle a greenish tint.
3) Changed shadow angle and resolution.

laurenth-unity:
Good suggestions, will update the unitypackage file

sebastienlagarde:
@iM0ve can you confirm all is good for you? thanks

PR #4739, opened 2019-09-30
Shadowmaps automation test

JMargevics:
### Purpose of this PR
Made a new PR instead of fixing numerous rebase conflicts on the previous branch. This PR adds an automated test scene for testing Shadow Map parameters (Normal Bias, Dimmer, Tint, Constant Depth Bias, Near Plane). 

[Screenshot](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/HDRP/Tests/ShadowMaps2/TestProjects/HDRP_Tests/Assets/ReferenceImages/Linear/WindowsEditor/Direct3D11/2309_ShadowMaps.png)

Yamato job link: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252FTests%252FShadowMaps2/.yamato%252Fupm-ci-hdrp.yml%2523HDRP_Windows64_playmode_DX11_trunk/518075/job


sebastienlagarde:
@JMargevics Yamato test fail, have you forget to update the screenshots?

TestProjects/HDRP_Tests/ProjectSettings/QualitySettings.asset
```diff
@@ -40,7 +40,8 @@ QualitySettings:
     asyncUploadBufferSize: 4
     asyncUploadPersistentBuffer: 1
     resolutionScalingFixedDPIFactor: 1
-    customRenderPipeline: {fileID: 0}
+    customRenderPipeline: {fileID: 11400000, guid: d7fe5f39d2c099a4ea1f1f610af309d7,
```
sebastienlagarde: arf, this shouldn't have been there

iM0ve:
1) Formatting issue
issue1

2) Grammar issue, also maybe the sentence could be simplified.
issue2

3) Formatting issue, suppose to be a link. One word is in [brackets], intended?
formatting3
formatting3_

4) Optional. For begginers maybe worth mentioning which scene they should open? They might open the sample scene(not viking) and be left staring at this screne after wizard finished.
optional5

5) If we now use HDRP wizard and not manual setup, this part is redundant
issue6

6) These steps should not be necessary, should be fixed by editing files user download. 
However, if we don't update the package, then we need to put this note after:
**Press the Play button and you will see the following in your Game window:**
Because this is where the issue will occur.
issue7

7) We cant have 2 directional lights casting shadows. Because of this, there will be no shadows or incorrect shadows.  Need to at least mention what's wrong with the shadows.
issue8

8) The image in the docs won't match. It will look like the image above(way brighter than docs version). This is caused by having 2 directional lights. The solution would be to either
a) Tell users to remove Fill_Directional light
b) Update the image in the docs
doc image for reference:
image

9) There is a new workflow of how to bake lights now. I think it should be mentioned here or a link to a detailed description, this part is a bit confusing but important for newcomers. That requires setting up this part of Lighting tab:
issue9_1

10) Optional: The order of these bits should be swapped. If completed as is, any light intensity changes will not be represented. Light cookies will be causing this. No fun of fiddling with lights if they don't do anything.
optional10

11) I'm not 100% certain about this. Last screenshot show "upgraded scene" after finishing tutorial with **working particles**. Would shuriken particles really just work in HDRP without any tweaks? By the end of the tutorial they did not work for me. 
endResult

We should either extend the tutorial explaining how to upgrade VFX. Or tell that it's complicated and link the HDRP particle sample for further investigation on their own.
endResult_

12) In upgrade HDRP template there is a step:
Create a Scene Settings GameObject (GameObject > Rendering > Scene Settings) and set the sky Exposure to 0 and the Multiplier to 20000.

This used to be true in unity 2019.1 but I don't see this menu anymore, are we planning to bring it back?
issue11

13) Optional. We mention that our HDRP has its own post-processing stack and there is barely anything else mentioned about it. At least a link would be nice, or brief explanation of how its based physical values (why bloom is completely different and etc).
optional12

14) Overall I could not finish upgrading HDRP sample because things were overexposed, Might need another check after the Scene settings step is fixed.

PR #4744, opened 2019-10-01
Removing references to Katana

sophiaaar:
### Purpose of this PR
Removing references to Katana in the Pull Request Template since SRP is no longer testable in Katana

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


pbbastian:
👍  Can we also remove the section on release notes? Those should be in the change log if there are any

sebastienlagarde:
Done remove release section

PR #4749, opened 2019-10-01
Added basic alpha compositing

Chman:
### Purpose of this PR
Added basic alpha compositing when HDRP is running in fp16 mode. Note that this only does a copy from the color buffer alpha to the target alpha and nothing else.

Also updated test 9007 by re-enabling post-processing on it (ref image are the same = it works).

Note: do we want to backport this?

---
### Testing status
See yamato.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low


sebastienlagarde:
I will recreate the PR,  I am unable to target hdrp/staging for whatever reasons.

PR #4751, opened 2019-10-01
Fixes for terrain heightmap default/null values and epsilons

nerites:
### Purpose of this PR
Fix for https://fogbugz.unity3d.com/f/cases/1187428/
For terrain that doesn't use all available layers, empty layers have their masks initialized to maximum height. This causes artifacting in the remaining layers. 

---
### Testing status

- [ ] Built a player
- Other: 
When I ran graphics tests locally, the only tests that failed were ones that also failed on HDRP/staging.

Visually confirmed that scaling height transition causes smoothly changing differences in appearance. 
In URP, for very small values of height transition, it seems like the edges of the splat map aren't smoothed out correctly, but I was unable to find a version of URP where this doesn't happen:
image

In HDRP, the edges of the splatmap are smooth as expected.

None, but maybe we should have a test for height transition?

---
### Overall Product Risks
**Technical Risk**: Low - single line change in HDRP, only initial value change in URP

**Halo Effect**: None

---
### Comments to reviewers
Landon, could you please test this branch?


LandonTownsendUnity:
Is this the bug you're trying to fix?
image
Because it still happens on the fix branch

nerites:
> Is this the bug you're trying to fix?
> image
> Because it still happens on the fix branch

Landon and I talked about this on slack. This issue appears for small values of transition, but is separate from the issue where scaling transition didn't seem to smoothly change the appearance of the painted terrain. Here is a quick comparison of behavior before and after the changes I made.

Before:
before

After:
after

There are still issues with terrain layering on URP, but Landon said he'd file a separate bug for that, since URP didn't have support for this prior so it's not a regression.

nerites:
Updated the HDRP fix per feedback from @parashar-unity - the line was initializing a default height value, similar to the lines I modified in URP. Setting it only to the offset (without the scale) sets it to the minimum possible height instead of the maximum.

sebastienlagarde:
@cdxntchou @zeroyao @russoh I am a bit worried about all the successive change in the weighting of layer of the terrain. In HDRP we have a blending mode that is used for our layered Lit and that is fine for various users but they complain that the weighting of the terrain system don't match.

I understand that the weight of the terrain should match between HD and UP, but I am worried about the successive change in the weighting and method. We should really define what we want with the tech art team, do the method and adds graphic test for it, so it will not change anymore. This should involve artists working on terrain like @SebastienLa .

Note: Maybe I missed some discussion and I am not aware of previous discussion about this topics.

LandonTownsendUnity:
> @cdxntchou @zeroyao @russoh I am a bit worried about all the successive change in the weighting of layer of the terrain. In HDRP we have a blending mode that is used for our layered Lit and that is fine for various users but they complain that the weighting of the terrain system don't match.
> 
> I understand that the weight of the terrain should match between HD and UP, but I am worried about the successive change in the weighting and method. We should really define what we want with the tech art team, do the method and adds graphic test for it, so it will not change anymore. This should involve artists working on terrain like @SebastienLa .
> 
> Note: Maybe I missed some discussion and I am not aware of previous discussion about this topics.
> 
> @cdxntchou can you validate that the fix here is what you expect?

The issue we're trying to fix is the fact that the way height blending works in HDRP has changed from 2019.2 to 2019.3+. The purpose of this PR is to change it back to the way it worked previously in 2019.2, as it appears some mistakes were introduced that changed the way it works (not only creating a drastic disparity for upgraded projects, but also, the way it was changed to work frankly looks bad compared to the old version). Universal RP is something that will need a separate PR for to make it work like HDRP did in 2019.2 (and the way HDRP now looks after this PR). We can consult a tech art team but I feel like restoring parity between the current version and 19.2 is the best option, not only so that people's upgraded projects aren't changed, but also because the way it worked in 19.2 was really well.

For people using the default terrain material this shouldn't change anything, this will only restore the behavior from 2019.2 for people using a custom terrain material with height blending enabled.

sebastienlagarde:
Ok, I have found the previous PR that was making blend change. It is this one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3870 and it was to fix 2 issues, so hope we don't enter into a loop :)

nerites:
Hey folks, Chris went over the changes with me just now and clarified some things about the various masks. I've now pushed new changes to reflect the following:
- When no mask is given for a terrain layer, height is initialized to half of the height range
- Splat control must be used to mask down height values before the max height calculation needed for normalization

URP now behaves almost identically to HDRP, except for a sudden drop in brightness at height transition value 0. 

urpupdate

com.unity.render-pipelines.high-definition/Runtime/Material/TerrainLit/TerrainLit_Splatmap.hlsl
```diff
@@ -88,12 +88,12 @@ void TerrainSplatBlend(float2 controlUV, float2 splatBaseUV, inout TerrainLitSur
     #define SampleNormal(i) float3(0, 0, 0)
 #endif
 
-#define DefaultMask(i) float4(_Metallic##i, _MaskMapRemapOffset##i.y + _MaskMapRemapScale##i.y, _MaskMapRemapOffset##i.z + _MaskMapRemapScale##i.z, albedo[i].a * _Smoothness##i)
+#define DefaultMask(i) float4(_Metallic##i, _MaskMapRemapOffset##i.y + 0.5 * _MaskMapRemapScale##i.y, _MaskMapRemapOffset##i.z + _MaskMapRemapScale##i.z, albedo[i].a * _Smoothness##i)
 
 #ifdef _MASKMAP
     #define MaskModeMasks(i, blendMask) RemapMasks(SAMPLE_TEXTURE2D_GRAD(_Mask##i, sampler_Splat0, splatuv, splatdxuv, splatdyuv), blendMask, _MaskMapRemapOffset##i, _MaskMapRemapScale##i)
 #define SampleMasks(i, blendMask) lerp(DefaultMask(i), MaskModeMasks(i, blendMask), _LayerHasMask##i)
-    #define NullMask(i)               float4(0, 1, _MaskMapRemapOffset##i.z + _MaskMapRemapScale##i.z, 0) // only height matters when weight is zero.
```
nerites: An unused layer is the same as splatControl/weight being 0 everywhere, so scale ought to be zeroed out (only offset affects height).

nerites:
Fixed the weird dropoff in brightness in URP for transition = 0. We needed the epsilon for active layer offset to be smaller than the epsilon for minimum height transition, as in HDRP. 
image

LandonTownsendUnity:
TerrainHeightBlendNoMaskUniversal
TerrainHeightBlendNoMaskHDRP
Top: universal
Bottom: hdrp
What happens when a layer with no mask map is mixed with a layer with a mask map.

In 19.2 there was a "channel default values" dropdown for channels missing a mask map; what happened to that?
image
image

nerites:
@LandonTownsendUnity I'm not sure why setting default values was removed, maybe @parashar-unity or @cdxntchou can comment on that? 

And thanks for catching that; I had applied the fix to only the AO channel and not the heightmap channel. Do you mind checking if it works now?

LandonTownsendUnity:
image
Looks like it defaults to 0.5 now, as expected

nerites:
Thanks! 

Also I talked to Parashar about the UI change. When he was converting the TerrainLayer object for SRP, there were properties already for metallic and smoothness, but not for height or AO. Adding a default height/AO property to the TerrainLayer object would require a PR to the engine code, so I don't think that's worth it. It'd probably be straightforward to just modify the SRP shaders and TerrainLitGUI to accept a default height and AO, but I don't see a major need and anyways that's a bit outside the scope of this PR. Maybe if the artists want it :)

LandonTownsendUnity:
Did this get backported?

PR #4753, opened 2019-10-02
Better behaviour when no normal map OS, but detail map is present

FrancescoC-unity:
Use a default normal, selected in TS and not OS, when we have detail maps and no normal map in object space. 

Fix for case https://fogbugz.unity3d.com/f/cases/1143352

Behaviour before: 

image

Behaviour after:
Unity_5jGEKhMZgj


sebastienlagarde:
Please, add changelog

FrancescoC-unity:
> Please, add changelog

Gah! Keep forgetting that, added

PR #4760, opened 2019-10-03
Hdrp/docs/getting started qa feedback

JordanL8:
### Purpose of this PR
Integrating feedback from the previous update to the getting started section.

---
### Comments to reviewers
@iM0ve to double check.


iM0ve:
The formatting and grammar issues are fixed and this could be merged. But the dock will still be far from usable. I think it needs another step by step evaluation or an overhaul.

The first project upgrade. Expected result:
image

Actual result:
image

Second. Expected result:
image

Actual result:
image (1)

sebastienlagarde:
We will handle the overbright case. Independent from the doc. I will merge the doc and we can update image later.

PR #4761, opened 2019-10-03
Added Wiki content to repo and did a quick edit

JordanL8:
### Purpose of this PR
Moving the SRP Core content from the Wiki into the package.

---

### Testing status
**Manual Tests**: Generated the package documentation and validated all of the links.


sebastienlagarde:
Note: for what is worse, the doc in Core will be far less discoverable that the one in HDRP or UP, which is already not very discoverable...

JordanL8:
I made sure to link to the package docs from every page that was on the Wiki. The few people who were aware of the docs in the Wiki will be able to get to the package. Kerry is working on the discoverability issue so hopefully we see more traffic on the package docs after the next Manual publish. I'll keep an eye on the page analytics and keep you updated.

PR #4768, opened 2019-10-04
Fixed sorting, sizing, naming, inactive objects, undo

chikuba:
Fixed:
Sorting for all custom columns
Undo
Listing of inactive objects (only works in 20.1)
Naming and sizing of columns
Cleaned up the code and added better error and null handling 

Screenshot 2019-10-04 at 16 10 18


sebastienlagarde:
Thanks for this. Please add Changelog too

sebastienlagarde:
>Listing of inactive objects (only works in 20.1)
Please, if there is code that only work in 20.1 can you guard it with  UNITY_2020_1_OR_NEWER

We want to backport the PR to 19.3 as it fix a lot of issue.

Thanks

chikuba:
@sebastienlagarde done!

laurenth-unity:
Works well for me, only note is on sorting of color temperature I think hidden values should be sorted as if they were equal to 0

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -109,19 +107,24 @@ public override LightingExplorerTab[] GetContentTabs()
...
         }
 
         protected virtual UnityEngine.Object[] GetHDLights()
         {
-            var lights = UnityEngine.Object.FindObjectsOfType<Light>();
+            #if UNITY_2020_1_OR_NEWER
+                var lights = Resources.FindObjectsOfTypeAll<Light>();
+            #else
+                var lights = UnityEngine.Object.FindObjectsOfType<Light>();
+            #endif
```
RSlysz: @sebastienlagarde I'm pretty sure we use this somewhere in HDRP. If I remember well it should be in the sceneView management for HDRP. If this have changed we must be sure that this is updated everywhere.

PR #4774, opened 2019-10-04
Fix offset in reflections caused by precision error

NetteVI:
### Purpose of this PR
Fixes https://fogbugz.unity3d.com/f/cases/1181639/
A precision error in the normals used by decals was causing reflections to be offset, due to being stored in unsigned 8 bit format. This fixes that precision error by setting it to (0,0,0) in the particular case when there are no decals (without changing it in other cases).

---
### Testing status

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-1181639-decals/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP/540903/job

---
### Overall Product Risks
**Technical Risk**:  Low

**Halo Effect**: Low


sebastienlagarde:
Hi, Thanks for investigating on this!

I guess this PR is not fixing the root cause of the issue.
We have specific implementation of ApplyDecalToSurfaceData for many of our material (hair, eye, Stacklit, Lit, HDLit, Fabric, Terrain etc...). So if we do such kind of patch we need to apply it everywhere. Better to try to fix the root issue which from your description seams to be the encoding of the normal?

Maybe the root cause is that instead of using
outDBuffer1 = float4(surfaceData.normalWS.xyz * 0.5 + 0.5, surfaceData.normalWS.w);

decode
surfaceData.normalWS.xyz = inDBuffer1.xyz * 2.0 - 1.0;

Maybe we should try:
 float4(surfaceData.normalWS.xyz * 0.5 +128.0 / 255.0, surfaceData.normalWS.w);

surfaceData.normalWS.xyz = inDBuffer1.xyz * 2.0  - (256.0 / 255.0);

could you check ? thanks

NetteVI:
Thanks for the suggestion! Actually, I had talked about this with Paul Melamed, because this bug can also be solved by just changing the line in the decode function (line 81) to 

    surfaceData.normalWS.xyz = inDBuffer1.xyz * 2.0 - (254.0 / 255.0);

To clarify, the issue is that when there is no normal decal, it should be set to (0, 0, 0), but since the normal is encoded in unsigned 8 bit format, 0 is 127 instead of 127.5. But Paul suggested only fixing this issue in the ApplyDecalToSurfaceData function since it's specific to when there is no decal normal, and adding the 0.5 to all normals results in that slight shift. Both solutions seem to work though and pass the local graphics test, so I guess the shift is trivial.

I can go ahead and commit the change to Decal.hlsl's decode function then, instead of in ApplyDecalToSurfaceData, if that slight shift is acceptable.

sebastienlagarde:
yes,  I will say between both solution it is a matter of convention ,what do we define as 0 , 127 or 128 and do we rely on rounding. 

I am ok with the surfaceData.normalWS.xyz = inDBuffer1.xyz * 2.0 - (254.0 / 255.0); version
as it allow to have -0.99607, 0 and 1.0039 (that we can convert to 1 via saturation)
my version give -1.0039, 0, 0.99607 . And we prefer to have 1 for positive value.

so go with it and you should saturate as we can now goes above one
surfaceData.normalWS.xyz = saturate(inDBuffer1.xyz * 2.0 - (254.0 / 255.0));

Thanks

sebastienlagarde:
Thanks for the update, sorry I say wrong thing about the saturation as it will kill the negative range, I have updated the PR, I am launching the test and will merge. Thanks

PR #4775, opened 2019-10-04
Universal/postprocessing fixes

hkr:
### Purpose of this PR

- Avoid using rendertargets with MSAA as temporary rendertarget for postprocessing
- Don't try to use the depthbuffer for the backbuffer as render target attachment, that doesn't work on some platforms
- Make use of load/store actions for SMAA

---
### Testing status

manual testing

**Automated Tests**: 

Universal tests are not passing. As far as I can see the errors are similar too master

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Medium

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -213,6 +213,8 @@ void Render(CommandBuffer cmd, ref RenderingData renderingData)
 
             // Don't use these directly unless you have a good reason to, use GetSource() and
             // GetDestination() instead
+            bool tempTargetUsed = false;
+            bool tempTarget2Used = false;
```
Chman: I feel like all the target management stuff could be made in a smarter way rather than having bools & if-checks here and there. We'll do that later in another PR though.

PR #4781, opened 2019-10-07
Hdrp/staging pt2

eturquin:
Follow-up to initial path tracing branch. Support for point, spot and directional lights has been added, as well as emission on Lit, and Unlit material, and Burley diffuse.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
```diff
@@ -1,117 +1,222 @@
...
 {
     value = 0.0;
     pdf = 0.0;
 
-    if (!lightList.count)
-        return;
-
-    for (uint i = 0; i < lightList.count; i++)
+    // First local lights
+    for (uint i = 0; i < lightList.localCount; i++)
```
sebastienlagarde: are you sorted by weight here? otherwise you can break instead of continue in the below condition

if (lightData.lightType != GPULIGHTTYPE_RECTANGLE)
            continue;

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
```diff
@@ -1,117 +1,222 @@
 // How many lights (at most) do we support at one given shading point
-#define MAX_LIGHT_COUNT 4
+// FIXME: hardcoded limits are evil, this LightList should instead be put together in C#
+#define MAX_LOCAL_LIGHT_COUNT 4
```
sebastienlagarde: increase to 20 at least. 4 is really low for a scene. (even if we change it later)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
```diff
@@ -1,117 +1,222 @@
...
+
+bool PickDistantLights(LightList list, inout float sample)
+{
+    return !PickLocalLights(list, sample);
+}
+
+float3 GetPunctualEmission(LightData lightData, float3 outgoingDir, float dist)
+{
+    float4 distances = float4(dist, Sq(dist), 1.0 / dist, -dist * dot(outgoingDir, lightData.forward));
+    return lightData.color * PunctualLightAttenuation(distances, lightData.rangeAttenuationScale, lightData.rangeAttenuationBias, lightData.angleScale, lightData.angleOffset);
```
sebastienlagarde: As a next step, you will need to support cookie. Not in this PR

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingMaterial.hlsl
```diff
@@ -1,87 +1,162 @@
...
+    return SampleBurley(mtl, inputSample, outgoingDir, value, pdf);
+#endif
+}
+
+void EvaluateDiffuse(Material mtl,
+                     float3 outgoingDir,
+                 out float3 value,
+                 out float pdf)
+{
+#ifdef USE_DIFFUSE_LAMBERT_BRDF
```
sebastienlagarde: it is a mistake to define this here (the USE_DIFFUSE_LAMBERT_BRDF), it must be per material. I will explain you.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
eturquin: Yep.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
eturquin: I don't see why, you still have to try all other lights that could be rectangles, and evaluate to something non-null.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/Shaders/PathTracingLight.hlsl
eturquin: Done.

PR #4787, opened 2019-10-07
SSAO Improvements and non temporal version

FrancescoC-unity:
Opening already as it is quite lengthy and missing bits are not big. 

Note that this is not necessarily all I wanted to try out, but it is getting big and substantial enough that it warrants a first PR. 

Working on some improvements (if interested in the changes should open the images separately)

- Some quality improvement to spatial filtering and with self intersection

Before (half res): 
Before

After (same settings)
After

Some _slight_  stability improvements (here shown in full res), there is still some temporal noise though

[These are gifs, probably should open separately if interested as difference might be subtler than visible at first glance with the small preview] 

Before
GTAO_FullRes_Before

After
GTAO_FullRes_After

Some significant improvement on ghosting

[The difference is even more visible with non resized gifs] 

Before
AO_Ghost_before_resized

After
AO_ghost_after_resized

Also added a non-temporal version.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1,4250 +1,4250 @@
-using System.Collections.Generic;
```
FrancescoC-unity: WTF, not sure why this all changed... will try to undo this and add just my few lines.

FrancescoC-unity:
The staging merge went horribly wrong... trying to figure out what's going on.

PR #4788, opened 2019-10-07
Fix URP post processing with XR single-pass methods

fabien-unity:
### Purpose of this PR
- add XR macros and functions to post processing shaders
- fixes for single-pass (instancing, multiview and double-wide)
- enable post-processing on Mobile devices
- effects tested on PC DX11 with Oculus Rift in all modes: bloom, Stop NaN, FXAA, SMAA, Depth-of-Field, Vignette

---
### Testing status

- [x] Built a player

**Yamato**: [running](https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/urp-xr-post-processing-fixes)

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
multi-pass is still not supported with URP
@DennisDeRykeUnity : can you cover testing? Oculus Quest in particular.


DennisDeRykeUnity:
I see that some of the Yamato tests failed, but I see comparable failures for the same pipelines in master three days ago so these failures cannot be attributed to this PR.  Some of the known SRP failures on Yamato are probably related to these two bugs: 
Case 1188407
Case 1116857

Chman:
Adding @hkr as a reviewer to double check the modifications on clear/load actions.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -350,13 +353,21 @@ int GetDestination()
 
         private BuiltinRenderTextureType BlitDstDiscardContent(CommandBuffer cmd, RenderTargetIdentifier rt)
         {
-            cmd.SetRenderTarget(rt, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store);
+            if (m_IsStereo)
+            {
+                // TODO: XR requires new method to set load/store actions AND depth slice to -1
+                cmd.SetRenderTarget(rt, 0, CubemapFace.Unknown, -1);
```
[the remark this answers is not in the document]
fabien-unity: oh! that would be perfect :) I'll try it

daves-unity:
Looks reasonable to me, although Github doesn't give me a way to officially say that by pressing a button.

mikesfbay:
Here's a screenshot of color bleeding between eyes with single-pass double-wide mode. Compare this branch with previous postfx v2:

branch: bloom color bleeding (highlighted area):
v3_bleed2

postfx V2: bloom has no color bleeding between eyes (using XR Interface method):
v2_noBleed

Occlusion mesh helps in some areas but with large blur sizes, color will bleed into other eye in double-wide mode. 
Since we already have a Unity developer page on how to fix such issues, we should fix them for v3. 
https://unity3d.com/how-to/XR-graphics-development-tips#post-processing-stack

"Stereo clamping
If you are doing neighborhood screen-space samples in a shader, you want to make sure you don’t accidentally sample the wrong eye in single-pass double-wide, or outside the valid RenderViewportScale area."

mikesfbay:
@fabien-unity For URP, our QA has to test over 14 combinations of platforms/stereo modes with over 10 effects each or over 140 combinations of postfx testing. It's better to get all stereo macro changes in and do a central testing rather than another round of testing.  

previous QA test matrix (without GO/Quest): https://docs.google.com/spreadsheets/d/1lSuvdZ-t6ePgQBbk3kv9qcsQ45Ilwc3QNKOZdhY_4Zo
@DennisDeRykeUnity @daves-unity

daves-unity:
@fabien-unity @mikesfbay In this case, we need to choose the most expedient solution to getting URP working with post-processing.  While in general, I agree with Mike's approach, let's just make a new issue for the color bleeding, and get these changes in.  We need to improve URP's public image.

DennisDeRykeUnity:
Thanks to Mike for catching the problem with bleeding across the eye boundary.  Mike is also correct that letting problems go until later PRs duplicates overhead work of reviewers including QA.  That said, I'll start manual testing on this today.  

I cannot test this as thoroughly as the matrix Mike posted.  

@fabien-unity Can I test on Go instead of Quest for this, or does it need to be Quest? I see that you already touched several effects with your own testing.  Are there specific effects that I should cover myself?

DennisDeRykeUnity:
On Rift I tested:
-  commit 7406d6d of urp-xr-post-processing-fixes 
- Unity Editor built from recent trunk changeset 19a26876029b
- Universal RP Template
-  Color Adjustments and Film Grain post processing effects
- HDR and MSAA enabled
- Stereo Instancing and Single Pass Stereo (double-wide)
- both Play mode and built player

The two post proc effects work properly and I see no new bugs.

DennisDeRykeUnity:
@fabien-unity Today on Oculus Go I tested: 
- commit 7406d6d of urp-xr-post-processing-fixes
- Unity Editor built from recent trunk changeset 19a26876029b
- Universal RP Template
- Single Pass Stereo

With the Post Proc Volume on or off, the scene is very light (i.e. pale, washed out) compared with the same project using released SRP.  If I disable all of the individual post effects except bloom I can make the tint and intensity very extreme to discern that it is doing something, but even unnaturally bright green bloom in Game view is hard to notice with the HMD so pale.  Please advise.

fabien-unity:
There's an extra conversion to sRGB happening during the final post due to a check on **requiresSrgbBlitToBackbuffer** even though the eye texture is set to sRGB.
@hoesing is aware of the issue but this will require a C++ update

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -209,7 +209,7 @@ half3 MixFog(real3 fragColor, real fogFactor)
 #if defined(UNITY_STEREO_INSTANCING_ENABLED) || defined(UNITY_STEREO_MULTIVIEW_ENABLED)
 
     // Only single-pass stereo instancing uses array indexing
-    #if defined(UNITY_STEREO_INSTANCING_ENABLED)
+    #if defined(UNITY_STEREO_INSTANCING_ENABLED) || defined(UNITY_STEREO_MULTIVIEW_ENABLED)
```
hoesing: I think this whole #if/#else can go away since it's already enclosed in the same conditional on line 209.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -192,13 +192,13 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-            => GetStereoCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, m_Descriptor.graphicsFormat);
+            => GetStereoCompatibleDescriptor(m_Descriptor.width, m_Descriptor.height, m_Descriptor.graphicsFormat, m_Descriptor.depthBufferBits);
 
-        RenderTextureDescriptor GetStereoCompatibleDescriptor(int width, int height, GraphicsFormat format)
+        RenderTextureDescriptor GetStereoCompatibleDescriptor(int width, int height, GraphicsFormat format, int depthBufferBits = 0)
         {
             // Inherit the VR setup from the camera descriptor
             var desc = m_Descriptor;
-            desc.depthBufferBits = 0;
+            desc.depthBufferBits = depthBufferBits;
```
phi-lira: out of curiosity. why is depth buffer required?

com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl
```diff
@@ -30,6 +30,27 @@ Varyings Vert(Attributes input)
     return output;
 }
 
+// ----------------------------------------------------------------------------------
+// Render fullscreen mesh by using a matrix set directly by the pipeline instead of
+// relying on the matrix set by the C++ engine to avoid issues with XR
+
+half4x4 _FullscreenProjMat;
+
+half4 TransformFullscreenMesh(half3 positionOS)
```
phi-lira: declaring this as half here is risky. it should work given this is fullscreen mesh, but if we end up reusing this for other meshes we might fall short on precision.

com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl
```diff
@@ -56,7 +77,13 @@ half GetLuminance(half3 colorLinear)
 
 half3 ApplyVignette(half3 input, float2 uv, float2 center, float intensity, float roundness, float smoothness, half3 color)
 {
+    center = UnityStereoTransformScreenSpaceTex(center);
     float2 dist = abs(uv - center) * intensity;
+
+#if defined(UNITY_SINGLE_PASS_STEREO)
+    dist.x /= unity_StereoScaleOffset[unity_StereoEyeIndex].x;
```
phi-lira: Can Single Pass stereo be used in mobile? Division by float in shader is expensive in mobile. If so, consider for future PRs to store 1/unity_StereoScaleOffset in the constant and do a multiplication here.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
fabien-unity: Right !
Fixed.
We'll need to reconcile this with HDRP at some later point.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
fabien-unity: The default value is 0 here. It's only used by SMAA for stencil as far as I can tell.

com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl
fabien-unity: I switched it to float

com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl
fabien-unity: Indeed, it's only used for mobile actually (double-wide). Noted for later.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where resetting a Freeform 2D Light would throw null reference exceptions. [case 1184536](https://issuetracker.unity3d.com/issues/lwrp-changing-light-type-to-freeform-after-clicking-on-reset-throws-multiple-arguementoutofrangeexception)
 - Fixed an issue where Freeform 2D Lights were not culled correctly when there was a Falloff Offset.
 - Fixed an issue where Tilemap palettes were invisible in the Tile Palette window when the 2D Renderer was in use. [case 1162550](https://issuetracker.unity3d.com/issues/adding-tiles-in-the-tile-palette-makes-the-tiles-invisible)
+- Fixed post processing with XR single-pass rendering modes
```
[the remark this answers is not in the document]
soraryu: These are relevant tests here:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4802
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4803

soraryu:
I tested locally on PC VR (Single-Pass Instanced) and after 2 minutes of testing have this with even a single ShaderGraph shader in the project (bottom is Lit, top is PBR Shader without changes).
image

Left / Right target eye in VR mode are broken in Single-Pass Instanced with Postprocessing on (works when postfx are off).


It's also broken in Single-Pass but looks different there (an area on the right side of the screen stays uncleared).


On Single-pass, Opaque texture seems to work on this PR with the exception of 4x Box Downsampling which is still double-eyed:

Depth texture seems to work.
I can't even test these on Single-Pass Instanced as shadergraphs are fully broken there (see above).

I really hope this PR goes into QA only once there is fixes to both #4802 and #4803 , otherwise @mikesfbay will have to do 2-3 more rounds of that QA matrix until there's even a chance that, as @daves-unity said, "URP's public image improves" in regards to actual usage for VR.

_(@mikesfbay , in case you're wondering I requested access to the testing matrix (f****.h*****@gmail.com) out of curiosity)_

fabien-unity:
@soraryu I pushed a fix for the single-pass double-wide issue with opaque texture.
The ShaderGraph issue about instancing is being worked on and is not related to post fx.
About the left/right manual selection : I'd say this is minor issue as not many users rely on it and we can address it later in priority order.

soraryu:
@fabien-unity that was fast, thanks :) I agree that the Left/Right stuff is minor, hope QA sees it the same.

ShaderGraph might not be a postprocessing issue, but that instancing issue
a) prevents testing of some things in this PR (like Opaque/Depth texture sampling) and 
b) maybe this is just my opinion, but putting this all through the big QA matrix without ShaderGraph support is not gonna reach that goal of better user confidence in URP. These are not isolated components as the past months have shown...

soraryu:
### Testing on Quest:

This is the scene:
image.

First test with postprocessing on crashed the Quest, had to reboot it. Subsequent launches of the app still did the same. Not sure, but might be running out of GPU memory (?). So I started breaking it down.

**Opaque/Depth Texture:**
- opaque and depth texture don't work. Their stereo position is wrong. Also, they get some weird dotted rendering (if you enlarge and look closely you can see the two cubes)
com prefrontalcortex urpfixestest-20191022-235551
- note that this pattern is headlocked (pixels stay the same)
- changing MSAA from 4x to None fixes the dot pattern:
com prefrontalcortex urpfixestest-20191023-003058
BUT keeps the distortion happening. It looks like the right eye is offset in camera space, but only for ShaderGraph objects (the two cubes that have Opaque/Depth texture sampling). The effect is that their geometry is "wobbling around".
image

**Post Processing**
- turning on Post Processing on the camera but doing nothing else results in wrong colorspace, I think this is the same @DennisDeRykeUnity mentioned earlier in this thread. 
image
Note that P3/sRGB (the two Quest colorspaces) change nothing here
Also note - potentially interesting for CI on the Quest - that "Capture Image" from Android LogCat (both Android Studio and the one in Unity) seems to be in a totally wrong colorspace in itself (the Screenshot feature on Quest correctly gives the same colors as Desktop).
- FFR doesn't work (to be fair it doesn't work without PostFX either, but it's important for Quest performance especially with fragment-bound post effects going on)
- all stack effects seem to work, but last effect I tried turning on is Motion Blur, which seems to cause the immediate crashes I was seeing on my first run. It crashes on "Medium" and "Simple" and I didn't bother trying more options. Guess the Quest just runs out of memory from the temporal buffer? Not sure though, as it would usually warn about that in the log, here it just crashes.

sharlenet:
Even though I only made a minor change to CHANGELOG.md, GitHub somehow thinks that I've changed the entire file. I used the web interface from a windows machine, and all the line breaks were changed.

phi-lira:
@sharlenet there's a conflict with master in the changelog and I can't figure how to resolve it since apparently format for changelog file was change in one of your commits.

sharlenet:
> @sharlenet there's a conflict with master in the changelog and I can't figure how to resolve it since apparently format for changelog file was change in one of your commits.

Chris Tchou showed me how to revert the commit and fix the issue! Let me know if you're still having problems.

fabien-unity:
I've disabled the stereo tests 13 and 16 for now because the test framework relies on RenderTexture which is not compatible with post fx at the moment.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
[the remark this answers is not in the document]
phi-lira: @fabien-unity could you look at this in a following PR?

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
```diff
@@ -81,17 +82,18 @@ Shader "Hidden/Universal Render Pipeline/Bloom"
...
+            half3 c0 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv - float2(texelSize * 4.0, 0.0)));
+            half3 c1 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv - float2(texelSize * 3.0, 0.0)));
+            half3 c2 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv - float2(texelSize * 2.0, 0.0)));
+            half3 c3 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv - float2(texelSize * 1.0, 0.0)));
+            half3 c4 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv                               ));
+            half3 c5 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv + float2(texelSize * 1.0, 0.0)));
+            half3 c6 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv + float2(texelSize * 2.0, 0.0)));
+            half3 c7 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv + float2(texelSize * 3.0, 0.0)));
+            half3 c8 = DecodeHDR(SAMPLE_TEXTURE2D_X(_MainTex, sampler_LinearClamp, uv + float2(texelSize * 4.0, 0.0)));
```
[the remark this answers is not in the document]
phi-lira: @fabien-unity could you look at this in a following PR?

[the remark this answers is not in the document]
ROBYER1: Is there any update on a fix for FFR with URP?
Even the latest Oculus Integration notes just mention URP breaks FFR here?! https://developer.oculus.com/downloads/package/unity-integration/

PR #4796, opened 2019-10-08
Fixed material leak + added lifetime feature

Chman:
### Purpose of this PR

All the materials allocated in Universal (as part of Forward & 2D renderers and post-processing) were never release and would leak on every domain reload. This PR fixes it and adds a `Cleanup()` method to `ScriptableRenderer` that users can extend.

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: None


com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -173,6 +173,13 @@ public virtual void FinishRendering(CommandBuffer cmd)
         {
         }
 
+        /// <summary>
+        /// Called when the render pipeline gets destroyed on quit or domain reload.
+        /// </summary>
+        public virtual void Cleanup()
```
phi-lira: Is there a reason we don't we make ScriptableRenderer IDisposable?

Chman: Yes, it would confuses users/devs into thinking that `ScriptableRenderer` supports the classic disposable pattern, which we can't use in Unity when dealing with native resources. So I just named it `Cleanup` (could also be named `Released` I guess, I have no preference).

orangy:
I suspect HDRP has the same issue, will this fix it as well?

Chman:
This PR only affects Universal, I haven't checked HDRP for leaks but I know someone has been doing a pass on it recently, so if it was in fact an issue over there maybe it's been fixed already? But if I remember correctly, HDRP already cleans up everything on domain reload.

orangy:
I have a case opened (1194289) with HDRP obviously leaking memory. Crash dump has failure on alloc, and it happens after editing scripts and Unity Editor reloading scripts. It definitely happens after a while, sometimes sooner, sometimes later. If I just work in Editor without touching scripts, it doesn't crash like this. May be it is not materials, but something else. It started to crash more often when I added Sky – with procedural it crashes very soon, with gradient it is a bit later.

Chman:
@sebastienlagarde ^

sebastienlagarde:
HDRP cleanup everything. I will reach our QA about your fogbugzz. Thanks

PR #4804, opened 2019-10-09
Remove special handling to set alpha of framebuffer to 1.0

hkr:
### Purpose of this PR

Don't force alpha to 1.0 on Android and iOS.
This is already handled at a lower level (on Android the native window is set to opaque when not using "preserve alpha").

This avoids and extra blit in some cases and also reduces complexity in URP.

Still untested on anything XR

---
### Testing status

Manual testing with project that writes alpha to the framebuffer.

**Automated Tests**: no

---
### Overall Product Risks
**Technical Risk**: Medium?

**Halo Effect**: Medium?


daves-unity:
Florian shared his project with me and I'm doing some testing on XR.  Will report back soon.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -24,7 +24,6 @@ namespace UnityEngine.Rendering.Universal
         public PostProcessingData postProcessingData;
         public bool supportsDynamicBatching;
         public PerObjectData perObjectData;
-        public bool killAlphaInFinalBlit;
```
alexey-unity: this changes public api (i guess?) i dont know how we handle this in SRP

hkr: @phi-lira any idea?
And is the variant of the blit shader that sets alpha to 1.0 also considered public API?

phi-lira: We should mark it as obsolete and always initialize/return false.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -345,7 +338,7 @@ static void InitializeCameraData(UniversalRenderPipelineAsset settings, Camera c
             cameraData.defaultOpaqueSortFlags = canSkipFrontToBackSorting ? noFrontToBackOpaqueFlags : commonOpaqueFlags;
             cameraData.captureActions = CameraCaptureBridge.GetCaptureActions(camera);
 
-			bool needsAlphaChannel = camera.targetTexture == null && Graphics.preserveFramebufferAlpha && PlatformNeedsToKillAlpha();
+			bool needsAlphaChannel = camera.targetTexture == null && Graphics.preserveFramebufferAlpha;
```
phi-lira: I think the Graphics.preserveFramebufferAlpha should also be removed in this case. @aleks01010101 do you know?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -252,13 +252,6 @@ static void SetSupportedRenderingFeatures()
 #endif
         }
 
-		static bool PlatformNeedsToKillAlpha()
-		{
-			return Application.platform == RuntimePlatform.IPhonePlayer ||
-                Application.platform == RuntimePlatform.Android ||
-                Application.platform == RuntimePlatform.tvOS;
```
phi-lira: The PR mentions this is handle in low level for Android and iOS. What about tvOS?

phi-lira:
Please also add changelog entry.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
aleks01010101: No, I don't think so. We can use RG11B10 when this is false.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
hkr: I will ask QA to check, but I assume it's same as iOS. @alexey-unity?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
hkr: 👍

hkr:
> This is great. Thanks for working on this. Is this supposed to work on all Android and iOS versions we support?

Yes, should work on all devices we support. Only thing that could hold this up could be QA pass on old weird devices that I don't have. But since Android 4.4 is the oldest version we support now I'm confident that this should work.

hkr:
> The public API cannot be removed. That's the only blocking change I'm requesting. Added more comments.
Done

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
alexey-unity: tvos = ios in this regard

sharlenet:
Even though I only made a minor change to CHANGELOG.md, GitHub somehow thinks that I've changed the entire file. I used the web interface from a windows machine, and all the line breaks were changed.

PR #4821, opened 2019-10-10
Revert VFXType & VFXUpdate

PaulDemeulenaere:
### Checklist for PR maker
Fix of current PR : https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4805

- VFXType change of structure has been reverted
- VFXBasicUpdate restored with a integration mode internally but displays a custom inspector

**Before** 
image

**After**
image

It even provides a multiselection update which wasn't available but possible using a custom inspector : 
_mulit_mode


com.unity.visualeffectgraph/Editor/Types/VFXTypes.cs
```diff
@@ -137,24 +137,30 @@ struct Cone
...
 
-        public static Cone defaultValue = new Cone { baseRadius = 1.0f, topRadius = 0.1f, height = 1.0f };
+        public static Cone defaultValue = new Cone { radius0 = 1.0f, radius1 = 0.1f, height = 1.0f };
     }
 
     [VFXType, Serializable]
     struct ArcCone
     {
-        [Tooltip("Sets the Cone shape input.")]
-        public Cone cone;
```
VladNeykov: Would this be eventually possible to fix/sanitize this? It is wrong for the Cone/Torus ArcShapes to not have their corresponding shape as the first input, my thought was that if we have to introduce a breaking change, best to do it while in preview.

If this can be addressed later, then sure, it's not a pressing change.

PaulDemeulenaere: Yes, of course, it could be adressed later to be consistent with ArcSphere but since it's a breaking change for exposed properties, maybe it has to be done only in a major version.

com.unity.visualeffectgraph/Editor/Models/Contexts/Implementations/VFXBasicUpdate.cs
```diff
@@ -4,9 +4,108 @@
...
 
 namespace UnityEditor.VFX
 {
+    [CustomEditor(typeof(VFXBasicUpdate))]
+    [CanEditMultipleObjects]
+    class VFXBasicUpdateEditor : VFXContextEditor
+    {
+        class UpdateStyles
+        {
+            public static GUIContent header = new GUIContent("Particle Update Options", "Particle Update Options");
```
iTris666: Better use EditorGUIUtility.TrTextContent("text","tooltip"

PaulDemeulenaere: Oh, I didn't know this method was reachable from user code, I'm doing new GUIContent in several, I will do a clean pass on vfx/staging, thanks !

PaulDemeulenaere:
Thanks, I've squashed merging it into `vfx/doc/tooltips`

PR #4823, opened 2019-10-10
Make LOD fade mask scale with the FoV

EvgeniiG:
### Purpose of this PR
Solve https://fogbugz.unity3d.com/f/cases/1146509/

---
### Testing status

**Manual Tests**: What did you do? I tested manually.
**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-lod-fade-fov/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP


sebastienlagarde:
Hi, there is multiple location where LODDitheringTransition is call (all our material).

Need to move ComputeFadeMaskSeed in shadervariables I guess and update all LODDitheringTransition  shadergraph and material call.

EvgeniiG:
Good point, forgot to check the template files.

EvgeniiG:
LODDitheringTransition appears to also used by Universal.

image

It doesn't appear to be used correctly, because you don't want to use z (depth) here, since it can be different for different LODs.

I will also change those instances.

EvgeniiG:
CI fails with 0 playmode tests run.

sebastienlagarde:
I will handle the LOD test

sebastienlagarde:
@RemyUnity I updated the test, however I get:
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/b15c8064-ed2b-4645-89a0-77a3bfd06703%2Flogs%2FTestProjects%2FHDRP_Tests%2Fupm-ci~%2Ftest-results/Assets/ActualImages/Linear/WindowsEditor/Direct3D11/9001_LODTransition.png
on yamato and
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/b15c8064-ed2b-4645-89a0-77a3bfd06703%2Flogs%2FTestProjects%2FHDRP_Tests%2Fupm-ci~%2Ftest-results/Assets/ActualImages/Linear/WindowsEditor/Direct3D11/9001_LODTransition.expected.png

locally.... 

Launching test locally is green, no idea why on yamato the head disappeard... could you look at it? thanks

sebastienlagarde:
>How can I debug on Yamato?

@EvgeniiG Last time we have get this kind of issue it was because of project settings (in standalone build we were having multiple config, we have removed all the config, but here there is no such a thing) or RT Handle (size of the windows wasn't the same on yamato than locally, as locally we don't have the same windows size). Possible that it come from RT handle; but just a guess. Maybe best is to try locally and using the windows size of the test, close the editor, open it and run the test, maybe we will have same condition?

EvgeniiG:
FYI this PR contains a suboptimal fix in terms of perf and quality: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3601
I implemented a better fix.

EvgeniiG:
Another note is that since we use the view direction V which we compute using positionWS, and the floating point math not being exact, sometimes both LODs end up with holes (when using the new rotation-invariant method). This does not happen often, but it's probably worth keeping in mind.

image

EvgeniiG:
If we use the old screen space mask we can solve this problem, but it is arguably less nice.

EvgeniiG:
I could also compute the view vector using the pixel position only, but that's probably going to be too expensive.

EvgeniiG:
The screen space mask is much faster though so perhaps we should just use that...

EvgeniiG:
Can't seem to break it locally, let's ask Yamato again.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -1100,19 +1100,15 @@ float4 GetQuadVertexPosition(uint vertexID, float z = UNITY_NEAR_CLIP_VALUE)
 // LOD0 must use this function with ditherFactor 1..0
 // LOD1 must use this function with ditherFactor -1..0
 // This is what is provided by unity_LODFade
-void LODDitheringTransition(uint3 fadeMaskSeed, float ditherFactor)
+void LODDitheringTransition(uint2 fadeMaskSeed, float ditherFactor)
```
phi-lira: Shouldn't we make an overload and make the uint3 obsolete? This might cause user shaders to fail compilation/thrown truncation warning.

phi-lira:
Added @Verasl as reviewer as this might affect BoatAttack. @Unity-Technologies/gfx-qa-foundation we need to validate speedtree is not regressing.

phi-lira:
@Unity-Technologies/gfx-qa-bellevue Could you validate this doesn't regress SpeedTree shaders in URP?

zadi0648:
How urgent is this PR? We may not get to it today and then we will be busy all of next week.

sebastienlagarde:
ok, we will ask @RemyUnity do to the QA pass then. Thanks Remy

RemyUnity:
I tried using the speedtree shaders in URP with version 7.1.2 ans this branch.
The behavior isn't affected, event if it's not really what I was expecting visually:
LOD fade in URP

EvgeniiG:
Very strange SpeedTree behavior... Then again, I don't think the PR could have caused this issue.

sebastienlagarde:
I have updated the reference screenshots. all good

PR #4844, opened 2019-10-22
Fix renderer features list giving null reference

martint-unity:
### Purpose of this PR
This is fixing this bug
https://fogbugz.unity3d.com/f/cases/1191896/

This also fixes a small bug where after you add 2 renderer features and change the settings on the first one the data was actually not saved. You had to deselect the renderer and re-select it for the first entry to work.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fui-ux-fix-renderers/.yamato%252Fupm-ci-universal.yml%2523All_Universal/613699/job


com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
```diff
@@ -22,10 +22,12 @@ class Styles
             public static readonly GUIContent PassNameField =
                 new GUIContent("Name", "This is the name for the current pass.");
 
-            public static GUIStyle BoldLabelSimple = new GUIStyle(EditorStyles.label);
+            public static GUIStyle BoldLabelSimple;
 
             static Styles()
             {
+                BoldLabelSimple = new GUIStyle(EditorStyles.label);
+//                BoldLabelSimple.stretchWidth = true;
```
phi-lira: remove unnecessary comment.

com.unity.render-pipelines.universal/Editor/ScriptableRendererDataEditor.cs
```diff
@@ -159,12 +93,81 @@ private void OnValidate()
...
             m_PassesList.onReorderCallbackWithDetails += ReorderPass;
-
+            
             m_PassesList.drawHeaderCallback = (Rect testHeaderRect) => {
-                EditorGUI.LabelField(testHeaderRect, Styles.RenderFeatures);
+                GUI.Label(testHeaderRect, Styles.RenderFeatures);
             };
         }
 
+        void DrawElementCallback(Rect rect, int index, bool isActive, bool isFocused)
```
phi-lira: Is there any change here to be reviewed in this function or is it just the same code as above refactored?

martint-unity: I think I just put it into a method instead of having inline as a lambda.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where Freeform 2D Lights were not culled correctly when there was a Falloff Offset.
 - Fixed an issue where Tilemap palettes were invisible in the Tile Palette window when the 2D Renderer was in use. [case 1162550](https://issuetracker.unity3d.com/issues/adding-tiles-in-the-tile-palette-makes-the-tiles-invisible)
 - Fixed issue where black emission would cause unneccesary inspector UI repaints [case 1105661](https://issuetracker.unity3d.com/issues/lwrp-inspector-window-is-being-repainted-when-using-the-material-with-emission-enabled-and-set-to-black-00-0)
+- Fixed an issue where deleting an entry from the Renderer List and undid that change, could cause a null reference. [case 1191896](https://issuetracker.unity3d.com/issues/nullreferenceexception-when-attempting-to-remove-entry-from-renderer-features-list-after-it-has-been-removed-and-then-undone)
```
reallyfancy: Change to: "Fixed an issue where deleting an entry from the Renderer List and then undoing that change could cause a null reference."

martint-unity: Should be fixed now. Next time, feel free to change it your self. We trust docs people :)

PR #4851, opened 2019-10-22
Fix a number of issues with prefab isolation mode.

JulienIgnace-Unity:
### Purpose of this PR
Fix a number of issues with prefab isolation mode:
- Volume from the main scene no longer leak in the prefab isolation
- Planar Reflection are properly handled

Also renamed ProbeSettings.camera to cameraSettings to avoid confusion.

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.core/Runtime/Volume/VolumeManager.cs
```diff
@@ -289,13 +289,20 @@ public void Update(VolumeStack stack, Transform trigger, LayerMask layerMask)
             // Sort the cached volume list(s) for the given layer mask if needed and return it
             var volumes = GrabVolumes(layerMask);
 
+            Camera camera = trigger.GetComponent<Camera>();
```
Chman: This will generate garbage on every frame, I know you just moved the code but I'm not sure why we didn't catch it in the previous version. Use TryGetComponent instead.

Also I'm not sure how `IsVolumeRenderedByCamera` is going to behave if `camera` is `null` (which will happen when the volume trigger isn't a camera but an object). I assume it'll be fine given that it worked before (or did it?). Worth a check.

PR #4164, opened 2019-07-17
UniversalRP XRSDK integration

thomas-zeng:
### Purpose of this PR
This PR contains the groundwork for Pure XRSDK integration. The whole system is disabled for now due to the reason projection, view and render target logic is not implemented yet.

- Added XRSDK System in UniversalRP (system is disabled until the projection matrix, view matrix, and render targets are hooked up with UniversalRP's render logic)
** see XRSystem.cs and XRPass.cs
- Added finalXRBlit in ForwardRenderer to support final blit to eye texture (this is disabled until XR System is enabled)
** see ForwardRenderer.cs
- Added XRSDK mirrorView support (this is disabled until XR system is enabled)
** see UniversalRenderPipeline.cs
---
### Release Notes
- None

---
### Testing status
**Katana Tests**: Running

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally

**Automated Tests**: 
- added XRSystemTests.ZeroGCMemoryPerFrame as unit test to validate we're not allocating garbage each frame

Launch Katana (Link on master here - update for your branch):
https://katana.bf.unity3d.com/projects/com.unity.render-pipelines/builders?ScriptableRenderLoop_branch=master&automation-tools_branch=master&unity_branch=trunk&sort=1-asc

---
### Overall Product Risks
**Technical Risk**: Low

**Halo Effect**: Low

---
### Comments to reviewers
This PR is trying to match HDRP's XRSDK implementation as close as possible. 
HDRP's XRSDK PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3393

This PR introduces XRSystem and XRPass to lay down the foundation pieces for further Pure XR integration. Pure XR Integration is WIP here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/compare/urp-xr-system-option3

PR #4857, opened 2019-10-23
URP Managed View&Proj

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Enabled URP to manage its own projection&view in URP-pure path
    - This opens up the opportunity of direct integration with XRSDK
- All URP build-in render passes are using pure path by default. 
- URP XR is on non-pure path, this will be addressed in the follow-up PRs (URP XRSDK integration work)
- Removed usage of cmd.Blit and cmd.SetViewAndProjection in URP-pure path, URP-pure now manages y-flip, blit explicitly in SRP land
---
### Testing status

Failed tests are:
  - URP: 044_Lighting_ReflectionProbe (3.098s) 
  - PostFX: 0213_Std_ColorGrading_LDR_Neutral (0.552s)
  - PostFX: 0214_Std_ColorGrading_HDR_Mix (0.104s)
  - PostFX: 0215_Std_ColorGrading_LDR_Mix (0.073s)
- [x] Built a player
- Other: 
Manually tested UniversalCustomRenderer examples

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
```diff
@@ -163,7 +163,29 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                viewMatrix = Matrix4x4.identity;
+                Matrix4x4 viewProjMatrix = projMatrix * viewMatrix;
+                Matrix4x4 invViewProjMatrix = Matrix4x4.Inverse(viewProjMatrix);
+
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewMatrix"), viewMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewMatrix"), Matrix4x4.Inverse(viewMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ProjMatrix"), projMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvProjMatrix"), Matrix4x4.Inverse(projMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewProjMatrix"), viewProjMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewProjMatrix"), Matrix4x4.Inverse(viewProjMatrix));
```
phi-lira: I'm concerned about the cost of setting all these constants in multiple passes. 
If it's a built-in shader/material and we need identity matrices we can actually change the shader code instead of uploading several constants per-pass.

Changing matrices also introduces another problem. You have to make sure to restore those end the pass end, as users can inject render passes pretty much anywhere they would expect the camera matrices to be correctly set.

com.unity.render-pipelines.universal/Runtime/Passes/SceneViewDepthCopy.cs
```diff
@@ -30,12 +30,43 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+            // Do logic using ScriptableRenderer.SetRenderTarget or override Configure to specify rt upfront.
             CoreUtils.SetRenderTarget(cmd, BuiltinRenderTextureType.CameraTarget);
+
             cmd.SetGlobalTexture("_CameraDepthAttachment", source.Identifier());
             cmd.EnableShaderKeyword(ShaderKeywordStrings.DepthNoMsaa);
             cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa2);
             cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa4);
-            cmd.Blit(source.Identifier(), BuiltinRenderTextureType.CameraTarget, m_CopyDepthMaterial);
+
+            if (renderingData.cameraData.isPureURPCamera)
```
phi-lira: I don't understand why this code has to change as well. This pass is only used in Editor. It doesn't affect XR.

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
```diff
@@ -69,12 +69,31 @@ public static bool ExtractSpotLightMatrix(ref CullingResults cullResults, ref Sh
...
+                viewMatrix = view;
+                Matrix4x4 viewProjMatrix = projMatrix * viewMatrix;
+                Matrix4x4 invViewProjMatrix = Matrix4x4.Inverse(viewProjMatrix);
+
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewMatrix"), viewMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewMatrix"), Matrix4x4.Inverse(viewMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ProjMatrix"), projMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvProjMatrix"), Matrix4x4.Inverse(projMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewProjMatrix"), viewProjMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewProjMatrix"), Matrix4x4.Inverse(viewProjMatrix));
```
phi-lira: I don't understand the reason for this. Here we need to set the light matrices, which are passed as arguments in the function. Shadow caster passes don't rely on camera matrices and don't affect XR.

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
```diff
@@ -0,0 +1,148 @@
...
+real4    BUILDIN_unity_SHBr;
+real4    BUILDIN_unity_SHBg;
+real4    BUILDIN_unity_SHBb;
+real4    BUILDIN_unity_SHC;
+real4    BUILDIN_unity_SpecCube0_HDR;
+float4   BUILDIN_unity_LightmapST;
+float4   BUILDIN_unity_DynamicLightmapST;
+float4   BUILDIN_unity_LODFade;
+float4   BUILDIN_unity_ProbesOcclusion;
+CBUFFER_END
```
phi-lira: Afaik you cannot change the layout of this CBUFFER as the SRP batcher specifically requires a certain layout.

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
```diff
@@ -0,0 +1,148 @@
...
+// y = far/near
+// z = x/far
+// w = y/far
+// or in case of a reversed depth buffer (UNITY_REVERSED_Z is 1)
+// x = -1+far/near
+// y = 1
+// z = x/far
+// w = 1/far
+float4 _ZBufferParams;
+CBUFFER_END
```
phi-lira: Why is this required to be redefined? It seems the same as in UnityInput.hlsl

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
```diff
@@ -0,0 +1,148 @@
...
+CBUFFER_START(UnityPerFrame)
+half4 _GlossyEnvironmentColor;
+half4 _SubtractiveShadowColor;
+// Time (t = time since current level load) values from Unity
+float4 _Time; // (t/20, t, t*2, t*3)
+float4 _SinTime; // sin(t/8), sin(t/4), sin(t/2), sin(t)
+float4 _CosTime; // cos(t/8), cos(t/4), cos(t/2), cos(t)
+float4 BUILDIN_unity_DeltaTime; // dt, 1/dt, smoothdt, 1/smoothdt
+float4 _TimeParameters; // t, sin(t), cos(t)
+float4 BUILDIN_unity_FogParams;
```
phi-lira: I'm not sure if I understand the reason to change these. When we remove SetupCameraProperties we will define those in URP with the same name.

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
```diff
@@ -0,0 +1,148 @@
...
+StructuredBuffer<int> _AdditionalLightsIndices;
+#else
+float4 _AdditionalLightsPosition[MAX_VISIBLE_LIGHTS];
+half4 _AdditionalLightsColor[MAX_VISIBLE_LIGHTS];
+half4 _AdditionalLightsAttenuation[MAX_VISIBLE_LIGHTS];
+half4 _AdditionalLightsSpotDir[MAX_VISIBLE_LIGHTS];
+half4 _AdditionalLightsOcclusionProbes[MAX_VISIBLE_LIGHTS];
+#endif
+real4 BUILDIN_unity_FogColor;
+CBUFFER_END
```
phi-lira: Why is this being changed? Please use the same as UnityInput.hsls. For instance, that one is not wrapped in a CBUFFER, because you are adding one here it will regress performance in GLES.

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
```diff
@@ -0,0 +1,148 @@
...
+
+// Descriptors
+TEXTURECUBE(BUILDIN_unity_SpecCube0);
+SAMPLER(BUILDIN_samplerunity_SpecCube0);
+TEXTURE2D(BUILDIN_unity_Lightmap);
+SAMPLER(BUILDIN_samplerunity_Lightmap);
```
phi-lira: Why changing these as well?

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesURPBuildinVarsDecl.hlsl
```diff
@@ -0,0 +1,39 @@
...
+#define BUILDIN_unity_SpecCube0            unity_SpecCube0;
+#define BUILDIN_samplerunity_SpecCube0     samplerunity_SpecCube0
+#define BUILDIN_unity_SpecCube0_HDR        unity_SpecCube0_HDR
+#define BUILDIN_unity_LightmapST           unity_LightmapST
+#define BUILDIN_unity_DynamicLightmapST    unity_DynamicLightmapST
+#define BUILDIN_unity_LODFade              unity_LODFade
+#define BUILDIN_unity_FogParams            unity_FogParams
+#define BUILDIN_unity_ProbesOcclusion      unity_ProbesOcclusion
+#define BUILDIN_unity_Lightmap             unity_Lightmap
+#define BUILDIN_samplerunity_Lightmap      samplerunity_Lightmap
```
phi-lira: Why can't people use them?

phi-lira:
The main thing I don't understand. For **non-stereo path**, nothing apart from the Camera matrix setup in ScriptableRenderer needs to change to make this "pure" on non-stereo path. However this PR adds a lot of overhead to sets camera matrices/state in several passes.

 I don't fully understand the stereo paths, but it seems a lot of those camera state changes could be removed if we change the built-in URP shaders (blit, copy color & depth) to be stereo aware.

com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -73,22 +75,63 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ProjMatrix"), projMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvProjMatrix"), Matrix4x4.Inverse(projMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewProjMatrix"), viewProjMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewProjMatrix"), Matrix4x4.Inverse(viewProjMatrix));
+
+                switch (m_DownsamplingMethod)
+                {
+                    case Downsampling.None:
+                        CoreUtils.SetRenderTarget(cmd, opaqueColorRT);
+                        cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, m_CopyColorMaterial);
```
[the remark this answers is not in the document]
thomas-zeng: I completely onboard with vertex shader full screen quad/triangle blit approach. That is simply faster and requires no vert/index stream prep.
Current URP still uses draw full-screen quad out of two reasons:
1. y-flip is handled through y-flip internal projection, full screen vert/quad bypass projection matrix and thus doesn't handle y-flip
    - Pure Mode in this PR handles y-flip in urp land thus we manage y-flip more explicitly, we can switch to vert shader blit approach with additional logic to have y-flip parity with legacy
2. some low end device does not support vertID in vert shader stage, in such case, full screen quad draw is needed. I believe URP still wants to support this case? We can have a blit helper function to use either full quad draw or vert blit based on runtime gfx device caps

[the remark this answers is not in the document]
thomas-zeng: Code path here is still needed. The reason is that Blit below calls into cmd.Blit
cmd.Blit is not an SRP solution because it uses build-in blit shaders in ImageFilter class and it uses build-in camera to drive that blit(eg, inject y flip logic into projection matrix data, alter blit src region and blit dst region etc).
I think it is better we replace cmd.Blit with srp managed blit(both shader and buffer binding)

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -86,6 +121,9 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                     m_TargetDimension);
 
                 Camera camera = cameraData.camera;
+
+                cmd.SetViewport(cameraData.camera.pixelRect);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, blitMaterial);
```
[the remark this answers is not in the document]
thomas-zeng: This is a mistake, will revert changes here

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -63,7 +63,42 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewMatrix"), Matrix4x4.Inverse(viewMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ProjMatrix"), projMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvProjMatrix"), Matrix4x4.Inverse(projMatrix));
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_ViewProjMatrix"), viewProjMatrix);
+                cmd.SetGlobalMatrix(Shader.PropertyToID("_InvViewProjMatrix"), Matrix4x4.Inverse(viewProjMatrix));
+
+                cmd.SetViewport(cameraData.camera.pixelRect);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, blitMaterial);
+            }
+            else if ( cameraData.isStereoEnabled || cameraData.isSceneViewCamera || cameraData.isDefaultViewport)
```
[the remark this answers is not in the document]
thomas-zeng: This is by design. 
The current plan is:
When stereo mode is on, XRSystem will configure blit target upfront to point to eye texture returned by xrsdk.
The config logic will happen in renderpass setup phase. In that case, during execution phase, no need to treat xr specially.

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -95,6 +95,7 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                     Vector4 cameraTranslation = viewMatrix.GetColumn(3);
                     viewMatrix.SetColumn(3, cameraTranslation + m_CameraSettings.offset);
 
+                    // XRTODO: remove cmd.SetViewProjectionMatrices
```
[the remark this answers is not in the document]
thomas-zeng: I think with URP managed camera data, maybe we want to consider moving all camera-related data management into srp land? 
cmd.SetViewProjectionMatrices is used to set build-in camera data and I think we should avoid using it

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesURPBuildinVarsDecl.hlsl
thomas-zeng: Right now we shouldn't prevent people from using them, in the future when we have srp equivalents for these, we should prevent people from using build-in vars because build-in vars require setCameraProperties(or other build-in function) call and in pure URP land, I think we can't guarantee all build-in functions are called.

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
thomas-zeng: I will make this the same as UnityInput.hlsl.

Out of curiosity, why CBuffer will regress perf in GLES?

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
thomas-zeng: UnityInput.hlsl contains lots of build-in stereo mode buffers that we try to get rid of. what I am trying to do here is to warp all cbuffers used in URP in ShaderVariablesMatrixDefsURP.hlsl 
I can include UnityInput.hlsl back and introduce ifdef there for camera & stereo buffer if that approach sounds better to you.

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
thomas-zeng: Shadow caster passes internally uses UNITY_MATRIX_* macro, which is rerouted to camera matrices. So matrices here need to be configured as light matrices.
The current URP uses cmd.SetViewProjectionMatrices() to wire light matrices to build-in unity_Matrix* buffers
In pure mode, I use the above code to wire light matrices to URP managed camera-related buffers

com.unity.render-pipelines.universal/ShaderLibrary/ShaderVariablesMatrixDefsURP.hlsl
phi-lira: Driver issue in a certain GPU vendor is part of the issue, but we are still investigating if there's something else in our code.

phi-lira:
I've taken a look at your comments and the cmd.Blit and cmd.SetViewProjection matrices. I understand better the issues now. I'm going to add some things we need to validate here:

For QA
 Features to test:
   - Text Rendering / UI
   - MSAA (ImageFilters::Blit makes sure to resolve MSAA targets)
   - Tesselation 
   - MSAA
   - Editor Rendering: Outline Selection, Gizmos, Fade Selection.
   - Instancing
   - Custom Render Passes
   - Terrain Speed Tree & Billboard  

  Projects to Test:
   - URP Graphics Test
   - Boat Attack
   - UniversalCustomRenderer examples

mikesfbay:
I've looked over the PR. It's unnecessary to remove command buffer cmd.Blit from Universal pipeline due to these reasons:
- Existing cmd.Blit handles both the monoscopic texture blit and stereo blit to and from eye texture on all platforms and API's. This shouldn't change whether URP uses XR shim display or Pure XRSDK. In the Pure XRSDK case, it must still handle the 3 stereo rendering modes and create eye render textures with VR flags. In fact, the new XR SDK display's XRTextureManager already sets up render texture similar to legacy XR so to work with our existing Blit. 

- Our existing Blit also handles y-flip at a platform/API dependent level. Once again this is below URP's renderer/command buffer level of abstraction, which should remain device/api-independent. URP shouldn't have to worry about device-dependent y-flips. 

Thus there is no good reason to replace all URP and postprocessing cmd.Blit's with drawMesh or make a Pure version.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -4,6 +4,11 @@
 #define MAX_VISIBLE_LIGHTS_SSBO 256
 #define MAX_VISIBLE_LIGHTS_UBO  32
 
+// Experimental Code Path: Pure URP
+#if UNITY_PURE_URP_ON 
+#define UNITY_PURE_URP_ENABLED
```
phi-lira: Suggestion: Figure out which platforms/APIs you want pure URP enabled and decide statically.

com.unity.render-pipelines.universal/Shaders/2D/Light2D-Point-Volumetric.shader
```diff
@@ -17,6 +17,9 @@ Shader "Hidden/Light2d-Point-Volumetric"
             #pragma multi_compile_local USE_POINT_LIGHT_COOKIES __
             #pragma multi_compile_local LIGHT_QUALITY_FAST __
 
+            // Enable Pure URP Camera Management
+            #pragma shader_feature _ UNITY_PURE_URP_ON
```
phi-lira: shader_feature is stripped in the build if no material defines it. It would have to be a multi_compile. However:
Adding a multi_compile to all shaders is not a good solution. 
 1) Doubles shader variants for all built-in shaders / shader graph
 2) Doesn't work for custom shaders. 

 We need to decide what platform/APIs will support pure and define this macro based on that so we don't depend on shader keyword.

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
```diff
@@ -9,15 +9,25 @@ Shader "Hidden/Universal Render Pipeline/Bloom"
...
+        // Enable Pure URP Camera Management
+        #pragma shader_feature _ UNITY_PURE_URP_ON
+
         #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
         #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Filtering.hlsl"
         #include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl"
         #include "Packages/com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl"
 
+#if defined(UNITY_PURE_URP_ENABLED)
+        TEXTURE2D_X(_BlitTex);
```
phi-lira: Suggestion: all of these shaders are stock shaders with URP. We can change already the code to bind _BlitText instead of _MainTex and avoid this define. Each code path that we add increases the cost and risk to maintain code and break per-platform. Whatever we can fix already we should.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -271,6 +272,22 @@ static void InitializeCameraData(UniversalRenderPipelineAsset settings, Camera c
...
+                Shader.DisableKeyword("UNITY_PURE_URP_ON");
+                URPCameraMode.isPureURP = false;
+            }
+            else
+            {
+                // We turn on pure mode on a per pass basis through cmd.EnableShaderKeyword in relevant passes
+                // This enables us to deprecate no-pure usage without breaking existing custom passes/render features
+                // XRTODO: Enable pure mode globally by uncomment the following line
+                // Shader.EnableKeyword("UNITY_PURE_URP_ON");
+                Shader.DisableKeyword("UNITY_PURE_URP_ON");
```
phi-lira: Choice of Pure or non-pure should be per-platform/device not per-pass.  It should work with custom render passes.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -60,6 +64,30 @@ half4 _AdditionalLightsSpotDir[MAX_VISIBLE_LIGHTS];
...
+#include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl"
+#if defined(UNITY_PURE_URP_ENABLED)
+// Define Getters
+// Note: In order to be able to define our macro/getter to forbid usage of unity_* build-in shader vars
+// We need to declare inline function.
+float4x4 GetRawUnityObjectToWorld() { return unity_ObjectToWorld; }
+float4x4 GetRawUnityWorldToObject() { return unity_WorldToObject; }
+// To get instancing working, we must use UNITY_MATRIX_M / UNITY_MATRIX_I_M as UnityInstancing.hlsl redefine them
+#define unity_ObjectToWorld        Use_Macro_UNITY_MATRIX_M_instead_of_unity_ObjectToWorld
+#define unity_WorldToObject        Use_Macro_UNITY_MATRIX_I_M_instead_of_unity_WorldToObject
```
phi-lira: We can't break user shaders.
Write shader upgrader to upgrade unity_ObjectToWorld -> UNITYM_MATRIX_M and unity_WorldtoObject -> UNITY_MATRIX_I_M

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -60,6 +64,30 @@ half4 _AdditionalLightsSpotDir[MAX_VISIBLE_LIGHTS];
 half4 _AdditionalLightsOcclusionProbes[MAX_VISIBLE_LIGHTS];
 #endif
 
+#include "Packages/com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl"
+#if defined(UNITY_PURE_URP_ENABLED)
+// Define Getters
+// Note: In order to be able to define our macro/getter to forbid usage of unity_* build-in shader vars
+// We need to declare inline function.
+float4x4 GetRawUnityObjectToWorld() { return unity_ObjectToWorld; }
+float4x4 GetRawUnityWorldToObject() { return unity_WorldToObject; }
```
phi-lira: Define these functions to the non pure path as well. Otherwise shaders that use them in non-pure path will not compile.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
thomas-zeng: Pure should be enabled for all platforms/APIs. I don't think pure is directly bound to API caps or platform caps.

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
thomas-zeng: I will try your suggestion tmrw

com.unity.render-pipelines.universal/Shaders/2D/Light2D-Point-Volumetric.shader
thomas-zeng: I think using shader feature is fine here. If no material defines it, why compile the variants?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: The code here is to disable Pure for URP XR.
URP XR need non-pure pass to work. All custom render passes need to run non-pure to ensure backward compat.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
thomas-zeng: This shouldn't break user shader. All custom user shader does not declare UNITY_PURE_URP_ON and should go through non-pure shader include. We can then guide them how to upgrade to pure.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
thomas-zeng: see comments above

thomas-zeng:
@phi-lira 
Hello felipe! Thank you for taking your time reviewing the PR! I like your internal shader VP suggestion and I will try it out soon. 
I think the PR is heading to the right direction as we clearly separate pure/non-pure on a per pass basis, By going that route, we can support all existing custom passes and guide them to upgrade to pure version. We can then deprecate cmd.Blit and cmd.SetViewAndProj and eventually move most of the rendering logic to srp land!

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
thomas-zeng: I think we need this define here because this shader is used by cmd.Blit in non-pure pass.  And cmd.Blit binds blit src to _MainTex internally. IMHO we can only remove this define once cmd.Blit usage is completely deprecated in the codebase.

com.unity.render-pipelines.universal/Shaders/2D/Light2D-Point-Volumetric.shader
thomas-zeng: You are correct felipe. We will need multi_compile here: 4fad0f7.
To address shader compiling time, I will add shader variants strip logic in URP shader preprocessor. Non-XR strip all shaders that are not UNITY_PURE_URP_ON and XR strips all shaders that are UNITY_PURE_URP_ON. I think by doing this, shader variant number should be the same as before.

thomas-zeng: You are correct! I switched to multi_compile in this commit: 4fad0f7.
To address shader compiling time, I will add shader variant stripping logic in URP shader preprocessor. Non-XR strip all shaders that are not UNITY_PURE_URP_ON and XR strips all shaders that are UNITY_PURE_URP_ON. I think by doing this, shader variant number should be the same as before.

mikesfbay:
I'm concerned a lot of these changes are unnecessary and confusing. Most of these changes stem from a rather shallow understanding of XR and Universal:

- There's no need to replace the current command buffer Blit in Universal for pure XRSDK since existing Blit will work with Pure XRSDK mode with minor modifications (trunk PR is coming).

- Most postprocessing blits aren't even XR related and shouldn't have been changed.

- There's no need to touch any of Universal's non-XR related camera view and projection matrices as in this PR since VR rendering doesn't even use them (provided by HMD).

- Handling Y-flip is platform/API dependent and exposing them all over Universal only adds complexity and compromises Universal's cross-platform/API simplicity. In addition, this PR also doesn't really handle XR related y-flips.

- There's no need to add multi-compile to all shaders as test code (which is more like a hack) and will only bloat and confuse Universal/XR code base.

At this point, there's little reason to add a bunch of test code (including hacks) of dubious value into production-ready code.

There's also no evidence from this PR to suggest pure XRSDK path for stereo rendering is better for Universal for all platforms/APIs without performance penalties and without obscuring and confusing the codebase when compared to simpler XR shim path.

com.unity.render-pipelines.universal/Shaders/2D/Light2D-Point-Volumetric.shader
phi-lira: Still you have the problem 2. Doesn't work for custom shaders and it's not reasonable to ask users to define this for their shaders if we want to remove it later.

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
phi-lira: Not true. The current FinalBlitPass uses command.Blit and binds _BlitTex.

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
```diff
@@ -3,6 +3,7 @@
 using UnityEditor;
 using UnityEditor.Build;
 using UnityEngine;
+using UnityEngine.XR;
```
phi-lira: This should be wrapped with XR_ENABLED (I don't actually know if this the proper name)

com.unity.render-pipelines.universal/Shaders/Utils/CopyColor.shader
```diff
@@ -0,0 +1,77 @@
+Shader "Hidden/Universal Render Pipeline/CopyColor"
```
phi-lira: Why not using the Universal Blit shader?

phi-lira:
Hi @thomas-zeng I agree with several points @mikesfbay brought up here.

 You seem to be missing a big factor here.: Users and the whole asset store developer ecosystem.
 
 It's not reasonable to expect to all users and devs that now they should add a multi_compile to all their shaders, handle y-flip themselves, and be mindful to set up camera matrices in their custom render passes because we don't guarantee that in any custom render pass (or at end camera rendering callbacks) they will have valid properly setup camera matrices. 

 Just to give one example on how this can break up things. We support Amplify Shader Editor: https://assetstore.unity.com/detail/tools/visual-scripting/amplify-shader-editor-68570

 Plus, with your code, you are setting camera matrices in multiple places for no reason. f.ex, the shadow caster pass doesn't need anything other than camera VP matrices.

com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
thomas-zeng: Addressed in f2654ec

com.unity.render-pipelines.universal/Editor/ShaderPreprocessor.cs
thomas-zeng: Removed in f2654ec

com.unity.render-pipelines.universal/Shaders/Utils/CopyColor.shader
thomas-zeng: Removed this shader and replaced its usage with blit shader in f2654ec

com.unity.render-pipelines.universal/Shaders/PostProcessing/Bloom.shader
thomas-zeng: Addresed in f2654ec

com.unity.render-pipelines.universal/Runtime/RenderingUtils.cs
```diff
@@ -142,5 +142,43 @@ public static bool SupportsRenderTextureFormat(RenderTextureFormat format)
...
+
+        /// <summary>
+        /// Helper function to set all view and projection related matricies
+        /// Interal usage only, function name and signature may be subject to change
+        /// </summary>
+        /// <param name="cmd">The command buffer to queue Gfx commands to.</param>
+        /// <param name="viewMatrix">Desired view matrix to set to</param>
+        /// <param name="projMatrix">Desired projection matrix to set to.</param>
+        /// <returns>Void</c></returns>
+        internal static void SetViewProjectionRelatedMatricesAll(CommandBuffer cmd, Matrix4x4 viewMatrix, Matrix4x4 projMatrix)
```
phi-lira: Suggestion: Combine these two in a single one and pass a boolean argument to tell if inverse matrices should be set.

phi-lira: `Related` name here is not adding much to the name signature imo. glstate_matrix_projection is implementation detail that should not matter for client code.

phi-lira:
@thomas-zeng current state of the pureURP is always true. so all the else code is dead code. Should we remove the else path at this point?

phi-lira:
Overall looks in a good shape. I need to wait for the other stakeholders involved but I'd like to start testing.

phi-lira:
@Unity-Technologies/gfx-qa-foundation Can we test this with the following projects:
  - Boat Attack
  - Universal Renderer Examples
  
 For performance tests I'd like to test the Unite polygon farm scene @ernestasKupciunas could you help us with performance testing this PR?

com.unity.render-pipelines.universal/Runtime/RenderingUtils.cs
thomas-zeng: Addressed in 587c2a1

TestProjects/UniversalGfxTestStereo/ProjectSettings/QualitySettings.asset
```diff
@@ -201,7 +201,7 @@ QualitySettings:
     skinWeights: 4
     textureQuality: 0
     anisotropicTextures: 2
-    antiAliasing: 0
```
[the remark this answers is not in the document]
thomas-zeng: Addressed in a1e18ff

mikesfbay:
Here are a few more points to consider on this PR:

- This PR is largely incomplete in terms of XR pure-path support. This PR also affects non-XR users since it's enabled when XR mode is off. One question to consider is: do we want to expose non-XR URP users to complicated XR logic so early before any evidence of Pure XR SDK is proven for URP? 

- This PR tries to replace command buffer Blit but a lot of functionality is missing in the PR. Existing command buffer Blit handles both normal color/depth texture blits and XR stereo blits to and from eye textures. It also handles texture array slice blits and partial texture blits. This PR doesn't support these nor does it handle XR blit cases. 

- MSAA resolve is handled by default in command buffer Blit. This PR doesn't support it. In fact, script users really shouldn't have to handle MSAA nor worry about it. This should be done at a graphics device-level.

- Y-flip is handled automatically in command buffer Blit and is platform/API dependent. In this PR, Y-flip is scattered throughout URP which only adds complexity and compromises Universal's cross-platform/API simplicity. In addition, this PR doesn't really handle all cases of Y-flips nor XR related ones.

- This PR also tries to replace command buffer setViewProjectionMatrices by setting view and projection matrices on its own. However, the command buffer's setViewProjectionMatrices maintains a consistent internal state with the graphics device and the rest of the engine. Have we really thought this through by looking at all engine dependencies in URP: Speedtree, Skybox, uses of clip planes, etc? Having the user set graphics view and projection matrices directly can lead to wrong internal state and hard-to-find rendering bugs. 

- This PR's view/projection changes are initially motivated by XR, but XR doesn't really use URP camera's view and projection matrices (HMD provides them). Thus, there's no XR need to replace URP's handling of camera view and projection.

- The PR is also inconsistent at replacing SetViewProjectionMatrices. In the PR, RenderShadowSlice still calls SetViewProjectionMatrices even though PR is supposed to have replaced them. This will lead to an inconsistent state.

- Lack of custom passes support and difficulty in extending/modifying URP. It's going to more difficult for users to add new passes and postfx since this PR exposed all the complex XR logic throughout URP. We can't expect any non-XR programmer to use them properly when extending URP. If they do change it or adapt from it, they are bound to make mistakes. This is a step back for users.

- This PR also adds a lot more DrawMesh to replace all Blits. DrawMesh is a slow function since it takes arbitrary mesh vertices and has more overhead in setup and rendering of general mesh. Blit always uses 4 vertex quad and doesn't call DrawMesh internally. 

My suggestion: Since this PR is largely incomplete, it should be confined to the experimental branch until it has shown evidence of feature and performance parity and API clarity with existing XR shim mode. It can't be a step back for users. Currently, it's too early and there's no proof supporting integration today. Also, command buffer blit should be left alone, and there's little need to replace it with something possibly worse in functionality, usability, and performance.

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -79,11 +67,55 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                bool isRenderToCameraTarget = true;
+                bool isCameraTargetIntermediateTexture = cameraData.camera.targetTexture != null || cameraData.camera.cameraType == CameraType.SceneView || cameraData.camera.cameraType == CameraType.Preview;
+                bool isRenderToTexture = !isRenderToCameraTarget || isCameraTargetIntermediateTexture;
+                Matrix4x4 projMatrix = GL.GetGPUProjectionMatrix(Matrix4x4.identity, isRenderToTexture);
+                RenderingUtils.SetViewProjectionMatrices(cmd, Matrix4x4.identity, projMatrix, true);
                 cmd.SetViewport(cameraData.camera.pixelRect);
-                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, blitMaterial);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, m_BlitMaterial);
+                RenderingUtils.SetViewProjectionMatrices(cmd, camera.worldToCameraMatrix, GL.GetGPUProjectionMatrix(camera.projectionMatrix, isRenderToTexture), true);
                 cmd.SetViewProjectionMatrices(camera.worldToCameraMatrix, camera.projectionMatrix);
```
phi-lira: This line should be removed.

phi-lira:
Hi @mikesfbay commenting in-place here.

> This PR is largely incomplete in terms of XR pure-path support. This PR also affects non-XR users since it's enabled when XR mode is off. One question to consider is: do we want to expose non-XR URP users to complicated XR logic so early before any evidence of Pure XR SDK is proven for URP?

Could you elaborate more on what XR logic we are exposing to users?

> This PR tries to replace command buffer Blit but a lot of functionality is missing in the PR. Existing command buffer Blit handles both normal color/depth texture blits and XR stereo blits to and from eye textures. It also handles texture array slice blits and partial texture blits. This PR doesn't support these nor does it handle XR blit cases.
> 
> MSAA resolve is handled by default in command buffer Blit. This PR doesn't support it. In fact, script users really shouldn't have to handle MSAA nor worry about it. This should be done at a graphics device-level.
> 
> Y-flip is handled automatically in command buffer Blit and is platform/API dependent. In this PR, Y-flip is scattered throughout URP which only adds complexity and compromises Universal's cross-platform/API simplicity. In addition, this PR doesn't really handle all cases of Y-flips nor XR related ones.

We've started testing. If you have some test scenes that could catch issues please let us know as we need to test and fix all possible issues.

> This PR also tries to replace command buffer setViewProjectionMatrices by setting view and projection matrices on its own. However, the command buffer's setViewProjectionMatrices maintains a consistent internal state with the graphics device and the rest of the engine. Have we really thought this through by looking at all engine dependencies in URP: Speedtree, Skybox, uses of clip planes, etc? Having the user set graphics view and projection matrices directly can lead to wrong internal state and hard-to-find rendering bugs.

We are still calling SetupCameraProperties that sets up the billboard + camera clip planes required for speedtree et al. Please help us validate with some test scenes.

> This PR's view/projection changes are initially motivated by XR, but XR doesn't really use URP camera's view and projection matrices (HMD provides them). Thus, there's no XR need to replace URP's handling of camera view and projection.

This not not true. Ideally we want to get rid of all hooks Unity have from camera class and instead pass it in camera data. URP depending on c++ matrices surfaces some current issues that we have hacks to fix. We can't fix all of them in this PR as it increases PR risk and complexity but this is a good step forward considering we don't regress quality and perf. That's why we need a good coverage of test projects. 

> The PR is also inconsistent at replacing SetViewProjectionMatrices. In the PR, RenderShadowSlice still calls SetViewProjectionMatrices even though PR is supposed to have replaced them. This will lead to an inconsistent state.

Can you add comments to the source code so we fix this?

> Lack of custom passes support and difficulty in extending/modifying URP. It's going to more difficult for users to add new passes and postfx since this PR exposed all the complex XR logic throughout URP. We can't expect any non-XR programmer to use them properly when extending URP. If they do change it or adapt from it, they are bound to make mistakes. This is a step back for users.

This was addressed in last iteration. We are testing with a few projects that use custom renderers and render passes. Could you elaborate more on what kind of piece of XR specific API you are concerned about?

> This PR also adds a lot more DrawMesh to replace all Blits. DrawMesh is a slow function since it takes arbitrary mesh vertices and has more overhead in setup and rendering of general mesh. Blit always uses 4 vertex quad and doesn't call DrawMesh internally.

Could you elaborate more of why DrawMesh is slower than built-in blit? We are passing a cached mesh with 4 vertices as well. For the future PR we can also expose triangle blit without vb/ib.

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
thomas-zeng: Addressed in f2010cc

mikesfbay:
@felipe I've elaborated in the follow-up doc:
https://docs.google.com/document/d/1JrZN7nP-c06nrIX-Oge3IjjqJWbw-gfPFBYq1uLsumg

ernestasKupciunas:
Done with the performance testing. On iOS side Unite polygon farm scene looks fine. I couldn't test it well on Android because of this bug [1202462](https://fogbugz.unity3d.com/f/cases/1202462/). Basically the app is crashing on launch. But it has nothing to do with this PR.
Testing document:
https://docs.google.com/spreadsheets/d/1tVLzRjF_bhDv04TVmLLDHkKRC-rNlhNaLNxPknstm1o/edit?usp=sharing

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
```diff
@@ -73,12 +73,27 @@ public static bool ExtractSpotLightMatrix(ref CullingResults cullResults, ref Sh
...
-            cmd.Clear();
-            context.DrawShadows(ref settings);
-            cmd.DisableScissorRect();
+            if (URPCameraMode.isPureURP)
+            {
+                // XRTODO: decouple SpeedTreeV7 billboard system updates with SetViewProjectionMatrices and replace
+                // cmd.SetViewProjectionMatrices with RenderingUtils.SetViewProjectionMatrices
+                // Matrix4x4 projMatrix = GL.GetGPUProjectionMatrix(proj, true);
+                // RenderingUtils.SetViewProjectionMatrices(cmd, view, projMatrix, false);
+                cmd.SetViewProjectionMatrices(view, proj);
```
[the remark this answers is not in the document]
thomas-zeng: I have reverted ShadowUtils.cs changes in 6008112.

thomas-zeng:
As discussed with @phi-lira, we keep cmd.Blit in final blit pass in this PR so that HDR in URP would work the same way as before. Note: HDR API is not supported in URP yet, proper HDR API support requires URP integration with the API.

com.unity.render-pipelines.universal/Runtime/ShadowUtils.cs
```diff
@@ -73,27 +73,12 @@ public static bool ExtractSpotLightMatrix(ref CullingResults cullResults, ref Sh
...
-            else
-            {
-                cmd.SetViewProjectionMatrices(view, proj);
-                context.ExecuteCommandBuffer(cmd);
-                cmd.Clear();
-                context.DrawShadows(ref settings);
-                cmd.DisableScissorRect();
-            }
+
+            cmd.SetViewProjectionMatrices(view, proj);
```
mikesfbay: Currently, the PR sets up view/proj in some places while still calls cmd.SetViewProjectionMatrices in others (such as RenderShadowSlice). This can lead to hard to find problems later. 
Shadow can still remove the cmd.SetViewProjectionMatrices if it calls URP-side billboards setup or have cpp side API expose billboard setup.

thomas-zeng:
@mikesfbay MSAA is supported by SetRenderTarget in this PR. @ryanhy-unity could you verify MSAA works

thomas-zeng:
@mikesfbay > It'll be better if there is a URP-side Blit utility that encapsulates the details of whether MSAA resolve is needed.
Yes, that would be ideal to control MSAA in SRP land. However, MSAA resolve is not in pure SRP land yet so it requires a bit more work to handle MSAA explicitly.  For the current PR, I would like to keep its current scope. 
I just want to make sure we don't regress MSAA handling with this PR. 
For the Pure SRP MSAA you brought up, I think it is a good idea. We should investigate it in future PR.

thomas-zeng:
Failing test in Universal CI - 005 Lit Baked Scene is known issue caused by new GI.
Failing test in VFX Graph - 101 is also known issue (new GI)

thomas-zeng:
@mikesfbay I will follow up with you on URP blit utility request on slack

mikesfbay:
Removing cmd.Blit and not having a URP replacement Blit is a major step backwards and inconvenient for people to use. Instead of having single blit call, there are up to 6 lines of extra code just to do a simple blit. The code repeated in many places is messy to read and potentially buggy to modify. People have to worry about whether MSAA resolve is handled if they do blit themselves, which is another step backwards.

thomas-zeng:
@mikesfbay Like I said, your concern will be addressed in follow up PR. We have to do things step by step.
I would say as long as PR didn't regress URP. we should proceed
I thank you for your opinion though. I appreciate your time writing down notes.  Your suggestions will be taken into consideration when working on Pure URP XRSDK integration.

mikesfbay:
Code quality and usability are also important. The same blit-replacing code is repeated 8 different places in this PR when it can be encapsulated in a URP utility function. As I mentioned before, code repeated in many places is messy to read and potentially buggy to modify. People now also have to worry about whether MSAA resolve is handled correctly, which is another step backwards. Code that's a step backwards in usability or messy to read should be fixed before check-in. 

There are also other URP PR's in pipeline (such as deferred rendering PR), and it doesn't make sense for them to modify their Blit code and only to be replaced again later.

Also our QA's time is valuable, and it's a waste of effort to do another QA pass later over the essentially same functionality with the added Blit utility function.

phi-lira:
@thomas-zeng if you merge master it should be green now.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -350,21 +368,40 @@ int GetDestination()
...
-                if (m_IsStereo)
-                {
-                    Blit(cmd, GetSource(), BuiltinRenderTextureType.CurrentActive, m_Materials.uber);
-                }
-                else
+                if (URPCameraMode.isPureURP)
                 {
-                    cmd.SetViewProjectionMatrices(Matrix4x4.identity, Matrix4x4.identity);
+                    bool isRenderToCameraTarget = m_Destination == RenderTargetHandle.CameraTarget;
+                    bool isCameraTargetIntermediateTexture = cameraData.camera.targetTexture != null || cameraData.camera.cameraType == CameraType.SceneView || cameraData.camera.cameraType == CameraType.Preview;
```
phi-lira: You need to change the following line to be able to work with camera stacking:
`cameraData.camera.targetTexture` => `cameraData.targetTexture`

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -350,21 +368,40 @@ int GetDestination()
...
+                    bool isRenderToTexture = !isRenderToCameraTarget || isCameraTargetIntermediateTexture;
+                    if (!isRenderToTexture)
```
phi-lira: This logic is a little bit convoluted. maybe add some commentaries?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -430,41 +467,67 @@ void DoSubpixelMorphologicalAntialiasing(ref CameraData cameraData, CommandBuffe
...
-            // Pass 3: Neighborhood blending
-            cmd.SetRenderTarget(new RenderTargetIdentifier(destination, 0, CubemapFace.Unknown, -1),
-                RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store,
-                RenderBufferLoadAction.DontCare, RenderBufferStoreAction.DontCare);
-            cmd.SetGlobalTexture(ShaderConstants._ColorTexture, source);
-            cmd.SetGlobalTexture(ShaderConstants._BlendTexture, ShaderConstants._BlendTexture);
-            cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 2);
+            if (URPCameraMode.isPureURP)
+            {
+                cmd.SetViewport(pixelRect);
```
phi-lira: Double checking. The `cmd.SetViewProjectionMatrices` just above the SetViewport was removed with no replacement intentionally?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -430,41 +467,67 @@ void DoSubpixelMorphologicalAntialiasing(ref CameraData cameraData, CommandBuffe
...
+                cmd.SetGlobalTexture(ShaderConstants._BlendTexture, ShaderConstants._BlendTexture);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 2);
 
-            // Cleanup
-            cmd.ReleaseTemporaryRT(ShaderConstants._EdgeTexture);
-            cmd.ReleaseTemporaryRT(ShaderConstants._BlendTexture);
-            cmd.SetViewProjectionMatrices(camera.worldToCameraMatrix, camera.projectionMatrix);
+                // Cleanup
+                cmd.ReleaseTemporaryRT(ShaderConstants._EdgeTexture);
+                cmd.ReleaseTemporaryRT(ShaderConstants._BlendTexture);
```
phi-lira: There's a restore matrices missing here. Is it intentional?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -618,22 +720,57 @@ void DoBokehDepthOfField(CommandBuffer cmd, int source, int destination, Rect pi
...
+                cmd.SetRenderTarget(ShaderConstants._PingTexture, RenderBufferLoadAction.Load, RenderBufferStoreAction.Store);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 1);
+
+                // Bokeh blur
+                cmd.SetRenderTarget(ShaderConstants._PongTexture, RenderBufferLoadAction.Load, RenderBufferStoreAction.Store);
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), ShaderConstants._PingTexture);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 2);
+
+                // Post-filtering
+                cmd.SetRenderTarget(ShaderConstants._PingTexture, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store);
```
phi-lira: Missing `BlitDstDiscardContent`?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -618,22 +720,57 @@ void DoBokehDepthOfField(CommandBuffer cmd, int source, int destination, Rect pi
...
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), ShaderConstants._PongTexture);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 3);
+
+                // Composite
+                cmd.SetGlobalTexture(ShaderConstants._DofTexture, ShaderConstants._PingTexture);
+                cmd.SetRenderTarget(destination, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store);
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), source);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 4);
```
phi-lira: Seems `BlitDstDiscardContent` is missing as well

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -92,26 +92,67 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
 
-                if (m_CameraSettings.overrideCamera && m_CameraSettings.restoreCamera)
+                if (URPCameraMode.isPureURP)
+                {
+                    if (m_CameraSettings.overrideCamera && m_CameraSettings.restoreCamera)
+                    {
+                        cmd.Clear();
+                        {
+                            bool isRenderToCameraTarget = colorAttachment == RenderTargetHandle.CameraTarget.id;
+                            bool isCameraTargetIntermediateTexture = cameraData.camera.targetTexture != null || cameraData.camera.cameraType == CameraType.SceneView || cameraData.camera.cameraType == CameraType.Preview;
```
phi-lira: `cameraData.targetTexture`

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -92,26 +92,67 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
                     ref m_RenderStateBlock);
 
-                if (m_CameraSettings.overrideCamera && m_CameraSettings.restoreCamera)
+                if (URPCameraMode.isPureURP)
+                {
+                    if (m_CameraSettings.overrideCamera && m_CameraSettings.restoreCamera)
+                    {
+                        cmd.Clear();
+                        {
+                            bool isRenderToCameraTarget = colorAttachment == RenderTargetHandle.CameraTarget.id;
```
phi-lira: Seems like you could have a helper method as this code is duplicated in multiple places. Problem of duplicating code is maintenance. If you have to fix something you have to remember to fix in multiple places. Like the cameraData.targetTexture below.

com.unity.render-pipelines.universal/Runtime/RenderingUtils.cs
```diff
@@ -152,6 +152,40 @@ public static bool SupportsRenderTextureFormat(RenderTextureFormat format)
...
+        /// <param name="projMatrix">Desired projection matrix to set to.</param>
+        /// <param name="setVPOnly">Set only VP matrix if true, set V,P,IV,VP,IVP matrices otherwise.</param>
+        /// <returns>Void</c></returns>
+        internal static void SetViewProjectionMatrices(CommandBuffer cmd, Matrix4x4 viewMatrix, Matrix4x4 projMatrix, bool setVPOnly = false)
+        {
+            Matrix4x4 viewProjMatrix = projMatrix * viewMatrix;
+
+            if (setVPOnly)
+            {
+                cmd.SetGlobalMatrix(UNITY_MATRIX_VP, viewProjMatrix);
```
phi-lira: Why don't we set UNITY_MATRIX_V and UNITY_MATRIX_P as well?

com.unity.render-pipelines.universal/Runtime/RenderingUtils.cs
```diff
@@ -152,6 +152,40 @@ public static bool SupportsRenderTextureFormat(RenderTextureFormat format)
...
+            }
+            else
+            {
+                cmd.SetGlobalMatrix(UNITY_MATRIX_V, viewMatrix);
+                cmd.SetGlobalMatrix(UNITY_MATRIX_IV, Matrix4x4.Inverse(viewMatrix));
```
phi-lira: Inverse projection should be set as well. Required for position reconstruction from depth.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -295,10 +295,18 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
...
             // Known Issue: billboard will not work with camera stacking when using viewport with aspect ratio different from default aspect.
-            if (cameraData.renderType == CameraRenderType.Overlay)
-            {
-                cmd.SetViewProjectionMatrices(cameraData.viewMatrix, cameraData.projectionMatrix);
-            }
+           if (cameraData.renderType == CameraRenderType.Overlay)
+           {
+               if (URPCameraMode.isPureURP)
+               {
+                   Matrix4x4 projMatrix = GL.GetGPUProjectionMatrix(cameraData.projectionMatrix, false);
```
phi-lira: This is wrong and will break camera stacking. Projection matrix of Overlay camera is patched because it inherits the viewport from base camera. Therefore the aspect ratio of the projection can change, thus why it needs patching. 

In fact, every time you restore projection and view matrices you should restore with cameraData.viewMatrix and cameraData.projectionMatrix.

phi-lira:
Adding @hkr as reviewer due to changes that might affect the LOAD/STORE in PostProcessing pass.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -505,30 +568,69 @@ void DoGaussianDepthOfField(Camera camera, CommandBuffer cmd, int source, int de
...
+                cmd.SetViewport(pixelRect);
+                cmd.SetGlobalTexture(ShaderConstants._ColorTexture, source);
+                cmd.SetGlobalTexture(ShaderConstants._FullCoCTexture, ShaderConstants._FullCoCTexture);
+                cmd.SetRenderTarget(m_MRT2, ShaderConstants._HalfCoCTexture, 0, CubemapFace.Unknown, -1);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 1);
+
+                // Blur
+                cmd.SetGlobalTexture(ShaderConstants._HalfCoCTexture, ShaderConstants._HalfCoCTexture);
+                cmd.SetRenderTarget(ShaderConstants._PongTexture, RenderBufferLoadAction.Load, RenderBufferStoreAction.Store);
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), ShaderConstants._PingTexture);
```
Chman: Don't use `Shader.PropertyToID()` in code that is executed on every frame, all these calls should be cached in `ShaderConstants` as they have quite a bad impact on CPU performance on lower-end platforms (mobile comes to mind).

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -250,7 +250,14 @@ int GetDestination()
 
             void Swap() => CoreUtils.Swap(ref source, ref destination);
 
-            // Setup projection matrix for cmd.DrawMesh()
+            // Use SRP managed View&Proj
+            if(URPCameraMode.isPureURP)
```
Chman: I see a lot of branching everywhere on `URPCameraMode.isPureURP` and a lot of code duplication... Is there anything we can do to make it nicer and easier to maintain? Because it seems like a very dangerous idea going forward. We could reduce it to a single branch in the render method and duplicate all the SetupXX/DoXX methods with their SetupXXPure/DoXXPure equivalents... That won't help with maintenance and code duplication but it might be cleaner.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -665,7 +802,18 @@ void DoMotionBlur(Camera camera, CommandBuffer cmd, int source, int destination)
...
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), source);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, (int)m_MotionBlur.quality.value);
+            }
+            else
+            {
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), source);
+                cmd.Blit(source, BlitDstDiscardContent(cmd, destination), material, (int)m_MotionBlur.quality.value);
+            }
 
             m_PrevViewProjM = viewProj;
```
Chman: Note: this breaks down as soon as there's more than one camera with motion blur enabled, this is a known issue.
Wasn't there a plan for this PR at some point to do like HDRP and store persistent per-camera datas? Or maybe I'm mis-remembering. Because eventually it will be needed, not only for motion blur but also for any kind of temporal effect.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -259,8 +266,19 @@ int GetDestination()
             {
                 using (new ProfilingScope(cmd, ProfilingSampler.Get(URPProfileId.StopNaNs)))
                 {
-                    cmd.Blit(GetSource(), BlitDstDiscardContent(cmd, GetDestination()), m_Materials.stopNaN);
-                    Swap();
+                    if (URPCameraMode.isPureURP)
+                    {
+                        BlitDstDiscardContent(cmd, GetDestination());
```
hkr: We should probably rename BlitDstDiscardContent. The original intention was that it's a workaround to specify load actions when using Blit. Here it's now used to set the render target without using blit

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
hkr: BlitDstDiscardContent was replaced by SetRenderTarget. I think that's actually better.

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
hkr: The cleanup is missing in the else branch

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -430,41 +467,67 @@ void DoSubpixelMorphologicalAntialiasing(ref CameraData cameraData, CommandBuffe
...
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 1);
+
+                // Pass 3: Neighborhood blending
+                cmd.SetRenderTarget(new RenderTargetIdentifier(destination, 0, CubemapFace.Unknown, -1),
+                    RenderBufferLoadAction.DontCare, RenderBufferStoreAction.Store,
+                    RenderBufferLoadAction.DontCare, RenderBufferStoreAction.DontCare);
+                cmd.SetGlobalTexture(ShaderConstants._ColorTexture, source);
+                cmd.SetGlobalTexture(ShaderConstants._BlendTexture, ShaderConstants._BlendTexture);
+                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, material, 0, 2);
+            }
```
hkr: The if and else branches do almost exactly the same, why duplicate them?

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -505,30 +568,69 @@ void DoGaussianDepthOfField(Camera camera, CommandBuffer cmd, int source, int de
...
+                cmd.Blit(ShaderConstants._PingTexture, ShaderConstants._PongTexture, material, 2);
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), ShaderConstants._PongTexture);
+                cmd.Blit(ShaderConstants._PongTexture, BlitDstDiscardContent(cmd, ShaderConstants._PingTexture), material, 3);
+
+                // Composite
+                cmd.SetGlobalTexture(ShaderConstants._ColorTexture, ShaderConstants._PingTexture);
+                cmd.SetGlobalTexture(ShaderConstants._FullCoCTexture, ShaderConstants._FullCoCTexture);
+                cmd.SetGlobalTexture(Shader.PropertyToID("_BlitTex"), source);
+                cmd.Blit(source, BlitDstDiscardContent(cmd, destination), material, 4);
+            }
```
hkr: I don't know why the old code sometimes use Blit and sometimes it didn't. 
Can we maybe just always use DrawMesh?
Otherwise maybe wrap DrawMesh and Blit in a separate function and do the if there. It would cause more branching but duplicating everything looks like a bad idea to me.

Chman: For clarification: the only times it used DrawMesh instead of Blit was because it either needed to do MRT (which the internal method Blit doesn't support) or use the Stencil buffer.

thomas-zeng: @hkr @Chman 
Thanks for providing feedback!
One reason we are using cmd.Blit is for XR support. XR Blit could change blit src/dst region and bind stereo cbuffer under the hood so without using it postFX could break XR. 
Regarding the code duplication, I completely agree that it is not ideal. Right now, the URP XR can't be in pure mode yet due to legacy dependencies and that's the reason why this PR branches a lot. This will be addressed once XR is in pure URP land. XR graphics will then make urp pure path the only path and remove all else branches.   
Draft PR is here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5672/files#diff-8c7fe0871ba3fca6282ea213dd728fc3R5232
(Still a draft, I will need to implement all XR features in urp land and reach feature parity with legacy XR shim before making a real PR. It's very close to feature parity now.)

Chman: > Right now, the URP XR can't be in pure mode yet due to legacy dependencies

I'm curious, what are these legacy dependencies? Do you mean on the engine side? Seeing as this PR dates from October (so initially targeting 19.3), are these dependencies still a concern with 20.1?

thomas-zeng: @Chman 
The current URP XR is very similar to the URP skybox: all logic is on the engine side. To move to pure URP land, I will need to surface all XR logic to SRP. In 19.3, it's not possible because we need to support legacyXR which does not provide enough c# facing API.

20.1+ XRSDK is the only option in the trunk(legacyXR c++ side is completely removed). Current XRSDK works in URP by talking to legacy XR c# API through the shim layer. In pure XRSDK mode, we will be using XRSDK API directly, bypassing shim. Eventually, we want to remove shim layer as well as legacy XR c# API.

Since we are surfacing XR logic to SRP and swap out the legacy XR, we need to make sure we have feature parity with whatever is working now. So the draft PR I am currently working on needs to get everything working in one go.

mikesfbay: Just to clarify, legacy XR engine didn't go away in trunk/20.1, but only the various XR vendor/provider code (i.e. VROculus) got shuffled into package plugins for XRSDK. 

The existing XR stereo cpp engine still works fine with URP via XRSDK shim path and is optimized on all platforms at gfx device level. For URP, I still believe this is the cleanest, simplest and optimal way for URP to work with.

The case for "pure XRSDK" for URP is still debatable, and I've listed a large set of unresolved issues here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4164#issuecomment-514044246 
and here: https://docs.google.com/document/d/1wcOaO50sNW8c8RNqGemdvSPp2G4wu-sra2bTF_68q94

phi-lira:
I see a lot of the concerns from @hkr and @Chman are because the isPureXRSDK code path. The reason is that we need to start moving to pure xr sdk but it's just too much to do in one PR. Maybe to remove some of the concerns we could have some more info on when a follow up PR can be done to remove the else/ non pure path.

Could you comment on this @thomas-zeng ?

thomas-zeng:
As discussed today between Felipe, Raman, Thomas. We decide to not move forward with this PR.
    - PR has if-else branch for XR, non-XR code. Added code maintenance cost
    - Pure XRSDK PR(currently in drafting stage) can have everything fixed in one go, and it is a superset of the URP Camera Matrices PR(this PR)

We decide to move forward with pure XRSDK PR, a superset of this PR, and land everything in one go.
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5672

Thanks everyone for taking your time reviewing/testing this PR.

PR #4871, opened 2019-10-23
Removal of additional camera data fix

martint-unity:
fix case 1189926
https://fogbugz.unity3d.com/f/cases/1189926/

Made it so that when deleting the camera it also deletes the additional camera component.
The additional camera component cannot be deleted on its own.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Ffix-removal-of-component/.yamato%252Fupm-ci-universal.yml%2523All_Universal/622105/job


com.unity.render-pipelines.universal/Editor/UniversalAdditionalCameraDataEditor.cs
```diff
@@ -10,12 +10,11 @@ class UniversalAdditionalCameraDataEditor : Editor
         public override void OnInspectorGUI()
         {
         }
-        [MenuItem("CONTEXT/LWRPAdditionalCameraData/Remove Component")]
+        [MenuItem("CONTEXT/UniversalAdditionalCameraData/Remove Component")]
         static void RemoveComponent(MenuCommand command)
         {
-            if (EditorUtility.DisplayDialog("Remove Component?", "Are you sure you want to remove this component? If you do, you will lose some settings.", "Remove", "Cancel"))
+            if (EditorUtility.DisplayDialog("Component Info", "You can not delete this component, you will have to remove the camera.", "OK"))
```
phi-lira: you dont' need the if () clause. Just call EditorUtility.DisplayDialog.

PR #4874, opened 2019-10-23
Fix triplanar node using camera relative position in HDRP

alelievr:
### Purpose of this PR
Fix triplanar node using camera relative position in HDRP

- [ ] Built a player
- Other: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/shadergraph%252Ffix%252Ftriplanar_node_absolute_world/.yamato%252Fupm-ci-abv.yml%2523all_project_ci


sebastienlagarde:
> has any testing been done on how this impacts upgraded nodes that were previously created? usually changing the default slot value has unexpected behavior when upgrading.

No upgrade testing. 
I will expect that when we upgrade the previous value remain, right?
In case it is not the case what is the process / script to do for upgrade? From what I understand there is no upgrade mechanism for shader graph for such a context?

alindmanUnity:
I mostly just want to double check what the actual behavior is so that we can try not to break any existing graphs, or to make sure we have good messaging to users about how the behavior in their graphs might change. This one shouldn't be too bad of a change though

alelievr:
Indeed, the old required world space slot was still here and not taken in account for the node requirements, I added both world and absolute world in space requirements for the node. So the behavior doesn't change

PR #4875, opened 2019-10-23
Case 1176497: Fix shader error spammed in the console when selecting terrain in HDRP/SRP

zeroyao:
---
### Purpose of this PR
Fix case 1176497 by adding HDRP/URP scene selection passes to terrain shaders.

---
### Testing status

- [ ] Built a player
- Other: 

None.

https://yamato.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/fix-terrainlit-srp-compatibility
---


zeroyao:
Changelogs updated for both HDRP and URP.

LandonTownsendUnity:
The bug report lacks repro steps, can you let me know how I can verify that this was fixed?

zeroyao:
> The bug report lacks repro steps, can you let me know how I can verify that this was fixed?

Try:

1. Create a terrain.
2. Click on it.
3. Go to "Terrain Settings" tab by clicking the gear icon.
4. Double click on the Material. In HDRP's case it's "DefaultHDTerrainMaterial".
5. In the new inspector window, right click on the title section, and select "Select Shader" (the whole inspector window is supposed to be disabled but doesn't matter). In HDRP's case it's "HDRP/TerrainLit".
6. If the bug is not fixed, you are supposed to see "Builtin cbuffer size is not the same for all stages ()" message after the "SRP Batcher" row.
7. If the bug is fixed, the message doesn't show.

Please verify it in both HDRP and URP (Universal should be using a different default terrain material and terrain shader).

zeroyao:
> The error seems to still be there for me, in both HDRP and Universal RP:
> image

Confirmed the message is gone with the Editor built from trunk version graphics/case-1176497:
* After double clicking the terrain Material field, the opening Inspector window is not necessarily disabled.
* On URP there will be another message reads `Material property is found in another cbuffer than "UnityPerMaterial" (_Metallic0)` replacing the old one. It would be best if we can open another bug report on it.

sebastienlagarde:
@phi-lira could you confirm it is ok for you? thanks

sebastienlagarde:
@zeroyao Yamato was red when this PR have been created, could you merge master and run Yamato again? Yamato is now green, except well identified test on universal. Thanks

zeroyao:
According to @sebastienlagarde the Yamato tests now have the same greenness with master.

phi-lira:
What is the other error @LandonTownsendUnity is mentioning?

PR #4896, opened 2019-10-25
Update all package min version to 2020.1.a11

sebastienlagarde:
Update all package min version to 2020.1.a11

This is required by a C++ change as otherwise we get error: 
error CS1061: 'ListView' does not contain a definition for 'onSelectionChange'


sionzeecz:
Is this mean that this error will be here till Unity 2020.1.a11 will be released? 
I'm getting this error with 2020.1.a9 (latest public version). 

Actually getting one more:
`com.unity.render-pipelines.core\Editor\LookDev\DisplayWindow.EnvironmentLibrarySidePanel.cs(89,31): error CS1061: 'ListView' does not contain a definition for 'onSelectionChange' and no accessible extension method 'onSelectionChange' accepting a first argument of type 'ListView' could be found (are you missing a using directive or an assembly reference?)`

`com.unity.render-pipelines.core\Editor\LookDev\DisplayWindow.EnvironmentLibrarySidePanel.cs(120,31): error CS1061: 'ListView' does not contain a definition for 'onItemsChosen' and no accessible extension method 'onItemsChosen' accepting a first argument of type 'ListView' could be found (are you missing a using directive or an assembly reference?)`

PR #4899, opened 2019-10-28
Basic MRT support in Universal RP

Wilfrid-Unity:
This pull requests contains cherry picks of the commits currently found in branch [universal/deferred](https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/universal/deferred) to support Multiple Render Target passes.

This is for example currently used by the GBuffer pass in branch [universal/deferred](https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/universal/deferred).

They have been tested from the [universal/deferred](https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/universal/deferred) branch on a limited subset of the [graphics test scenes](https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/universal/deferred/TestProjects/UniversalGraphicsTest/Assets/Scenes).

It might be interesting to make these changes available in master already.

Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fmrt


com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -360,51 +363,115 @@ internal void Clear()
...
             renderPass.Configure(cmd, renderingData.cameraData.cameraTargetDescriptor);
 
-            RenderTargetIdentifier passColorAttachment = renderPass.colorAttachment;
-            RenderTargetIdentifier passDepthAttachment = renderPass.depthAttachment;
-            ref CameraData cameraData = ref renderingData.cameraData;
-
-            // When render pass doesn't call ConfigureTarget we assume it's expected to render to camera target
-            // which might be backbuffer or the framebuffer render textures.
-            if (!renderPass.overrideCameraTarget)
+            if(IsMRT(renderPass.colorAttachment))
```
phi-lira: Potential issue: Stereo code is not called in MRT clause. I don't have enough VR/Stereo knowledge to properly suggest it. Adding @thomas-zeng as reviewer.

@Wilfrid-Unity Could you test with MockHMD if stereo in editor is working?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -95,6 +107,17 @@ public void ConfigureTarget(RenderTargetIdentifier colorAttachment, RenderTarget
...
+            ConfigureTarget(new RenderTargetIdentifier[]{ colorAttachment });
+        }
+
+        /// <summary>
+        /// Configures render targets for this render pass. Call this instead of CommandBuffer.SetRenderTarget.
+        /// This method should be called inside Configure.
+        /// </summary>
+        /// <param name="colorAttachment">Color attachment identifier.</param>
+        /// <seealso cref="Configure"/>
+        public void ConfigureTarget(RenderTargetIdentifier[] colorAttachment)
```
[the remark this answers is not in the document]
hkr: `SystemInfo.supportedRenderTargetCount`

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -427,7 +494,7 @@ void EndXRRendering(ScriptableRenderContext context, Camera camera)
 
         internal static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier colorAttachment, RenderTargetIdentifier depthAttachment, ClearFlag clearFlag, Color clearColor)
         {
-            m_ActiveColorAttachment = colorAttachment;
+            m_ActiveColorAttachment = new RenderTargetIdentifier[]{colorAttachment};
```
hkr: Can we avoid this allocation?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -484,6 +551,14 @@ internal static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier c
             }
         }
 
+        static void SetRenderTarget(CommandBuffer cmd, RenderTargetIdentifier[] colorAttachment, RenderTargetIdentifier depthAttachment, ClearFlag clearFlag, Color clearColor)
+        {
+            m_ActiveColorAttachment = colorAttachment;
+            m_ActiveDepthAttachment = depthAttachment;
+
```
hkr: This should handle load/store actions similar to other overloads

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
hkr: Maybe `m_ActiveColorAttachment` should be fixed size with separate length and make deep copy in MRT path?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
hkr: Comment why MRT path is separate would be helpful

thomas-zeng: Stereo code clause is needed to get the current XR working. So they should be called in MRT clause.
Note: XR Graphics is working on a Pure XR+URP solution and in that solution, we can remove specially XR calls. But for now, we will need those special XR calls(StartMultiEye & DrawOcclusionMesh).

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Yes, however there is no version of CommandBuffer.SetRenderTarget supporting this combination yet.
https://docs.unity3d.com/2020.1/Documentation/ScriptReference/Rendering.CommandBuffer.SetRenderTarget.html

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -35,7 +35,7 @@ namespace UnityEngine.Rendering.Universal
     {
         public RenderPassEvent renderPassEvent { get; set; }
 
-        public RenderTargetIdentifier colorAttachment
```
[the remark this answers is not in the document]
Wilfrid-Unity: OK, to avoid breaking client code I restored that member in revision d882acc

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -35,7 +35,7 @@ namespace UnityEngine.Rendering.Universal
...
+        public RenderTargetIdentifier[] colorAttachment
```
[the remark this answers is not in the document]
Wilfrid-Unity: Yes, I pushed commit 573921b that has both previous colorAttachment and new colorAttachments member

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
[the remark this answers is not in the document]
Wilfrid-Unity: Thanks, I added a Debug.LogError in that case, with commit b59753f

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -99,14 +99,14 @@ static class RenderPassBlock
         const string k_SetRenderTarget = "Set RenderTarget";
         const string k_ReleaseResourcesTag = "Release Resources";
 
-        static RenderTargetIdentifier m_ActiveColorAttachment;
+        static RenderTargetIdentifier[] m_ActiveColorAttachment;
```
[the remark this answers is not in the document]
Wilfrid-Unity: Thanks, I updated the name in 573921b

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: I added code for stereo similarly to the non-MRT code path, in commit 942d429
I also tried to enable XR and use the MockHMD, and it looked OK (i.e left and right sides render similarly).
Does that sound OK @thomas-zeng ?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Thanks for the suggestion @hkr I removed a few occurrences of "new" in commit 5c4d39d1

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
phi-lira: SetRenderTarget supports RenderTargetBinding (which in turn supports color load/store actions) https://docs.unity3d.com/ScriptReference/Rendering.RenderTargetBinding.html

However I'd like to start a discussion on this as we need to think RenderPass way for future. I'll discuss in slack.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -360,51 +369,130 @@ internal void Clear()
...
+                    Camera camera = cameraData.camera;
+                    ClearFlag clearFlag = GetCameraClearFlag(camera.clearFlags);
+
+                    // Overlay cameras composite on top of previous ones. They don't clear.
+                    // MTT: Commented due to not implemented yet
+//                    if (renderingData.cameraData.renderType == CameraRenderType.Overlay)
+//                        clearFlag = ClearFlag.None;
+
+                    //SetRenderTarget(cmd, renderPass.colorAttachment, m_CameraDepthTarget, clearFlag,
+                    SetRenderTarget(cmd, renderPass.colorAttachments, renderPass.depthAttachment, clearFlag,
```
thomas-zeng: Here we clear all colorAttachments based on camera clear flag. Maybe we should only clear camera target in this case? m_FirstCameraRenderPassExecuted bool name also suggests clear logic here is camera target only.

Wilfrid-Unity: @thomas-zeng  Do you mean replacing above line by something like that?
```
SetRenderTarget(cmd, renderPass.colorAttachments[0], renderPass.depthAttachment);
ClearRenderTarget(cmd, clearFlag, CoreUtils.ConvertSRGBToActiveColorSpace(camera.backgroundColor));
SetRenderTarget(cmd, renderPass.colorAttachments, renderPass.depthAttachment);
```

thomas-zeng: I don't have a preference because I feel I might need to better understand the logic here.
It is just unclear to me why we use camera clear flag to clear render targets that might not be camera target.
Let's say here we have a render targets array that contains camera rt and couple other rts, and camera.clearflag is set to don't clear. Logic here seems to skip clear not only for camera rt but also for other rts?

Wilfrid-Unity: Yes.

This change was originally introduced in 92bb414c2
It allowed in subsequent revision 291a9efa24 to remove @kaychang-unity 's temporary code + comment from 18d1730595af "Temporary: m_LightingTexture is not yet part of the GBuffer, so we must clear it before accumulating lighting results into it."

This code path is currently only used to set GBuffer RenderTargets. 
With that change we clear all GBuffer RenderTargets (including the Lighting Buffer) at that timing.

@thomas-zeng 
Would you prefer more granularity regarding the Clear options? i.e support per-RenderTarget clear parameters in Camera.clearFlags?
Or do you think it is a potentially dangerous side-effect to apply the Camera.clearFlags to all RenderTargets during ExecuteRenderPass? in that case we could revert it and just add back the explicit clear from 18d1730

thomas-zeng: @phi-lira what's the difference between camera.clearflag and renderpass.clearflag? in case they conflict(renderpass set target to cameraTarget and configure clear flag differently than camera.clearflag), which one should we respect?

thomas-zeng: I think targets other than cameraTarget should use renderpass.clearFlag instead of camera.clearFlag here.

Wilfrid-Unity: @thomas-zeng  Thanks for the help !

After chatting with @kaychang-unity it turns out that I misunderstood comment from 18d1730 ; GBuffer / LightingBuffer should not be cleared since all their pixels should be rewritten by geometry or the Skybox.

I pushed commit a1b2275 that implements your suggestion:

> targets other than cameraTarget should use renderpass.clearFlag instead of camera.clearFlag

I checked this change with the universal/deferred branch and test scene #008 , and no unnecessary Clears were present in a standalone player (only depthBuffer is cleared).

kaychang-unity: My suggestion about this:
At the moment, the code applies camera.clearFlag when m_CameraColorTarget is first bound. However, camera.clearFlag may also contain Clearflag.Depth. Deciding the apply the clear of the depth buffer at the moment m_CameraColorTarget is bound the first time is NOT GOOD, because in our deferred renderer scenario, we render depth-prepass first, then bind m_CameraColorTarget later as part of the gbuffer config. This has the effect of wiping out the depth-prepass when we bind the gbuffer... ouch!
Instead, I suggest we separately track the first time m_CameraColorTarget and m_CameraDepthTarget are bound, then use camera.clearFlag on them individually when that happens.
If m_CameraColorTarget is bound the first time as part of a gbuffer config, then I propose we still clear it using camera.clearFlag, while the other color RTs are cleared using renderpass.clearFlag.

phi-lira: @kaychang-unity solution is good for me. to confirm understanding: we track first time rendering to CameraColor and CameraDepth separately, and we allow to have ClearFlag that can be (camera.clearFlag.Color | renderPass.clearFlag.Depth) (or vice-versa, in case of depth pre-pass).

thomas-zeng: I think @kaychang-unity solution is good for me. I like the idea to track first-time color/depth separately. 

Another thought that is not related to this PR: is it possible we do clear logic and set rt logic in render passes compiling phase where we have full picture of the frame? (URP has a render passes sorting function that is doing some pre-execution work) 
In the execution phase, we are taking in render passes iteratively and never know what comes next. In the compiling phase, we have full picture of the frame.

kaychang-unity: @thomas-zeng I didn't know there was a compiling phase for the renderpasses, so my opinion may not account to much...
But I don't think it is possible to decide the clear logic for each RTs even when you have the full picture of the frame. The problem is that for some RTs, the user may want them to be cleared before their use (RenderBufferLoadAction.Clear), while for some other, the RenderBufferLoadAction.DontCare is good enough (ex: gbuffer color RTs don't need to be cleared because geometry + skybox overwrite all pixels). Then persistent RTs accross frames (usually use for temporal effects) need RenderBufferLoadAction.Load. But the full picture of the frame will not give you that information. You would need the user to inject that contextual information somehow.
What I heard being done though during this kind of "compilation phase", is optimal allocation/deallocation of temporary RTs.

thomas-zeng: @kaychang-unity Yup, compiling phase could analyze the frame and develop temp RTs strategy. I think compiling phase should also build GPU work's dependency graph for explicit Graphics API(DX12, Metal2 and Vulkan). Right now we are not doing that in URP. In XR we were getting couple requests for such thing mostly because our XR partners are on explicit API(oculus, msft, magicleap..). So this is definitely something XR graphics is looking for.

Compiling phase happens after render pass setup phase and before execution phase. In URP case, clear flag is configured in setup phase so during compiling phase we know the clear info for all passes. Compiling phase could then inject clear and setRT logic on a per pass basis that will be executed in execution phase.
This is beneficial because maybe one day we want to execute render passes using multiple threads or jobs(eg. dots renderer world), it is important passes are self-contained and can build their own cmd buffer simultaneously.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -84,7 +89,30 @@ public ScriptableRenderPass()
...
+        }
+
+        /// <summary>
+        /// Configures render targets for this render pass. Call this instead of CommandBuffer.SetRenderTarget.
+        /// This method should be called inside Configure.
+        /// </summary>
+        /// <param name="colorAttachment">Color attachment identifier.</param>
+        /// <param name="depthAttachment">Depth attachment identifier.</param>
+        /// <seealso cref="Configure"/>
+        public void ConfigureTarget(RenderTargetIdentifier[] colorAttachments, RenderTargetIdentifier depthAttachment)
```
phi-lira: We should handle case users pass `null` as color colorAttachments.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -97,7 +125,27 @@ public void ConfigureTarget(RenderTargetIdentifier colorAttachment, RenderTarget
...
+        /// </summary>
+        /// <param name="colorAttachment">Color attachment identifier.</param>
+        /// <seealso cref="Configure"/>
+        public void ConfigureTarget(RenderTargetIdentifier[] colorAttachments)
+        {
+            overrideCameraTarget = true;
+
+            uint nonNullColorBuffers = ScriptableRenderer.NonNullColorBuffersCount(colorAttachments);
+            if( nonNullColorBuffers > SystemInfo.supportedRenderTargetCount)
+                Debug.LogError("Trying to set " + nonNullColorBuffers + "renderTargets, which is more than the maximum supported:" + SystemInfo.supportedRenderTargetCount);
```
phi-lira: minor: insert space at "renderTargers

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -84,7 +89,30 @@ public ScriptableRenderPass()
...
+        /// <param name="colorAttachment">Color attachment identifier.</param>
+        /// <param name="depthAttachment">Depth attachment identifier.</param>
+        /// <seealso cref="Configure"/>
+        public void ConfigureTarget(RenderTargetIdentifier[] colorAttachments, RenderTargetIdentifier depthAttachment)
+        {
+            overrideCameraTarget = true;
+
+            uint nonNullColorBuffers = ScriptableRenderer.NonNullColorBuffersCount(colorAttachments);
+            if( nonNullColorBuffers > SystemInfo.supportedRenderTargetCount)
+                Debug.LogError("Trying to set " + nonNullColorBuffers + "renderTargets, which is more than the maximum supported:" + SystemInfo.supportedRenderTargetCount);
```
phi-lira: minor: insert space at "renderTargets

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -360,51 +369,153 @@ internal void Clear()
                 context.Submit();
         }
 
+        public static uint NonNullColorBuffersCount(RenderTargetIdentifier[] colorBuffers)
```
phi-lira: Suggestion: `GetValidColorBufferCount`
Add static rendering utility functions to RenderingUtils.cs

phi-lira: Please add API docs to new public API.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -84,7 +89,30 @@ public ScriptableRenderPass()
         public void ConfigureTarget(RenderTargetIdentifier colorAttachment, RenderTargetIdentifier depthAttachment)
         {
             overrideCameraTarget = true;
-            m_ColorAttachment = colorAttachment;
+
+            m_ColorAttachments[0] = colorAttachment;
+            for (int i = 1; i < m_ColorAttachments.Length; ++i)
+                m_ColorAttachments[i] = 0;
```
phi-lira: This piece of code is duplicated in `ConfigureTarget(RenderTargetIdentified colorAttachment)`. Consider calling that funciton here instead to avoid code duplication and improve maintanability.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -360,51 +369,153 @@ internal void Clear()
...
+//                        clearFlag = ClearFlag.None;
 
-            if (passColorAttachment == m_CameraColorTarget && !m_FirstCameraRenderPassExecuted)
+                    // apply "clearFlag" to m_CameraColorTarget (if it is not included by renderPass.clearFlag)
+                    if(!Contains(renderPass.clearFlag, clearFlag))
+                        SetRenderTarget(cmd, renderPass.colorAttachments[cameraTargetColorIndex], renderPass.depthAttachment, clearFlag,
+                            CoreUtils.ConvertSRGBToActiveColorSpace(camera.backgroundColor));
+
+                    //SetRenderTarget(cmd, renderPass.colorAttachment, m_CameraDepthTarget, clearFlag,
+                    SetRenderTarget(cmd, renderPass.colorAttachments, renderPass.depthAttachment, renderPass.clearFlag,
```
phi-lira: This SetRenderTarget is overriding the previous one above.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Thanks for all the feedback.

I started implementing the solution suggested in a separate test branch because there are still points not clear to me.
Please take a look at commit a15a73b.

Some questions:

1) In the single-renderTarget code path, which test should be used to check that m_CameraDepthTarget is being bound?
if (passColorAttachment == m_CameraDepthTarget && m_FirstTimeCameraDepthTargetIsBound)
or
if (passDepthAttachment == m_CameraDepthTarget && m_FirstTimeCameraDepthTargetIsBound)
(or both?)


2) @phi-lira About this comment
`we allow to have ClearFlag that can be (camera.clearFlag.Color | renderPass.clearFlag.Depth) (or vice-versa, in case of depth pre-pass).`
Do you mean that the value that should be passed has the "clearFlag" parameter to function SetRenderTarget is called the first time m_CameraColorTarget/m_CameraDepthTarget is bound should be (camera.clearFlag.Color | renderPass.clearFlag.Depth)?
In that case what do you mean by "or vice-versa"?


3) Regarding the part "clear m_CameraColorTarget using camera.clearFlag and clear other render targets of the MRT setup using another renderpass.clearFlag", I am not sure which API I can use to do that.
Should I add 2 separate clear passes (one for m_CameraColorTarget and one for the other targets) to the code block, and then only call SetRenderTarget (without clear flags) with the MRT set-up?
If you have an example of implementation that would be great.


4) Another concern: could it happen that some user depthTargets are configured as ScriptableRenderPass.colorAttachment? Could it be the case with m_CameraDepthTarget?


5) @kaychang-unity about comment:
`we render depth-prepass first, then bind m_CameraColorTarget later as part of the gbuffer config. This has the effect of wiping out the depth-prepass when we bind the gbuffer`
I was not able to reproduce this case (what I did: locally rebase branch universal/mrt on top of universal/deferred, and execute test-scene#008-AdditionalLights in Windows Editor and on Console).
Could you please give a concrete example of this happening (which test scene + settings changed should I use to reproduce and test this case)? 



6) Also if there are test scenes that I could use to verify the changes in this PR (including all the code paths involved regarding the MRT support in function ExecuteRenderPass), please tell me, that should be more straightforward than only reasoning about the code.

kaychang-unity: 5. I messaged you a workaround to get depth-prepass running in test-scene#008-AdditionalLights (enable shadows on directional lights).

I'm having a look at the code in more details now. Since neither Wilfrid nor me are knowledgeable enough about this code base, it is indeed difficult to get it right. I do have questions on several locations in the code which I cannot understand theirs purpose/why they are written in that way.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
Wilfrid-Unity: Thanks, I factorized some code in 0380ec07f

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
Wilfrid-Unity: In commit 0380ec07f I added the null check to function RenderingUtils.GetValidColorBufferCount, a null array will trigger  the `Debug.LogError` in this function.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
Wilfrid-Unity: Fixed in 0380ec0

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
Wilfrid-Unity: Fixed in 0380ec0 too

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Thanks, I renamed and moved the function as you suggested.
I added ///<sumary<>... documentation to the APIs, is it enough or do other documents need to be updated?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: This code will change, once we find how to solve the issues discussed in https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899#discussion_r342490642

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: I added to the temp branch **mrt-clearflag-test** the changes @kaychang-unity sent me (based on the discussion https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/942d42946befc1acc9c049ace8ab9e7d1e866c7d#r35822502).

So the latest candidate fix for this issue is here: bdd63fc
How does this version of ExecuteRenderPass look to you all? (I could not do extensive testing yet)

Wilfrid-Unity: I have merged the solution provided by @kaychang-unity in branch **universal/mrt-clearflag-test**, into this branch **universal/mrt** ; this should fix the issues discussed in this thread.

@thomas-zeng Does everything look OK to you now in this branch **universal/mrt**?

thomas-zeng: I think the logic looks good. Any chance we can combine if-else block to one code block? Now I feel there are lots of duplicated code and it's unclear why we have if-else for mrt and non-mrt.

kaychang-unity: @thomas-zeng Not sure which if-else block you refers too. Could you comment on the latest iteration of the file?

Wilfrid-Unity: @thomas-zeng Yes please, the latest iteration of function **ScriptableRenderer.ExecuteRenderPass** is [here](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/70afa21a13771f0767954ec0e5b3e0bb1a2efbcd/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L376).

The MRT and non-MRT code paths are quite different:
- In the MRT case we make the assumption that all renderPass.colorAttachments refer to actual Color targets (and thus do not require ClearFlag.Depth), which keeps the code a bit simpler
- In the non-MRT case we maintain the support for cases where m_CameraColorTarget is actually a Depth target. In that case, ClearFlag.Depth will be applied to m_CameraColorTarget (but not to m_CameraDepthTarget since in that case it cannot be set as renderPass.depthAttachment, is that correct @kaychang-unity ? )

kaychang-unity: Yes, in non-MRT case, m_CameraColorTarget can point to a depth surface (ex: depth-prepass), so it must be cleared with ClearFlag.Depth. However, we don't have a use case when m_CameraDepthTarget points to a color surface.
Long term, the renderpass API will have to be fixed to not allow binding depth surfaces as m_CameraColorTarget. Then the logic in ScriptableRenderer.ExecuteRenderPass wouldn't be convoluted as it is now.

Wilfrid-Unity:
I added tests to check the code related to camera/renderPass clearFlags in ScriptableRenderer.ExecuteRenderPass.

They are simple custom renderers that look like this in the Editor / Standalone players, and illustrate that camera.clearColor (green) is picked when cameraTarget is one of the targets in the MRT setup, and that renderPass.clearColor (yellow) is picked otherwise:

Test 105
image

Test 106
image


I added the reference images, but I only have access to Windows.

I also noticed that the screenshots performed during the tests seemed to fail to capture camera.clearColor.
Not sure if there is an issue with the test, or something else:

image

phi-lira:
@Unity-Technologies/gfx-qa-foundation this is a high-priority PR that is needed for Deferred rendering. Could you put some priority to validate it?

 We are looking to validate for any regression only. Validating for regressions in BoatAttack and the UniversalRendererExamples is enough coverage imo.

Wilfrid-Unity:
Hi @mikesfbay , this pull request contains a [refactoring of the timing of call to context.StartMultiEye during ScriptableRenderer.ExecuteRenderPass](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899/files#diff-fdb79535867edca9e61fb34f63128b92R584), that @kaychang-unity and @thomas-zeng previously discussed [here](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/942d42946befc1acc9c049ace8ab9e7d1e866c7d?diff=unified)

I see that you modified this area of code [recently](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5061/files#diff-fdb79535867edca9e61fb34f63128b92R403). Do the changes in this PR look OK to you?

The tests in https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/master/TestProjects/UniversalGfxTestStereo are passing (I checked locally and am now again waiting for C.I results).
Do you further have a more complex scene that you would like to test, to check that no regression is introduced by this PR?
(I can personally only test in the Windows Editor).

mikesfbay:
@Wilfrid-Unity Sure, I'll go over your changes soon. We may need our XR QA to do a pass on this.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +414,184 @@ internal void Clear()
...
+                if (passColorAttachment != m_ActiveColorAttachments[0] || passDepthAttachment != m_ActiveDepthAttachment || finalClearFlag != ClearFlag.None)
+                    SetRenderTarget(cmd, passColorAttachment, passDepthAttachment, finalClearFlag, finalClearColor);
+            }
+
+            // We must execute the commands recorded at this point because potential call to context.StartMultiEye(cameraData.camera) below will alter internal renderer states
+            // Also, we execute the commands recorded at this point to ensure SetRenderTarget is called before RenderPass.Execute
             context.ExecuteCommandBuffer(cmd);
             CommandBufferPool.Release(cmd);
 
+            if (firstTimeStereo && cameraData.isStereoEnabled && (m_FirstTimeStereoTargetIsBound || (cameraData.isXRMultipass && m_XRRenderTargetNeedsClear)))
```
mikesfbay: The test (cameraData.isXRMultipass && m_XRRenderTargetNeedsClear) is in the wrong place. Compare where it is in master branch to see the missing clear. See the forwarded screenshots of what went wrong during VR when using your branch.

Wilfrid-Unity:
@mikesfbay 
I added to this branch a test scene that reproduced the issue you were mentioning (see revision 5f70c3f). I am not sure how to make the TestRunner switch to XR Multipass automatically, so for now I did not add the scene to the build settings.
I added a fix (revision 8bd5011) that fixed this issue ; I confirmed locally that the left eye texture was cleared before rendering the right eye texture.
Could you please check again if everything looks OK for your cases?


@caryhaynie 
It looks like your pull request https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4397 might cover similar cases.
Please could you confirm if the [current version of ScriptbleRenderer.ExecuteRenderPass in this universal/mrt branch](https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/8bd501137918d67a400e36ece4b149b12a5b203e/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L523) covers the issues your PR attempts to fix?
Otherwise could you please provide a test/repro scene illustrating the issue that needs to be cleared?


@kaychang-unity 
I did not port above changes to the MRT code path yet ; and did not check yet the case of depth re-bound after pre-pass ; does something look incorrect in the last change from 8bd5011? One thing I was wondering is if there should be two separate flags for m_XRRenderTargetNeedsClear (like there are two separate flags m_FirstTimeCameraColorTargetIsBound/m_FirstTimeCameraDepthTargetIsBound )?

Wilfrid-Unity:
> I think the non-MRT path looks fine. If it passes unit tests, it should be good.
> Does applying the same logic for the MRT pass causes some issue in the MRT test scene?

Unit tests for "All Universal CI" and "All Universal_Stereo CI" are passing:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fmrt/.yamato%252Fupm-ci-universal.yml%2523All_Universal_trunk/867679
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fmrt/.yamato%252Fupm-ci-universal_stereo.yml%2523All_Universal_Stereo_trunk/867683

And MRT test scene#105 is still OK.


However I just noticed that when using UnityEditor revision 1b24727fa96d MRT test scene#106 looks broken locally ; but this is not related to the recent changes in this branch (it also happens when opening revision b4b985c8). I'm investigating what's happening there.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
-                cmd.Clear();
+                // Note: if we have to give up the assumption that no depthTarget can be included in the MRT colorAttachments, we might need something like this:
+                // int cameraTargetDepthIndex = IndexOf(renderPass.colorAttachments, m_CameraDepthTarget);
+                // if( !renderTargetAlreadySet && cameraTargetDepthIndex != -1 && m_FirstTimeCameraDepthTargetIsBound)
+                // { ...
+                // }
 
-                if (cameraData.isStereoEnabled)
+                if (renderPass.depthAttachment == m_CameraDepthTarget && m_FirstTimeCameraDepthTargetIsBound)
+                // note: should be split m_XRRenderTargetNeedsClear into m_XRColorTargetNeedsClear and m_XRDepthTargetNeedsClear and use m_XRDepthTargetNeedsClear here?
```
mikesfbay: For multipass mode, if m_XRRenderTargetNeedsClear == true, then both color and depth buffer need clearing (not just color). Thus, you need to change the depth clear logic here to reflect this. 

For comparison, in the multipass PR, the XR clear was for both color and depth, thus this PR should preserve that.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
+                    //                        clearFlag = ClearFlag.None;
 
-                SetRenderTarget(cmd, m_CameraColorTarget, m_CameraDepthTarget, clearFlag,
-                    CoreUtils.ConvertSRGBToActiveColorSpace(camera.backgroundColor));
+                    // We need to specifically clear the camera color target.
+                    // But there is still a chance we don't need to issue individual clear() on each render-targets if they all have the same clear parameters.
+                    needCustomCameraColorClear = (cameraClearFlag & ClearFlag.Color) != (renderPass.clearFlag & ClearFlag.Color)
+                                                || CoreUtils.ConvertSRGBToActiveColorSpace(camera.backgroundColor) != renderPass.clearColor;
+
+                    m_XRRenderTargetNeedsClear = false; // register that the XR camera multi-pass target does not need clear any more (until next call to BeginXRRendering)
```
mikesfbay: Move this line down to the depth section below.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
+                if (renderPass.depthAttachment == m_CameraDepthTarget && m_FirstTimeCameraDepthTargetIsBound)
+                // note: should be split m_XRRenderTargetNeedsClear into m_XRColorTargetNeedsClear and m_XRDepthTargetNeedsClear and use m_XRDepthTargetNeedsClear here?
                 {
-                    context.StartMultiEye(cameraData.camera, eyeIndex);
-                    XRUtils.DrawOcclusionMesh(cmd, cameraData.camera);
+                    m_FirstTimeCameraDepthTargetIsBound = false;
+                    //firstTimeStereo = true; // <- we do not call this here as the first render pass might be a shadow pass (non-stereo)
+                    needCustomCameraDepthClear = (cameraClearFlag & ClearFlag.Depth) != (renderPass.clearFlag & ClearFlag.Depth);
+
+                    //m_XRRenderTargetNeedsClear = false; // note: is it possible that XR camera multi-pass target gets clear first when bound as depth target?
```
mikesfbay: Uncomment this line (see above reason).

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
+                    m_XRRenderTargetNeedsClear = false; // register that the XR camera multi-pass target does not need clear any more (until next call to BeginXRRendering)
+                }
+                else
+                {
+                    finalClearFlag |= (renderPass.clearFlag & ClearFlag.Color);
+                    finalClearColor = renderPass.clearColor;
+                }
+
+                if (passDepthAttachment == m_CameraDepthTarget && m_FirstTimeCameraDepthTargetIsBound )
+                // note: should be split m_XRRenderTargetNeedsClear into m_XRColorTargetNeedsClear and m_XRDepthTargetNeedsClear and use m_XRDepthTargetNeedsClear here?
```
mikesfbay: For multipass mode, if m_XRRenderTargetNeedsClear == true, then both color and depth buffer need clearing (not just color). Thus, you need to change the depth clear logic here to reflect this.

For comparison, in the multipass PR, the XR clear was for both color and depth, thus this PR should preserve that.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
+                {
+                    m_FirstTimeCameraDepthTargetIsBound = false;
+
+                    finalClearFlag |= (cameraClearFlag & ClearFlag.Depth);
+
+                    //firstTimeStereo = true; // <- we do not call this here as the first render pass might be a shadow pass (non-stereo)
+
+                    // finalClearFlag |= (cameraClearFlag & ClearFlag.Color);  // <- m_CameraDepthTarget is never a color-surface, so no need to add this here.
+
+                    //m_XRRenderTargetNeedsClear = false; // note: is it possible that XR camera multi-pass target gets clear first when bound as depth target?
```
mikesfbay: Uncomment this line (see above reason).

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,193 @@ internal void Clear()
...
+                    finalClearFlag |= (cameraClearFlag & ClearFlag.Color);
+                    finalClearColor = CoreUtils.ConvertSRGBToActiveColorSpace(camera.backgroundColor);
+                    firstTimeStereo = true;
+
+                    // m_CameraColorTarget can be an opaque pointer to a RenderTexture with depth-surface.
+                    // We cannot infer this information here, so we must assume both camera color and depth are first-time bound here (this is the legacy behaviour).
+                    m_FirstTimeCameraDepthTargetIsBound = false;
+                    finalClearFlag |= (cameraClearFlag & ClearFlag.Depth);
+
+                    m_XRRenderTargetNeedsClear = false; // register that the XR camera multi-pass target does not need clear any more (until next call to BeginXRRendering)
```
mikesfbay: Move this line down to the depth section below.

mikesfbay:
Looping in our XR QA.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Thank you. I addressed this issue in revision e5036314 by also applying the depth clear flag if m_XRRenderTargetNeedsClear == true

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Revision e5036314 fixes this case.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Revision e5036314 fixes this case.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: As commented in 
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/e503631426b70349c0c542e618a5afe91f5974ba/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L560 , depth flag is already applied when m_CameraColorTarget and m_XRRenderTargetNeedsClear == true, therefore this case is already covered.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: Thank you, I checked that this case is covered, see explanation https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899/files#r356460177

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
Wilfrid-Unity: I tested that this case is covered, see explanation https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899/files#r356460177

mikesfbay:
@Wilfrid-Unity I've confirmed that multipass color/depth buffers are cleared from renderdoc. One thing is that in Unity Debug, I'm seeing some texture surface asserts during SetRenderTarget of graphics device worker (see screenshots on slack). This happens on exiting Editor playmode with VR modes.

Wilfrid-Unity:
> @Wilfrid-Unity I've confirmed that multipass color/depth buffers are cleared from renderdoc. One thing is that in Unity Debug, I'm seeing some texture surface asserts during SetRenderTarget of graphics device worker (see screenshots on slack). This happens on exiting Editor playmode with VR modes.

Thanks for having checked the multipass mode.
It seems that when built in Debug mode, UnityEditor asserts many more messages than the actual Release version ; do you have more detailed repro steps about that? I tried to run the full UniversalGfxTestStereo tests suite locally from Editor revision 1b24727fa96d ; and did not get that message.

DennisDeRykeUnity:
@Wilfrid-Unity The issue that I reported occurs without XR enabled, or on Rift using Single Pass Instancing, but not on Rift using Multi Pass Stereo.  Hopefully that can help you narrow down the broken code branch.

Wilfrid-Unity:
> Portions of the geometry in Game view clears to blue as you move. 
> Using a built Standalone player (...) the background does not clear so after-images of the geometry soon cover the entire viewport.

Thanks for the detailed repro steps Dennis.
I'll investigate how to avoid this regression and ensure the scene looks the same as with 7.1.6 / master branch.

DennisDeRykeUnity:
Here is my test matrix for this PR, showing what I have tested so far and what else I plan to test.  So far my tests have been on Rift but I'm moving to Go next.

Wilfrid-Unity:
@DennisDeRykeUnity I added a fix for issue https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899#pullrequestreview-332944287 with revision 441c9d6

I checked locally that the issue did not happen any more, in Windows Editor 2020.1.0a15.1847 (revision c34b21e125d7).

I also tried to add a test scene reproducing the issue (to avoid future regressions). When played in the Editor it showed the incorrect blue pixels (however I noticed that those did not appear in captures performed by automated tests).

Does this fix the "Fail" entries in the test grid https://docs.google.com/spreadsheets/d/1rmnSbcWIC65aHYWlKWQSQtuW-GMUJdVUu2rixIPTztg/edit#gid=0 ?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -384,56 +413,200 @@ internal void Clear()
...
+                }
+                else
+                {
+                    finalClearFlag |= (renderPass.clearFlag & ClearFlag.Color);
+                    finalClearColor = renderPass.clearColor;
+                }
+
+                if ( (m_CameraDepthTarget!=BuiltinRenderTextureType.CameraTarget ) && (passDepthAttachment == m_CameraDepthTarget || passColorAttachment == m_CameraDepthTarget) && m_FirstTimeCameraDepthTargetIsBound)
```
mikesfbay: @Wilfrid-Unity This additional test (m_CameraDepthTarget!=BuiltinRenderTextureType.CameraTarget ) was not present in master branch and master worked in Dennis's tests.  Can you explain why this is needed here?

Wilfrid-Unity: @mikesfbay As mentioned in the commit message and in above comment, this is required to solve [this issue](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899#pullrequestreview-332944287) reported by Dennis yesterday (without this depth buffer is not cleared at the correct timing, causing pixels to remain blue when moving the camera).

mikesfbay: Adding this extra test would imply any non-camera depth target (i.e. render texture targets) would be cleared without checking if it has been cleared already. It may fix Dennis's depth clear issue but isn't this too general?

Wilfrid-Unity: On the opposite, change 441c9d6 adds a more specific condition: `(m_CameraDepthTarget!=BuiltinRenderTextureType.CameraTarget ) && ...` required to clear the camera depth target the first it is bound.

This was required because some passes (Shadow pass, Color Grading Lut pass) can be executed before the scene's actual camera depth target is bound. During these passes `m_CameraDepthTarget` and `passDepthAttachment` are both defined, and set to BuiltinRenderTextureType.CameraTarget. Without condition `(m_CameraDepthTarget!=BuiltinRenderTextureType.CameraTarget ) && ...`, flag `m_FirstTimeCameraDepthTargetIsBound` would be reset during those passes ; so the actual camera depth target would not be cleared during the next pass (Depth Pre-pass or DrawObjects pass).

DennisDeRykeUnity:
I verified that commit 441c9d6 resolves my [issue.](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899#pullrequestreview-332944287) 

Here is my updated test [matrix.](https://docs.google.com/spreadsheets/d/1rmnSbcWIC65aHYWlKWQSQtuW-GMUJdVUu2rixIPTztg/edit#gid=0)

@Wilfrid-Unity Consider this comment an approval from me if: 
- Mike approves
- no additional commits are required 
- Universal CI and Universal Stereo CI are green on Yamato for the latest commit 

But if further commits occur, then please contact me and I will repeat some of my tests with the latest code prior to approving.  Makes sense?

Wilfrid-Unity:
> I verified that commit [441c9d6](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/441c9d659ac9d57c1e0c67b0ca56ba902fec7fe3) resolves my [issue.](https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4899#pullrequestreview-332944287)
> 
> Here is my updated test [matrix.](https://docs.google.com/spreadsheets/d/1rmnSbcWIC65aHYWlKWQSQtuW-GMUJdVUu2rixIPTztg/edit#gid=0)

@DennisDeRykeUnity Thanks for confirming the fix.


> @Wilfrid-Unity Consider this comment an approval from me if:
> 
> * Mike approves
> * no additional commits are required
> * Universal CI and Universal Stereo CI are green on Yamato for the latest commit
> 
> But if further commits occur, then please contact me and I will repeat some of my tests with the latest code prior to approving. Makes sense?

Yes, it makes perfect sense. I will notify you if more commits need to be added to the branch, so that you can repeat those tests.

phi-lira:
I cannot merge if there are requested changes. @DennisDeRykeUnity and @mikesfbay can you approve the PR if not changes are required?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
mikesfbay: @Wilfrid-Unity Ok, can you add a comment above the line so others don't have the same question next time? Something along the lines, "the camera target test prevents non-camera passes from prematurely setting the firstTimeDepthBound flag and causing camera depth not to be cleared"

Wilfrid-Unity: Good point, I added such a comment in 0d6b9b2.

Wilfrid-Unity:
> I cannot merge if there are requested changes. @DennisDeRykeUnity and @mikesfbay can you approve the PR if not changes are required?

@phi-lira All the approvals have been added, could you please merge now?

PR #4902, opened 2019-10-28
Replace constant depth bias with slope-scale bias for shadows

FrancescoC-unity:
Replaced constant depth bias with slope-scale bias for shadows, this simplify our handling of the bias and improve quality throughout. 

Lewis added as I changed the docs. 

Replaced lots of visual tests as shadows are ubiquitous :-) 

NOTE: I also updated the test that currently fails on staging since I had to because of shadows, please let me know if needs reverting.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1650,7 +1622,9 @@ internal int UpdateShadowRequest(HDCamera hdCamera, HDShadowManager manager, Vis
                             }
                             break;
                     }
-                    
+					
+					shadowRequest.slopeBias = HDShadowUtils.GetSlopeBias(slopeBias);
```
sebastienlagarde: ident

PR #4908, opened 2019-10-28
Add specific settings for RTX2080 test projects

RemyUnity:
We need to have DXR graphic tests to check that DXR doesn't break with HDRP or Editor code change.
This is also a merge of the tests made by @anisunity in the "LocalTestProjects" folder and some made by @iM0ve .

Automation is set tu run on Yamato, but we are waiting for an update of the bokken windows image and graphic driver to be able to use DXR raytracing there.

In the meantime, they can at least be run locally, and this will also prevent to have to maintain multiple branches.


com.unity.visualeffectgraph/Editor/Models/Operators/Implementations/VoroNoise2D.cs
```diff
@@ -9,15 +9,15 @@ class VoroNoise2D : VFXOperator
     {
         public class InputProperties
         {
-            [Tooltip("The coordinate in the noise field to take the sample from.")]
+            [Tooltip("Sets the coordinate in the noise field to take the sample from.")]
```
anisunity: This does not below in this PR as well as all similr changes

PR #4911, opened 2019-10-28
Fix lighting attenuation precision for Vulkan on Adreno GPUs

rjonaitis:
### Purpose of this PR
Fixes light attenuation precision for Vulkan running on Android Adreno GPUs
AttenuationPrecision

Previous PR for GLES https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/3172

---
### Testing status
Manually tested on Galaxy Note9 (Adreno 630)
Project attached to fogbugz case: https://issuetracker.unity3d.com/product/unity/issues/guid/1131387/

---

### DIFF
--- a/com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
+++ b/com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
@@ -41,7 +41,7 @@ struct Light
 {
     half3   direction;
     half3   color;
-    half    distanceAttenuation;
+    float   distanceAttenuation;
     half    shadowAttenuation;
 };
 
@@ -142,7 +142,7 @@ Light GetAdditionalPerObjectLight(int perObjectLightIndex, float3 positionWS)
     float distanceSqr = max(dot(lightVector, lightVector), HALF_MIN);
 
     half3 lightDirection = half3(lightVector * rsqrt(distanceSqr));
-    half attenuation = DistanceAttenuation(distanceSqr, distanceAndSpotAttenuation.xy) * AngleAttenuation(spotDirection.xyz, lightDirection, distanceAndSpotAttenuation.zw);
+    float attenuation = DistanceAttenuation(distanceSqr, distanceAndSpotAttenuation.xy) * AngleAttenuation(spotDirection.xyz, lightDirection, distanceAndSpotAttenuation.zw);
 
     Light light;
     light.direction = lightDirection;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -516,9 +516,9 @@ void MixRealtimeAndBakedGI(inout Light light, half3 normalWS, inout half3 bakedG
 ///////////////////////////////////////////////////////////////////////////////
 //                      Lighting Functions                                   //
 ///////////////////////////////////////////////////////////////////////////////
-half3 LightingLambert(half3 lightColor, half3 lightDir, half3 normal)
+float3 LightingLambert(half3 lightColor, float3 lightDir, float3 normal)
```
phi-lira: Curious: is that problem that lightDir and normal are not normalized? If normalized they should fit half precision.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -599,7 +599,7 @@ half4 UniversalFragmentBlinnPhong(InputData inputData, half3 diffuse, half4 spec
     MixRealtimeAndBakedGI(mainLight, inputData.normalWS, inputData.bakedGI, half4(0, 0, 0, 0));
 
     half3 attenuatedLightColor = mainLight.color * (mainLight.distanceAttenuation * mainLight.shadowAttenuation);
-    half3 diffuseColor = inputData.bakedGI + LightingLambert(attenuatedLightColor, mainLight.direction, inputData.normalWS);
+    float3 diffuseColor = inputData.bakedGI + LightingLambert(attenuatedLightColor, mainLight.direction, inputData.normalWS);
```
phi-lira: same here. This should fit half precision no?

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
rjonaitis: normalization does not help. It most likely caused by some kind of optimization in GPU drivers, as it only affects Adreno

phi-lira:
@Unity-Technologies/gfx-qa-foundation Please validate there are no regression.

rjonaitis:
Added additional changes to fix lighting precision for PBR path

phi-lira:
@ernestasKupciunas can we do some performance test validation. 
 https://github.com/Unity-Technologies/PotatoBenchmark
 The scene `002 - FillRate (Lit Shader - Multiple Lights).unity` should be enough to benchmark this.

ernestasKupciunas:
@phi-lira here are the performance results:
https://docs.google.com/spreadsheets/d/1cppR7sKWsEswqmnnFiCjNZtQyN7eEaGa5KT1JYBh1Mk/edit?usp=sharing
There is a small performance regression, which takes around ~10%

ellioman:
What's the status of this PR?

rjonaitis:
PR is not being merged due to 10% performance regression. Currently it's low priority to investigate it in more details.

PR #4915, opened 2019-10-29
Fix the render queue not updated when switching between HDRP and a custom shader

alelievr:
### Purpose of this PR
Fix the render queue not updated when switching between HDRP and a custom shader


sebastienlagarde:
@alelievr add changelog please

PR #4925, opened 2019-10-30
Fix vfx standalonetests

JulienIgnace-Unity:
### Purpose of this PR
Fix vfx standalonetests
- Removed Migrate called from OnValidate (was causing issues with UnityObject.Destroy() being called even though it's not allowed from there)
- Restored commented out code from Seb in a earlier attempt to fix the issue.
- Updated wrong screenshot for standalone
- Temporarily removed an assert causing tests to wrongly fail

---
### Testing status

- [ ] Built a player


JulienIgnace-Unity:
I called validate just after migrate so data should be fine. ie: It goes Validate (wrong result) => Awake => Validate (correct result)

PR #4930, opened 2019-10-30
Fixed a number of issues with full screen debug modes not reseting correctly when setting another mutually exclusive mode

JulienIgnace-Unity:
### Purpose of this PR
- Fixed a number of issues with full screen debug modes not reseting correctly when setting another mutually exclusive mode
- Fixed SSAO debug showing garbage when SSAO is not enabled.


TomasKiniulis:
Hey @JulienIgnace-Unity, went through Material, Lighting and Rendering sections and didn't find any issue introduced with this. Setting another mutually exclusive mode now properly cancels out the previously selected debug mode, also SSAO correctly shows white when disabled.

PR #4932, opened 2019-10-30
Hdrp/decal ignore zero normal

Paulm-Unity:
### Checklist for PR maker

---
### Purpose of this PR
Because decal normals are signed and stored as R8G8B8, 0 is stored as 127 and not 127.5 so if some pixels in 8x8 tile have decal normals and some don't, the ones that don't, end up incorrectly affecting material normal since they are not 0. This change checks to see if there is decal contribution per pixel and ignores pixels with no contribution within an 8x8 tile.  

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Material/Lit/LitDecalData.hlsl
```diff
@@ -8,7 +8,10 @@ void ApplyDecalToSurfaceData(DecalSurfaceData decalSurfaceData, inout SurfaceDat
 
     if (decalSurfaceData.HTileMask & DBUFFERHTILEBIT_NORMAL)
     {
-        surfaceData.normalWS.xyz = normalize(surfaceData.normalWS.xyz * decalSurfaceData.normalWS.w + decalSurfaceData.normalWS.xyz);
+        if (decalSurfaceData.normalWS.w < 1.0f)
```
sebastienlagarde: Hey, I don't understand why we require this test.
isn't it rather a precision issue?
like in this PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4774? 

Note that we have a lot of location (all our shaders) where we do this thing, so if it is the correct fix, we need to do it for all materials. But I am worry about adding a test like this here

sebastienlagarde: Hey, 
isn't this bug already fix by this PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4774?
Guess the root cause of the issue is that we were not being able to encode 0 perfectly before 4774. Or it is about something else?

Note that we have a lot of location (all our shaders) where we do this thing, so if it is the correct fix, we need to do it for all materials.

Paulm-Unity: Yes that is the same issue, I didn't know this PR actually went in, thanks!

sebastienlagarde:
Fix already available for this.

PR #4949, opened 2019-11-04
PCSS reparameterization 

FrancescoC-unity:
This changes few things: 

 - PCSS Min filter size is now minimum blur intensity (functionally same thing) and instead of being in the range [0 .. 0.0001] is in the range [0 .. 1] (just remapped)

- Shadow softness is now gone, it is instead controlled by angular diameter for directional light or shape radius for punctual light. This is also now resolution independent: 

radius-min

- Shadow bias for PCSS now scales up if radius of filtering gets higher (as now the softness can be much higher) 

Had to tweak a visual test (the PCSS one) due to parametrization change. 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Fpcss-ui-update/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP_trunk/664755/job


sebastienlagarde:
@FrancescoC-unity can you merge with latest HDRP/Staging? thanks

PR #4952, opened 2019-11-04
Fix crash of prefab mode in 2019.2

FrancescoC-unity:
So backporting the different handling of the matcap view fixes the crash in prefab mode with 2019.2 . The issue was that for whatever reason we'd run _all_ variants for _all_ pixels in matcap mode before. 

Fix for https://fogbugz.unity3d.com/f/cases/1185492/


TomasKiniulis:
Checked the PR. GPU never went above 71%, before it was almost 100% when moving objects following with a crash. There's a bit of noticeable stuttering with or without matcap view but no crashes occur, and performance is way better than before in matcap view.

PR #4956, opened 2019-11-05
Allow setting decal material to none

EvgeniiG:
## Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1196129/

---
### Testing status

**Manual Tests**: No
**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-decal-null-mat/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP_trunk/683993/job


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/DecalProjector.cs
```diff
@@ -250,19 +250,24 @@ internal void OnValidate()
                 if (m_OldMaterial != m_Material)
                 {
                     DecalSystem.instance.RemoveDecal(m_Handle);
```
Paulm-Unity: Should probably only call RemoveDecal here if m_OldMaterial wasn't null

EvgeniiG: Fair enough.

sebastienlagarde:
Please add changelog

PR #4957, opened 2019-11-05
Add XR multipass rendering for Universal pipeline

mikesfbay:
### Purpose of this PR
Added XR multipass rendering for Universal pipeline. Multipass rendering is a requirement on many VR platforms and allows graceful fallback when single-pass rendering isn't available. 
This PR extends brandon's work from last year.

Fix for fogbugz case: https://fogbugz.unity3d.com/f/cases/1182204/

---
### Testing status

Tested on Oculus Quest/GO, Rift.

- [ x] Built a player


phi-lira:
Adding the multipass logic in the render pass has multiple issues (same issues present in Brandon's PR). Since brandon's PR we introduced the concept of ScriptableRenderers and execution blocks in that renderer. I wonder if it's possible to wrap the eye loop inside the main and after rendering execution blocks (new logic since brandon's PR) (https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L241-L261). This would solve the following problems that you have with your current approach:

 - Multipass would work for all renderers (forward, deferred, 2D, custom). By doing the logic in the setup, you make this feature only available in the ForwardRenderer. We would have to add support to deferred renderer and users would need to understand renderers will have to be aware of XR. We want to abstract XR rendering from users. 
 - You want to duplicate the render pass execution logic only. By doing it in the Setup you are potentially including duplicated logic, f.ex some passes call setup logic before enqueing: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/universal/xr/multipass/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs#L211
- With your current design it will break if someone introduces a custom stereo-aware render pass before the execution of BeginStereo (render pass before rendering opaques). By doing the logic in the blocks we guarantee no stereo-aware render pass will execute before the BeginStereo logic and we sort the passes to be executed in the correct order.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -188,6 +192,14 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+          {
+
+            if (cameraData.isXRMultipass)
+            {
+                EnqueuePass(m_BeginXRRenderPass);
+            }
+
             if (requiresDepthPrepass)
             {
                 m_DepthPrepass.Setup(cameraTargetDescriptor, m_DepthTexture);
```
phi-lira: By doing this multi-eye logic in the ForwardRenderer.Setup we can have some other logic being executed twice, which might be undesirable.

com.unity.render-pipelines.universal/Runtime/Passes/ScreenSpaceShadowResolvePass.cs
```diff
@@ -68,6 +68,15 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
             context.ExecuteCommandBuffer(cmd);
+
+            if (renderingData.cameraData.isXRMultipass)
+            {
+                cmd.Clear();
+                RenderTargetIdentifier screenSpaceOcclusionTexture = m_ScreenSpaceShadowmap.Identifier();
+                cmd.SetGlobalTexture(m_ScreenSpaceShadowmap.id, screenSpaceOcclusionTexture);
+                context.ExecuteCommandBuffer(cmd);
+                cmd.Clear();
+            }
```
phi-lira: This could be moved to before the context.ExecuteCommandBuffer above and avoid one extra execution.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -131,7 +134,9 @@ public virtual void Configure(CommandBuffer cmd, RenderTextureDescriptor cameraT
         /// </summary>
         /// <param name="cmd">Use this CommandBuffer to cleanup any generated data</param>
         public virtual void FrameCleanup(CommandBuffer cmd)
-        {}
+        {
+            eyeIndex = 0;
```
phi-lira: Users can override this method. You can't guarantee a custom render pass will call this.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -230,15 +233,14 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
...
-            bool stereoEnabled = renderingData.cameraData.isStereoEnabled;
             context.SetupCameraProperties(camera, stereoEnabled);
 
             // Override time values from when `SetupCameraProperties` were called.
             // They might be a frame behind.
             // We can remove this after removing `SetupCameraProperties` as the values should be per frame, and not per camera.
             SetShaderTimeValues(time, deltaTime, smoothDeltaTime);
 
-            if (stereoEnabled)
+            if (stereoEnabled && !renderingData.cameraData.isXRMultipass)
```
phi-lira: I'm wondering if we can't move the eye loop logic here. By doing this it seems we would fix many issues. (Detail of the issues added in the comment at the end of PR)

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -402,15 +405,15 @@ void ExecuteRenderPass(ScriptableRenderContext context, ScriptableRenderPass ren
                 context.ExecuteCommandBuffer(cmd);
                 cmd.Clear();
 
-                if (cameraData.isStereoEnabled)
+                if (cameraData.isStereoEnabled && !cameraData.isXRMultipass)
                 {
-                    context.StartMultiEye(cameraData.camera);
+                    context.StartMultiEye(cameraData.camera, renderPass.eyeIndex);
                     XRUtils.DrawOcclusionMesh(cmd, cameraData.camera);
```
phi-lira: Can we move the StartMultiEye and DrawOccusionMesh to execute before the main rendering block in ScriptableRenderer (stereo aware block)?

mikesfbay:
I have a new design and PR based on the suggestions above:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5061

I'll leave this one as-is for comparison (will close soon).

mikesfbay:
I have a new PR based on the design suggestions above:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5061

I'll leave this one as-is for comparison (will close soon).

phi-lira:
Closed. Merged in #5061

PR #4960, opened 2019-11-05
Adding a new refraction mode and changing refraction UI

anisunity:
All in the title


sebastienlagarde:
@anisunity could you provide more info in the PR description for the @Unity-Technologies/gfx-qa-hdrp ? thanks

sebastienlagarde:
Note that Yamato refraction test are broken. I expect it is due to the cahnge, so we need to provide some migration or redo the test.

anisunity:
This PR adds a new refraction mode called Thin refraction. Basically, it is just a special case of the previously existing box mode where the medium is thin (this is intended mainly for windows and other thin glass). We also decided to remove the distortion option from the lit shader.

PR #4978, opened 2019-11-06
Constant buffer usage prototype (not to be merged as is)

joeldevahl:
### Purpose of this PR
This PR presents a way to use constant buffers instead of separate properties to bind data to shaders.
Over all this reduces main thread cost in many ways:
- less C# API calls
- less work when replaying command buffers
- less work when extracting shader properties

All APIs used in this PR have landed to 20.1, with only GL and PS4 backends for constant buffer binding to compute shaders to land soon.

---
### Testing status

Manual testing using the SRP batcher stress test, with materials modified to use sub surface scattering.

Manual testing in separate (non HDRP) stress test to see performance impact. Can attach test project.


joeldevahl:
Stress test: 
[ComputeConstantBuffersStress.zip](https://github.com/Unity-Technologies/ScriptableRenderPipeline/files/3818583/ComputeConstantBuffersStress.zip)

sebastienlagarde:
This PR is now obsolete, so closing.

PR #4979, opened 2019-11-06
Bugfix for transparent objects and shadow receiving

ellioman:
### Purpose of this PR
Fixing issue #1116936
https://issuetracker.unity3d.com/issues/lwrp-cascaded-shadows-do-not-appear-on-alpha-blended-objects

---
### Testing status

- [ ] Built a player
- [X ] Checked new UI names with UX convention
- Other: 

Added a automatic test with a reference image.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fbugfix-1116936

The "105_TransparentReceiveShadows" scene in the UniversalGraphicsTest project has spheres using unlit, lit, simple lit, particles lit, particles simple lit and shadergraph shaders to easily see if things are working or not.

---
### Comments to reviewers
Transparent objects were not receiving shadows when using shadow cascades.

Things done:
1. Fixed the problem so now transparent objects should receive shadows with cascades both enabled and disabled.
2. Added a toggle "Transparent Receive Shadows" in the renderer asset which enables/disables globally that transparent object can receive shadows. Default setting is set to true.
3. Particles are now able to receive shadows. It was disabled because it didn't work with screen space shadows but since that is no longer used we were able to reenable the feature.


com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
```diff
@@ -185,9 +178,9 @@ half ComputeCascadeIndex(float3 positionWS)
     float3 fromCenter1 = positionWS - _CascadeShadowSplitSpheres1.xyz;
     float3 fromCenter2 = positionWS - _CascadeShadowSplitSpheres2.xyz;
     float3 fromCenter3 = positionWS - _CascadeShadowSplitSpheres3.xyz;
-    float4 distances2 = float4(dot(fromCenter0, fromCenter0), dot(fromCenter1, fromCenter1), dot(fromCenter2, fromCenter2), dot(fromCenter3, fromCenter3));
+    float4 distances = float4(dot(fromCenter0, fromCenter0), dot(fromCenter1, fromCenter1), dot(fromCenter2, fromCenter2), dot(fromCenter3, fromCenter3));
```
phi-lira: 2 here means pow(distances, 2)

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
```diff
@@ -7,14 +7,6 @@
 
 #define MAX_SHADOW_CASCADES 4
 
-#ifndef SHADOWS_SCREEN
```
phi-lira: Thinking about it, I'm worried users might be relying on this defined. We should always defined it to 0 now instead of removing it.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -432,7 +432,7 @@ static void InitializeShadowData(UniversalRenderPipelineAsset settings, NativeAr
             shadowData.supportsMainLightShadows = SystemInfo.supportsShadows && settings.supportsMainLightShadows && mainLightCastShadows;
 
             // we resolve shadows in screenspace when cascades are enabled to save ALU as computing cascade index + shadowCoord on fragment is expensive
-            shadowData.requiresScreenSpaceShadowResolve = shadowData.supportsMainLightShadows && supportsScreenSpaceShadows && settings.shadowCascadeOption != ShadowCascadesOption.NoCascades;
+            shadowData.requiresScreenSpaceShadowResolve = false;// shadowData.supportsMainLightShadows && supportsScreenSpaceShadows && settings.shadowCascadeOption != ShadowCascadesOption.NoCascades;
```
phi-lira: Let's add a comment here explaining the reasoning.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -129,30 +129,37 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
             }
 
+            bool postProcessEnabled = cameraData.postProcessEnabled;
+            bool isSceneViewCamera = cameraData.isSceneViewCamera;
+            bool requiresDepthTexture = cameraData.requiresDepthTexture;
+            bool isStereoEnabled = cameraData.isStereoEnabled;
+
             bool mainLightShadows = m_MainLightShadowCasterPass.Setup(ref renderingData);
             bool additionalLightShadows = m_AdditionalLightsShadowCasterPass.Setup(ref renderingData);
             bool resolveShadowsInScreenSpace = mainLightShadows && renderingData.shadowData.requiresScreenSpaceShadowResolve;
```
phi-lira: We don't resolve shadows in screenspace anymore. We can remove this and it's usages.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -129,30 +129,37 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
             // Depth prepass is generated in the following cases:
-            // - We resolve shadows in screen space
             // - Scene view camera always requires a depth texture. We do a depth pre-pass to simplify it and it shouldn't matter much for editor.
             // - If game or offscreen camera requires it we check if we can copy the depth from the rendering opaques pass and use that instead.
-            bool requiresDepthPrepass = renderingData.cameraData.isSceneViewCamera ||
-                (cameraData.requiresDepthTexture && (!CanCopyDepth(ref renderingData.cameraData)));
+            // - We resolve shadows in screen space
+            bool requiresDepthPrepass = isSceneViewCamera;
+            requiresDepthPrepass |= (requiresDepthTexture && !CanCopyDepth(ref renderingData.cameraData));
             requiresDepthPrepass |= resolveShadowsInScreenSpace;
```
phi-lira: This should be removed as this is always false.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -33,6 +33,10 @@
 #define BUMP_SCALE_NOT_SUPPORTED !SHADER_HINT_NICE_QUALITY
 #endif
 
+#if defined(_ADDITIONAL_LIGHTS) || defined(_MAIN_LIGHT_SHADOWS_CASCADE)
+#define REQUIRES_WORLD_SPACE_POS_INTERPOLATOR
+#endif
```
phi-lira: Thinking about this. We should move out the REQUIRES_WORLD_SPACE_POS_INTERPOLATOR out of core. It doesn't make sense to be core as the interpolator struct is declared for each shader. It should be moved to the files that declare the interpolators.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -570,6 +570,10 @@ half4 UniversalFragmentPBR(InputData inputData, half3 albedo, half metallic, hal
     BRDFData brdfData;
     InitializeBRDFData(albedo, metallic, specular, smoothness, alpha, brdfData);
 
+#if defined(REQUIRES_WORLD_SPACE_POS_INTERPOLATOR)
+    inputData.shadowCoord = TransformWorldToShadowCoord(inputData.positionWS);
+#endif
```
phi-lira: We should consider the InputData to be readonly here for simplicity. We should change compute ShadowCoord like this:

 If not using cascades:
    - We compute in vertex
    - declare an interpolator for shadowCoord
    - in fragment we initialize the inputData.shadowCoord as the interpolator

If using cascades:
    - we don't declare an interpolator, as we will compute in frag.
    - compute inputData.shadowCoord in the place we initialize InputData in frag.

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
ellioman: Got it. Will revert back to the previous naming

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
ellioman: Done

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
ellioman: Done. Removed the resolveShadowsInScreenSpace and its uses.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
ellioman: Removed.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
ellioman: Makes sense. Fixed.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -31,6 +33,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed user LUT sampling being done in Linear instead of sRGB.
 - Fixed an issue when trying to get the Renderer via API on the first frame [case 1189196](https://issuetracker.unity3d.com/product/unity/issues/guid/1189196/)
 - Fixed a material leak on domain reload.
+- Fixed an issue with transparent objects not receiving shadows when using shadow cascades. [case 1116936] (https://issuetracker.unity3d.com/issues/lwrp-cascaded-shadows-do-not-appear-on-alpha-blended-objects)
```
phi-lira: remove the space between [case xxx] and (url). The space will cause the link to break.

com.unity.render-pipelines.universal/Runtime/Passes/TransparentSettingsPass.cs
```diff
@@ -0,0 +1,40 @@
...
+            return !m_shouldReceiveShadows;
+        }
+
+        public override void Execute(ScriptableRenderContext context, ref RenderingData renderingData)
+        {
+            // Get a command buffer...
+            CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
+
+            // Toggle light shadows enabled based on the renderer setting set in the constructor
+            CoreUtils.SetKeyword(cmd, ShaderKeywordStrings.MainLightShadows, m_shouldReceiveShadows);
```
phi-lira: Should we also disable _MAIN_LIGHT_SHADOW_CASCADES keyword?

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
```diff
@@ -246,7 +255,7 @@ half AdditionalLightRealtimeShadow(int lightIndex, float3 positionWS)
 
 float4 GetShadowCoord(VertexPositionInputs vertexInput)
 {
-#if SHADOWS_SCREEN
+#if defined(_MAIN_LIGHT_SHADOWS_CASCADE)
     return ComputeScreenPos(vertexInput.positionCS);
```
phi-lira: Since we are not using screenspace shadow resolve shadowCoord should not return the screen space uv shadow coord anymore.

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesSimpleLit.shader
```diff
@@ -96,7 +96,6 @@ Shader "Universal Render Pipeline/Particles/Simple Lit"
             #pragma shader_feature _EMISSION
             #pragma shader_feature _ _SPECGLOSSMAP _SPECULAR_COLOR
             #pragma shader_feature _GLOSSINESS_FROM_BASE_ALPHA
-            #pragma shader_feature _RECEIVE_SHADOWS_OFF
```
phi-lira: Is this international? Seems like we are removing from SimpleLit the ability to not render shadows.

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesLit.shader
```diff
@@ -89,7 +89,6 @@ Shader "Universal Render Pipeline/Particles/Lit"
             #pragma shader_feature _NORMALMAP
             #pragma shader_feature _EMISSION
             #pragma shader_feature _METALLICSPECGLOSSMAP
-            #pragma shader_feature _RECEIVE_SHADOWS_OFF
```
phi-lira: Is this intentional? Seems like we are removing the option to select in the material if particle lit shaders should receive shadows or not.

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
```diff
@@ -5,14 +5,25 @@
...
+    #if defined(_MAIN_LIGHT_SHADOWS)
+        #define MAIN_LIGHT_CALCULATE_SHADOWS
+    #endif
+
+    #if defined(_ADDITIONAL_LIGHT_SHADOWS)
+        #define ADDITIONAL_LIGHT_CALCULATE_SHADOWS
+    #endif
 #endif
+
+#if !defined(_MAIN_LIGHT_SHADOWS_CASCADE)
```
phi-lira: seems like this should be if defined(_MAIN_LIGHT_SHADOWS) and !defined(_MAIN_LIGHT_SHADOWS_CASCADES). When we don't have shadows at all we don't need the interpolator.

phi-lira: Seems like this could also be inside the #if !defined(_RECEIVE_SHADOWS_OFF) block. If that is defined we dont' have shadwos and then don't require the interpolator as well.

com.unity.render-pipelines.universal/CHANGELOG.md
ellioman: Done

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesLit.shader
ellioman: No this is a mistake on my part. We took this feature out before realising that we were able to enable shadows on particles again since we removed screen space shadows. Fixed!

com.unity.render-pipelines.universal/Shaders/Particles/ParticlesSimpleLit.shader
ellioman: No this is a mistake on my part. We took this feature out before realising that we were able to enable shadows on particles again since we removed screen space shadows. Fixed!

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
ellioman: Right! Fixed.###

com.unity.render-pipelines.universal/ShaderLibrary/Shadows.hlsl
ellioman: Makes total sense. Fixed.

com.unity.render-pipelines.universal/Runtime/Passes/TransparentSettingsPass.cs
ellioman: Disabling it might take out the REQUIRES_WORLD_SPACE_POS_INTERPOLATOR define so that's a good thing. Fixed

ellioman:
@Unity-Technologies/gfx-qa-foundation 
- Please test boat attack to not have regressions
- Please test https://github.com/Unity-Technologies/UniversalRenderingExamples
-- Take a look at the gun example, shadows should work when you enable shadow cascades

PR #4980, opened 2019-11-07
Analytics for shadergraph

martint-unity:
### Purpose of this PR
Adding the base for our analytics.

---

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fanalytics/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk/700547/job


martint-unity:
Same issue failing as in Master
[10:38:56.155 INF] An unexpected error happened while running tests.
	Library/PackageCache/com.unity.test-framework@1.1.5/UnityEditor.TestRunner/Api/CallbacksDelegator.cs:64

Nothing this PR introduced

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
```
pbbastian: Get rid of this and use `JsonUtility` instead

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
...
+        static bool EnableAnalytics()
+        {
+            AnalyticsResult result = EditorAnalytics.RegisterEventWithLimit(k_EventName, k_MaxEventsPerHour, k_MaxNumberOfElements, k_VendorKey, k_Version);
+            if (result == AnalyticsResult.Ok)
+                s_EventRegistered = true;
+
+            return s_EventRegistered;
+        }
+
+        private struct AnalyticsData
```
pbbastian:
```suggestion
        struct AnalyticsData
```

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
...
+
+                if (!nodeTypeAndCount.ContainsKey(nType))
+                {
+                    nodeTypeAndCount[nType] = 1;
+                }
+                else
+                {
+                    nodeTypeAndCount[nType] += 1;
+                }
+            }
```
pbbastian: Maybe extract this into a method, and add a unit test for it

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        private struct AnalyticsData
+        {
+            public string nodes;
+            public int node_count;
+            public string main_guid;
+            public int subgraph_count;
+        }
+
+        public static void SendShaderGraphEvent(string mainGuid, IEnumerable<AbstractMaterialNode> nodes)
```
pbbastian:
```suggestion
        public static void SendShaderGraphEvent(string assetGuid, IEnumerable<AbstractMaterialNode> nodes)
```

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
...
+                s_EventRegistered = true;
+
+            return s_EventRegistered;
+        }
+
+        private struct AnalyticsData
+        {
+            public string nodes;
+            public int node_count;
+            public string main_guid;
```
pbbastian:
```suggestion
            public string asset_guid;
```

com.unity.shadergraph/Editor/ShaderGraphAnalytics.cs
```diff
@@ -0,0 +1,79 @@
...
+                    nodeTypeAndCount[nType] += 1;
+                }
+            }
+
+            var jsonRepr = JsonConvert.SerializeObject(nodeTypeAndCount);
+            var data = new AnalyticsData()
+            {
+                nodes = jsonRepr,
+                node_count = nodes.Count(),
+                main_guid = mainGuid,
```
pbbastian:
```suggestion
                asset_guid = assetGuid,
```

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -339,6 +339,10 @@ public void UpdateAsset()
                 if (string.IsNullOrEmpty(path) || graphObject == null)
                     return;
 
+                var nodes = graphObject.graph.GetNodes<AbstractMaterialNode>();
```
marctem: Why get the nodes here instead of inside the Analytics class? Seems like it limits our potential to record more data and could be unnecessary if analytics is off.

GrantLamb-Unity:
Adding Favro card for reference: https://favro.com/organization/c564ede4ed3337f7b17986b6/d06b8de3e40da6d340b5b063?card=Uni-101544

I'm unfamiliar with Analytics testing so I have a few questions:
1. What exactly is the event send trigger? - I saw there was some discussion within the documentation [here](https://docs.google.com/document/d/1oS18PTLdUdSTgHqNWtrMo6ehix_sp_kbh8ahi5Lym1M/edit#) but It's not clear as to what was implemented.
2. What URL are analytics being sent too? (So I can observe it happening in real time via the Charles Network Proxy Tool)
3. I tried messing around with Shadergraph, closed the editor, then checked "C:\Users\<User>\AppData\LocalLow\Unity\Editor for data file containing your event name". I didn't see a reference to the "uShaderGraphUsage" event in any of these files. Although I may not be triggering the event properly due to question 1.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
martint-unity: You are right, passing in the graph now.

martint-unity:
It is triggered when clicking save asset. Will be sent next cycle the editor send it. It is around 5-10 minutes. 
I used fiddler to check it is sending. 
But I don't think you should test if it is sending or not. The database people on this side sees the data so I think that is good enough. 
Don't really know what you should test except that it is not crashing :)

GrantLamb-Unity:
Tested the following: Created various Shadergraphs and Sub-Graphs, Saved Assets periodically in various states. No warnings or Errors related to Analytics occurred during testing.

I wasn't able to verify the Analytics Messaging trigger or the data being sent is the expected data; but due to Martint-unity's comments it sounds like that was covered already.

Approving this PR.

PR #4981, opened 2019-11-07
Re-enable optional clears of GBuffers

sebastienlagarde:
Re-enable the optional clears of GBuffer.
Seems that someone have commented it few month ago. 

Reviewevers, is anyone know why this code have been commented? is there any reasons as there is no comment? Maybe a mistake?


anisunity:
I did not do this. I guess not clearing the gbuffers is fine as long as the usage of the gbuffers is conditional on the depth and stencil buffers. However, i am not sure it is the case everywhere, i guess we must check before merging (potentially ray tracing)

sebastienlagarde:
This was like this since the beginning of HDRP, The question isn't should we enable it, the question is why someone have commented it :)

PR #4985, opened 2019-11-07
Properties starts as collapsed

martint-unity:
### Purpose of this PR
This pr will make so that the properties starts collapsed.
---
### Testing status

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fremember-expanded-inputs/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk/701226/job


LandonTownsendUnity:
Can you go into more detail about what this is supposed to fix? What it seems to do is make it so that all of the properties are collapsed when you restart the editor, whereas previously they all expand when you restart the editor. It doesn't seem to actually remember the expanded/collapsed state of any property between editor restarts.

martint-unity:
@LandonTownsendUnity 
That is correct. It starts collapsed and it will remember the state during this session. So if you close the graph and re-open it, it will remember it but only during this session. Restarting the editor wipes that "memory"

LandonTownsendUnity:
I guess what has me confused is, that Shadergraph already seems to remember this between closing and reopening the tab, even on the master branch:
SGBlackboardCollapsedAlreadyRemembers
So to me it seems like all this branch adds is making them collapsed by default.
Is there something I'm missing?

PR #4990, opened 2019-11-08
Refactor of Profiling in SRP to prepare for GPU profiling.

JulienIgnace-Unity:
### Purpose of this PR
This PR as several goals:
- Rename ProfilingSample to ProfilingScope to better represent what it actually does (ProfilingSample kept as obsolete)
- Replaced calls to CommandBuffer.Begin/EndSample(string) by CommandBuffer.Begin/End(CustomSampler) as advised by the profiling team.
- Prepared code for upcoming PR adding support for GPU timing (https://ono.unity3d.com/unity/unity/pull-request/96099/_/graphics/gpu-recorder)
- Now a profiling scope will do several things: 
     - Add a CommandBuffer profiling scope
     - Add a Event Marker (for RenderDoc and such)
     - Add an inline profiling scope (to profile code directly from the caller and not just the command buffer)
     - Add a GPU profiling scope
- Added profiling timings to the HDRP Display Statistics debug window

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                int max = 0;
+                foreach (var value in Enum.GetValues(typeof(ProfilingSamplerId)))
+                {
+                    max = Math.Max(max, (int)value);
+                }
+                s_Samplers = new ProfilingSampler[max + 1];
+            }
+
+            int index = (int)(object)samplerId;
```
phi-lira: I'm not sure if I follow why casting to object is necessary here. samplerId is an enum. By casting to object you cause boxing.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    public struct ProfilingScope : IDisposable
+    {
+        CommandBuffer   m_Cmd;
+        bool            m_Disposed;
+        CustomSampler   m_Sampler;
+        CustomSampler   m_InlineSampler;
+
+        public ProfilingScope(CommandBuffer cmd, in ProfilingSampler sampler)
```
phi-lira: `in` modifier is redundant. sampler is a reference type.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+            m_InlineSampler?.Begin();
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        // Protected implementation of Dispose pattern.
+        void Dispose(bool disposing)
```
phi-lira: Not really expert on C# but shouldn't this be `virtual` method? Is it implicit?

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                if (m_Cmd != null && m_Name != "")
+                    m_Cmd.EndSample(m_Name);
+                m_Sampler?.End();
+            }
+
+            m_Disposed = true;
+        }
+    }
+}
```
phi-lira: Please revert the deletion ProfilingScore.cs.meta file.

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphProfileId.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace UnityEngine.Experimental.Rendering.RenderGraphModule
+{
+    internal enum RenderGraphProfileId
+    {
+        RenderGraphClear,
+        RenderGraphClearDebug,
+    }
+
+    internal static class HDCustomSamplerExtension
```
phi-lira: Even though this is internal, it's really weird you had `HD` prefix on a core class. If this is specific to HDRP shouldn't it live in HDRP package?

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+#endif
+                m_InlineSampler?.End();
+            }
+
+            m_Disposed = true;
+        }
+    }
+
+    [System.Obsolete("Please use ProfilingScope")]
+    public struct ProfilingSample : IDisposable
```
phi-lira: Why does this appear as new files? Was it moved from another class? This class should not change appart from the Obsolete attribute.

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphProfileId.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static ProfilingSampler Get(this RenderGraphProfileId samplerId)
```
phi-lira: Add descriptive name for static function.

phi-lira:
Adding foundation QA to test that this PR doesn't introduce breakage or regression in URP.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
```diff
@@ -0,0 +1,195 @@
...
+        public int cpuSampleCount => m_Recorder.enabled ? m_Recorder.sampleBlockCount : 0;
+        public float inlineCpuElapsedTime => m_InlineRecorder.enabled ? m_InlineRecorder.elapsedNanoseconds / 1000000.0f : 0.0f;
+        public int inlineCpuSampleCount => m_InlineRecorder.enabled ? m_InlineRecorder.sampleBlockCount : 0;
+    }
+
+    public class ProfileSamplerList<ProfilingSamplerId> where ProfilingSamplerId : Enum
+    {
+        static ProfilingSampler[] s_Samplers = null;
+
+        static public ProfilingSampler Get(ProfilingSamplerId samplerId)
```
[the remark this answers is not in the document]
JulienIgnace-Unity: The idea in this particular case is to avoid users to have to write a lot of boilerplate code just to add a profiling scope, so the shorter the better.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
JulienIgnace-Unity: It does not compile without this, I'm not sure why :/ (workaround provided by Thomas Hourdel)

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
JulienIgnace-Unity: Leftover from a time where ProfilingSampler was a struct. Will remove

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
JulienIgnace-Unity: I think it's implicit (all our disposeable are implemented that way)

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
JulienIgnace-Unity: The file was renamed because I added the new API in it directly (wasn't sure I wanted to keep an old class in a separate file forever)

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
JulienIgnace-Unity: Not sure I'm following, it's just a rename?

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphProfileId.cs
JulienIgnace-Unity: That's a copy/paste mistake, will rename it to something RenderGraph specific

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphProfileId.cs
JulienIgnace-Unity: Same comment as above.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
phi-lira: I couldn't comment on the .meta file so I commented here. For some reason I had seen an delete .meta file, but now it shown as a rename. I'll resolve this.

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphProfileId.cs
phi-lira: `RenderGraphProfileId.GetSampler` It's more descriptive and I don't see why it increases code complexity.

com.unity.render-pipelines.core/Runtime/Debugging/ProfilingScope.cs
phi-lira: might be worth investigating why. Have you checked GC Allocs in the Profiler?

com.unity.render-pipelines.core/Runtime/RenderGraph/RenderGraphResourceRegistry.cs
```diff
@@ -377,10 +377,10 @@ internal void CreateAndClearTexturesForPass(RenderGraphContext rgContext, int pa
                     if (resource.desc.clearBuffer || m_RenderGraphDebug.clearRenderTargetsAtCreation)
                     {
                         // Commented because string.Format causes garbage
-                        //using (new ProfilingSample(rgContext.cmd, string.Format("RenderGraph: Clear Buffer {0}", resourceDescMoved.desc.name)))
+                        //using (new ProfilingScope(rgContext.cmd, string.Format("RenderGraph: Clear Buffer {0}", resourceDescMoved.desc.name)))
```
phi-lira: Why not remove this?

com.unity.render-pipelines.universal/Runtime/Passes/AdditionalLightsShadowCasterPass.cs
```diff
@@ -256,7 +257,7 @@ void RenderAdditionalShadowmapAtlas(ref ScriptableRenderContext context, ref Cul
 
             bool additionalLightHasSoftShadows = false;
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
-            using (new ProfilingSample(cmd, m_ProfilerTag))
+            using (new ProfilingScope(cmd, m_ProfilingSampler))
```
phi-lira: Isn't this causing GC.Alloc?

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -55,7 +56,7 @@ public override void Configure(CommandBuffer cmd, RenderTextureDescriptor camera
         public override void Execute(ScriptableRenderContext context, ref RenderingData renderingData)
         {
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
-            using (new ProfilingSample(cmd, m_ProfilerTag))
+            using (new ProfilingScope(cmd, m_ProfilingSampler))
```
phi-lira: Same here

com.unity.render-pipelines.universal/Runtime/Passes/DrawObjectsPass.cs
```diff
@@ -39,7 +41,7 @@ public DrawObjectsPass(string profilerTag, bool opaque, RenderPassEvent evt, Ren
         public override void Execute(ScriptableRenderContext context, ref RenderingData renderingData)
         {
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
-            using (new ProfilingSample(cmd, m_ProfilerTag))
+            using (new ProfilingScope(cmd, m_ProfilingSampler))
```
phi-lira: and here

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -253,4 +253,16 @@ void SortCameras(Camera[] cameras)
 #endif
         };
     }
+
+    enum URPProfileId
```
phi-lira: please make internal.

phi-lira:
Adding @pbbastian for coverage

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2123,29 +2110,20 @@ void Callback(CommandBuffer c, HDCamera cam)
                 }
                 else
                 {
-                    using (new ProfilingSample(cmd, "Build Light list", CustomSamplerId.BuildLightList.GetSampler()))
```
sebastienlagarde: is it expected to removed Build Light list counter ?

sebastienlagarde:
Yamato is green (with the revision that fix issue from Juho). What fail is one test on VFX LW (same as master), and one test of universal litbake (same as master).
Also playmode HDRP DX11 failed on GC alloc. This is already on master but hidden by another issue. This PR also fix this issue (a missing if editor)

PR #4996, opened 2019-11-09
Improve history buffer management

EvgeniiG:
### Purpose of this PR
Fix invalid assertion.
Do not allocate history buffers if they are not going to be used.
Invalidate history buffers on reallocation.
Document code.
Fix https://fogbugz.unity3d.com/f/cases/1196636/

---
### Testing status

**Manual Tests**: Yes
**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-vl-rt-alloc/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP_trunk/711472/job


TomasKiniulis:
Assertion errors do not appear anymore entering playmode with new hdrp project. Importing DotsShooter to make sure.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -821,8 +799,14 @@ void VolumetricLightingPass(HDCamera hdCamera, CommandBuffer cmd, int frameIndex
             {
                 var parameters = PrepareVolumetricLightingParameters(hdCamera, frameIndex);
 
-                var historyRT = hdCamera.GetPreviousFrameRT((int)HDCameraFrameHistoryType.VolumetricLighting);
-                var feedbackRT = hdCamera.GetCurrentFrameRT((int)HDCameraFrameHistoryType.VolumetricLighting);
+                RTHandle historyRT  = null;
```
JulienIgnace-Unity: This change is not necessary, those two buffers are not used in the next function if reprojection is not enabled. Let's not add more logic than necessary.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -482,6 +462,8 @@ void PushVolumetricLightingGlobalParams(HDCamera hdCamera, CommandBuffer cmd, in
             // The lighting & density buffers are shared by all cameras.
             // The history & feedback buffers are specific to the camera.
             // These 2 types of buffers can have different sizes.
+            // Additionally, history buffers can have different sizes, since they are not resized at the same time
+            // (every frame, we resize the feedback buffer but not the history buffer, and then we swap them).
```
JulienIgnace-Unity: This comment is partly wrong, we actually swap then resize the current buffer for all history RTs (which does result in the history buffer for next frame not being resized at the current frame)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -359,7 +366,15 @@ public void Update(FrameSettings currentFrameSettings, HDRenderPipeline hdrp, MS
         public void BeginRender()
         {
             RTHandles.SetReferenceSize(m_ActualWidth, m_ActualHeight, m_msaaSamples);
-            m_HistoryRTSystem.SwapAndSetReferenceSize(m_ActualWidth, m_ActualHeight, m_msaaSamples);
+
+            bool reallocated = m_HistoryRTSystem.SwapAndSetReferenceSize(m_ActualWidth, m_ActualHeight, m_msaaSamples);
```
JulienIgnace-Unity: What's the reason behind this?
Is this supposed to avoid having an invalid history buffer? If this is the case then this is wrong since the history buffer will still be valid since it's not resized at this frame (we never resize history, we swap, then resize the current buffer). If not, is this because the reprojection does not support buffers with different resolutions?

JulienIgnace-Unity: Besides, for color buffer it's clearly wrong for the reason stated in the previous comment. So we need to revert this.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
EvgeniiG: These buffers may not have been allocated. Why is it safe to access them?

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
EvgeniiG: That's what the comment says?

EvgeniiG: I am not sure I follow.
Are you saying that the history buffer we write to during the current frame (and read during the next frame) is NOT resized during the current frame?

EvgeniiG: Better now?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -275,7 +282,7 @@ public void Update(FrameSettings currentFrameSettings, HDRenderPipeline hdrp, MS
             {
                 bool isCurrentColorPyramidRequired = m_frameSettings.IsEnabled(FrameSettingsField.RoughRefraction) || m_frameSettings.IsEnabled(FrameSettingsField.Distortion);
                 bool isHistoryColorPyramidRequired = m_frameSettings.IsEnabled(FrameSettingsField.SSR) || antialiasing == AntialiasingMode.TemporalAntialiasing;
-                bool isVolumetricHistoryRequired = m_frameSettings.IsEnabled(FrameSettingsField.Volumetrics) && m_frameSettings.IsEnabled(FrameSettingsField.ReprojectionForVolumetrics);
+                bool isVolumetricHistoryRequired   = IsVolumetricReprojectionEnabled();
```
EvgeniiG: > This PR did not change that, am I missing something?
Yes, you did.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
EvgeniiG: My bad, will remove.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
JulienIgnace-Unity: Please enlighten me. Do you refer to the part where you protected hdCamera.GetPreviousFrameRT by a condition? Those functions just return null if it's not allocated.
The allocation is done in InitializeVolumetricLightingPerCameraData and if the buffer count is 0 it does not allocate anything already?

EvgeniiG: Weird, I left the comment explaining where but can't find it.

EvgeniiG: Ah, wait, this is it, it's in the right spot. The condition above changed.

EvgeniiG: (see the code line the comment is attached to)

EvgeniiG: >Those functions just return null if it's not allocated.
That's good to know so I can remove the if. Thanks.

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
EvgeniiG: Removed the if based on the discussion in another thread.

PR #5003, opened 2019-11-12
Remove unnecessary dependencies from SRP test projects

sophiaaar:
On the 1st of November, the Unity Ads package made a change that broke projects using the Unity Test Framework. Since Ads is included as a default package, this broke all our test projects. We should never be in a position where we are prevented from landing code because of a change in a package that we don't use in the projects.

However, we also don't want to remove the coverage we have right now (ie testing that a "default" Unity project works with SRP), so I've also added a new test project that will serve as a "smoke test". This will be run nightly to ensure that the combination of default packages + SRP doesn't break.

---
### Purpose of this PR
* Removes unnecessary packages from the project manifests
* Adds `"disableProjectUpdate": true` to the manifests so that the "default" dependencies don't get added back in
* Adds a "Smoke Test" project that will be run nightly to ensure that a fresh/default Unity project that uses the SRP packages will still work. This is so we can ensure that if something like this Ads issue happens again, and our users come across it, we're already aware of it and we can help them with a workaround if required

---
### Testing status
**Yamato**: 
Smoke test pipelines - it is expected that the Standalone ones will fail while this Ads issue is still present: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/automation%252Fremove-unneeded-dependencies/.yamato%252Fupm-ci-abv.yml%2523all_smoke_tests/719416/job/pipeline

Standalone tests are now passing after removing the unneeded packages: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/automation%252Fremove-unneeded-dependencies/.yamato%252Fupm-ci-shadergraph.yml%2523ShaderGraph_Windows64_DX11_Standalone_fast-trunk/714990/job


TestProjects/HDRP_Tests/Packages/manifest.json
```diff
@@ -1,15 +1,20 @@
...
     "com.unity.render-pipelines.high-definition": "file:../../../com.unity.render-pipelines.high-definition",
     "com.unity.render-pipelines.high-definition-config": "file:../../../com.unity.render-pipelines.high-definition-config",
     "com.unity.shadergraph": "file:../../../com.unity.shadergraph",
-    "com.unity.test-framework": "1.1.3",
+    "com.unity.test-framework": "1.1.5",
     "com.unity.testframework.graphics": "file:../../../com.unity.testframework.graphics",
     "com.unity.testing.hdrp": "file:../../../com.unity.testing.hdrp",
+    "com.unity.ugui": "1.0.0",
     "com.unity.visualeffectgraph": "file:../../../com.unity.visualeffectgraph",
+    "com.unity.xr.legacyinputhelpers": "1.3.7",
```
sebastienlagarde: guess we have too much package added here, we should just have the SRP package + test framework and rely on auto update to get other packae, no ?

sophiaaar: These all got added when the tests would run on Yamato anyway

TestProjects/ShaderGraphUniversalStereo/Packages/manifest.json
```diff
@@ -1,25 +1,16 @@
...
     "com.unity.ide.visualstudio": "1.0.11",
-    "com.unity.ide.vscode": "1.1.2",
-    "com.unity.multiplayer-hlapi": "1.0.4",
-    "com.unity.purchasing": "2.0.6",
+    "com.unity.ide.vscode": "1.1.3",
     "com.unity.render-pipelines.core": "file:../../../com.unity.render-pipelines.core",
     "com.unity.render-pipelines.universal": "file:../../../com.unity.render-pipelines.universal",
     "com.unity.shadergraph": "file:../../../com.unity.shadergraph",
-    "com.unity.test-framework": "1.1.2",
+    "com.unity.test-framework": "1.1.3",
```
sebastienlagarde: should  be 1.1.5 ?

TestProjects/HDRP_Tests/Packages/manifest.json
sophiaaar: I specifically had to turn off auto update otherwise the Ads (etc) packages got added back in, even if they werent in the manifest

TestProjects/ShaderGraphUniversalStereo/Packages/manifest.json
sophiaaar: yep

TestProjects/VisualEffectGraph/Packages/manifest.json
```diff
@@ -1,26 +1,51 @@
...
     "com.unity.modules.physics2d": "1.0.0",
+    "com.unity.modules.screencapture": "1.0.0",
+    "com.unity.modules.terrain": "1.0.0",
+    "com.unity.modules.terrainphysics": "1.0.0",
+    "com.unity.modules.tilemap": "1.0.0",
     "com.unity.modules.ui": "1.0.0",
-    "com.unity.modules.vr": "1.0.0"
+    "com.unity.modules.uielements": "1.0.0",
+    "com.unity.modules.umbra": "1.0.0",
+    "com.unity.modules.unityanalytics": "1.0.0",
```
sebastienlagarde: guess we shouldn't add all those pacakges

sophiaaar: this is intended

sophiaaar: i'll elaborate - these were getting added before anyway when yamato ran. the modules, if i understand correctly, the modules are packages that live in trunk, so we can think of them as just unity code.

TestProjects/HDRP_RuntimeTests/Packages/manifest.json
```diff
@@ -1,24 +1,15 @@
 {
   "dependencies": {
-    "com.unity.2d.sprite": "1.0.0",
-    "com.unity.2d.tilemap": "1.0.0",
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.3.2",
-    "com.unity.collab-proxy": "1.2.16",
-    "com.unity.ide.rider": "1.1.0",
+    "com.unity.ide.rider": "1.1.1",
```
phi-lira: we should remove this.

TestProjects/HDRP_RuntimeTests/Packages/manifest.json
```diff
@@ -1,24 +1,15 @@
...
     "com.unity.ide.visualstudio": "1.0.11",
-    "com.unity.ide.vscode": "1.1.0",
-    "com.unity.multiplayer-hlapi": "1.0.4",
-    "com.unity.purchasing": "2.0.6",
+    "com.unity.ide.vscode": "1.1.3",
```
phi-lira: This should be removed as well.

TestProjects/HDRP_RuntimeTests/Packages/manifest.json
```diff
@@ -1,24 +1,15 @@
...
     "com.unity.shadergraph": "file:../../../com.unity.shadergraph",
-    "com.unity.test-framework": "1.1.1",
+    "com.unity.test-framework": "1.1.5",
     "com.unity.testframework.graphics": "file:../../../com.unity.testframework.graphics",
     "com.unity.testing.hdrp": "file:../../../com.unity.testing.hdrp",
-    "com.unity.textmeshpro": "2.0.1",
-    "com.unity.timeline": "1.2.0",
     "com.unity.ugui": "1.0.0",
     "com.unity.visualeffectgraph": "file:../../../com.unity.visualeffectgraph",
     "com.unity.xr.legacyinputhelpers": "1.3.7",
```
phi-lira: This should be removed as well.

TestProjects/SRP_SmokeTest/Assets/CommonAssets/Scripts/SmokeGraphicsTestSettings.cs
```diff
@@ -0,0 +1,14 @@
+using UnityEngine.TestTools.Graphics;
+
+public class SmokeGraphicsTestSettings : GraphicsTestSettings
```
phi-lira: This is a duplicate of UniversalGraphicsTestSettings. Consider making this generic and moving to graphics test framework so we don't duplicate code. It's fine to do in follow up PR.

TestProjects/SRP_SmokeTest/Assets/Tests/SmokeTests.cs
```diff
@@ -0,0 +1,28 @@
...
+using UnityEngine;
+using UnityEngine.TestTools;
+
+namespace Tests
+{
+    public class SmokeTests
+    {
+        // A Test behaves as an ordinary method
+        [Test]
+        public void SmokeTestsSimplePasses()
```
phi-lira: These tests will always pass? Why adding them?

TestProjects/VisualEffectGraph/Packages/manifest.json
phi-lira: I'm not sure if this applied to modules, but there was an issue that if you declared explicitly (instead of letting them get added implicictly) something breaks because it gets confused about packages that depend on each other.

TestProjects/UniversalGraphicsTest/Packages/manifest.json
```diff
@@ -2,22 +2,15 @@
   "dependencies": {
     "com.unity.2d.sprite": "1.0.0",
     "com.unity.2d.tilemap": "1.0.0",
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.3.2",
-    "com.unity.collab-proxy": "1.2.16",
     "com.unity.ext.nunit": "1.0.0",
-    "com.unity.ide.rider": "1.1.0",
+    "com.unity.ide.rider": "1.1.1",
```
phi-lira: these are not required.

TestProjects/UniversalGraphicsTest/Packages/manifest.json
```diff
@@ -2,22 +2,15 @@
...
     "com.unity.ide.visualstudio": "1.0.11",
-    "com.unity.ide.vscode": "1.1.2",
-    "com.unity.multiplayer-hlapi": "1.0.4",
-    "com.unity.purchasing": "2.0.6",
+    "com.unity.ide.vscode": "1.1.3",
```
phi-lira: not required.

TestProjects/HDRP_RuntimeTests/Packages/manifest.json
sophiaaar: which part? im not sure which line this comment is targeting

TestProjects/SRP_SmokeTest/Assets/Tests/SmokeTests.cs
sophiaaar: just so that they're there for checking builds work and theres no console errors etc

TestProjects/UniversalGraphicsTest/Packages/manifest.json
sophiaaar: with auto update turned off, the version numbers wont bump automatically - this change was checked by me locally to ensure that we could safely update this one

sophiaaar:
spoke to felipe about his comments via dm

PR #5005, opened 2019-11-12
Changing the behavior for ray tracing in case a submesh has both opaque and transparent sub meshes

anisunity:
All in the title, like we discussed


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassRaytracingGBuffer.hlsl
```diff
@@ -48,6 +48,9 @@ void ClosestHitGBuffer(inout RayIntersectionGBuffer rayIntersectionGbuffer : SV_
 [shader("anyhit")]
 void AnyHitGBuffer(inout RayIntersectionGBuffer rayIntersectionGbuffer : SV_RayPayload, AttributeData attributeData : SV_IntersectionAttributes)
 {
+#ifdef _SURFACE_TYPE_TRANSPARENT
+    IgnoreHit();
```
sebastienlagarde: Is it possible to just not declare the AnyHitGBuffer at all with the #ifdef?

anisunity:
@sebastienlagarde is your question the one we discussed?

anisunity:
A use case example would be to take a mesh renderer with multiple submeshes some of them transparent and some of them opaque. Enable ray traced reflection on a mirror an observe what appears/disappears with this change. It would be also good to see how the behavior changes as all the objects become transparent or opaque.

anisunity:
the bug reported by @remi-chapelain happening when the number of materials is lower than the number of submeshes has been included in this PR

PR #5013, opened 2019-11-13
Fix cubemap convolution first frame

alelievr:
### Purpose of this PR
Fixed black cubemap convolution with baked reflection probe at the first frame

This PR disables the async compilation of a bunch of shaders needed for the rendering during the first frame of HDRP.

- [ ] Built a player
- Other:


sebastienlagarde:
Merged those commit in another PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5010

PR #5016, opened 2019-11-14
Add Bellevue QA to SG code owners

pbbastian:
### Purpose of this PR
Adds Bellevue QA as code co-owner of Shader Graph folder.


zadi0648:
So basically whenever code changes to Shader Graph package happens we will be one of the "default" reviewers, correct? Sounds good. As well as this means our reviews alone will grant merge access? Anything else this entails we should know about?

sebastienlagarde:
Yes you understand it correctly. No other effect.

PR #5017, opened 2019-11-14
add UnityStereoScreenSpaceUVAdjust function to Core.hlsl

arumons:
---
### Purpose of this PR
Core.hlsl doesn't include UnityStereoScreenSpaceUVAdjust function.
I copy it from UnityCG.cginc.

---
### Testing status

- [x] Works on android device(aquos r3) + cardboard 
- [x] Built a player

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
+++ b/com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
@@ -246,6 +246,18 @@ half3 MixFog(real3 fragColor, real fogFactor)
 
 #endif
 
+inline float2 UnityStereoScreenSpaceUVAdjustInternal(float2 uv, float4 scaleAndOffset)
+{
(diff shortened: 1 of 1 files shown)


phi-lira:
Hi @arumons we have API to sample stereo textures. I'm wondering if this API is needed. Could you elaborate on the use case? I'm adding XR developers as reviewers as they have more knowledge about this than me.

phi-lira:
@arumons seems like we already support these functions. Look at `TransformStereoScreenSpaceTex`. We are moving to support macros so you can sample XR textures without worrying to transform uv to correct eye space.

arumons:
@phi-lira 
thanks for reply!

ok. I will try to use TransformStereoScreenSpaceTex.

I looked this Blit Section ( https://docs.unity3d.com/2017.4/Documentation/Manual/SinglePassStereoRendering.html )

PR #5020, opened 2019-11-14
Adding an optional bilateral pass to filter the volumetric lighting and a test scene for it

anisunity:
This PR add an option (on the volume settings) to blur the volumetric lighting with a gaussian filter after it has been generated.
This help to reduce under sampling artifact in volumetric.

Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252FVolumetricFiltering/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP_fast-trunk


com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -747,7 +749,8 @@ VolumetricLightingParameters PrepareVolumetricLightingParameters(HDCamera hdCame
 
             parameters.volumetricLightingCS = m_VolumetricLightingCS;
             parameters.volumetricLightingKernel = (parameters.tiledLighting ? 1 : 0) | (parameters.enableReprojection ? 2 : 0) | (enableAnisotropy ? 4 : 0) | (highQuality ? 8 : 0);
-
+            parameters.volumetricFilteringKernelX = 16;
```
sebastienlagarde: why is this hardcoded?

sebastienlagarde:
Please add a changelog. thanks

anisunity:
will do

anisunity:
@sebastienlagarde done

EvgeniiG:
No screenshots?

EvgeniiG:
What's the cost?

sebastienlagarde:
Cost is around 1ms on PS4. So not cheap, but in ratio of the cost of volumetric, not bad. 
Yes screenshots would have been nice. I have seen it on test scene with tree and directional shadow, and few other example. Discuss with Laurent and it is a valuable improvement. It is optional, so users still say if it want it or not, as it is a blur you can lose details with density texture, but overall it is ok. A bit what was done for Megacity.

EvgeniiG:
No reprojection, no filtering.
You see objectionable artifacts due to the fact that we (temporarily) use white noise.

image

No reprojection, but with filtering.
Removes high frequency artifacts, but low frequency artifacts remain.

image

Reprojection, no filtering - artifacts are essentially gone.

image

EvgeniiG:
If aliasing is due to shadows in your scene with the tree, the correct solution is to prefilter the shadows, not blur the results.

EvgeniiG:
Also, it appears to be a separable Gaussian filter rather than a bilateral filter (which makes sense in this case).

EvgeniiG:
Adding Fabian to make sure this works in VR.

sebastienlagarde:
@EvgeniiG FYI there is now test on Yamato to validate that it work in VR. It is named:
HDRP on Windows64_DX11_playmode XR

And indeed it doesn't work :), the filter only correct one eye, not the other.

anisunity:
I'll be looking why the test fails with the vr config

anisunity:
I checked, even without filtering, the test does not have the same result as the non XR version. I disabled the test and i'll add it to the confluence page.

PR #4802, opened 2019-10-09
URP/Shadergraph/XR GPU Instancing Tests

soraryu:
### Purpose of this PR
Currently, enabling "GPU Instancing" on shadergraph shaders breaks them. Sometimes on Desktop, sometimes in VR. This is not tested against yet.
This PR adds
- a test inside ShaderGraph testing with materials that enable "GPU Instancing"
- a test inside ShaderGraphUniversalStereo testing which is essentially the same but in XR

---
### Testing status

**Manual Tests**: What did you do?
- [X] Opened test project + Run graphic tests locally
- [X] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [X] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

_InstancedRendering test inside ShaderGraph project:_
InstancedRendering
Note that this fails when you disable SRP Batcher on the URPAsset:
image
Also note that even when it works the shaders have errors and a lot of warnings:
image

_InstancedRendering test inside ShaderGraphUniversalStereo project:_
002_xr_InstancedRendering
Note that his currently fails as well on all XR rendering modes:
image

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low (stuff needs to be fixed so these tests pass again)

---
### Comments to reviewers
I'd like to point out that I'd consider the underlying issues **high priority**. Those shader errors even prevent making builds.

PR #4948, opened 2019-11-02
Fixed XR shader failure in URP ShaderGraph

thomas-zeng:
### Checklist for PR maker
---
### Purpose of this PR
- https://fogbugz.unity3d.com/f/cases/1185272/
- https://fogbugz.unity3d.com/f/cases/1188487/
- https://fogbugz.unity3d.com/f/cases/1195207/
---
### Testing status

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [x] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [x] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/urp%252Fxr%252Ffix-sg-compile-err

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
- HDRP might have similar issues with Shader Graph. HDSubShaderUtilities.cs will require similar changes to be XR aware.
- QA: please save Shader graph asset and clear shader cache before testing.

PR #4803, opened 2019-10-09
Added Test for sampling OpaqueTexture in XR through ShaderGraph

soraryu:
### Purpose of this PR
Opaque Texture sampling in ShaderGraph currently breaks in XR. It's used for a wide variety of effects, e.g. water, blurry refractions, translucency...
This PR adds a test for that.

### Testing status

**Manual Tests**: What did you do?
- [X] Opened test project + Run graphic tests locally
- [X] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo
- [X] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)
New test scene in ShaderGraphUniversalStereo, 003_xr_OpaqueTexture (numbering makes sense with my previous PR which has 002).

_Reference Image_
003_xr_OpaqueTexture

_Multi Pass_
191009-111316-Unity

_Single Pass_

_Single Pass Instanced_

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

---
### Overall Product Risks
**Technical Risk**: None

**Halo Effect**: Low (bugs need to be fixed to make these tests pass)

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #5022, opened 2019-11-14
Workaround for Single-Pass and Single-Pass Instanced testing issues in #4802

soraryu:
### Purpose of this PR
#4802 is currently blocked because the Graphics Test Framework is not able to actually test Single Pass and Single Pass Instanced due to the method used for capturing actual frame images.

This PR fixes that by using ScreenCapture.CaptureScreenshot, which should actually be the thing used for all graphics tests as it captures the final result and does not do some hard-to-fix reconstruction based on a list of active cameras.

Once this PR is approved (or before if you think that makes sense) I'd like to move the functionality that I added and which is currently only available in the ShaderGraphUniversalStereo project into the Graphics Testing Framework and make it available for all tests.

Also, hopefully there is a broader discussion around what exactly the Test Framework is supposed to test (e.g. how are reference images for Quest captured - this should be exactly the images from Quest, not images captured in Editor, to enable accurate tests on output and not on arbitrary RenderTextures). I think testing the output of ScreenCapture.CaptureScreenshot makes sense, as this is what should actually be tested, even if it makes setup (making sure window sizes match etc.) a bit more complicated.

---
### Testing status

- [X] Built a player
- Other: this enables more tests to run correctly but requires some additional changes. Notably, built players currently would need to produce exactly the image size as the reference image. The current code takes care of resizing the player to match the reference image, but ideally this means ref images should be captured from target platforms (e.g. for Quest, where player is not resizable).

I tested this by merging it into #4802 where tests currently (incorrectly) fail, with this PR they succeed.

#4802 is a good testcase of the workarounds / improvements here.

---
### Comments to reviewers

### DIFF
--- a/TestProjects/ShaderGraphUniversalStereo/Assets/CommonAssets/Scripts/LWGraphicsTests.cs
+++ b/TestProjects/ShaderGraphUniversalStereo/Assets/CommonAssets/Scripts/LWGraphicsTests.cs
@@ -6,6 +6,7 @@ using UnityEngine.TestTools;
 using UnityEngine.XR;
 using UnityEngine.TestTools.Graphics;
 using UnityEngine.SceneManagement;
+using System.IO;
 
 public class LWGraphicsTests
 {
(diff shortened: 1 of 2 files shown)


soraryu:
@phi-lira this PR is supposed to be the base for merging in #4802 and also for adding more tests around #4948.

DennisDeRykeUnity:
I made dev branch xr_sg_urp_dennisd from SRP master then merged in three dev branches from prefrontalcortex remote per three PRs to be reviewed:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4803
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5022
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4802

On Yamato I have launched "_ABV for SRP repository - fast-trunk" against branch xr_sg_urp_dennisd per link below.  Tomorrow I will compare these test results against SRP master. Also tomorrow in the Editor I plan to examine the new tests introduced by these PRs, and tinker with the screen capture workaround.  
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/xr_sg_urp_dennisd/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/754717/job

@soraryu Thank you for your contributions.  Sorry that it has taken me a while to start reviewing these.

DennisDeRykeUnity:
@soraryu I am closing three of your PRs (#4803, #5022, and#4802) because I merged them into PR #5215 which has been merged into master.  See discussion above.  Thanks again for your contributions!

PR #5024, opened 2019-11-14
Custom pass fixes 3

alelievr:
### Purpose of this PR

+ Add another culling call dedicated for custom passes so we can draw objects that are culled by the camera.
+ Added two new injection points: `AfterDepthAndNormal` and `BeforePreRefraction`
+ Remove the default material override in DrawRenderers
+ Make custom fullscreen passes work in VR
+ Added an enum ShaderPass to select which shader passes to render when there is  no material override in a DrawRenderers. Currently available: `DepthPrepass` and `Forward` (This changes the list of shader tags sent to DrawRenderers).
image
+ Added PerObject data for DrawRenderers (this allows to draw Lit objects)
+ Added the custom pass volume in the camera volume culling flags
image
+ Added a None option when selecting target buffers
image
+ Fix override material indices broken in build players due to shader stripping  (some objects weren't drawn in build due to that)
+ Added Lazy allocation of custom buffers so memory is allocated only if it's referenced from the UI or get from script.
+ Added `Custom Pass` in the create volume entry
image

Added @RemyUnity for awareness but you should already know all these features.

Bonus:
[Liquid2](https://github.com/alelievr/HDRP-Custom-Passes#liquid)


com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass.md
```diff
@@ -23,6 +23,18 @@ Here you can see an example of a custom pass with a box collider (solid transpar
...
 
+## Injection Points
+
+Custom passes can be injected at 6 different places. Note that these injection points **don't tell you exactly where you pass will be executed**, the only guarantee we give is that a certain list of buffers will be available for **Read** or **Write** and will contain a certain subset of objects rendered before your pass. Unlike Universal Render Pipeline, we don't give you the guarantee that your passes will be executed before or after any HDRP internal operation (also due to our async rendering). However we guarantee that the injection points will be triggered from top to bottom in this order in a frame:
+Name | Available Buffers | Description
+--- | --- | ---
+BeforeRendering | Depth (Write) | Just after the depth clear, you can write to the depth buffer so Z-Tested opaque objects won't be rendered. It's useful to mask a part of the rendering. Here you can also clear the buffer you allocated or the `Custom Buffer`
+AfterOpaqueDepthAndNormal | Depth (Read \| Write), Normal (Read \| Write) | Buffers will contain all opaque objects. Here you can modify the normal and depth buffer, it will be taken in account in the lighting and the depth pyramid.
```
sebastienlagarde: Add that there is also the roughness at this step with the normal

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass.md
```diff
@@ -23,6 +23,18 @@ Here you can see an example of a custom pass with a box collider (solid transpar
...
+BeforePreRefraction | Color (no pyramid \| Read \| Write), Depth (Read \| Write), Normal (Read) | Buffers will contain all opaque objects. Use this pass to render any transparent objects that you want to be in the color pyramid (and in the refraction).
+BeforeTransparent | Color (Pyramid \| Read \| Write), Depth (Read \| Write), Normal (Read) | Here you can sample the color pyramid. It's useful to do some blur effects but note that all objects rendered at this point won't be in the color pyramid. You can also draw some transparent objects here that needs to refract the scene.
```
sebastienlagarde: Caution because we have 2 color pyramid in HDRP, one for refraction, the other for distortion and SSR. Better to say that an object will be visible in refraction or not (but still talk about color pyramid as it may be of interest for users)

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass.md
```diff
@@ -199,6 +211,9 @@ To code your custom pass, you have three entry point:
...
 
+> **Important:** if the a is never referenced in any of your scenes it won't get built and the effect will not work when running the game outside of the editor. Either add it to a [Resources folder](https://docs.unity3d.com/Manual/LoadingResourcesatRuntime.html) or put it in the **Always Included Shaders** list in `Edit -> Project Settings -> Graphics`. Be careful with this especially if you load shaders using `Shader.Find()` otherwise, you'll end up with a black screen.
```
sebastienlagarde: " if the a" ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2123,10 +2132,6 @@ void Callback(CommandBuffer c, HDCamera cam)
 
                 RenderSky(hdCamera, cmd);
 
-                // To allow users to fetch the current color buffer, we temporarily bind the camera color buffer
-                cmd.SetGlobalTexture(HDShaderIDs._ColorPyramidTexture, m_CameraColorBuffer);
-                RenderCustomPass(renderContext, cmd, hdCamera, cullingResults, CustomPassInjectionPoint.BeforeTransparent);
-
                 m_PostProcessSystem.DoUserBeforeTransparent(cmd, hdCamera, m_CameraColorBuffer);
```
sebastienlagarde: naming for before transparent for postprocess need to be change

sebastienlagarde: rename afterOpqueandsky

PR #5025, opened 2019-11-14
Invalidating the history for directional, point and spot lights for ray traced screen space shadows when the light has moved

anisunity:
Fixes ghosting issues with these lights.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -862,6 +864,11 @@ void CleanupLightLoop()
             CoreUtils.Destroy(m_CopyStencilForSSR);
         }
 
+        void LightLoopNewRender()
```
sebastienlagarde: We already have a LightLoopNewFrame please use it instead of LightLoopNewRender

PR #5027, opened 2019-11-15
Nov-2019 Shader Graph Bug Fixes

zadi0648:
Just fixin' some bugs

_Misc Info:_
* I appear to use a different IDE than all of you (JetBrains Rider) which automatically removes unnecessary white space in files I alter. Pretty nice but potentially annoying for code reviewers? (Let me know)
* Changelog: There appeared to be two Fixes sections for this release, presumably a mistake?

---
**New:**

* Fixed an unreported bug where changing anything about a Keyword Entry causes its ID to be reset!
Repro:
1) Create a new Enum Keyword.
2) Remove Entry B.
3) Drag the Enum Keyword into the graph.
4) Plug another node into Port C.
5) Back on the blackboard, Rename C, C to B, B
> The Nodes becomes disconnected, because the KeywordEntry's ID will have been changed to 2 (was 3 before).

---
**Fix 1:**
Changing a Shader Input's Precision didn't properly update node colors inside of the graph (when using Precision for the Color Mode).

https://fogbugz.unity3d.com/f/cases/1181557/

Before:
1181557_Before

After:
1181557_Fix

**Notes:**
* BlackboardFieldPropertyView.cs now informs the color manager to update on DirtyNodes

---
**Fix 2:**
Enum Keyword Nodes didn't properly update after the # of keyword entries got altered, which would crash the entire graph on the addition of a new keyword entries.

https://fogbugz.unity3d.com/f/cases/1188890/

Before:
1188890_Before

After:
1188890_After

**Notes:**
* _Automation:_ I wanted to add a regression test for this, however as we know add/removing through the Reorderable List wasn't feasible (in general to the blackboard as made clear by the lack of automation testing for the blackboard), so I added test for adding & removing Enum Keyword enteries to a Keyword. Hopefully one day we can have better automation for the blackboard!
* I also discovered & fixed an unreported bug where new entries could result in duplicate entry being added, causing the Keyword Node's input ports to not match the number of Enum Entries:
KeywordViewIDBug_Before
For example this could be cause by an enum having it's 2nd entry being removed, resulting in two entries with ids 1 & 3. Adding a new entry would make another entry that also has id 3.
KeywordViewIDBug_After
Now a helper method helps find the first used id to use (last port being pink is a known preexisting issue).
* Created a const inside of KeywordNode to remove magic numbers.
* On removal of Enum Keyword Entries, the reorderable list now remains selected allowing for quick rapid removal of entries.

---
**Fix 3:**
Convert to subgraphs could result in duplicate names.

https://fogbugz.unity3d.com/f/cases/1150630/

Before:
1150630_Before

After:
1150630_After

**Notes:**
* Simple fix, the names now get sanitized.
* The order of them is messed up because of preexisting regression that doesn't allow blackboard reordering, I checked that it is on order once fixed (a fix exists on another branch).

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252F11-2019-bug-fixes/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk/1327389/job
Master:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/master/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk/1327409/job

OLD:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252F11-2019-bug-fixes/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk


LandonTownsendUnity:
#1 confirmed fixed by PR

martint-unity:
Have you run Yamato on this?

com.unity.shadergraph/Editor/Drawing/Blackboard/BlackboardFieldKeywordView.cs
```diff
@@ -190,19 +190,41 @@ private void AddEntry(ReorderableList list)
...
 
-            // Update GUI
+            // Update Blackboard & Nodes
+            DirtyNodes();
             Rebuild();
             graph.OnKeywordChanged();
             m_SelectedIndex = list.list.Count - 1;
         }
 
+        private int GetFirstUnusedIndex()
```
marctem: This function is pretty hard to follow. Zack and I discussed and realized we need to think more about how to serialize enum entry IDs so they're stable in the UI ordering across deletes and re-ordering.

zadi0648:
Ready to go 👍

PR #5029, opened 2019-11-15
Fix issue with contact shadow when max distance is close to far clip plane

FrancescoC-unity:
We march outside the range of the depth buffer if contact shadow max distance is bigger than far clip plane (or close to)


sebastienlagarde:
please add changelog

PR #5031, opened 2019-11-15
Fixing meta that are conflicting between Universal and HDRP

sebastienlagarde:
Fixing meta that are conflicting when both Universal and HDRP package are present in a project

Note: No yamato launch as this isn't cover by Yamato

For graphic QA: Please load a regular scene on 2019.3. And reference in the manifest both UP and HDRP (from Github reposiroty). You shouldn't see any conflicting meta file (but still have RemoveComponent error)


sebastienlagarde:
Validated by QA, André is aware, merging.

PR #5042, opened 2019-11-15
Shadergraph/fix/stencil

kaychang-unity:
---
### Purpose of this PR
StencilOverride parameters for shadergraph shaders was not injected into the final shaderLab code. It is now fixed.
This feature is currently not used, but will be used once deferred renderer for UniversalRP is merged into trunk (coming soon).

---
### Testing status
Checked TestProjects\UniversalGraphicsTest\Assets\Scenes\050_Shader_Graphs.unity: OK.
Ran UniversalGraphicsTest graphics tests locally: OK.

- [ X ] Built a player

**Yamato**: (shadergraph/fix/stencil):
Started job: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/shadergraph%252Ffix%252Fstencil/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/742595

---


phi-lira:
I think we should backport this to 19.3/19.4. If so please add the tag `needs-backport-2019.3`

kaychang-unity:
Will the backport to 2019.3 be done automatically or is there something else I need to do to make it happen?

marctem:
> I think we should backport this to 19.3/19.4. If so please add the tag `needs-backport-2019.3`

Why would we need to backport this if it only affects deferred which won't land until 20.1?

sebastienlagarde:
>Will the backport to 2019.3 be done automatically or is there something else I need to do to make it >happen?
Nothing is automatic, we check tag then backport manually or ask people to backport

regarding GenerateShaderPass in GenerationUtils.cs that this PR affect, is it used at all ? 

Both HDRP and UP have their own GenerateShaderPass  (in HDSubShaderUtilities) and in 3 different files in UP. Or there is common code (but specific to UP) in ShaderGraph ?

kaychang-unity:
@marctem I think it is technically a bug. Since some external teams could try to use the StencilOverride interface with no effect.

kaychang-unity:
> regarding GenerateShaderPass in GenerationUtils.cs that this PR affect, is it used at all ?
@sebastienlagarde  UniversalRP uses this code path.
`GenerateShaderPass` in GenerationUtils.cs ultimately calls `ShaderSpliceUtil.TemplatePreprocessor()`, so does HDRP. I don't know why the code in GenerationUtils.cs is not re-used in HDRP.

`ShaderSpliceUtil.TemplatePreprocessor()` is the one that will insert the stencil states using the template file PassMesh.template.

I saw HDRP also has code that set `StencilOverride`. Not sure if it has been working so far.

alelievr:
@kaychang-unity HDRP have it's own shader passes and so does not uses `PassMesh.template`. We directly uses `ShaderSpliceUtil.TemplatePreprocessor` on our pass template files to replace the blend modes, and as you can see here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template#L15 we have all the blend modes (stencil included) in our templates.
That's what StencilOverride is replacing at the end and HDRP ShaderGraphs would be working without it.

kaychang-unity:
@alelievr I see thanks!
So, this PR will not break anything in HDRP.

kaychang-unity:
@gfx-qa-bellevue, waiting on you!

LandonTownsendUnity:
I added the StencilOverride lines to the m_forwardpass like you said
but the lines didn't end up in the PBR generated shader
image
image

PR #5043, opened 2019-11-15
Fixed missing normalize operation on vertex normals.

kaychang-unity:
### Purpose of this PR
For some shader variants, speedtree shader is missing a normalize operation on vertex normals.
The bug was discovered because when deferred pipeline is used, the normals are implicitely normalized (when stored into the gbuffer). As a result, the lighting between the forward renderer and the deferred renderer was slightly different. This fixes the issue.

The deferred renderer currently lives in its own branch but will be merged into master branch soon.

NormalizeNormalPerPixel() actually only _optionally_ normalize the normal depending on platforms. It will not do it on mobile/Switch platforms, but will do it on other platforms. Anyhow, this fix will allow the editor/desktop platforms to have identical results between forward and deferred renderer.

Interestingly, the result images for scenes 057 and 058 in UniversalGraphicsTests do not change with this fix, because the trees are too far from the camera. A change in lighting (lighting is now correct) is only visible if the camera gets closer.

### Testing status

**Manual Tests**:  Ran UniversalGraphicsTests locally: OK

**Yamato**: (universal/fix/speedtree):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Ffix%252Fspeedtree/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_trunk/744192

---
### Comments to reviewers
-


phi-lira:
I think we should backport this to 2019.3/4. If so please add the `needs-backport-2019.3` tag.

phi-lira:
@russoh did you had the chance to look at this PR?

PR #5046, opened 2019-11-18
Exposed Serialized

zadi0648:
Main PR: #5045

A removed part of 20% branch because I'd figured it was like a no go due to Martin's work on the topic 😅, but perhaps worth discussing.

Makes Exposed Serialized.
Pros:
* Exposed status is now duplicated, something I often find I want.
* Remembered always; between closing & open files, opening & closing, unity sessions, moving between graphs, ect.
* Kept when copying and pasting between graphs too.
Cons:
* Changing the exposed status of a property will now renders the graph dirty (trigging needs-save), although this field could be ignored in the function that handles that.
* More data == bad ?

Misc thoughts:
* If we don't roll with an ability to duplicate a Shader Input's exposed status, I think it may be best to set exposed to false by default. Of course this may also be resolved by the internal inspector.
* Otherwise, could be worth looking into Duplication remembering Exposed status after your changes land Martin.

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Graphs/BooleanShaderProperty.cs
+++ b/com.unity.shadergraph/Editor/Data/Graphs/BooleanShaderProperty.cs
@@ -46,6 +46,7 @@ namespace UnityEditor.ShaderGraph.Internal
             {
                 displayName = displayName,
                 hidden = hidden,
+                expanded = expanded,
                 value = value
             };
         }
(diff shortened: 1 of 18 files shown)


martint-unity:
I don't think we want anything like this serialized. 
If we need this save between sessions we could save it to prefs instead of just the session.
Also we are going over to a Inspector workflow which means these foldies will no longer be needed.

PR #5049, opened 2019-11-18
Update changelog date

sebastienlagarde:
Update changelog date


sebastienlagarde:
>this can just be that when the PR author adds a line to the changelog, they also update the date next >to the version

we just need to do it when we release a package Otherwise we will constantly conflict when merging PR.

sophiaaar:
> > this can just be that when the PR author adds a line to the changelog, they also update the date next >to the version
> 
> we just need to do it when we release a package Otherwise we will constantly conflict when merging PR.

But isn't updating the changelog part of the PR process anyway?

sebastienlagarde:
>But isn't updating the changelog part of the PR process anyway?
yes. Indeed updating the date could be done at the same time. I just bet that a lot of people will forget it, and in itself we just need to update it before releasing the package. In particular it cause conflict when PR are waiting for approval and so are merge at a later date.

PR #5059, opened 2019-11-18
Replace deprecated XR code

fabien-unity:
### Purpose of this PR
PlayerSettings.virtualRealitySupported is getting deprecated with 2019.3 and removed in 2020.1

---
### Testing status

**Yamato**: [running](https://yamato.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/xr-replace-vr-supported/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/752729/job)


sebastienlagarde:
Is it this PR that is depenednt on new MR windows package?

com.unity.render-pipelines.core/Runtime/Common/XRGraphics.cs
```diff
@@ -47,7 +47,14 @@ public static float renderViewportScale
         // TryEnable gets updated before "play" is pressed- we use this for updating GUI only.
         public static bool tryEnable
         {
-            get { return PlayerSettings.virtualRealitySupported; }
+            get
+            {
+            #if UNITY_2020_1_OR_NEWER
+                return false;
```
phi-lira: For context. Why in 2020.1 we always return false?

fabien-unity: It's because the underlying code "legacy VR" will be removed from 2020.1.
XR SDK will be the only way to enable VR with 2020.1

phi-lira:
Had offline discussion about this with @fabien-unity . Current understanding is that this doesn't break URP because it supports XR shim path.

PR #5061, opened 2019-11-19
Add XR multipass rendering for Universal pipeline (version 2)

mikesfbay:
### Purpose of this PR
Added XR multipass rendering for Universal pipeline (version 2). Multipass rendering is a requirement on many VR platforms and allows graceful fallback when single-pass rendering isn't available. 

Fix for fogbugz case: https://fogbugz.unity3d.com/f/cases/1182204/

---
### Testing status

Tested on Oculus Quest and Rift using Template Scene.

- [ x] Built a player

multipass_rift


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -263,6 +263,14 @@ static void InitializeCameraData(UniversalRenderPipelineAsset settings, Camera c
...
             cameraData.camera = camera;
             cameraData.isStereoEnabled = IsStereoEnabled(camera);
+            cameraData.isXRMultipass = false;
+            cameraData.numberOfXRPasses = 1;
+
+            if (cameraData.isStereoEnabled && !cameraData.isSceneViewCamera && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass)
+            {
+                cameraData.numberOfXRPasses = 2;
+                cameraData.isXRMultipass = true;
+            }
```
phi-lira: Is there a case that a camera might need different multipass/xr passes settings? Planar reflection camera, etc?
 The way these are initialized here seems to me we don't really have to expose these in the CameraData (per camera settings) and instead we could simplify logic by exposing it in the ScriptableRenderer.

 If you expose it in ScriptableRenderer it also allows users creating renderers and do something like:
 `ScriptableRenderer.SetXREeyePassCount(int passCount)`

 Another option would be expose in XRSettings?

 What do you think?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -245,5 +248,18 @@ public DrawingSettings CreateDrawingSettings(ShaderTagId shaderTagId, ref Render
             else
                 CoreUtils.SetRenderTarget(cmd, colorAttachment, colorLoadAction, colorStoreAction, clearFlags, clearColor);
         }
+
+        public void SetupXRMultipassState(ScriptableRenderContext context, CommandBuffer cmd, RenderingData renderingData)
```
phi-lira: This is really just settings up camera matrices. I wonder if at this point we shouldn't expose an URP SetupCameraMatrices in RenderingUtils.cs. This could also help abstract matrix setup for the pure XR SDK. (as we want to move CommandBuffer.SetCameraMatrices -> URP.

 Also please add API docs when you do it. We are adding api docs to all URP public methods.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -245,5 +248,18 @@ public DrawingSettings CreateDrawingSettings(ShaderTagId shaderTagId, ref Render
...
+        {
+            if (renderingData.cameraData.isXRMultipass)
+            {
+                Camera.StereoscopicEye eye = (Camera.StereoscopicEye)eyeIndex;
+                Camera camera = renderingData.cameraData.camera;
+                cmd.SetViewProjectionMatrices(camera.GetStereoViewMatrix(eye), camera.GetStereoProjectionMatrix(eye));
+                context.ExecuteCommandBuffer(cmd);
```
phi-lira: We should consider having the option to not execute the command buffer as the client code might want to do more work with the same command buffer before executing it.

We could have overrides, one that takes a context and one that takes a command buffer. The context one creates a cmd buffer and execute it, the command buffer one defers the execution to client code. (as if they are passing already a command buffer they might be wanting to batch command buffer work).

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -114,5 +116,10 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             context.ExecuteCommandBuffer(cmd);
             CommandBufferPool.Release(cmd);
         }
+
+        public override void FrameCleanup(CommandBuffer cmd)
+        {
+            eyeIndex = 0;
```
phi-lira: Seems like this eyeIndex logic is based on some assumptions that are risky. 
 - If a user adds a stereo aware custom render pass and overrides this method to clean up some pass data, they need to call base.FrameCleanup otherwise the eyeIndex will not be reset properly.
 - The SetupXRMultipassState function below increments the eyeIndex. There's nothing guaranteeing that user will not call that function twice for a render pass (for whatever reason).
 
Suggestion. The ScriptableRenderer controls the render pass execution and knows exactly which eye is currently rendering to. Make `eyeIndex` internal prop. Then inside ScriptableRenderer.ExecuteRenderPass we make sure to set the `eyeIndex` in the pass to the correct one. This way we encapsulate `eyeIndex` logic from user and prevents any faulty use of that property to break us.

phi-lira:
Please add an entry to changelog.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
mikesfbay: It's better to leave this in cameraData since XR passes can be configured differently in XR SDK.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
mikesfbay: This function is no longer needed in latest revision 8f0bd8a

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
mikesfbay: This function (and the code block) is no longer needed in latest revision 8f0bd8a

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
mikesfbay: Yes, letting ScriptableRenderer control eyeIndex is a good idea and solves the above issues. I've changed this in latest revision 8f0bd8a

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -245,5 +248,18 @@ public DrawingSettings CreateDrawingSettings(ShaderTagId shaderTagId, ref Render
...
+                cmd.Clear();
+                ++eyeIndex;
```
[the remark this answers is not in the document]
mikesfbay: ScriptableRenderer controls eyeIndex so this isn't an issue.

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -67,7 +67,10 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                 ref CameraData cameraData = ref renderingData.cameraData;
                 Camera camera = cameraData.camera;
                 if (cameraData.isStereoEnabled)
-                    context.StartMultiEye(camera);
+                {
+                    context.StartMultiEye(camera, eyeIndex);
+                    SetupXRMultipassState(context, cmd, renderingData);
```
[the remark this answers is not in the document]
mikesfbay: SetupXRMultipassState is no longer used in latest revision 8f0bd8a. As for the reason why DepthOnlyPass calls StartMultiEye, I believe Robs added it to fix shadows.

com.unity.render-pipelines.universal/Runtime/Passes/DepthOnlyPass.cs
```diff
@@ -67,7 +67,9 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
                 ref CameraData cameraData = ref renderingData.cameraData;
                 Camera camera = cameraData.camera;
                 if (cameraData.isStereoEnabled)
-                    context.StartMultiEye(camera);
+                {
+                    context.StartMultiEye(camera, eyeIndex);
```
thomas-zeng: I suspect this is a redundant call. Do you think you have time to investigate this? I am trying to understand why this is needed for DepthOnlyPass but not for other render passes. Other than this, PR LGTM

mikesfbay: Robs added this call to DepthOnlyPass when he hooked up XR for LWRP. I think it was to fix a mobile shadow issue. We can look into removing it later if QA find no issues.

phi-lira: I also wondered about that for a while. Can we look at it in a following PR? I guess of this PR it's fine.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
phi-lira: Ok. Could we make them internal? I'm concerning about documenting exposed this to users since it's only required for one XR specific mode.

thomas-zeng:
@ryanhy-unity Hey ryan, could you test this PR when you have time? We would like to land this ASAP and backport this to 19.3

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
mikesfbay: Yes, they are changed to internal in revision 1475878

ryanhy-unity:
my URP XR automated tests passed but for one that failed due to a known issue with cascade shadows.  The URP template scene looked good too.  Checked Oculus on mobile (Quest) and Desktop.

PR #5062, opened 2019-11-19
Fix for: "Properties from the Blackboard can be dragged into the Graph when Blackboard is disabled"

zadi0648:
**Summary:**
Fix for https://fogbugz.unity3d.com/f/cases/1187828/

**Update:**
Also fixes this issue as well: https://fogbugz.unity3d.com/f/cases/1211132/

Before:
NoMoreInvisibleBlackboardProperties1

After:
NoMoreInvisibleBlackboardProperties0

Implementation Notes:
* Because uielements.visible could sometimes result in elements still being selectable / draggable, these toggle-able windows are now removed & readded to the view upon toggle.
* ^ This is the same as how VFX works.
* Helper method ensures that the Main Preview & Blackboard are always in the same order (and any new sub windows), they will appear in the reverse order that they appear in ToggleSubWindows(). Fyi using Insert() on something that has already been inserted will simply move it and using RemoveFromHierarchy() on something that has already been removed does nothing, which is why ToggleSubWindows() doesn't bother checking for its current status.

QA Notes:
* This bug is super finicky, and really is a workaround for a bug in Graph View that may never get fixed. Sometimes it was other parts of the blackboard that is select-able (which you will be able to debug via UI Elements Debugger).

Manually tested:
* All combinations of toggle.
* Resizing the graph while the elements are removed, then readding them: location still makes sense.
* That the view order of the sub windows is always consistent with what it was before.
* Using the UI Elements Debugger to confirm removal and order.

---

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Ffix%252F1187828/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk


com.unity.shadergraph/Editor/Drawing/Views/GraphEditorView.cs
```diff
@@ -280,6 +280,19 @@ public GraphEditorView(EditorWindow editorWindow, GraphData graph, MessageManage
             Add(content);
         }
 
+        void ToggleSubWindows()
```
pbbastian: This isn't really toggling, but rather making sure that the sub-windows are shown/hidden. I suggest renaming to this: `UpdateSubWindowsVisibility`

martint-unity:
Have you run Yamato on this?

PR #5072, opened 2019-11-19
Add "Save context" to Shader Graph to update dependent materials only when needed

alelievr:
### Purpose of this PR

When changing a shader graph settings, it could be require to update the dependent materials (Keyword, RenderState etc...). In HDRP this happen for example with subsurface scattering or alpha masking. 
We have previously introduce a callback that load all the material in the project and check if they are matching the current modified shader graph (as there is no depency tracking for material in Unity :( ). 
However, this hack cause a large slowdown when saving a shadergraph in case of large project, sometimes several minutes in big projects, as loading all material can be heavy.

This PR introduce a mechanism only to trigger the HDRP update material when the relevant shader graph settings are changed instead of every time we save the sahdergraph. Don't solve the fundamental issue, but save a large amount of iteration time for 2019.3 LTS.

QA reviewer: Goal is to create a Material from a shader graph, change settings that affect render state on it (SSS, alpha mask), then validated that Material are correctly updated. Then do the same test with a lot of material. See the update time, and do the same test with just changing a node in the shader graph and see that it should be instant (as not material update are trigger).

Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix%252Fimprove_update_materials/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk


com.unity.shadergraph/Editor/Data/Nodes/MasterNode.cs
```diff
@@ -45,6 +45,8 @@ protected virtual VisualElement CreateCommonSettingsElement()
             return null;
         }
 
+        public virtual SaveContext GetSaveContext() => new SaveContext();
```
pbbastian:
```suggestion
        public virtual object saveContext => null;
```

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -316,7 +316,7 @@ public AbstractMaterialNode outputNode
 
         public bool didActiveOutputNodeChange { get; set; }
 
-        internal delegate void SaveGraphDelegate(Shader shader);
+        internal delegate void SaveGraphDelegate(SaveContext shader);
```
pbbastian: I'd rather have the API look like this, and then get rid of the base `SaveContext`. Then we remove the need for a HDRP/URP class inheriting a Shader Graph class.
```suggestion
        internal delegate void SaveGraphDelegate(Shader shader, object context);
```

com.unity.shadergraph/Editor/Data/Nodes/MasterNode.cs
alelievr: Fine for me

PR #5086, opened 2019-11-20
Cookie warning

alelievr:
### Purpose of this PR
- Adds a warning on the cookie field of the light if the texture type is Cookie:
image

Note that there is no warning for point lights (as there is no cookie cube option in the texture importer).  
I also added a button that will automatically fix the setting on the texture.


iM0ve:
The warning does appear if texture type is set to cookie
Fix button does change the cookie type to **Texture**  
However on staging the Light cookies dont work at all(works fine on 7.1.5).
Screen Shot 2019-11-21 at 16 18 23

iM0ve:
With latest HDRP/staging cookies are working now
Screen Shot 2019-11-21 at 18 32 05

PR #5112, opened 2019-11-21
Fix nullref when path tracing is enabled

FrancescoC-unity:
Light cluster was not available when path tracing is evaluated


anisunity:
No it was moved due to a dependecy on other passes and given  path tracing is orthogonal

PR #4931, opened 2019-10-30
Universal/docs/post processing docs

reallyfancy:
---
### Purpose of this PR
Added docs for the following post-processing effects: Bloom, Chromatic Aberration, Depth of Field, Tonemapping, Vignette. Also updated table of contents, and fixed a couple of typos/old names.

Changes reviewed by Thomas Hourdel.

PR #5129, opened 2019-11-22
Universal/2019.3/backports

phi-lira:
- Backport of fixes to yamato. 
 - Backport of #4931


sebastienlagarde:
Yamato is green, the erro with mac is the same than on release/2019.3. (and we need a way to fix it :))

PR #5151, opened 2019-11-25
Fix for corrupt node to sub graph conversion

zadi0648:
Fix for https://fogbugz.unity3d.com/f/cases/1200214/

---
### Implementation Notes

* The problem was caused because Reload wouldn't update the number of slots if the Sub Graph has any error at all. Afterwards during string building the # of Sub Graph asset outputs did not match the # of outputs the node contained which caused the crash.
* Since UpdateSlots() will return if asset is null (meaning a more serious error has occurred), calling it on Sub Graphs that do contain errors seems worry free.
* Reload still doesn't not attempt to re-validate for nodes that do contain an error: only update ports.
* Changing validation seemed risky, or rather solid as it currently is?

---
### Testing status

* Repro steps from the Bug
* Tested add/removing inputs to other Sub Graph w/ errors, including:
- Recursive (using the ones from this case https://fogbugz.unity3d.com/f/cases/1138262/)
- Corrupted Sub Graphs (although this often still yielded in a crash, or even Unity crash, which is fine because deliberately corrupting your sub graphs is not supported lol...)


sebastienlagarde:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Ffix%252F1200214/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_fast-trunk

sebastienlagarde:
test are green

PR #5011, opened 2019-11-13
CoreRP/Catch srp loading error while assetdatabase not available [Need C++ PR]

RSlysz:
## Purpose of this PR
Remove error while SRP load resources at creation and the AssetDatabase is already importing
Need trunk PR: 
- https://ono.unity3d.com/unity/unity/pull-request/96486/_/graphics/catch-srp-loading-error-while-assetdatabase-not-available (2020.1)
- https://ono.unity3d.com/unity/unity/pull-request/96488/_/2019.3/graphics/catch-srp-loading-error-while-assetdatabase-not-available (2019.3)

Linked FogBugz: https://fogbugz.unity3d.com/f/cases/1188429/

---
### Testing status

**Manual Tests**: 
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- Other: 
Manually tested that while HDRP package is being upgraded to newest version (7.1.1 to last HDRP/staging where this branch where before rebasing) remove every exception regarding the AssetDatabase not ready.
Also temporarily edited trunk code to verify that filtered exception where the ones that we wanted to remove.

**Automated Tests**: 

---
### Comments to reviewers
@phi-lira can you check that the last commit cover your issue and there is no side effect on universal? (And 2D)

PR #5172, opened 2019-11-25
Lw/fix lazy resources reloading

RSlysz:
### Purpose of this PR
Fix lazy init in Universal:
- lazy init of editor resources
- resources for Forward and 2D are not created when constructing Universal but later
Added lazy init pattern in CoreRP for these cases.

---
### Comments to reviewers
Previous the first PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5011 , the try catch catched resources but never rethrow exception. So the comportment should be still the same except that wrong path set in the attribute will be thrown. (better for debuggability)


phi-lira:
I'll force merge this PR as it gets enough coverage and we need it to fix an issue asap to release a package.

PR #5178, opened 2019-11-25
 Added callback to SRP to hook up pipeline MSAA settings.

phi-lira:
### Purpose of this PR
Removed dependency from QualitySettings.antiAliasing from URP by exposing a delegate to allow Unity to query antialiasing from SRP.

This fixes 1195272 (https://issuetracker.unity3d.com/issues/lwrp-the-anti-alias-quality-settings-value-is-changing-without-user-interaction)

---
### Testing status

- [ ] Built a player
- Other: 

**Automated Tests**:  Already covered by tests.

---
### Comments to reviewers
 This requires engine PR that exposes `antiAliasingSampleCount` in the RenderPipeline. All places that engine relies on QualitySettings.antiAliasing have been updated to call this new API in SRP to return pipeline msaa samples. Default values for any SRP is 1 (no msaa).

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
@@ -192,6 +192,11 @@ namespace UnityEngine.Rendering.Universal
 
             SortStable(m_ActiveRenderPassQueue);
 
+#if ENABLE_VR
+            if (renderingData.cameraData.isStereoEnabled)
+                XR.XRDevice.UpdateEyeTextureMSAASetting();
+#endif
+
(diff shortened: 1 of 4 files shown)


com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -192,6 +192,11 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
 
             SortStable(m_ActiveRenderPassQueue);
 
+#if ENABLE_VR
+            if (renderingData.cameraData.isStereoEnabled)
+                XR.XRDevice.UpdateEyeTextureMSAASetting();
+#endif
```
ellioman: Do we need to add the VR_MODULE check here as well?

PR #5191, opened 2019-11-25
Fix package test: jpg are not valid in package , convert to png

sebastienlagarde:
Fix package test: jpg are not valid in package , convert to png


sebastienlagarde:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-package/.yamato%252Fupm-ci-packages.yml%2523test_hdrp_win_trunk/792150/job

PR #5193, opened 2019-11-26
Fix recursive sub graph import

pbbastian:
### Purpose of this PR
Fixes FB-1198653. The are two issues at hand here.

When a sub graph references itself, it ends up trying to load the asset that is currently being imported. This doesn't happen if there is 1 sub graph in-between due to how the sub graph importer works. Anyway, this causes an assertion in ADB, because you obviously can't do that. But then we hit a bug in ADB, where it doesn't mark the asset as failed-to-import, so it tries again, and goes into an infinite loop. This has been fixed by moving the circular dependency check up earlier.

But, that caused us to hit the same bug in another way. Before the actual import, ADB gathers dependencies by calling `GatherDependenciesFromSourceFile` on importers that support it, like ours. It then orders things correctly, and does the actual import. This also has a circular dependency check, but like before it fails to mark the assets as failed-to-import, and so goes into an infinite loop. This should be fixed in ADB, but the bug is there now, and we need to handle it.

So the workaround I've made is to fully gather dependencies in that callback, and if we detect circular dependencies, return an empty list instead, and when that then goes to import we mark up the dependencies the old-fashioned way using `AssetImportContext.DependsOnSourceAsset`, so that we still get a re-import if any of those dependencies change. (The latter part is needed because we're returning an empty list, so we don't get that for free anymore.)

This is somewhat slower, as `GatherDependenciesFromSourceFile` now has to recursively gather dependencies, whereas before it only had to go 1 level deep. So when this eventually gets fixed in ADB, we should `#ifdef` the workaround based on the version that the fix lands in.

### Testing status

**Manual Tests**:
- Followed repro steps from bug, and verified the fix locally:
  1. Create 2 subgraphs, sub1 and sub2
  2. Drag and drop sub1 into sub2 and connect it to the output
  3. With sub2 open, click "Save As" and replace sub1.

**Automated Tests**: We don't want to add a test for this, since it would cause the agent to go into an infinite loop on failure.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Sticky Notes can now be grouped properly.
 - Fixed an issue where nodes couldn't be copied from a group.
 - Fixed an issue where adding the first output to a Sub Graph without any outputs prior caused Shader Graphs containing the Sub Graph to break.
+- Self-referencing Sub Graphs no longer causes importing to go into an infinite loop.
 
 ## [7.1.1] - 2019-09-05
 ### Added
--- a/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderSubGraphImporter.cs
@@ -22,7 +22,18 @@ namespace UnityEditor.ShaderGraph
         {
             try
             {
+// TODO: When the fix for circular dependencies in ADBv2 lands, replace `true` with appropriate Unity version.
+#if false
                 return MinimalGraphData.GetDependencyPaths(assetPath);
+#else
+                var dependencyMap = new Dictionary<string, string[]>();
+                using (var tempList = ListPool<string>.GetDisposable())
+                {
+                    GatherDependencies(assetPath, dependencyMap, tempList.value);
+                    var containsCircularDependency = ContainsCircularDependency(assetPath, dependencyMap, tempList.value);
+                    return containsCircularDependency ? new string[0] : dependencyMap[assetPath];
+                }
+#endif
             }
             catch (Exception e)
             {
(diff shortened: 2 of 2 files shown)


LandonTownsendUnity:
I'm running into all kinds of bugs when I try to create circular dependencies using 3 subgraphs... bugs like click dragging on the invalid subgraph nodes undocking the shadergraph window, and clicking on the subgraph node getting it stuck to the mouse and leaving me unable to click anything else. I'm trying to find repro steps for these bugs now.

LandonTownsendUnity:
https://drive.google.com/open?id=1REJMDP04E4Txvi8JXeqOO9_1_PuRIpNr
video of both bugs
the bug where moving the subgraph undocks the tab is much more common and the one where the subgraph node gets stuck and can't be deleted is a bit harder to reproduce, but if you do exactly what I do in the video it should reproduce (remember every little thing counts, even something as simple as moving the subgraph node when I do or switching tabs is the difference between the bug reproducing and not reproducing).

Other bugs are surfacing as well, such as compilation errors being stuck on the output even after all the references to all the subgraphs have been removed, only going away when you close and reopen the subgraph. I get the feeling this fix is too spotty and unstable to be considered finished.

pbbastian:
@LandonTownsendUnity Thanks for testing it out :) I'll have a look at those issues and see if I can get rid of them. @martint-unity got a reply from the asset import pipeline team, and they found out that the main problem is that the array we return in `GatherDependenciesFromSourceFile` includes the path of the asset being imported. They're still going to fix the issue, so that it writes out a warning instead of infinite looping, but I'll try modifying the fix here :)

LandonTownsendUnity:
The shadergraph being undocked is a separate bug that I've reported here:
https://fogbugz.unity3d.com/f/cases/1204206/
fixing that bug is not necessary. However, it is still needed to fix the bug where shadergraph corruption can occur using the steps I followed in the video I linked.

LandonTownsendUnity:
This bug is fixed in the latest version of 20.1 and 19.3 (in the engine's asset importer), do we still need this PR?

pbbastian:
I think we should be good without it :)

PR #5195, opened 2019-11-26
Fix issues with conflicting internal names

RSlysz:
### Purpose of this PR

Currently, it often happens that a small API change in trunk breaks code that we gather by reflection.
See last https://fogbugz.unity3d.com/f/cases/1201116/

To fix it fully, we need to expose internals to HDRP and CoreRP editor packages. This is done in this C++ PR https://ono.unity3d.com/unity/unity/pull-request/97182/_/graphics/expose-internal-to-hdrp-and-corerp-editor which cannot land without a package that include conflicting namespace
So here we solve two things. The above tickets and also solving the conflicting namespace for this to land in a package and for the above C++ PR to land at some point.

---
### Comments to reviewers
Full reflection fixes can be found here on the branch  HDRP/remove-reflection-on-internals. This should be done after we have a package that do not have conflicting namespace to be used in test validation 💣 

Note: In EnvironmentLibrary, the removed reflection code was not used anymore.

@fredericv-unity3d for awarness


com.unity.render-pipelines.core/Editor/LookDev/EnvironmentLibrary.cs
```diff
@@ -169,12 +166,6 @@ public static void CreateAndAssignTo(ObjectField field)
 
     static class EnvironmentLibraryLoader
     {
-        public static void Load(Action onInspectorRedrawRequested)
```
RSlysz: this is dead code not used anymore since we show the field in the library line of the lookdev

com.unity.render-pipelines.core/Editor/LookDev/EnvironmentLibrary.cs
```diff
@@ -169,12 +166,6 @@ public static void CreateAndAssignTo(ObjectField field)
...
-        {
-            UnityEngine.Object target = LookDev.currentContext.environmentLibrary;
-            UIElementObjectSelectorWorkaround.Show(target, typeof(EnvironmentLibrary), LoadCallback(onInspectorRedrawRequested));
-        }
-
```
RSlysz: this is dead code not used anymore since we show the field in the library line of the lookdev

com.unity.render-pipelines.core/Editor/LookDev/EnvironmentLibrary.cs
```diff
@@ -183,36 +174,5 @@ static Action<UnityEngine.Object> LoadCallback(Action onUpdate)
                 onUpdate?.Invoke();
             };
         }
-
-
-        // As in UIElement.ObjectField we cannot support cancel when closing window
-        static class UIElementObjectSelectorWorkaround
```
RSlysz: this is dead code not used anymore since we show the field in the library line of the lookdev

com.unity.render-pipelines.core/Editor/LookDev/EnvironmentLibrary.cs
RSlysz: this is dead code not used anymore since we show the field in the library line of the lookdev

PR #5199, opened 2019-11-26
Disable custom pass debug mode

alelievr:
### Purpose of this PR
- Fixed rendering errors when enabling debug modes with custom passes


sebastienlagarde:
@alelievr don't we have the same issue with custom postprocess?

PR #5200, opened 2019-11-26
Improved the logic for depth texture management and scene view handling of post-processing

Chman:
### Purpose of this PR

- Improved the logic around the depth texture, it used to be always created when post-processing was enabled but is now only created if it's actually needed.
- The scene view now mirrors the volume layer mask setting from the MainCamera instead of using the `Default` layer mask.
- Some cleanup in `SetupPerCameraShaderConstants` (redundant code).


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -63,7 +63,9 @@ namespace UnityEngine.Rendering.Universal
 
         public IEnumerator<Action<RenderTargetIdentifier, CommandBuffer>> captureActions;
 
+        [Obsolete("This field isn't used anymore and will return a default value.")]
         public LayerMask volumeLayerMask;
```
phi-lira: This should return -1.

Chman: Done & merged master to fix the conflicts.

phi-lira:
Added backport tag. We can only backport this in minor update. So 7.2.0 seems like a good candidate.

Chman:
This PR is too old by now (merge hell). I'll redo the changes against master in a new PR and do a specific PR for 2019.3 as it needs special handling for the PPv2 support in there.

PR #5206, opened 2019-11-27
Fix debug menu reset when entering playmode

alelievr:
### Purpose of this PR
Fix debug menu reset when entering playmode

- [x] Built a player


alelievr:
Update: There are too many issues left due ti the current design of the debug window, so we'll postpone the task to an undefined milestone.
Closing the PR for now as we probably start from scratch when we continue to investigate this issue.

PR #5208, opened 2019-11-27
Fix stencil read on Xbox

FrancescoC-unity:
SHADER_API_D3D11 is not always defined on Xbox One and I was given wrong info regarding D3D12 and stencil data.


sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp Fix need to be validated on xbox

PR #5209, opened 2019-11-27
Ray Traced Shadows Fixes

anisunity:
Fixing ghosting issues when screen space shadow indexes change
Fixing a bug whit ray traced shadow histories when n>4
Including objects in ray traced shadows if their cast shadow mode is shadow only or double sided
Only tracing rays for shadow if the point is inside the cone for spotlight shadows
Only tracing rays if the point is inside the range for point lights
Creating a new test scene for ray traced shadows


sebastienlagarde:
Waiting for approval from QA

anisunity:
@remi-chapelain

anisunity:
fixed typos in the description

PR #5210, opened 2019-11-27
Bloom thresholding

Chman:
### Purpose of this PR

Per request, added a threshold to bloom. Moving the value higher than 0 will break the energy conservation rule.

This has only been tested in simple cases and should be tested on bigger projects.


Chman:
Documentation updated. We'll need a new screenshot of the inspector.

PR #5213, opened 2019-11-27
Scale PCSS filter size with light distance occluder

FrancescoC-unity:
Added a scale to the filter size to account to scale filter size based on light source to occluder distance. This matches better the ray traced reference as the light changes. 

The fitted curve is not great, but to get better matches I'd need higher order approx and this is already borderline too much.  

Also, contains a fix for a warning I added with a previous PR (oops :) )


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowSampling.hlsl
```diff
@@ -266,7 +266,7 @@ float SampleShadow_MSM_1tap(float3 tcs, float lightLeakBias, float momentBias, f
 //
 //                  PCSS sampling
 //
-float SampleShadow_PCSS(float3 tcs, float2 posSS, float2 scale, float2 offset, float shadowSoftness, float minFilterRadius, int blockerSampleCount, int filterSampleCount, Texture2D tex, SamplerComparisonState compSamp, SamplerState samp, float depthBias)
```
sebastienlagarde: Have in mind that now we are out of preview, this kind of change isn't allowed anymore... (without providing a fallback). Guess as we are so close to release we will say it is fine but have that in mind.

PR #5222, opened 2019-11-28
Recursive Rendering not working anymore

anisunity:
Recursive rendering was not working anymore after a refactor, it can be observed on the recursive rendering test scene.

@Unity-Technologies/gfx-qa-hdrp


iM0ve:
Recursive rendering is now working on staging
image

PR #5223, opened 2019-11-28
Add accumulation motion blur [Draft]

sebastienlagarde:

### Checklist for PR maker

---
### Purpose of this PR

---
### Testing status

- [ ] Built a player
- Other: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252FAccumulation-MB

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/AccumulationMotionBlur.shader
```diff
@@ -0,0 +1,70 @@
...
+
+        void FragAccumulation(Varyings input, out float3 outColor : SV_Target0)
+        {
+            UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(input);
+            
+            float2 uv = input.texcoord;
+            
+            //TODO: Remove Fetch
+            float3 color   = Fetch(_InputTexture,        uv, 0.0, _RTHandleScale.xy);
+            float3 history = Fetch(_InputHistoryTexture, uv, 0.0, _RTHandleScale.xy); //TODO: History Scale
```
sebastienlagarde: should be _RTHandleScaleHistory

com.unity.render-pipelines.high-definition/Editor/PostProcessing/MotionBlurEditor.cs
```diff
@@ -38,6 +40,27 @@ public override void OnInspectorGUI()
         {
             bool advanced = isInAdvancedMode;
 
+            if(advanced)
```
FrancescoC-unity: We need to check with UX about how they feel with having something in the "advanced settings" popping up on top. 
I do agree that this should be the first thing in the list, but it is an unknown pattern. 

I wonder if it would be better at this point to have a different volume component for this, after all I don't see many users switching between the two and the algorithm are different enough.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -465,6 +471,16 @@ void PoolSource(ref RTHandle src, RTHandle dst)
                         }
                     }
 
+                    if (m_MotionBlur.IsActive() && m_MotionBlur.algorithm == MotionBlurAlgorithm.Accumulation && !isSceneView)
```
FrancescoC-unity: Does the preview camera count as scene view? If not, I assume we don't want to risk it running when in scene view selecting the camera

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/MotionBlurCommon.hlsl
```diff
@@ -44,6 +45,9 @@ CBUFFER_END
 #define _SampleCount                        uint(_MotionBlurParams2.x)
 #define _TileSize                           uint(_MotionBlurParams2.y)
 #define _DepthScale                         _MotionBlurParams2.z
+#define _AccumulationSampleCount            _MotionBlurParams3.x
```
FrancescoC-unity: Uhm I don't think we need this (the whole _MotionBlurParams3 thing) in here as it is not pertinent to the other motion blur nor is used in those shaders. I would argue we should have a different CBuffer for Accumulation MB and include that in Uber and the accumulation mb shader (instead of bringing this whole file over)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -1422,6 +1447,53 @@ RTHandle Allocator(string id, int frameIndex, RTHandleSystem rtHandleSystem)
...
+                1f // Always enabled for dev
+            );
+
+            m_AccumulationMotionBlurPropertyBlock.SetTexture(HDShaderIDs._InputTexture, source);
+            m_AccumulationMotionBlurPropertyBlock.SetTexture(HDShaderIDs._InputHistoryTexture, prevHistory);
+            m_AccumulationMotionBlurPropertyBlock.SetVector(HDShaderIDs._MotionBlurParams3, accumulationMotionBlurParams);
+
+            CoreUtils.SetRenderTarget(cmd, destination, depthBuffer);
+            cmd.SetRandomWriteTarget(1, nextHistory);
+            cmd.DrawProcedural(Matrix4x4.identity, m_AccumulationMotionBlurMaterial, 0, MeshTopology.Triangles, 3, 1, m_AccumulationMotionBlurPropertyBlock);
```
FrancescoC-unity: Could this pass be compute instead? Is just that the UAV in PS handling makes me a bit uncomfortable in general, so unless necessary I tend not do that.

I assume this was modelled around TAA, but the reason TAA is in pixel is because of Stencil usage, which we don't need here

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Chman: The preview camera concept is a bit weird, in the sense that the preview you see in the material inspector is reported as "preview camera" but the one in the scene view when you select a camera is reported as "game view" (last I checked) :/

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Chman: +1 for moving it to compute. We need to keep everything between AA and the final pass as compute for the upcoming async postfx job.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/AccumulationMotionBlur.shader
```diff
@@ -0,0 +1,70 @@
...
+            color = (color * rcp(asfloat(_AccumulationSampleCount))) + (history * step(1, _AccumulationSampleIndex));
```
Chman: I'm still weirded out by the use of `asfloat` here.

com.unity.render-pipelines.high-definition/Editor/PostProcessing/MotionBlurEditor.cs
johnpars: I can go ahead and organize all of this under its own volume component if that makes sense for now, I definitely agree that adding this in the advanced field only makes an existing UX challenge for the accumulation even more tricky

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/MotionBlurCommon.hlsl
johnpars: Got it- I went ahead and moved this into its own Cbuffer and left these velocity motion blur settings alone. Additionally, I realized that the normalization (done in the uber pass) can also be consolidated to the compute pass. So everything is now consolidated there.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
johnpars: Moved to compute! Also, I realized that the normalization step done in the uber pass could be consolidated to the one accumulation compute pass as well, so everything gets done in there.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/AccumulationMotionBlur.shader
johnpars: Rightly so! This line has been broken up into something more readable during the move to compute, and all the divs have been precomputed.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/AccumulationMotionBlur.shader
johnpars: Fixed during the move from PS -> CS

sebastienlagarde:
hey, poking people here to get an update on progress :)

johnpars:
hey Seb, I've been set back on completing this PR due to no known holistic plan (afaik) for HDRP sub-frame yet (pointed out by Pavlos). Until then, Thomas, Pavlos, & I agreed it makes sense to move it to a color accumulation pass in a Post-FX Custom Pass. 
No update on this yet, I have had trouble carving out that time due to recent holiday/new year project demands. We have new engineering headcount on the team which will help greatly in opening availability back up soon

sebastienlagarde:
Replace by https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6114

PR #5224, opened 2019-11-28
Shader use for Runtime Debug Display are now correctly strippep in release

sebastienlagarde:
### Purpose of this PR

When a user is doing a release build, we disable the debug menu windows. But we weren't disabling the generation of shader variant for the debug. This PR fix this issue.

---
### Testing status
To test the PR, you should build the template of HDRP in debug build, check the build time for it.
Then suppress your library folder (or maybe only the shadercache folder), and do it again in release. The time should be well improve (as there is a lot of shader variant that are discard).

---


JulienIgnace-Unity:
These visual artefacts are unrelated and caused by a bug in compute shader async compilation. A fix should be on the way soon (C++ side)

PR #5227, opened 2019-11-28
Fixed camera inverse view-proj matrix in XR mode.

kaychang-unity:
Fixed depthCopy and colorCopy passes not working in XR mode.

---
### Purpose of this PR
Fixed a few bugs for XR support. _InvCameraViewProj cannot be used in XR mode as it is not correctly updated.
Depth-copy pass did not copy the correct portion of the depth buffer. The bug did not happen in Forward Renderer because the depth-prepass is not reused for the main forward pass at this time.
I replicated the same fix for color-copy pass, although I don't have a use-case for it yet.

These fixes are needed for the deferred renderer.

---
### Testing status

**Yamato**: (fix/xr):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/fix%252Fxr/.yamato%252Fupm-ci-universal.yml%2523All_Universal_fast-trunk

---
### Comments to reviewers


com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -65,9 +65,9 @@ half4 _AdditionalLightsOcclusionProbes[MAX_VISIBLE_LIGHTS];
 #define UNITY_MATRIX_V     unity_MatrixV
 #define UNITY_MATRIX_I_V   unity_MatrixInvV
 #define UNITY_MATRIX_P     OptimizeProjectionMatrix(glstate_matrix_projection)
-#define UNITY_MATRIX_I_P   ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED
+#define UNITY_MATRIX_I_P   unity_CameraInvProjection
```
phi-lira: This is going to be confusing and it seems wrong. There's a distinction between unity_Camera* and unity_Matrix*.
 unity_Camera* will always be defined to camera matrices, while unity_Matrix* depends on the pass. One particular example is shadows. unity_Matrix{*} will be define to light matrices while unity_Camera* it's still defined to camera.

 so, the confusion is that UNITY_MATRIX_I_P should be instead `unity_MatrixInvP` to be consistent. However is now returning a camera matrix instead of the current shader bound matrix. My suggestion is that we add either macros or functions for unity_Camera* raw data.

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
```diff
@@ -188,6 +189,7 @@ float4x4 glstate_matrix_projection;
 float4x4 unity_MatrixV;
 float4x4 unity_MatrixInvV;
 float4x4 unity_MatrixVP;
+#define unity_MatrixInvVP _InvCameraViewProj
```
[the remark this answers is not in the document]
kaychang-unity: Ok, I can make that change.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
kaychang-unity: `unity_MatrixInvP` is part of the group of matrices set by C++ code and is missing from that group. Instead I will revert to `#define UNITY_MATRIX_I_P   ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED`.

PR #5233, opened 2019-11-29
Fixed an issue where cached sky contexts were being destroyed wrongly, breaking rendering in the LookDev

JulienIgnace-Unity:
### Purpose of this PR
- Fixed an issue where cached sky contexts were being destroyed wrongly, breaking the lighting in the LookDev.
- Also fixed another issue where LookDev Volumes would override those from the main scene and break sky rendering in the scene view for example.
- Moved some LookDev stuff to internal in the process as well.

Fixes https://fogbugz.unity3d.com/f/cases/1202498/

---
### Testing status

- [ ] Built a player

- Other: 
For QA: Before the fix, LookDev would have a black ambient lighting if shadows were enabled. This PR fixes this (there is no fogbugz logged for this issue). It also fixes test 2211 where ambient was wrong.


iM0ve:
Tried opening a2 and toggle LookDev shadows. They had no effect and it was still rendering black.
Lookdev works fine on a regular project though.

com.unity.render-pipelines.core/Editor/LookDev/Stage.cs
```diff
@@ -225,6 +225,18 @@ public void SetGameObjectVisible(bool visible)
                 light.enabled = visible;
         }
 
+        public void BeginRendering(IDataProvider dataProvider)
```
RSlysz: Should not be public
OnBeginRendering is a better name as it is in fact a hook in the rendering process.

com.unity.render-pipelines.core/Editor/LookDev/Stage.cs
```diff
@@ -225,6 +225,18 @@ public void SetGameObjectVisible(bool visible)
...
+        {
+            SetGameObjectVisible(true);
+            dataProvider.BeginRendering(runtimeInterface);
```
RSlysz: Should be OnBeginRendering in the SRI too for matching (should remain public in the SRI)

com.unity.render-pipelines.core/Editor/LookDev/Stage.cs
```diff
@@ -225,6 +225,18 @@ public void SetGameObjectVisible(bool visible)
...
+        }
+
+        public void EndRendering(IDataProvider dataProvider)
```
RSlysz: Should not be public
OnEndRendering is a better name as it is in fact a hook in the rendering process.

com.unity.render-pipelines.core/Editor/LookDev/Stage.cs
```diff
@@ -225,6 +225,18 @@ public void SetGameObjectVisible(bool visible)
...
+        {
+            SetGameObjectVisible(false);
+            dataProvider.EndRendering(runtimeInterface);
```
RSlysz: Should be OnEndRendering in the SRI too for matching (should remain public in the SRI)

RSlysz:
Also add missing documentation in SRI interface public new hook :)

iM0ve:
PASS:
* Assigning HDRIs env settings in lookdev no longer affects the scene view and volume
* Objects are now lit properly (not too dark anymore)
* No more console errors

OTHER NOTES:
* Did not test this for A2 project
* There is still an issue with lookdev when clicking New Scene (planned to be addressed on another PR)

Video:
https://drive.google.com/file/d/1f67cW7XxxjIYD3bD6c24HsliKXhftOZ4/view?usp=sharing

PR #5235, opened 2019-11-29
Fixing an inconsistency between ray tracing shadows and shadow maps for culling

anisunity:
Shadow maps and ray tracing shadows for point and spot lights used to take culling into account differently, this unifies the behavior. by doing that, it solves this bug: https://fogbugz.unity3d.com/f/cases/1201887#BugEvent.1071826620


sebastienlagarde:
Please Add changelog

PR #5238, opened 2019-11-29
Contrast Adaptive Sharpen Integration

jsholdice-amd:
### Purpose of this PR

Adding the CAS upscaling effect to the HDRP. 

---
### Testing status

Upscaling effect has been tested with Unity Version 2020.1.0a13. Tested in editor and standalone player. Renderdoc used to confirm effect as it is very subtle and hard to see with the naked eye sometimes.

Effect is accessable by changing the dynamic resolution type on the HDRP Asset to 'Contrast Adaptive Sharpen'


Chman:
Note: we'll need to add a changelog entry before we merge it.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
```diff
@@ -0,0 +1,97 @@
...
+}
+
+AF4 CasOutput(AF4 pix)
+{
+	return pix;
+}
+
+float CalcLuminance(float3 color)
+{
+	return dot(color, float3(0.299f, 0.587f, 0.114f));
```
sebastienlagarde: We have a function to calulate luminance in color.hlsl, please use it - Luminance so we have consistant constant use (in our case constant for rec 709.

TomasKiniulis:
Testing on 2020.1.0a16 Contrast Adaptive Sharpen does not get scaled correctly.
2020.1.0a16:
hfwHqVpFhG
2020.10a15:
AIdirEPePf

Chman:
Adding Fabien for VR.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -588,6 +592,42 @@ void PoolSource(ref RTHandle src, RTHandle dst)
...
+                            cmd.SetComputeBufferParam(cs, kMain, "CasParameters", m_ContrastAdaptiveSharpen);
+
+                            cmd.DispatchCompute(cs, kInit, 1, 1, 1);
+
+                            int dispatchX = (int)System.Math.Ceiling(destination.rt.width / 16.0f);
+                            int dispatchY = (int)System.Math.Ceiling(destination.rt.height / 16.0f);
+
+                            cmd.DispatchCompute(cs, kMain, dispatchX, dispatchY, 1);
+
+                            PoolSource(ref source, destination);
```
Chman: This should be outside of the if statement or it's going to break subsequent passes.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -588,6 +592,42 @@ void PoolSource(ref RTHandle src, RTHandle dst)
...
+
+                            cmd.SetComputeBufferParam(cs, kInit, "CasParameters", m_ContrastAdaptiveSharpen);
+                            cmd.SetComputeBufferParam(cs, kMain, "CasParameters", m_ContrastAdaptiveSharpen);
+
+                            cmd.DispatchCompute(cs, kInit, 1, 1, 1);
+
+                            int dispatchX = (int)System.Math.Ceiling(destination.rt.width / 16.0f);
+                            int dispatchY = (int)System.Math.Ceiling(destination.rt.height / 16.0f);
+
+                            cmd.DispatchCompute(cs, kMain, dispatchX, dispatchY, 1);
```
Chman: We need some VR specific handling on that dispatch but we can take care of it on our side.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
Chman: FYI our `Luminance` function is in `Packages/com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl`

TomasKiniulis:
> Testing on 2020.1.0a16 Contrast Adaptive Sharpen does not get scaled correctly.
> 2020.1.0a16:
> hfwHqVpFhG
> 2020.1.0a15:
> AIdirEPePf

Issue occurs only when

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
jsholdice-amd: This was primarily for debugging while we were developing it, removed the local CalcLuminance.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
jsholdice-amd: To clarify, the PoolSource needs to be moved to make sure it's in the same block as the m_Pool.get? If so, easy fix will be in next commit.

Chman: Yep pretty much, right now in the rare case where the if statement doesn't get executed then `PoolSource` will never get called and the next effect in line will get a recycled render target with potential garbage in it.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -588,6 +592,42 @@ void PoolSource(ref RTHandle src, RTHandle dst)
...
+                    using (new ProfilingSample(cmd, "Contrast Adaptive Sharpening", CustomSamplerId.ContrastAdaptiveSharpen.GetSampler()))
+                    {
+                        var destination = m_Pool.Get(Vector2.one , k_ColorFormat);
+                        
+                        var cs = m_Resources.shaders.contrastAdaptiveSharpenCS;
+                        int kInit = cs.FindKernel("KInitialize");
+                        int kMain = cs.FindKernel("KMain");
+                        if (kInit >= 0 && kMain >= 0)
+                        {
+                            cmd.SetComputeFloatParam(cs, "Sharpness", 1);
```
FrancescoC-unity: We should really use the same pattern as elsewhere, defining the names in HDStringConstants.cs

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -21,6 +21,11 @@ float3 Bilinear(TEXTURE2D_X(_InputTexture), float2 UV)
     return Sample(TEXTURE2D_X_ARGS(_InputTexture, s_linear_clamp_sampler), UV);
 }
 
+float3 ContrastAdaptiveSharpen(TEXTURE2D_X(_InputTexture), float2 UV)
```
FrancescoC-unity: Is this because bilinear sampling is beneficial? Because if a load is ok, I don't see why we need an extra multi compile here

jsholdice-amd: What I've tried to do is not set any keyword if it's using CAS and let it just go through the default non-upscaled path in the shader, however when I try to do this the image gets upscaled a second time (images below).

I agree, ideally there wouldn't be the sample the way there is in the other upscale methods, no multi compile, and it would just go through the normal path in the shader but I've not been able to make it work that way.

With multicompile: 
Unity_TlAFv3Z1KH

Without:
Unity_RF5RU9mcmW

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
jsholdice-amd: Done, will be in the next commit.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
jsholdice-amd: Found a way to simplify it a bit but I think it still needs the multi compile. I removed the extra sample but there does still need to be a little bit different between that and no upsampling in the finalpass shader.

fabien-unity:
@jsholdice-amd : I just tried this branch and it seems broken when using any forced percentage below 100% : the final image covers only a subset of the viewport.
Once this is addressed, I'll test the VR path.

jsholdice-amd:
> @jsholdice-amd : I just tried this branch and it seems broken when using any forced percentage below 100% : the final image covers only a subset of the viewport.
> Once this is addressed, I'll test the VR path.

Do you have any additional information? I haven't been able to reproduce this issue locally. I've tested it with the 2020.1.0a16 version in a brand new HDRP project, with the forced resolution set to 50 and am not seeing what you describe.

fabien-unity:
Here's what I get by making a new project from the HDRP template using 2020.1.0a16 with this branch:
image

michal-z-amd:
> Here's what I get by making a new project from the HDRP template using 2020.1.0a16 with this branch:
> image

@fabien-unity I can't repro this. What HW/driver are you using? Do other filters work fine for you? Thanks!

michal-z-amd:
I was debugging this bit more and the problem seems to be:
Unity does not load/compile CAS compute shader when we select CAS Filter from drop down menu.
Unity loads/compiles CAS computer shader only when Editor is closed and then opened again. And after this restart CAS upscaling works perfectly fine.

Is there a way to force load CS? Shouldn't the CS be compiled when we call FindKernel?

Chman:
This seems related to a bug we currently have with the compute shader compilation pipeline. I noticed the same behavior, restarting Unity fixed it. So this issue shouldn't block this PR.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
fabien-unity: For VR single-pass, this needs to be replaced by:
`cmd.DispatchCompute(cs, kMain, dispatchX, dispatchY, camera.viewCount);`

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
```diff
@@ -0,0 +1,92 @@
...
+
+#include "ffx_cas.h"
+
+
+[numthreads(64, 1, 1)]
+void KMain(uint3 LocalThreadId : SV_GroupThreadID, uint3 WorkGroupId : SV_GroupID, uint3 dispatchThreadId : SV_DispatchThreadID)
+{
+	uint4 c0 = CasParameters[0];
+	uint4 c1 = CasParameters[1];
+
```
fabien-unity: For VR single-pass rendering, the following code needs to be added here:
`UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);`

jsholdice-amd: Unable to test this myself since, but the changes are exactly as you described.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
```diff
@@ -0,0 +1,95 @@
...
+Doing this on the GPU despite the fact that CAS provides a CPU version
+of CasSetup(). This is done to prevent us from having to rewrite a C# 
+version every time CAS changes. This is not called from KMain to reduce overhead
+during main render.
+*/
+[numthreads(1,1,1)]
+void KInitialize()
+{
+
+	UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
```
Chman: This cannot compile, `dispatchThreadId` doesn't exist in the scope (missing a `SV_DispatchThreadID` parameter in the kernel declaration). And shouldn't this be in `KMain` instead?

TomasKiniulis: Getting the following due to this: i2hKXwhQwf

Rechecked on 2020.1.0a19, not getting properly upscaled CAS result even when editor is reopened. Selecting any other upscale filter breaks rendering due to the CAS compute shader

jsholdice-amd: Fixed, did a quick test as well, still no VR test but the shaders compiled and the effect seemed to work.

TomasKiniulis: The filter works great now! Although, I'm still getting a list of "implicit truncation of vector type at kernel" shader warnings in the console. They only appear once - updating the project or rebuilding the library. Could these be addressed?

image

jsholdice-amd: Made the change, warnings haven't been coming up in the editor but the shadercompiler-UnityShaderCompliler.exeX.log files look clean with the following result.

"Cmd: preprocessCompute
  insize=2306 outsize=2307 ok=1"

jsholdice-amd: Made the change, warnings haven't been coming up in the editor but the shadercompiler-UnityShaderCompliler.exeX.log files look clean with the following result.

"Cmd: preprocessCompute
insize=2306 outsize=2307 ok=1"

sebastienlagarde:
@fabien-unity please can you validate the PR. THanks

sebastienlagarde:
Note:  the PR miss a changelog and need to be merge with latest master. Otherwise, all good to go.

jsholdice-amd:
Merged and resolved conflicts, added CAS line in the changelog as well as requested. Tested under same conditions as before briefly to make sure it's all still working as expected with no problems.

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute
sebastienlagarde: Hey @jsholdice-amd I don't this the change despite your comment saying that you have done it. Have you forget to push the commit? thanks

PR #5239, opened 2019-12-02
[SkipCI] Shader Graph saving-menus improvements/fixes

zadi0648:
Two things:

* Fixed a bug where _Save As..._ frequently brings up the wrong Project when using multiple Unity Projects at once https://fogbugz.unity3d.com/f/cases/1201012/
* Implemented this Favro card: Sub Graphs now remember the path of the last converted Sub Graph (**NEW:** see below gif for implementation info): (https://favro.com/organization/c564ede4ed3337f7b17986b6/d06b8de3e40da6d340b5b063?card=Uni-80411)

---
_Convert-to-Sub-Graph conversion info:_

Talked to Grant and Landon about this, after talking over the most common use cases we decided on "option 3", which I expanded upon a little:

* State 1: by default, Convert-to-Subgraph save-file menu navigation will open up in the directory of the SG ("origin") that contained the nodes. _It will continue to do so_ for as long as the user saves Sub Graphs inside the same directory as the origin SG.
* State 2: once the user saves a Sub Graph to a folder outside of the origin's directory (for example /Assets/Subgraphs/), the save-file menu navigation will then open up in this new directory the next time a conversion-to-subgraph happens. This happens for all Shader Graphs and Sub Graphs within the same project.
* If in State 2, closing Unity will return the user to State 1 upon reopening the project (would require a hacky solution to fix, and we agreed that ideally that's how it ought to work anyhow).
* If in State 2 and the user does save a Sub Graph to the origin's directory, it will return to State 1.

---
Halo Effect: 0/4
Technical Risk: 1/4

---
**Notes:**
* Changed _Save As..._ to use `SaveFilePanelInProject`, which enforces that the file be saved inside the same project. Perhaps that was done so that the user could save graphs outside the Project? (which technically worked but Unity would throw a little fit) However wasn't how Convert to Subgraphs worked, and we should have them work the same way. So if desired we can change both to `SaveFilePanel`.
* So basically the "Save As..." bug was caused by the fact that path included the file in it which SaveFilePanel would assume is a folder I presume. However as such it seems like we found a bug with the EditorUtility API, which I'll file a bug for.

---
Testing status

**Manual Tests**:
- Saving sub graphs over and over in different directories of different names
- Doing that with multiple unity projects open at once
- (for _Save As..._) Tried feeding various strings combinations in, including Assets/SGs/0/SubFix.shadergraph Assets/SGs/0/ Assets12345678901234567890
- _"Save As..."_ multiple unity projects open at once
- _"Save As..."_ for Shader and Sub Graphs

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fmenu-navigation-storing/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1535320/job/pipeline
Master compare:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/master/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1512933/job/pipeline


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -358,13 +358,22 @@ public void SaveAs()
...
             if (selectedGuid != null && graphObject != null)
             {
-                var path = AssetDatabase.GUIDToAssetPath(selectedGuid);
-                if (string.IsNullOrEmpty(path) || graphObject == null)
+                var pathAndFile = AssetDatabase.GUIDToAssetPath(selectedGuid);
+                if (string.IsNullOrEmpty(pathAndFile) || graphObject == null)
                     return;
 
+                // The asset's name needs to be removed from the path, otherwise SaveFilePanel assumes it's a folder
+                int lastIndex = pathAndFile.LastIndexOf("/");
```
marctem: Should use Path.GetDirectoryName() here.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -396,7 +405,9 @@ public void ToSubGraph()
         {
             var graphView = graphEditorView.graphView;
 
-            var path = EditorUtility.SaveFilePanelInProject("Save Sub Graph", "New Shader Sub Graph", ShaderSubGraphImporter.Extension, "");
+            if (m_PrevPath == null || m_PrevPath.Length == 0) m_PrevPath = Application.dataPath;
```
marctem: Can you do the initialization when you declare m_PrevPath, or in the initialization method?

zadi0648: Okay, I did that to be safe in case a bug below cause path to become null, but after a 2nd review I see that can't happen. Also should have been String.IsNullOrEmpty.

com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
zadi0648: ah, much easier! TIL

zadi0648:
Hold off on testing & reviewing this; Peter and I are talking about potentially a way to make this happen cross SGs in the same project

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added support for vertex skinning when you use the DOTS animation package.
 - You can now use the right-click context menu to set the precision on multiple selected nodes.
 - Updated the zoom level to let you zoom in further.
+- Converting nodes to Sub Graph now opens up in the directory of the original graph (that contained the nodes), if the path of the new Sub Graph is outside of this directory it will also remember that path to reduce folder navigation for next time.
```
sharlenet: Suggestion:
When you convert a node to a Sub Graph, it now opens up in the directory of the original graph that contained the node. If the new Sub Graph is outside this directory, it also remembers that path to ease folder navigation.

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -29,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an error where the UV channel drop-down menu on nodes had clipped text. [1188710](https://issuetracker.unity3d.com/issues/shader-graph-all-uv-dropdown-value-is-clipped-under-shader-graph)
 - Sticky Notes can now be grouped properly.
 - Fixed an issue where nodes couldn't be copied from a group.
+- Fixed a bug where "Save As.." for a Shader or Sub Graph would sometimes open up in the wrong project when multiple Unity projects are open simultaneously.
```
sharlenet: Suggestion:
Fixed a bug where __Save As..__ for a Shader or Sub Graph sometimes appeared in the wrong Project when you had multiple Unity Projects open simultaneously.

zadi0648: Actually looking over it again I think we should be more specific; that is we should elaborate that it is the file-saving-dialog (or what is it that we want to call it?) opens up in the wrong project, as opposed to Save As... opening up in the wrong project which doesn't make literal sense

sharlenet: Fixed a bug where the **Save Graph As** dialog for a Shader or Sub Graph sometimes appeared in the wrong Project when you had multiple Unity Projects open simultaneously.

zadi0648: 👍

martint-unity:
Have you run Yamato on this?

PR #5240, opened 2019-12-02
Fix reflection probe texture missing message

sebastienlagarde:
Fix reflection probe texture missing message


com.unity.render-pipelines.high-definition/Editor/Lighting/Reflection/HDReflectionProbeEditor.Preview.cs
```diff
@@ -48,7 +48,8 @@ public override void OnPreviewGUI(Rect position, GUIStyle style)
                 GUILayout.FlexibleSpace();
                 Color prevColor = GUI.color;
                 GUI.color = new Color(1, 1, 1, 0.5f);
-                GUILayout.Label("Reflection Probe not baked yet");
+                if (m_CubemapEditor.targets)
+                GUILayout.Label("No Texture for Reflection Probe available yet, please bake in Baked type, assign a texture in Custom type or enable Realtime type");
```
JordanL8: There is no Texture available for the Reflection Probe. Either use Baked and bake a Texture in, use Custom and assign a Texture, or enable Realtime.

PR #5259, opened 2019-12-04
Fix issues with decals flickering or having missing parts on PS4 especially in VR.

robinb-u3d:
### Purpose of this PR

Fixes the wave optimized rendering path that is used on PS4 for decals.

* `WaveIsFirstLane` would sometimes return a helper lane as the first lane in a wave and could cause the decals InterlockedOr() to never write as helper lanes have their writes masked out. Now it will return the first non helper lane.
* `WaveActiveBitOr` was inside the `WaveIsFirstLane` and so would only produce an `or` across the lane not across the tile as the if reduces the active lanes to 1. Moved into the tile scope to allow the correct processing.
* Work around for some optimization issues that showed up when trying to fix the above issues till the shader compiler gets appropriate fixes.

---
### Testing status

- [x] Opened test project + Run graphic tests locally - results look to match current PS4 status for HDRP runtime test.
- [x] Built a player

- Other: 


sebastienlagarde:
Added @EvgeniiG and @Paulm-Unity for awareness of the issue on PS4.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassDecal.hlsl
```diff
@@ -84,22 +84,23 @@ void Frag(  PackedVaryingsToPS packedInput,
...
-    uint minTileCoord1d = WaveActiveMin(tileCoord1d);   
-    while (minTileCoord1d != -1) 
+    uint minTileCoord1d = WaveActiveMin(tileCoord1d);
+    while (minTileCoord1d != -1)
     {
         if ((minTileCoord1d == tileCoord1d) && (!clipped))// if this is the current tile and not a helper lane
-        {               
+        {
+            // calculate the mask across the current tile
+            mask = WaveActiveBitOr(surfaceData.HTileMask);
```
Paulm-Unity: Wouldn't it affect performance because this was moved to be outside of the if(WaveIsFirstLane()) ? The idea is to only do it once per wave.

com.unity.render-pipelines.core/ShaderLibrary/API/PSSL.hlsl
```diff
@@ -49,7 +56,13 @@ uint WaveGetLaneIndex()
 
 bool WaveIsFirstLane()
 {
+#ifdef SHADER_STAGE_FRAGMENT
+    // For fragment shaders we have to make sure to return the first non helper lane that is valid.
+    ulong mask = __s_read_exec() & __s_read_initialpixelvalidmask();
```
EvgeniiG: I am not sure I understand.
Why can't the helper lane be the first lane? Isn't this specific to your particular use case?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassDecal.hlsl
EvgeniiG: All lanes of the wavefront participate in wave operations. So predicating it by WaveIsFirstLane should not affect performance.

com.unity.render-pipelines.core/ShaderLibrary/API/PSSL.hlsl
EvgeniiG: From the spec:
WaveIsFirstLane - True only for the active lane in the current wave with the smallest index.
Active Lane - A lane for which execution is being performed. In pixel shaders, it may include any helper pixel lanes.

EvgeniiG: More information:

__s_read_initialpixelvalidmask - Reads exec mask  before any s_wqm_b64 is applied and without taking into account any discard instruction.

So __s_read_initialpixelvalidmask will return the exec mask before the whole quad mode (helper lanes) are activated and before all discard information happens. So it's a subset of __s_read_exec (I don't see why we need to & it).

More importantly, I think this code here conflicts with the definition of WaveIsFirstLane, which should take both helper lanes and discard into account.

robinb-u3d: So from the following (please correct if there is a better location for this) https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics 

> Helper Lane: A lane which is executed solely for the purpose of gradients in pixel shader quads. The output of such a lane will be discarded, and so not rendered to the destination surface.

> Inactive Lane: a lane which is not being executed at this point in the code, e.g. due to flow control, or insufficient work to fill this lane in the wave, or because it is a helper lane.

> Active Lane: A lane for which execution is currently being performed as determined by the specified flow control.

robinb-u3d: So here a Helper Lane is inactive. 
> Wave Query Intrinsics:
bool WaveIsFirstLane()
This result returns true only for the active lane in the current wave with the smallest index. It can be used to identify operations that are to be executed only once per wave.

So here WaveIsFirstLane() is true for active lanes and helper lanes are defined to be inactive (however the text here doesn't require active and inactive to be disjoint I suppose).

robinb-u3d: So the problem here is simply that if a helper lane is selected as the only lane for writing output then it will attempt to write output but helper lanes are not allowed to write results so it's output will be masked and no output will be produced. By using __s_read_initialpixelvalidmask  we exclude helper lanes from our version of WaveIsFirstLane() so that we never pick a helper lane as the sole lane to write the results out (which being a helper lane it cannot do).

The definition of WaveIsFirstLane() appears to be a little fuzzy over several different descriptions of the spec. But if it can pick helper lanes and given that the final spec appears to disallow discovery of helper lanes then it's use in pixel shaders for picking a single writer will not work.

EvgeniiG: Here's my source: https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/waveisfirstlane

I think a good guiding principle is that most wave intrinsics only take the active lanes into account (even if the name does not reflect it).

EvgeniiG: "So the problem here is simply that if a helper lane is selected as the only lane for writing output then it will attempt to write output but helper lanes are not allowed to write results so it's output will be masked and no output will be produced."

I believe you identified the bug correctly. :-)

My suggestion would be to add and then use WaveIsHelperLane() to mask out helper lanes before writing the output.

So, something along these lines:

if (!WaveIsHelperLane()) // Mask out helper lanes
{
    if (WaveIsFirstLane()) // Get the first active lane that is not a helper lane
    {
        /* Store the results */
    }
}

The generated code should be similar to your change.

EvgeniiG: The way I see it, active lanes are just all lanes that are executing the same piece of code. This includes both helper lanes and discarded lanes because they need to execute the code to make derivative calculation possible.

EvgeniiG: Note: this is apparently not true on Metal (adding @jessebarker so that he can comment).

FrancescoC-unity: Well, the MS sources are diverging, wonderful :D 
I think definition is a bit fuzzy. My personal view tends to agree with Evgenii's source, as long as a lane as its bit set in the exec mask, it is active.  

However, things are a bit more confusing. Helper lanes are not considered valid for (at least some) wave intrinsics. For example, if you look at the disassembly of a WaveActiveMin called in  abranch, you will notice that it will enter the intrinisc with the exec mask stored at shader *entry* before helper lanes are enabled by WQM, so it is excluding helper lanes so the specs once again contradict this as they mention WaveActiveMin  only acts on *active* lanes. 

In this specific case regarding WaveIsFirstLane, would be curious to see what the disassembly on SM6 does. Does it take the & of current exec mask with the exec mask pre WQM activation or just the current exec mask? The disassembly should show that and give the answer for what we want to replicate here.

If we can't do the above, I like @EvgeniiG suggestion as it is clear regardless of what definition we give to active lane.

EvgeniiG:
My understanding seems to mirror what is happening at the HW level on the PS4.
You can find it in the docs - Wave Shader Compiler User's Guide, 9: Tips and Tricks.

EvgeniiG:
Also, I believe it makes sense that helper lanes are not taken into account by some wave intrinsics, such as WaveActiveMin. Why would you ever want helper lanes (that includes discarded lanes) to pollute your results? It's fuzzy, but results in the more natural behavior.

Paulm-Unity:
would it be helpful to us a sample that draws a 2d triangle and test this? i assume Ps4 SDK comes with a 'hello world' triangle sample?

FrancescoC-unity:
> Also, I believe it makes sense that helper lanes are not taken into account by some wave intrinsics, such as WaveActiveMin. Why would you ever want helper lanes (that includes discarded lanes) to pollute your results? It's fuzzy, but results in the more natural behavior.

Yes, of course make sense, my point about WaveActiveMin was to question the definition of active. 
The doc mentions that WaveActiveMin acts on _active_ lanes, but also it ignores _helper_ lanes leading to the conclusion that _helper_ lanes are not _active_ .

FrancescoC-unity:
> My understanding seems to mirror what is happening at the HW level on the PS4.
> You can find it in the docs - Wave Shader Compiler User's Guide, 9: Tips and Tricks.

More than HW or even PS4 level here, the point is understanding the definition of active in SM6.0 (which is what we should use as baseline), hence why I suggested looking at SM6.0 at the existing  WaveIsFirstLane and mirror that with the equivalent instructions on PS4.

com.unity.render-pipelines.core/ShaderLibrary/API/PSSL.hlsl
```diff
@@ -49,7 +56,18 @@ uint WaveGetLaneIndex()
 
 bool WaveIsFirstLane()
 {
+#ifdef SHADER_STAGE_FRAGMENT
+    // For fragment shaders we have to make sure to return the first non helper lane that is valid as helper lanes are
+    // described as inactive according to the HLSL docs here: https://github.com/Microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics
+    // the MSDN docs appear to be out of date on the Wave functions: https://github.com/Microsoft/DirectXShaderCompiler/issues/699
+    // where they describe WaveIsFirstLane as including helper lanes. Inspecting the GCN disassembly from DX12 a DX12 shader with the OpenGPU
+    // RGA tool shows that on DX12 WaveIsFirstLane() does indeed appear to exclude helper lanes picking the lowest non-helper lane.
```
EvgeniiG: Nice, thank you!

PR #5260, opened 2019-12-04
Universal Camera Stacking

phi-lira:
Creating Draft PR to start collecting feedback. 

---
### Purpose of this PR
 Introduces ability to stack cameras. This exposes an overlay camera render type that can be stacked on top of base cameras.

---
### Testing status

- [x] Built a player
- [x] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

**Automated Tests**: Added test scene for camera stacking.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fcamera-stacking

 Extending Universal Rendering Examples to test it.

---
### Comments to reviewers


Chman:
That's a big PR. It looks good overall, I ran a few manual tests and seems to behave as expected (at least with postfx).

Note: `015_CameraMulti_FPSCam` currently fails as it doesn't draw any shadow.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -157,23 +166,31 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             bool createColorTexture = RequiresIntermediateColorTexture(ref renderingData, cameraTargetDescriptor) || camera.forceIntoRenderTexture;
 
             // If camera requires depth and there's no depth pre-pass we create a depth texture that can be read later by effect requiring it.
-            bool createDepthTexture = requiresDepthTexture && !requiresDepthPrepass;
+            bool createDepthTexture = cameraData.requiresDepthTexture && !requiresDepthPrepass;
+            createDepthTexture |= (renderingData.cameraData.renderType == CameraRenderType.Base && !renderingData.resolveFinalTarget);
```
martint-unity: Would this be easier to understand if it were like this?
createDepthTexture = createDepthTexture || (renderingData.cameraData.renderType == CameraRenderType.Base && !renderingData.resolveFinalTarget);
I assume it is the same thing.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -157,23 +166,31 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+
+            // Configure all settings require to start a new camera stack (base camera only)
+            if (cameraData.renderType == CameraRenderType.Base)
+            {
+                m_ActiveCameraColorAttachment = (createColorTexture) ? m_CameraColorAttachment : RenderTargetHandle.CameraTarget;
+                m_ActiveCameraDepthAttachment = (createDepthTexture) ? m_CameraDepthAttachment : RenderTargetHandle.CameraTarget;
```
martint-unity: Why are we having () here?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -157,23 +166,31 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
 
-            m_ActiveCameraColorAttachment = (createColorTexture) ? m_CameraColorAttachment : RenderTargetHandle.CameraTarget;
-            m_ActiveCameraDepthAttachment = (createDepthTexture) ? m_CameraDepthAttachment : RenderTargetHandle.CameraTarget;
-            bool intermediateRenderTexture = createColorTexture || createDepthTexture;
-            
-            if (intermediateRenderTexture)
-                CreateCameraRenderTarget(context, ref cameraData);
+                // Doesn't create texture for Overlay cameras as they are already overlaying on top of created textures.
+                bool createTextures = intermediateRenderTexture;
+                if (createTextures)
```
martint-unity: why not use intermediateRenderTexture?

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -78,8 +78,12 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
 
-            Camera camera = renderingData.cameraData.camera;
-            float cameraAspect = (float) camera.pixelWidth / (float) camera.pixelHeight;
+            ref CameraData cameraData = ref renderingData.cameraData;
+            Camera camera = cameraData.camera;
+
+            // In case of camera stacking we need to take the viewport rect from base camera
+            Rect pixelRect = renderingData.cameraData.pixelRect;
+            float cameraAspect = (float) pixelRect.width / (float) pixelRect.height;
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
```
martint-unity: I think our standard is without space between (float)pixelRect.width

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -327,6 +363,11 @@ protected static ClearFlag GetCameraClearFlag(CameraClearFlags cameraClearFlags)
             // RenderBufferLoadAction.DontCare in Vulkan/Metal behaves as DontCare load action
             // RenderBufferLoadAction.DontCare in GLES behaves as glInvalidateBuffer
 
+            // Overlay cameras composite on top of previous ones. They don't clear color.
+            // For overlay cameras we check if depth should be cleared on not.
+            if (cameraData.renderType == CameraRenderType.Overlay)
+                return (cameraData.clearDepth) ? ClearFlag.Depth : ClearFlag.None;
+
```
martint-unity: Do we need the () here?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -541,14 +575,23 @@ void FillBlockRanges(NativeArray<RenderPassEvent> blockEventLimits, NativeArray<
...
             for (int i = 0; i < m_ActiveRenderPassQueue.Count; ++i)
                 m_ActiveRenderPassQueue[i].FrameCleanup(cmd);
 
-            FinishRendering(cmd);
+            // Happens when rendering the last camera in the camera stack.
+            if (resolveFinalTarget)
+            {
+                for (int i = 0; i < m_ActiveRenderPassQueue.Count; ++i)
+                    m_ActiveRenderPassQueue[i].OnFinishCameraStackRendering(cmd);
+
```
martint-unity: foreach(var passQue in m_ActiveRenderPassQueue)
    passQue.OnFinishCameraStackRendering(cmd)

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
phi-lira: Fixed.

phi-lira:
Fix for BoatAttack issue submitted. VFX_LWRP test is not passing due to known issue in master (already fixed but not merged to this PR)

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
mikesfbay: @martin-unity3d Don't use foreach. It causes garbage collection: https://gamedev.stackexchange.com/questions/149914/why-is-the-for-loop-more-efficient-than-the-for-each-loop-in-unity

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -238,6 +270,105 @@ public static void RenderSingleCamera(ScriptableRenderContext context, Camera ca
...
+                            else
+                            {
+                                lastActiveOverlayCameraIndex = i;
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    Debug.LogWarning("Camera Stacking is not supported in VR. Only Base cameras will render.");
```
mikesfbay: Currently, this log adds a lot of console message spamming during VR modes and covers up other debug messages during VR. We can try these alternative:
1. not print this and add a note in the documentation on lack of VR support, 
2. print only once during playmode
3. print only when there's stacked cameras (currently, every scene has this)

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -238,6 +270,105 @@ public static void RenderSingleCamera(ScriptableRenderContext context, Camera ca
...
+
+            // We need to know the last active camera in the stack to be able to resolve
+            // rendering to screen when rendering it. The last camera in the stack is not
+            // necessarily the last active one as it users might disable it.
+            int lastActiveOverlayCameraIndex = -1;
+            if (cameraStack != null)
+            {
+                // TODO: Add support to camera stack in VR multi pass mode
+                // TODO: Add support to camera stack in VR multi view mode
+                if (!IsStereoEnabled(baseCamera))
```
mikesfbay: Since XR QA confirmed that camera stack works in VR desktop mode, we should enable it for VR, and XR will add fogbugz cases for multiview and multipass as things to do. This way, we can at least test for any camera stack feature regression on desktop when new XR PRs are QA-ed.

mikesfbay: I've confirmed that the FPS camera stacking demo works on both Quest and Rift. We can undo the stereo disable above and let QA reconfirm.

phi-lira:
Failing test in Universal CI - 005 Lit Baked Scene is known issue caused by new GI. 
Failing test in VFX Graph - 101 is also known issue (new GI)

PR #5267, opened 2019-12-06
Add Android texture overrides for EXR texture

tomzig16:
### Purpose of this PR
This PR updates templates for all active post processing tests that are using .exr format texture (`Mini_SyntheticChart.exr` file).
Android does not support .exr format and by default it changes format to RGBM (RGBM range is up to 8 bits, while EXR -65k to 65k). In result, template textures are rendered with artifacts. While logically they are correct, they might confuse people into thinking that something is wrong, so I have added Android texture importer overrides for it.
image
Notice `Format` setting in the screenshot. This makes the result look more like on Windows PC, not exactly the same, but close enough to not be confused as broken due to wrong colors and artifacts. I have added image to compare old version, new version and version used on DX11 below.

Also, while working on this PR I have noticed that 2 other tests started failing because of bad reference images, so I fixed them as well:
* Test 035 - updated reference image
* Test 105 - newly added test, no reference image was used. Copied from DX11 and there were no issues.

**Edit: forgot to mention that this PR adds .meta files for reference images which I accidentally missed the last time**

image
From left to right: Android old reference image, Android new reference image, reference image used in DirectX11. All three images are from 085 test

---
### Testing status
**Manual Tests**: 
- Ran test project locally via test runner
  - on PC
  - several Android devices

<br/>

**Automated Tests**: 
- Yamato Android test run
- Windows DX11 Editor test run
- Windows DX11 Playmode test run
- Windows DX11 Standalone test run

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/android%252Ftomaszi%252Ffix-tests

### Any test projects to go with this to help reviewers?
Changes are made only to UniversalGraphicsTest project


phi-lira:
Was the other PR that introduced Android tests flagged to be backported to 2019.3? If so, please add needs-backport-2019.3 tag on this one as well.

PR #5268, opened 2019-12-06
Fixing exposure for path tracing

anisunity:
Changing the maximum intensity volume component to a maximum exposure to be able to support scenes with high range. Changing the exposure application to the path tracing main so that it happens after the intensity clamp process that implies that we do not need to retrigger an accumulation because of the exposure.


anisunity:
removed the color pyramid generation

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/PathTracing/PathTracing.cs
```diff
@@ -21,8 +21,8 @@ public sealed class PathTracing : VolumeComponent
         [Tooltip("Defines the maximum number of bounces for each path.")]
         public ClampedIntParameter maximumDepth = new ClampedIntParameter(3, 1, 10);
 
-        [Tooltip("Defines the maximum intensity value computed for a path.")]
-        public ClampedFloatParameter maximumIntensity = new ClampedFloatParameter(10f, 0f, 100f);
+        [Tooltip("Defines the maximum exposure value computed for a path.")]
+        public ClampedFloatParameter maximumExposure = new ClampedFloatParameter(1.0f, 0.0f, 32.0f);
```
sebastienlagarde: As we use EV100 eveywhere in the UI please use EV100 here as well, thanks

anisunity:
After discusison, I am going to change the fix

PR #5269, opened 2019-12-06
Fixing ray traced reflections exposure issue

anisunity:
Fix a bug that broke the pre-exposition for ray traced reflections.
https://fogbugz.unity3d.com/f/cases/1202239/


sebastienlagarde:
Waiting for QA

PR #5270, opened 2019-12-09
Ray Traced Shadow Denoising Issues

anisunity:
Wrong amount of history buffers allocated for ray traced shadows
Wrong type for the history validation buffer

This meant that history was never used and denoising did not work when we had more than 4 ray trace shadows.


seblagarde:
Add changelog

PR #5278, opened 2019-12-09
Add ENABLE_VR_MODULE to #if ENABLE_VR

PixelJJ:
Add ENABLE_VR_MODULE to #if ENABLE_VR to allow building when built-in package VR Module is disabled. Fix https://fogbugz.unity3d.com/f/cases/1202892/

EDIT: Felipe. Turns out another PR broke again what this PR is trying to fix. I fixed it and disabled VR module from UniversalGraphicsTest so we catch these issues. This required me to update the GraphicsTest to also add the VR module defines.

---
### Testing status

- [x] Built a player


phi-lira:
Please add changelog.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -266,11 +266,13 @@ static void InitializeCameraData(UniversalRenderPipelineAsset settings, Camera c
             cameraData.isXRMultipass = false;
             cameraData.numberOfXRPasses = 1;
 
+#if ENABLE_VR && ENABLE_VR_MODULE
```
mikesfbay: Adding "ENABLE_VR_MODULE" to this line caused multipass to render only left eye with MockHMD or Oculus. The project was from template scene and had worked fine before. The package\manifest.json does contain define for VR package.

Changing to #if ENABLE_VR && !UNITY_XBOXONE worked fine.

DennisDeRykeUnity: @mikesfbay What you describe sounds like the following active bug: 
https://fogbugz.unity3d.com/f/cases/1182204/

mikesfbay: @DennisDeRykeUnity My multipass changes already went in master, so the fogbugz case should be closed soon.
The issue I brought up above is new to this PR.

DennisDeRykeUnity: My testing shows that Mike is correct.  His fix for case 1182204 went into master via the following PR prior to PixelJJ establishing this dev branch.
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5061

In Multi Pass stereo mode I see one eye entirely black in the three following configurations: 
Unity 2020.1a15, Oculus Go, OpenGLES3, XR SDK
Unity 2019.3.0f2, Oculus Rift, D3D11, XR SDK
Unity 2019.3.0f2, MockHMD, D3D11, Legacy XR 

The MockHMD test can be done using any scene on Windows without any VR headset.  Just remove XR Management via Package Manager, then enable XR at the bottom of Player settings.  Then in Player settings for the VR SDK choose MockHMD, and pick Multi Pass as the Stereo Mode.  In Game view choose Both Eyes.  In the Editor you can then enter Play mode and see left and right eye views in Game view, except one is black due to the bug.  For correct behaviour change to one of the other Stereo Modes.  Bugs that can be seen in MockHMD usually appear on multiple actual headsets.  

Ping me when this has been resolved and I'll resume testing of this PR.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -402,10 +402,10 @@ bool RequiresIntermediateColorTexture(ref RenderingData renderingData, RenderTex
             bool isOffscreenRender = cameraData.camera.targetTexture != null && !cameraData.isSceneViewCamera;
             bool isCapturing = cameraData.captureActions != null;
 
-#if ENABLE_VR
+#if ENABLE_VR && ENABLE_VR_MODULE
```
mikesfbay: Is adding ENABLE_VR_MODULE necessary in SetupBackbufferFormat? Single-pass mode desktop seems to work but I haven't tested when MSAA samples is changing for Quest, since the ENABLE_VR_MODULE will mask out this code if not defined. 

@DennisDeRykeUnity  we need to check case of msaa/antialiasing samples changing per frame for Quest/GO, if this ENABLE_VR_MODULE is used.

fabien-unity:
@DennisDeRykeUnity , @phi-lira : the defines were missing for URP. I fixed the asmdef. Please retest.
I've confirmed it fixes the case for MockHMD + DX11

mikesfbay:
@DennisDeRykeUnity we also need to check for dynamic MSAA/antialiasing samples changing per frame for Quest/GO since ENABLE_VR_MODULE is enabled for that part of code.

DennisDeRykeUnity:
@fabien-unity When you said you have confirmed that your asmdef changes  fixes the case for MockHMD + DX11, do you mean both eyes work in MultiPass on MockHMD with DX11?  

@mikesfbay I'll do another test round on this today.  If Fabien's commit re-enables MultiPass and if I can repeatedly change MSAA sampling at runtime via script, do you see other issues with this PR?

FYI, another PR which is intended to be a subset of this PR has already been merged: 
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5302

mikesfbay:
@DennisDeRykeUnity Other than msaa samples changing and multipass tests, I would also suggest testing MSAA + multiview for Quest/GO and MSAA + SPI on Rift

DennisDeRykeUnity:
A simple test scene on Rift works in both MultiPass and Stereo Instancing modes using this PR’s dev branch.  

Comparing this PR’s dev branch (universal/bugfix/1202892 commit 2ac1ff8) versus master (commit 464a721) I see no disparity in my test scene using:
Unity 2020.1.0a15 
Oculus Go 
MultiPass and MultiView stereo modes 

Some anomalies that I encountered: 
1) In MultiPass stereo mode I encounter the crash described in case 1203699, both on master and this PR’s dev branch. 
2) The script that I have typically used to adjust MSAA samples during runtime has no noticeable effect, both on master and this PR’s dev branch. 

Prior to approving I need to test PostProc, and I need to do more MSAA testing on both Rift and Go.

phi-lira:
@DennisDeRykeUnity and @mikesfbay I'll revert the PR #5302 until we land this one.
Please let me know. Thanks a lot for looking at it.

TestProjects/UniversalGraphicsTest/Packages/manifest.json
```diff
@@ -41,7 +41,6 @@
     "com.unity.modules.unitywebrequestwww": "1.0.0",
     "com.unity.modules.vehicles": "1.0.0",
     "com.unity.modules.video": "1.0.0",
-    "com.unity.modules.vr": "1.0.0",
```
thomas-zeng: Hi @PixelJJ This module is required for the current URP XR implementation. Without the module, URP XR is forced off. Is there any specific reason why we are removing this module? We were seeing XR forced off in URP UniversalGraphicsTest(XR).

phi-lira: Yes, there is. To catch XR compilation issues if users remove the module. This should affect only UniveralGraphicsTest not the GfxStereoTests.

PR #5283, opened 2019-12-09
Delay discard of fragments during decal projection

jessebarker:
Metal Shading Language declares that fragment discard invalidates
derivatives for the rest of the quad, so we need to reorder when
we discard during decal projection, or we get artifacts along the
edges of the projection (any partial quads get bad partial derivatives
regardless of whether they are computed implicitly or explicitly).

### Checklist for PR maker

---
### Purpose of this PR

See the description.  MacOS may not encounter this problem on all platforms, but there is no guarantee as it is very driver (and/or version) dependent.  The specification of this feature is, however, unambiguous for _all_ Metal platforms.  It is also worth mentioning that without a special extension (VK_EXT_shader_demote_to_helper_invocation), Vulkan will also have this behavior by default (unclear which GPU vendors actually support it).

---
### Testing status

- [x ] Built a player
- [ x] C# and shader warnings (supress shader cache to see them)
- Other: 

The HDRP Shader Graph sample scene was used to find this problem and confirm the fix.  The decal projection (puddle) will show a "seam" artifact around the edge of the projection volume if the issue is present (so the scene should look "normal")

---
### Comments to reviewers

N.b., Much of the diffs here are based upon the whitespace config of the checked-in file.  Actual diffs are much smaller.


sebastienlagarde:
@TomasKiniulis Because of the latest change that I have made, we need to do another QA pass. And we should do it for PS4 and Mac. Can you do it? Thanks!

TomasKiniulis:
> @TomasKiniulis Because of the latest change that I have made, we need to do another QA pass. And we should do it for PS4 and Mac. Can you do it? Thanks!

Mac displays the build correctly
Screenshot 2020-01-10 at 12 46 23

iPhone 11 also still display decals correctly
IMG_0006 (added additional shader graph puddle decal, just to make sure there are no artifacts)

@sebastienlagarde there are issues on PS4, mesh decals are displayed in pink and projectors are not rendered at all

sebastienlagarde:
@TomasKiniulis I have fixed the test on PS4. Was a simple mistake. no need for retesting as I have done it.

TomasKiniulis:
> @TomasKiniulis I have fixed the test on PS4. Was a simple mistake. no need for retesting as I have done it.

Great! Thanks!

PR #5295, opened 2019-12-10
Fix issue with vbufferparams not being initialized

FrancescoC-unity:
In case anything in IsVolumetricReprojectionEnabled() returns false while volumetrics are on, the vbufferParams would never be initialized. 

For QA: A repro for the previous issue is to have custom frame setting on a planar reflection probes and enable volumetrics. Without this fix, you should have a nullreference error


EvgeniiG:
Why wasn't I added as a reviewer? I have bug assigned to me, and trying to figure out why it's not reproducing on staging is a PITA.

In any case, the fix is not good, because it breaks (or doesn't fix) the fog toggle on Editor restart. This one: 

image

I will try to to make a proper fix.

EvgeniiG:
Note: the problem is even more complicated than I thought.

EvgeniiG:
See https://unity.slack.com/archives/GHD5LADU7/p1576547420026800

FrancescoC-unity:
Hey apologies for not adding you as reviewer, I was convinced I did :/

PR #5301, opened 2019-12-11
Fix a typo in VFX Shader Graph output code

keijiro:
Fixed a typo (missing closing brace) that causes a shader compilation error when using a View Direction node in a shader graph and assigning it to a particle output context of a visual effect graph.

Fogbugz case: https://fogbugz.unity3d.com/f/cases/1204027

Here is a test project that simply applies view direction as per-pixel particle color in a VFX graph. It causes a shader compilation error before applying this fix: https://github.com/keijiro/SrpTest


julienf-unity:
This has already been fixed in vfx/staging and backported to 2019.3

PR #5303, opened 2019-12-11
Hdrp/fix search preference settings

fredericv-unity3d:
### Purpose of this PR
Fixes https://fogbugz.unity3d.com/f/cases/1190223/.

There are several causes for this bug.
1. The appropriate keywords were not provided during the creation of the `SettingsProvider`
1. `UIElements` based settings panel implementations don't support highlighting fields.

These is the strategy used to fix the issue:
1. Refactored the quality settings and default settings panel with an IMGUI implementation
1. Refactored the `GUIContent` used as a label in a static class so it can be used with `SettingsProvider.GetSearchKeywordsFromGUIContentProperties<T>()` to fetch the keywords
1. Added a static property `OverridableFrameSettingsArea.frameSettingsKeywords` to provide the keywords used by the frame settings UI.

---
### Testing status

**Manual Tests**: Manual testing of the search field in the project settings window.

No automated test for this UI only feature.

No public documentation changed nor required in this PR

---
### Comments to reviewers

For QA: I suggest a manual testing of the search field for both project settings panels 'HDRP Default Settings' and 'Quality/HDRP'.


fredericv-unity3d:
>     * Search edit field element is very unresponding to user events. Needs a lot of clicks to focus on it or select text.

I agree, but this is more a trunk issue common to all panels. It should be reported as a separate bug.

>     * Can't highlight property categories even though search "finds" them

That is correct and is the behavior of builtin panels (For instance, in 'Quality' the 'rendering' title is not highlighted, only fields are highlighted).

Although, I feel the same, but this should be discussed more broadly.

PR #5313, opened 2019-12-11
Deleted SG Asset Dialog

zadi0648:
Implementation of this old card "_Do Something (Anything) when the Shader Graph file is deleted while the window into the graph is still open_":
https://favro.com/organization/c564ede4ed3337f7b17986b6/d06b8de3e40da6d340b5b063?card=Uni-49784

DeletionSaveAsMenu

---
**General Notes**
- Stops users from entering in a still open Shader Graph that is in a buggy soft lock mode; they either ought to close the Shader Graph or last chance to Save As.
- Works with all graphs, allowing users a last chance Save As if they regret the deletion of the Shader Graph
- This popup does not happen until the window has been selected.
- Note that "Save As" is the secondary, not the primary, option. Please feel free to give feedback on that. Although that's abnormal as far as saving stuff goes, I did that because I figured that primary should be the default use case, after all the user will have deleted the file on purpose 96% of the time.

---
**Notes to QA**
- ^ stuff above
- Should work with various window arrangements
- If you have the bug where Save As opens in the wrong project, that should be fixed in another PR of mine. Here is a combined branch to test that: sg/combined/deletion-dialog-menu-navigation
- More ways to delete stuff you care to test? Simple script that deletes Assets? That would be more testing unity's asset deletion detection system, but might be good nonetheless if you care to do that.
- V See manually tested.

---
**Notes to Docs:**
Since this has a new Dialog popup, is the wording okay?:
_Note:_ this has been changed since PR has been first created.

"_____ Graph Asset Missing!"
"_Path-to-file-&-file-name_ has been deleted or moved outside of Unity."
_space_
"What do you wish to do?"
1) "Close Window"
2) "Save As"
Screen Shot 2019-12-20 at 5 40 42 PM

---
**Testing:**

Manually Tested:
- Delete -> Don't Save (Close), Delete -> Save As, Delete -> Save As the same file, many times and ways
- Deleting inside of Unity and outside of Unity via File Explorer
- Shader Graphs and Sub Graphs
- Various window / tab arrangements. The popup (should) only popup once the tab or Shader Graph window has been selected, as opposed to immediately when the file has been deleted.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fdeletion-dialog/.yamato%252Fupm-ci-packages.yml%2523all_package_ci_fast-2020.1/1539611/job

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fdeletion-dialog/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk


zadi0648:
^ Yeah we resolved that, was lacking a commit

zadi0648:
Landon and I talked about the first, decided that default action should be what other dialogs do

Nice catch on the moving outside of the project

zadi0648:
Options have been switched.

Unfortunately we can't have a reliable solution to moving stuff outside of Unity. I thought there might be more info to go off such that we could determine if the asset that deleted is the same as the one that got imported, but all we are given is a string and there is no way to tell if the user added a different sg file of the same name somewhere else while deleting this one for example. Afaik Unity doesn't handle that situation as a move anywhere else in Unity, and as such is a greater Unity problem; hopefully documented somewhere too(?): you need to also move the meta files for that to work with any Unity file.

zadi0648:
Sharlene I requested a review because the dialog text has been changed and I just wanted to confirm that the new dialog is okay 😄

martint-unity:
Have you run Yamato on this?

com.unity.shadergraph/Editor/Importers/ShaderGraphAssetPostProcessor.cs
```diff
@@ -43,15 +43,36 @@ static void UpdateAfterAssetChange(string[] newNames)
             }
         }
 
+        static void DisplayDeletionDialog(string[] deletedAssets)
+        {
+            // This will change the title of the window.
```
martint-unity: I don't see it changing the title?

martint-unity:
What about just putting an * on the tab and let the user them selves decide what to do with that? We have "Save Asset" and "Save As" as options already there in the menu bar.

pbbastian:
> Stops users from entering in a still open Shader Graph that is in a buggy soft lock mode

Can you elaborate on this?

> they either ought to close the Shader Graph or last chance to Save As.

Why? The window should be able to stay open without a file, so I don't see why we should force the user to close or save the file.

pbbastian:
`*` on the tab would be nice, and then have "Save Asset" act like "Save As" if the original asset (based on GUID, not file name) is gone. Also, that should happen immediately, and not just when you focus the window.

pbbastian:
BTW, these types of dialogs should be more in the form of continue/cancel. "What do you want to do?" is too vague, and doesn't imply what the ongoing action is. See [EditorUtility.DisplayDialog](https://docs.unity3d.com/ScriptReference/EditorUtility.DisplayDialog.html) and [EditorUtility.DisplayDialogComplex](https://docs.unity3d.com/ScriptReference/EditorUtility.DisplayDialogComplex.html) for good examples of this. Also, avoid exclamation marks in the title :)

zadi0648:
@pbbastian @martint-unity 

Yeah so we discussed this here and agreed that allowing the user to continue to work on the Shader Graph while the underneath Graph file underneath is deleted _would be nice_; the problem is that it's not possible. That was according to Marc's, Alex's, and my current understanding. However I didn't dive into that, perhaps it is an easy fix to allow them to do that, just a bunch of null checks.

So to clarify, I'm not stopping users from continuing to work on a SG that has no file, it's already impossible to do so. Editing a SG with no file is unsupported atm. I'm simply stopping them before they enter that broken state (& also allowing them to continue to work on the SG if they chose to save).
So yeah, currently the Shader Graph cannot be altered in any way once the associated file is gone. It's in a broken soft lock state: nodes stick to the cursor and it cannot be saved, ect.

GraphObject is null if there is no file underneath the object. I was informed that the file needs to exist in order to get Sub Graph information. Maybe that's not true, maybe there needs to be just some null checks.

Also, while most software allows users to continue to edit a file with it's associate file gone, we found other software does act like how this PR acts. For example Visual Studio.

Supporting allowing the user to edit the SG with no file I'd imagine requires a larger change. Once they happens, we can remove this dialog. Or you can ask me to look into it instead of this PR, but I probably won't be getting around to that for some time and in the meantime users will encounter the broken state.

Also I'll fix the dialog wording concerns

pbbastian:
It's definitely not impossible for Shader Graph to work like this. The fact that it doesn't is a bug. I think it used to work. It can be broken now, but still supported. That's one reason why we should fix it :)

But, after talking with @marctem about this, I agree that we should land your fix now, and then track the issue on the backlog. Because it does fix an issue here and now.

IMO it is a much nicer behaviour to have Shader Graph keep working on deleted files (just * the title), and I think the fact that it currently doesn't represents some general nastiness in that part of the code base, which could pop up as other bugs or make development harder => very worthy of fixing.

zadi0648:
> IMO it is a much nicer behaviour to have Shader Graph keep working on deleted files (just * the title), and I think the fact that it currently doesn't represents some general nastiness in that part of the code base
Agreed 😺

zadi0648:
bors +m

zadi0648:
bors +r

zadi0648:
Passed 💚 :

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fdeletion-dialog/.yamato%252Fupm-ci-packages.yml%2523all_package_ci_fast-2020.1/1539611/job/pipeline

PR #5322, opened 2019-12-12
Fix NPE when using scalable settings and first editor start

fredericv-unity3d:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1192046/.

A scalable settings' value is required while providing a null reference to the source setting.

There are two issues:
1. The template's `HDRenderPipelineAsset` is incorrect: `m_RenderPipelineSettings.lightSettings.useContactShadow` is not filled properly.
1. For some reasons, Unity's serialization framework has filled this struct with a null pointer for the `useContactShadow` property. Even after the `OnAfterDeserialize` call.

The solution here is to avoid the NPE when resolving the scalable setting and instead forcing to use the custom value even if a level is requested (but the level values provider is `null`).

This issue happens only for the very first frame the editor starts, so it is fine to ignore silently the serialization error for those frames.

---
### Testing status

**Manual Tests**: Tested manually
1. On a fresh project by following the stated steps
1. Verifying that the property work as expected before first play mode
1. Verifying that the property work as expected before / after play mode, after closing opening unity

**Automated Tests**: -

**Yamato**:-

---


sebastienlagarde:
Merging for 7.1.7 package, but please QA do test.

PR #5337, opened 2019-12-13
Fix/xr2

kaychang-unity:
---
### Purpose of this PR
Reverted some change in matrix for XR mode, as they were incorrect.

---
### Testing status

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/CHANGELOG.md
+++ b/com.unity.render-pipelines.universal/CHANGELOG.md
@@ -61,7 +61,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue where using a ForwardRendererData preset would cause a crash. [case 1201052](https://issuetracker.unity3d.com/product/unity/issues/guid/1201052/)
 - Fixed an issue where particles had dark outlines when blended together [case 1199812](https://issuetracker.unity3d.com/issues/urp-soft-particles-create-dark-blending-artefacts-when-intersecting-with-scene-geometry)
 - Fixed an issue with deleting shader passes in the custom renderer features list [case 1201664](https://issuetracker.unity3d.com/issues/urp-remove-button-is-not-activated-in-shader-passes-list-after-creating-objects-from-renderer-features-in-urpassets-renderer)
-- Fixed camera inverse view-projection matrix in XR mode, depth-copy and color-copy passes.
+- Fixed depth-copy and color-copy passes.
 - Fixed an issue with the null check when `UniversalRenderPipelineLightEditor.cs` tries to access `SceneView.lastActiveSceneView`.
 - Fixed an issue where the 'Depth Texture' drop down was incorrectly disabled in the Camera Inspector. 
 
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -37,7 +37,7 @@ namespace UnityEngine.Rendering.Universal
         static internal class PerCameraBuffer
         {
             // TODO: This needs to account for stereo rendering
-            public static int unity_MatrixInvVP;
+            public static int _InvCameraViewProj;
             public static int _ScaledScreenParams;
             public static int _ScreenParams;
             public static int _WorldSpaceCameraPos;
@@ -127,7 +127,7 @@ namespace UnityEngine.Rendering.Universal
             PerFrameBuffer.unity_DeltaTime = Shader.PropertyToID("unity_DeltaTime");
             PerFrameBuffer._TimeParameters = Shader.PropertyToID("_TimeParameters");
 
-            PerCameraBuffer.unity_MatrixInvVP = Shader.PropertyToID("unity_MatrixInvVP");
+            PerCameraBuffer._InvCameraViewProj = Shader.PropertyToID("_InvCameraViewProj");
             PerCameraBuffer._ScreenParams = Shader.PropertyToID("_ScreenParams");
             PerCameraBuffer._ScaledScreenParams = Shader.PropertyToID("_ScaledScreenParams");
             PerCameraBuffer._WorldSpaceCameraPos = Shader.PropertyToID("_WorldSpaceCameraPos");
@@ -581,7 +581,7 @@ namespace UnityEngine.Rendering.Universal
             Matrix4x4 viewMatrix = camera.worldToCameraMatrix;
             Matrix4x4 viewProjMatrix = projMatrix * viewMatrix;
             Matrix4x4 invViewProjMatrix = Matrix4x4.Inverse(viewProjMatrix);
-            Shader.SetGlobalMatrix(PerCameraBuffer.unity_MatrixInvVP, invViewProjMatrix);
+            Shader.SetGlobalMatrix(PerCameraBuffer._InvCameraViewProj, invViewProjMatrix);
         }
 
 
--- a/com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
+++ b/com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
@@ -42,7 +42,7 @@ struct InputData
 half4 _GlossyEnvironmentColor;
 half4 _SubtractiveShadowColor;
 
-#define _InvCameraViewProj unity_MatrixInvVP
+float4x4 _InvCameraViewProj;
 float4 _ScaledScreenParams;
 
 float4 _MainLightPosition;
@@ -67,7 +67,11 @@ half4 _AdditionalLightsOcclusionProbes[MAX_VISIBLE_LIGHTS];
 #define UNITY_MATRIX_P     OptimizeProjectionMatrix(glstate_matrix_projection)
 #define UNITY_MATRIX_I_P   ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED
 #define UNITY_MATRIX_VP    unity_MatrixVP
-#define UNITY_MATRIX_I_VP  unity_MatrixInvVP
+#if defined(USING_STEREO_MATRICES)
+#define UNITY_MATRIX_I_VP  unity_MatrixInvVP // only defined in stereo-mode
+#else
+#define UNITY_MATRIX_I_VP  _InvCameraViewProj // Technically incorrect as per-pass matrix is a different set from camera set matrix.
+#endif
 #define UNITY_MATRIX_MV    mul(UNITY_MATRIX_V, UNITY_MATRIX_M)
 #define UNITY_MATRIX_T_MV  transpose(UNITY_MATRIX_MV)
 #define UNITY_MATRIX_IT_MV transpose(mul(UNITY_MATRIX_I_M, UNITY_MATRIX_I_V))
--- a/com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
+++ b/com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
@@ -189,7 +189,6 @@ float4x4 glstate_matrix_projection;
 float4x4 unity_MatrixV;
 float4x4 unity_MatrixInvV;
 float4x4 unity_MatrixVP;
-float4x4 unity_MatrixInvVP;
 float4 unity_StereoScaleOffset;
 int unity_StereoEyeIndex;
 #endif


kaychang-unity:
I reverted the change for _InvCameraViewProj: it is never equivalent to unity_MatrixInvVP because the former applies z adjustments, while the later does not.
However, original code reverted defines UNITY_MATRIX_I_VP as _InvCameraViewProj (which is wrong), so I just added a comment since there is no possible fix to it without adding C++ code.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -65,9 +65,13 @@ half4 _AdditionalLightsOcclusionProbes[MAX_VISIBLE_LIGHTS];
...
 #define UNITY_MATRIX_I_V   unity_MatrixInvV
 #define UNITY_MATRIX_P     OptimizeProjectionMatrix(glstate_matrix_projection)
-#define UNITY_MATRIX_I_P   ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED
+#define UNITY_MATRIX_I_P   unity_CameraInvProjection // workaround: technically per-pass matrix set is different from camera matrix set
 #define UNITY_MATRIX_VP    unity_MatrixVP
-#define UNITY_MATRIX_I_VP  unity_MatrixInvVP
+#if defined(USING_STEREO_MATRICES)
+#define UNITY_MATRIX_I_VP  unity_MatrixInvVP // only defined in stereo-mode
+#else
+#define UNITY_MATRIX_I_VP  _InvCameraViewProj // incorrect: _InvCameraViewProj applies platform-specific z adjustments (reversed-z and [0,1] range), while unity_MatrixInvVP does not.
```
phi-lira: The apply of platform-specific adjustments should not matter because of name of the var. So I don't understand why we can't bind as unity_MatrixInvVP.

From what I look it seems to be like this. 
If we call CommandBuffer.SetViewProjectionMatrix(view, proj) -> Unity will apply platform-specific z adjustments in c++ here.

If we call SetGlobalMatrix() it won't. However, the API GL.GetGPUMatrix(proj) will apply these.
So from what I look from the code, maybe we should remove the GL.GetGPUMatrix from here as the we might been applying twice?

https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs#L582-L586

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -61,7 +61,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed issue where using a ForwardRendererData preset would cause a crash. [case 1201052](https://issuetracker.unity3d.com/product/unity/issues/guid/1201052/)
 - Fixed an issue where particles had dark outlines when blended together [case 1199812](https://issuetracker.unity3d.com/issues/urp-soft-particles-create-dark-blending-artefacts-when-intersecting-with-scene-geometry)
 - Fixed an issue with deleting shader passes in the custom renderer features list [case 1201664](https://issuetracker.unity3d.com/issues/urp-remove-button-is-not-activated-in-shader-passes-list-after-creating-objects-from-renderer-features-in-urpassets-renderer)
-- Fixed camera inverse view-projection matrix in XR mode, depth-copy and color-copy passes.
+- Fixed depth-copy and color-copy passes.
+- Fixed ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED shader error.
```
phi-lira: This changelog should be in shader graph as well.

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
kaychang-unity: I don't know who added the call to GL.GetGPUMatrix(proj) but was under the impression removing it would be "bad" for existing user projects that may rely on this parameter. Anyhow, I can definitely remove it. In that case _inCameraViewProj will be equal to unity_MatrixInvVP indeed.

kaychang-unity: Fixed.

com.unity.render-pipelines.universal/CHANGELOG.md
kaychang-unity: Fixed.

kaychang-unity:
@phi-lira Next bug I found is in XR Multi-pass.
_InvCameraViewProj/unity_MatrixInvVP is updated from C# code at
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/621e0f29bab9435f66f4f42a07916ccd4ddcd14f/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs#L568 
However, this is outside of the loop for rendering each eyes in multi-pass:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/621e0f29bab9435f66f4f42a07916ccd4ddcd14f/com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs#L234

As a result, the right eye is not correctly rendered. This issue does not affect "Single pass" and "Single Pass Instanced" because in these modes, unity_MatrixInvVP is setup from C++ code.

To sum up, the bug is that using UNITY_MATRIX_I_VP in XR Multi-pass will generate incorrect result for the right eye.

kaychang-unity:
@phi-lira  I committed a fix to have correct UNITY_MATRIX_I_VP in XR Multi-pass with revision 
41f491d. Please advise if you want this fix or have it reverted.

ellioman:
@kaychang-unity Can you fix the conflicts?

@phi-lira Can you re-review this and check the two comments above this one?

kaychang-unity:
I fixed the conflicts and it compiles again.
I'll test XR multi-pass with the MockHMD tomorrow, once I figure out how to enable it again (the old UI has disappeared, now asking XR channel).

kaychang-unity:
Multipass cannot be tested at the moment. UniversalGfxTestStereo regressed in master branch to the point only SinglePass is working (SinglePass instanced and Multipass are broken). I was told UniversalGfxTestStereo will be soon updated from using LegacyXR to the new XRSDK. 
UniversalGraphicsTests is also currently broken for XR support (using new XR SDK).

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -766,24 +765,31 @@ static void SetupPerFrameShaderConstants()
             Shader.SetGlobalVector(PerFrameBuffer._SubtractiveShadowColor, CoreUtils.ConvertSRGBToActiveColorSpace(RenderSettings.subtractiveShadowColor));
         }
 
-        static void SetupPerCameraShaderConstants(in CameraData cameraData)
+        internal static void SetupPerCameraShaderConstants(ScriptableRenderContext context, CameraData cameraData, bool stereoEnabled, int eyeIndex)
```
phi-lira: Check if the ScripableRenderer already creates a command buffer we could reuse. 
Also, we might as well move this to ScriptableRenderer.

phi-lira:
PR seems good to me, I'd like to get 
@thomas-zeng opinion/approval before merging. 

Thomas is also changing how to setup camera matrix logic here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4857

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -289,6 +289,7 @@ public void Execute(ScriptableRenderContext context, ref RenderingData rendering
             /// * Setup global time properties (_Time, _SinTime, _CosTime)
             bool stereoEnabled = renderingData.cameraData.isStereoEnabled;
             context.SetupCameraProperties(camera, stereoEnabled, eyeIndex);
+            UniversalRenderPipeline.SetupPerCameraShaderConstants(context, renderingData.cameraData, stereoEnabled, eyeIndex);
```
phi-lira: This change will introduce a regression because some of these properties are required to be set before, f.ex, _WorldSpaceCameraPos.

Maybe we need to split this into two functions. One for setting up just camera matrices, other for everything else.

kaychang-unity: I split it to 2 functions:
- original function is back where it was with most of the original code.
-  a new SetupPerCameraMatrices() that only update unity_MatrixInvVP.
Do you still want this function to be moved to ScriptableRenderer (with different name)?

phi-lira: "Do you still want this function to be moved to ScriptableRenderer" That's fine to be like it is now. It's internal so we can always change later if neeed.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -834,12 +834,18 @@ static void SetupPerCameraShaderConstants(in CameraData cameraData)
...
+        }
+
+        internal static void SetupPerCameraMatrices(CommandBuffer cmd, ScriptableRenderContext context, CameraData cameraData, bool stereoEnabled, int eyeIndex)
+        {
+            // The following camera properties must be set per eye for XR support.
+            Camera camera = cameraData.camera;
 
-            Matrix4x4 projMatrix = GL.GetGPUProjectionMatrix(camera.projectionMatrix, false);
-            Matrix4x4 viewMatrix = camera.worldToCameraMatrix;
+            Matrix4x4 projMatrix = stereoEnabled ? camera.GetStereoProjectionMatrix((Camera.StereoscopicEye)eyeIndex) : camera.projectionMatrix;
```
thomas-zeng: In SPI and MultiView pass, camera inv projection is matrix4x4Array type. Shader indexed into this array based on eye idx (instance id in SPI mode and view id in MultiView mode).

The logic here of setting stereo proj and view only works for multipass.

thomas-zeng: I think we should either be clear this is not supported in XR at all(throw a warning and communicate XR is not supported to user)
or support multipass, SPI and Multiview fully.

Let me know what you think! :)

kaychang-unity: Yes, this code only make sense if we are in multipass mode. I can add extra logic to only run this code if we are in that mode.
I don't know what are the plans as to which modes XR wants to support long term though.

kaychang-unity:
@thomas-zeng Does the latest commit addresses your concern?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: URP long term will support:
Multiview, SPI, Multipass 
Android platforms that supports multiview extension eg, Quest, MagicLeap run Multiview
PC&Console eg WMR, rift run SPI

Multi-pass is a fallback only.

thomas-zeng:
I am still having concerns with the latest commit. What if the user is on SPI/Multiview which I expect most XR users will be on? Does the shader using inv camera projection work for them?
If not, we need to emit not supported message to them otherwise they will see broken rendering result and wonder what's going wrong.

kaychang-unity:
Please confirm whether my understanding is correct. When SPI/Multiview is enabled:
- `IsMultiPassStereoEnabled(camera)` return false
- `USING_STEREO_MATRICES` is defined in shader, so the following define will be used:
`#define unity_MatrixInvVP mul(unity_StereoMatrixInvV[unity_StereoEyeIndex], unity_StereoCameraInvProjection[unity_StereoEyeIndex])`

If that's correct, then I think it should be fine.
I am not really able to test XR multipass myself at the moment (I don't have an actual XR device and new mockHMD is still not working in URP, right?).

thomas-zeng:
Hey, thanks kay for your explanation.
With the shader ifdef USING_STEREO_MATRICES logic, I believe XR should be fine.

MockHMD should be working in URP now. 
We should also have XR QA coverage on effect that is using unity_MatrixInvVP.
Could you suggest which test case we should cover here?

thomas-zeng:
Adding XR QA as reviewer

kaychang-unity:
Thanks, I will try running the mockHMD again tomorrow.

For testing, I don't have a naturally occurring test case. However, I can artificially make the deferred renderer use `UNITY_MATRIX_I_VP` in XR Multipass.
If you replace the line at https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/289e6f5658587029f8bedeb74ff8585f45c70b97/com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader#L103 with `#if 0`, you can force deferred shading code to take the pass that use `UNITY_MATRIX_I_VP`.

phi-lira:
@DennisDeRykeUnity any update on this?

DennisDeRykeUnity:
I see that "_ABV for SRP repository - fast-trunk" is green on Yamato (branch: fix/xr2, commit: 3176f753): 
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/fix%252Fxr2/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/1325921/job

I tested this build using: 
- XRSDK
- URP
- URP Template scene, with exaggerated Bloom Tint and Intensity
- Oculus Rift in Play mode
- MultiPass and Stereo Instancing
- Unity 2020.1.0a23

I intend to do a similar test on Oculus Go prior to approving this PR.  

@kaychang-unity Beyond this, is there any additional XR testing that you recommend?  Is there some specific shader that I should use to exercise your code changes?

kaychang-unity:
@DennisDeRykeUnity Not particularly. But unrelated to XR support, this should fix this fogbugz case: https://fogbugz.unity3d.com/f/cases/1115116

phi-lira:
Missing shader graph reviewers.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -76,6 +76,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where particles had dark outlines when blended together [case 1199812](https://issuetracker.unity3d.com/issues/urp-soft-particles-create-dark-blending-artefacts-when-intersecting-with-scene-geometry)
 - Fixed an issue with deleting shader passes in the custom renderer features list [case 1201664](https://issuetracker.unity3d.com/issues/urp-remove-button-is-not-activated-in-shader-passes-list-after-creating-objects-from-renderer-features-in-urpassets-renderer)
 - Fixed camera inverse view-projection matrix in XR mode, depth-copy and color-copy passes.
+- Fixed _InvCameraViewProj shader parameter to not be calculated from GL.GetGPUMatrix(proj) anymore.
```
pbbastian: Unmatched underscore messes up the markdown formatting for the rest of the file.
```suggestion
- Fixed `_InvCameraViewProj` shader parameter to not be calculated from GL.GetGPUMatrix(proj) anymore.
```

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated the zoom level to let you zoom in further.
 - If Unity Editor Analytics are enabled, Shader Graph collects anonymous data about which nodes you use in your graphs. This helps the Shader Graph team focus our efforts on the most common graph scenarios, and better understand the needs of our customers. We don't track edge data and cannot recreate your graphs in any form.
 - The Create Node Menu now has a tree view and support for fuzzy field searching.
+- Fixed a `ERROR_UNITY_MATRIX_I_P_IS_NOT_DEFINED` shader error.
```
pbbastian: Since this doesn't have any changes to Shader Graph itself, the change log item should go into the Universal change log.

kaychang-unity: I reverted that line of code.

kaychang-unity:
@LandonTownsendUnity I confirm that  UNITY_MATRIX_I_VP **IS NOT** the inverse of UNITY_MATRIX_VP. As to how it came out to be this way .... I dont know.
I think UNITY_MATRIX_I_VP  does not include platform specific adjustment (UNITY_REVERSED_Z, UNITY_UV_STARTS_AT_TOP and NDC z half/full unit cube), while UNITY_MATRIX_VP does, which is why these multiplying these matrices together does not give identity.
@phi-lira we have similar issue with UNITY_MATRIX_P and UNITY_MATRIX_I_P, where if we define 
UNITY_MATRIX_I_P == unity_CameraInvProjection, then UNITY_MATRIX_I_P **IS NOT** the inverse of UNITY_MATRIX_P.

phi-lira:
@kaychang-unity while looking at y-flip issues in URP I've realized it's an oddly hard situation to fix if we don't handle camera matrices fully in URP side (due to engine SetRenderTarget, SetViewProjectionMatrices, Blit and SetCameraProperties having logic to patch projection matrix.)
So I started working on it #5913

I could take this PR and merge into #5913 or we wait until #5913 lands to land this one. What do you think it's best?

kaychang-unity:
@phi-lira I would prefer if you take this PR and merge it into #5913 . As you said we need URP to fully handle the camera matrices to solve the issue properly. It also means that #5913 will also have the fix for https://fogbugz.unity3d.com/f/cases/1115116 .

PR #5339, opened 2019-12-13
Discard invalid pixels when doing deferred lighting in compute (for camera stacking)

pmavridis:
When configuring HDRP to do camera stacking (by not clearing color/depth and clearing the stencil), an extra check is required in deferred compute lighting to ensure we don't write on pixels that have a valid depth from a previous camera in the stack, but were not covered by the current camera.

---

### Testing status

I have tested this on a compositor that is setting up camera stacking on-top of HDRP. Poke me if you want a copy of the project (it is a few GBs).

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
@@ -112,6 +112,7 @@ CBUFFER_END
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoopDef.hlsl"
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl"
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl"
+#include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStencilUsage.cs.hlsl"
 
 #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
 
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
```diff
@@ -176,6 +176,13 @@ void SHADE_OPAQUE_ENTRY(uint3 dispatchThreadId : SV_DispatchThreadID, uint2 grou
         return;
     }
 
+    // This is required for camera stacking: Since the depth is preserved between layers, we might have a valid depth but invalid normal
+    // Note: Renderdoc does not show any extra fetch for thic check, since the same value is also used/fetched later
+    if (all(FetchRawNormalBuffer(posInput.positionSS) == UNITY_INVALID_NORMALBUFFER_VALUE))
```
sebastienlagarde: With the change of stencil that is coming, we will have a stencil bit: isRequiringDeferredLighting, that we could test here and thus don't have any artifact. Tought?

sebastienlagarde:
Waiting for stencil management PR

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/Deferred.compute
pmavridis: Yes, this makes sense. But I'm wondering if we can do it without introducing a new memory fetch (especially on platforms that don't pack stencil with depth). I will have a look again after the stencil refactor.

sebastienlagarde: Indeed on PS4 it maybe a new fetch (on GCN in general), however it solve currently a real issue, like what you fixed in this PR that we have sometime with custom pass as well. i.e when a tile is accepted because there is some pixel valid in it but it shouldn't process those interal pixel. So I think this is desirable in the end. We will check after the stencil refector.

pmavridis:
Note: in the non-compute case, this issue is resolved with this PR:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5627

PR #5342, opened 2019-12-13
[skip CI] Add visible in scene view property in Custom Post Process API

alelievr:
### Purpose of this PR
Add a public virtual bool in the custom post processes API to specify if a post processes should be executed in the scene view.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -2441,15 +2441,18 @@ bool RenderCustomPostProcess(CommandBuffer cmd, HDCamera camera, ref RTHandle so
 
                 if (customPP is IPostProcessComponent pp && pp.IsActive())
                 {
-                    var destination = m_Pool.Get(Vector2.one, k_ColorFormat);
-                    CoreUtils.SetRenderTarget(cmd, destination);
-                    using (new ProfilingSample(cmd, customPP.name))
+                    if (!(camera.camera.cameraType == CameraType.SceneView && !customPP.visibleInSceneView))
```
Chman: Could be simplified a bit ;)
```
if (camera.camera.cameraType != CameraType.SceneView || customPP.visibleInSceneView)
```

alelievr: done

PR #5343, opened 2019-12-13
Fix invalid near and far plane values

fredericv-unity3d:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1182017/

There are several causes for this issue:
1. A managed camera was destroyed incorrectly
1. The camera settings allowed invalid values for near and far clip planes

To solve this, I:
1. Removed the invalid destroy call
1. Updated the camera settings to enforce the use of valid near and far clip plane values
1. Updated the UI to constraint the range of valid values

Include also a fix on `FixWarningGameObjectSceneIDMapIsMissingInLoadedScenes`.
If the scene was not already dirty, the fixed result won't be commited on the disk.
Marking the edited scene as dirty appropriately address this issue.
Fix provided by @danielrf

---
### Testing status

**Manual Tests**: Manual testing of the range of far and near clip plane values
1. When editing the UI
1. When deserializing invalid values by modifying the scene asset by hand

**Automated Tests**: -

**Yamato**: -

---
### Comments to reviewers
XML documentation was updated for some public APIs.


com.unity.render-pipelines.high-definition/Runtime/Utilities/CameraSettings.cs
```diff
@@ -150,10 +158,44 @@ public enum Mode
...
-            /// <summary>Far clip plane distance.</summary>
-            public float farClipPlane;
-            /// <summary>Near clip plane distance.</summary>
-            public float nearClipPlane;
+
+            /// <summary>
+            /// Far clip plane distance.
+            ///
+            /// Value that will be stored for the far clip plane distance.
+            /// IF you need the effective far clip plane distance, use <see cref="farClipPlane"/>.
```
reallyfancy: To be clear, do you mean that the value used for calculations is farClipPlane, and this may be clamped?

com.unity.render-pipelines.high-definition/Runtime/Utilities/CameraSettings.cs
```diff
@@ -150,10 +158,44 @@ public enum Mode
...
+            /// </summary>
+            [FormerlySerializedAs("farClipPlane")]
+            public float farClipPlaneRaw;
+            /// <summary>
+            /// Near clip plane distance.
+            ///
+            /// Value that will be stored for the near clip plane distance.
+            /// IF you need the effective near clip plane distance, use <see cref="nearClipPlane"/>.
```
reallyfancy: Do you mean that the value used for calculations is nearClipPlane, and this may be clamped?

com.unity.render-pipelines.high-definition/Runtime/Utilities/CameraSettings.cs
fredericv-unity3d: The value stored on disk is `farClipPlaneRaw` and the value used in calculations is `farClipPlane`.
`farClipPlane` is the clamped value of `farClipPlaneRaw`

com.unity.render-pipelines.high-definition/Runtime/Utilities/CameraSettings.cs
fredericv-unity3d: Exactly like `farClipPlane`

sebastienlagarde:
Note: The PR was,'t compiling. CameraSettingsUtilitiesTest.cs was not compiling due to name change in nearClipPlaneRaw/farClipPlaneRaw.

I am wondering how this could have been tested?

PR #5349, opened 2019-12-13
Custom pass materials editor in the Custom Pass Volume component

alelievr:
### Purpose of this PR
Added a material editor for every material referenced in the custom pass volume.

image


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPass.cs
```diff
@@ -305,6 +305,13 @@ protected RTHandle GetNormalBuffer()
             return currentRTManager.GetNormalBuffer(IsMSAAEnabled(currentHDCamera));
         }
 
+        /// <summary>
+        /// List all the materials that needs to be displayed with at the bottom of the component.
```
iM0ve:
```suggestion
        /// List all the materials that need to be displayed at the bottom of the component.
```

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPass.cs
```diff
@@ -305,6 +305,13 @@ protected RTHandle GetNormalBuffer()
...
+        /// All the materials gathered by this method will be used to create a Material Editor and then can be edited directly on the custom pass.
+        /// </summary>
+        /// <returns>an enumerable of materials to show in the inspector. These materials can be null, the list is cleaned afterwards</returns>
```
iM0ve:
```suggestion
        /// <returns>An enumerable of materials to show in the inspector. These materials can be null, the list is cleaned afterwards</returns>
```

PR #5352, opened 2019-12-16
Fix Redo for Shader Graph

zadi0648:
Fixes high user-pain bug "[ShaderGraph] Cannot redo undone actions when certain nodes are added to a graph"
https://fogbugz.unity3d.com/f/cases/1186225/

Actually fixes a whole lot more than that, from my testing Redo wasn't working >95% of the time.

Was caused by each node clearing graph view selection and then setting itself as selected each undo, unnecessarily, which Graph View would record the selection clearing & adding for each one. This would cause Graph View to lose recorded undos if the graph was even remotely medium sized.

---
**Notice:** 

Bigger & better branch (which fixes a few more fixes) can be found here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5353

Please at least have a review for that PR before merging this one.

---
**Notes for QA:** 

- _Test all the things to Undo -> Redo!_
- See Extended Version PR for list of known bugs that you might run into (both ones that that one fixes and  doesn't).
- Please test combinations of expanding and collapsing nodes, including single node, multiple nodes that are selected, and popup menus. See Zack if confused on all the options.
- I'm not sure which PR you want to test (yet), or if you are okay with testing both. Unless if instructed otherwise, you may want to wait a little before testing this.

---

Yamato: (Select your branch):


zadi0648:
was going to say tests looks good 🙂

TrunkVersusRedoFix

PR #5353, opened 2019-12-16
Fix Redo for Shader Graph [Extended Version] v2

zadi0648:
**Version 2:**

Standing known issues have been fixed! Also this PR has now been changed to target master. 

(btw #5352 has already gone in) 

**Additional Fixes:**
- Expanding and Collapsing multiple nodes (not the previews) simultaneously now guarantees that  they will all be in the same state upon Redo/Undo.
- Fixed a bug where Expanding and Collapsing a node (not the preview) did not properly cause the Shader Graph's title to update with an *.
- Fixed a code smell where Undo was potentially being recorded several times for the same action.
(**Changed:** _The below bugs were marked as known bugs, but are now also fixed_)
- Undoing and Redoing Node Expansion (aka hiding unused ports) won't record the first instance.
- If you press down on the Node Expansion but before releasing the mouse move the cursor off the button, it causes our side to trigger without Graph View's side having any change, resulting in a possibly inconsistent node.

**Additional Changes:**
- Cleared up confusing method names (OnSelection should be reserved for methods that do something when Selection happens)

**Controversial Change?:**
Expanding a node(s)'s port, expanding/collapsing a node(s)'s preview, and opening a node's settings no longer selects the node if the node isn't already selected. This was done because:
- This was actually a Graph View Parity issue: VFX and such don't do this.
- The user experience seems better this way. Sometimes they will want to see settings, and expand/collapse stuff without loosing their current selection.
- Contributed to one of the above fixed bugs / is buggy in general.

---

**Halo Effect:** 1/3
**Technical Risk:** 2/3 

---

**Yamato**:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Ffix-redo-extended/.yamato%252Fupm-ci-shadergraph.yml%2523All_ShaderGraph_trunk/1009590/job


com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -426,13 +431,25 @@ void CollapsePreviews(DropdownMenuAction action)
...
-            graph.owner.RegisterCompleteObjectUndo("Expand Previews");
+            RecordPreviewExpandUndo(true);
+
             foreach (AbstractMaterialNode node in graph.GetNodes<AbstractMaterialNode>())
             {
                 node.previewExpanded = true;
             }
         }
 
+        internal void RecordNodeExpandUndo(bool state)
```
pbbastian: I don't think this and `RecordPreviewExpandUndo` should exist. They're hiding the call to undo registration, which IMO should never be wrapped away, but rather always be there directly with the code modifying the data, because it has such big side effects.

Also, the message is being obscured in a bool parameter (not very intuitive at call site), and they wrap over only a single method. I assume the intention is to de-duplicate the messages? If so, put them into `const string`s instead :)

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -376,17 +376,21 @@ public void RemoveFromGroupNode()
...
             foreach (MaterialNodeView selectedNode in selection.Where(x => x is MaterialNodeView).Select(x => x as MaterialNodeView))
             {
-                if(selectedNode.CanToggleExpanded())
+                if (selectedNode.CanToggleExpanded() && selectedNode.expanded != state)
                     selectedNode.expanded = state;
             }
+
+            // TODO: need to replace Node minimize manipulator in GV (requires trunk change), for now (for consistency's sake)
+            // we record after so that all the nodes are at the same state after the undo.
+            RecordNodeExpandUndo(state);
```
pbbastian: Undo actions must be registered _before_ the action, as the scope of the undo action is anything that happens until the next undo action.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -376,17 +376,21 @@ public void RemoveFromGroupNode()
...
-            graph.owner.RegisterCompleteObjectUndo("Toggle Expansion");
             foreach (MaterialNodeView selectedNode in selection.Where(x => x is MaterialNodeView).Select(x => x as MaterialNodeView))
             {
-                if(selectedNode.CanToggleExpanded())
+                if (selectedNode.CanToggleExpanded() && selectedNode.expanded != state)
                     selectedNode.expanded = state;
             }
+
+            // TODO: need to replace Node minimize manipulator in GV (requires trunk change), for now (for consistency's sake)
+            // we record after so that all the nodes are at the same state after the undo.
```
pbbastian: Can you elaborate on this? I'm not really following

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -423,56 +421,47 @@ void UpdateSettingsExpandedState()
...
-        {
-            m_GraphView.ClearSelection();
-            m_GraphView.AddToSelection(this);
-        }
-
         void SetNodeExpandedStateOnSelection(MouseUpEvent evt)
         {
-            if (!selected)
-                SetSelfSelected();
+            MaterialGraphView graphView = m_GraphView as MaterialGraphView;
```
pbbastian: Let's instead change the type of `m_GraphView` to `MaterialGraphView`, since that will always be the case.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -423,56 +421,47 @@ void UpdateSettingsExpandedState()
...
+
+            // If selected, expand/collapse the other applicable nodes that are also selected
```
pbbastian: This wasn't the behavior before, was it? I find it really weird that clicking the expand/collapse button on 1 node will apply that change to multiple nodes.

zadi0648: It is! It has been since a PR this early summer.
When we first introduced it I too was unsure about it too, I thought it had poor discoverability and might seem like a bug when first discovered. I think it's useful now, imo it should be documented somewhere if its not. But we can discuss it more at a future meeting.

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
zadi0648: Sounds good to me.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
zadi0648: Ah, I see! I was wondering why they weren't in a method or something; I thought it was rough that we have the same RegisterUndo+string combo over and over in the code, for example BlackboardFieldPropertyView has `graph.owner.RegisterCompleteObjectUndo("Change property value");` 24 times 😮. So yes it was about "de-duplication"; I thought either they should be wrapped, we should start varying our undo messages more, or a some const strings.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
zadi0648: I realized something & will fix this

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
alindmanUnity: just a note -- when you have all selected the option does become "Collapse Previews" plural, which while not awesome is at least not nothing. Probably worth changing to either "Collapse Selected Previews" etc or try to have them be two separate options?

zadi0648:
Moved this PR to target master

The known issues should now be fixed 😄, and code review comments addressed

glamberous:
Confirmed the issues listed in the PR have been fixed. 

Did some additional exploratory halo testing and didn't find any issues caused by this PR. Tested Undo/Redo in the following areas:
- Properties
- SubGraphs
- Hooking up Nodes
- Deleting/Creating Nodes

GrantLamb-Unity:
Confirmed the issues listed in the PR have been fixed.

Did some additional exploratory halo testing and didn't find any issues caused by this PR. Tested Undo/Redo in the following areas:

Properties
SubGraphs
Hooking up Nodes
Deleting/Creating Nodes

martint-unity:
Have you run Yamato on this?

com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -423,60 +418,46 @@ void UpdateSettingsExpandedState()
...
+            node.owner.owner.RegisterCompleteObjectUndo(!expanded ? "Expand Nodes" : "Collapse Nodes");
+            expanded = !expanded;
 
-        void SetNodeExpandedStateOnSelection(MouseUpEvent evt)
-        {
-            if (!selected)
-                SetSelfSelected();
-            else
+            // If selected, expand/collapse the other applicable nodes that are also selected
+            if (selected)
```
pbbastian: Hmm, if it's not selected, should it not become the only selected node when clicking on it?

zadi0648: Yeah so that's what the **Controversial Change?:** section is about, which is definitely up to debate. From my experience it can be nice to right click and expand/expand previews without the node selection changing in the process. As mention the alternative is buggy atm.

marctem: I agree with Zack on this on - explicitly selecting it seems to cause more problems than benefits. It's pretty common in other programs as well - even OSX finder does this (they have a "sort-of-selected hilight" when you right-click a file).

PR #5354, opened 2019-12-16
Universal/PPv2 compatibility layer

Chman:
### Purpose of this PR

Added an option to temporarily support PPv2 in URP to ease the porting of LW projects to URP. It's a project-wide setting, meaning you can't use builtin & ppv2 effects at the same time.

When the PPv2 package is installed, you'll find a new option in the URP Asset:

image

That this option is hidden and reverted to "Integrated" when the package is uninstalled.

Note: #5462 has been merged into this PR in preparation for the final merge so it's probably best to wait for it to be merged into `release/2019.3` before reviewing or testing it.


com.unity.render-pipelines.universal/Editor/Overrides/TonemappingEditor.cs
```diff
@@ -16,15 +16,17 @@ public override void OnEnable()
...
+
             PropertyField(m_Mode);
 
             // Display a warning if the user is trying to use a tonemap while rendering in LDR
-            var asset = UniversalRenderPipeline.asset;
-            if (asset != null && !asset.supportsHDR)
-            {
+            if (UniversalRenderPipeline.asset?.supportsHDR == false)
                 EditorGUILayout.HelpBox("Tonemapping should only be used when working in HDR.", MessageType.Warning);
-                return;
```
phi-lira: is this removal intentional?

com.unity.render-pipelines.universal/Editor/Unity.RenderPipelines.Universal.Editor.asmdef
```diff
@@ -17,5 +17,11 @@
     "precompiledReferences": [],
     "autoReferenced": true,
     "defineConstraints": [],
-    "versionDefines": []
-}
\ No newline at end of file
+    "versionDefines": [
```
phi-lira: We should  not add hard dependency, instead we force ppv3 if ppv2 is not present.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
+            public static GUIContent postProcessingFeatureSet = EditorGUIUtility.TrTextContent("Feature Set", "Sets the feature set to use for the render pipeline. Builtin should always be preferred over the other compatibility mode.");
```
phi-lira: What name are we using in docs for PPv3? I'm worried builtin will cause confusion. Maybe we should use `Integrated post-processing stack`?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
...
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
+            public static string postProcessingFeatureSetWarning = "Post-processing V2 support for the Universal Render Pipeline will be deprecated in a near future. We recommend only using this mode for compatibility purpose.";
+            public static string postProcessingFeatureSetError = "The Post-processing V2 package isn't installed, the feature set will fallback on Builtin.";
```
phi-lira: Integrated? PPv3?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -345,16 +351,32 @@ void DrawPostProcessingSettings()
...
-                else if (isHdrOn && m_ColorGradingMode.intValue == (int)ColorGradingMode.HighDynamicRange)
-                    EditorGUILayout.HelpBox(Styles.colorGradingModeSpecInfo, MessageType.Info);
+                EditorGUILayout.PropertyField(m_PostProcessingFeatureSet, Styles.postProcessingFeatureSet);
 
-                EditorGUILayout.DelayedIntField(m_ColorGradingLutSize, Styles.colorGradingLutSize);
-                m_ColorGradingLutSize.intValue = Mathf.Clamp(m_ColorGradingLutSize.intValue, UniversalRenderPipelineAsset.k_MinLutSize, UniversalRenderPipelineAsset.k_MaxLutSize);
-                if (isHdrOn && m_ColorGradingMode.intValue == (int)ColorGradingMode.HighDynamicRange && m_ColorGradingLutSize.intValue < 32)
-                    EditorGUILayout.HelpBox(Styles.colorGradingLutSizeWarning, MessageType.Warning);
+                bool ppv2Enabled = false;
+                if (m_PostProcessingFeatureSet.intValue == (int)PostProcessingFeatureSet.PostProcessingV2)
```
phi-lira: Maybe we should wrap the `POST_PROCESSING_STACK_2_0_0_OR_NEWER` define in this if as well. We should force ppv3/integrated when no ppv2 package is installed instead of having it as hard dependency. 

This is also a good way to know if we can "upgrade" a project straight away to v3.

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
```diff
@@ -12,6 +12,10 @@ internal class Renderer2D : ScriptableRenderer
         FinalBlitPass m_FinalBlitPass;
         PostProcessPass m_FinalPostProcessPass;
 
+#if POST_PROCESSING_STACK_2_0_0_OR_NEWER
+        PostProcessPassCompat m_PostProcessPassCompat;
+#endif
```
phi-lira: Was PPv2 supported in 2D renderer before? I wonder if we should add compatibility for it.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -98,6 +98,12 @@ public enum ColorGradingMode
         HighDynamicRange
     }
 
+    public enum PostProcessingFeatureSet
+    {
+        Builtin,
```
phi-lira: Integrated/v3?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -221,60 +243,106 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             bool requiresFinalPostProcessPass = postProcessEnabled &&
                                      renderingData.cameraData.antialiasing == AntialiasingMode.FastApproximateAntialiasing;
 
-            // if we have additional filters
-            // we need to stay in a RT
-            if (afterRenderExists)
+#if POST_PROCESSING_STACK_2_0_0_OR_NEWER
```
phi-lira: Add a #region here so we can fold out this easily.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -221,60 +243,106 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+                // we need to stay in a RT
+                if (afterRenderExists)
                 {
-                    if (requiresFinalPostProcessPass)
+                    bool willRenderFinalPass = (m_ActiveCameraColorAttachment != RenderTargetHandle.CameraTarget);
+                    // perform post with src / dest the same
+                    if (postProcessEnabled)
                     {
-                        m_PostProcessPass.Setup(cameraTargetDescriptor, m_ActiveCameraColorAttachment, m_AfterPostProcessColor, m_ActiveCameraDepthAttachment, m_ColorGradingLut, true, false);
+                        m_PostProcessPass.Setup(cameraTargetDescriptor, m_ActiveCameraColorAttachment, m_AfterPostProcessColor, m_ActiveCameraDepthAttachment, m_ColorGradingLut, requiresFinalPostProcessPass, !willRenderFinalPass);
```
phi-lira: Seems like some additional changes in PostProcessing not just related to ppv2? Is it fixing additional issue? If so we should make a separate PR to master if we need those as this one will only be available in 2019.3

com.unity.render-pipelines.universal/Editor/Overrides/TonemappingEditor.cs
Chman: Yep, it's at the end of the function so it's redundant :)

com.unity.render-pipelines.universal/Editor/Unity.RenderPipelines.Universal.Editor.asmdef
Chman: I'm not sure I understand, this is not a dependency, it's just a define set if the PPv2 package is installed. The same is done in the Runtime asmdef, I just copied it over to the Editor asmdef because I needed it there as well.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Chman: Yeah I wasn't sure to be honest. This is where I'd like to get some input from doc people :)

com.unity.render-pipelines.universal/Runtime/2D/Renderer2D.cs
Chman: It was, I added it back for the sake of consistency.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
Chman: I can add a region, but as my IDE supports folding out #if/#else I didn't think it would be necessary.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Chman: Integrated works for me. No v3 though, as it would make it confusing (it's not a package anymore, and "v3" in HDRP isn't the same either).

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Chman: I'm still not sure what you mean about having a hard dependency, there isn't any. Do you mean that we should just hide everything in the UI if PPv2 isn't installed?

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
Chman: Hmm right, it's part of a fix that reached master but isn't in 2019.3. I'll remove it.

Chman: Turns out it's not a change, it's the diff tool failing at figuring out which line was actually modified. You can check the source files manually, the code hasn't changed.

Chman:
For the doc I will have to add a section about PPv2 compatibility in the final PR (this one is just a draft). In the meantime, can we check that the 4 messages/tooltips added in `UniversalRenderPipelineAssetEditor.cs` are good?

com.unity.render-pipelines.universal/Editor/Unity.RenderPipelines.Universal.Editor.asmdef
phi-lira: Ah ok. I misunderstood it.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
+            public static GUIContent postProcessingFeatureSet = EditorGUIUtility.TrTextContent("Feature Set", "Sets the feature set to use for the render pipeline. Integrated should always be preferred over the other compatibility mode.");
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
+            public static string postProcessingFeatureSetWarning = "Post-processing V2 support for the Universal Render Pipeline will be deprecated in a near future. We recommend only using this mode for compatibility purpose.";
```
JordanL8: Unity plans to deprecate Post-processing V2 support for the Universal Render Pipeline in the near future. You should only use this mode for compatibility purposes.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
+            public static GUIContent postProcessingFeatureSet = EditorGUIUtility.TrTextContent("Feature Set", "Sets the feature set to use for the render pipeline. Integrated should always be preferred over the other compatibility mode.");
```
JordanL8: Sets the feature set to use for the render pipeline. To future proof your application, use Integrated instead of the comparability mode. Only use compatibility mode if your project still uses the Post-processing V2 package, but be aware that Unity plans to deprecate Post-processing V2 support for the Universal Render Pipeline in the near future.

This is a bit big for a tooltip, but I think its quite crucial information.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
...
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
+            public static string postProcessingFeatureSetWarning = "Post-processing V2 support for the Universal Render Pipeline will be deprecated in a near future. We recommend only using this mode for compatibility purpose.";
+            public static string postProcessingFeatureSetError = "The Post-processing V2 package isn't installed, the feature set will fallback on Integrated.";
```
JordanL8: Your Project does not include the Post-processing V2 package. The Universal Render Pipeline will use its Integrated post-processing solution.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -55,11 +55,15 @@ internal class Styles
...
             public static string colorGradingModeWarning = "HDR rendering is required to use the high dynamic range color grading mode. The low dynamic range will be used instead.";
             public static string colorGradingModeSpecInfo = "The high dynamic range color grading mode works best on platforms that support floating point textures.";
             public static string colorGradingLutSizeWarning = "The minimal recommended LUT size for the high dynamic range color grading mode is 32. Using lower values will potentially result in color banding and posterization effects.";
+            public static string postProcessingGlobalWarning = "The Post-processing Feature Set in the render pipeline asset is set to Post-processing V2. This volume component will not have any effect.";
```
JordanL8: The Post-processing Feature Set in the URP Asset is set to Post-processing V2. This Volume component will not have any effect.

JordanL8:
Tooltips look good to me. Also, I agree that we should use 'integrated' for all the docs and tooltips in the future.

Chman:
Fixed, thanks for the input!

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -98,6 +98,12 @@ public enum ColorGradingMode
         HighDynamicRange
     }
 
+    public enum PostProcessingFeatureSet
```
phi-lira: Either make this internal or add API docs to any new public API.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -583,6 +592,26 @@ public bool useSRPBatcher
             set { m_UseSRPBatcher = value; }
         }
 
+        public PostProcessingFeatureSet postProcessingFeatureSet
```
phi-lira: Either make this internal or add API docs to any new public API.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -244,60 +255,111 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+                // we need to stay in a RT
+                if (afterRenderExists)
                 {
-                    if (renderingData.cameraData.captureActions != null)
+                    // perform post with src / dest the same
+                    if (applyPostProcessing)
                     {
-                        m_CapturePass.Setup(m_ActiveCameraColorAttachment);
-                        EnqueuePass(m_CapturePass);
+                        m_PostProcessPassCompat.Setup(cameraTargetDescriptor, m_ActiveCameraColorAttachment, m_ActiveCameraColorAttachment);
```
phi-lira: If there's exactly one effect in the stack this will fail iirc.

Chman: This will work fine here, it's handled by PPv2. But saying this actually made me realize this PR was missing the support for opaque-only PPv2 effect, I added it back (and yeah, if there's only one opaque-only effect I believe it will fail as it did on LWRP)...

PR #5369, opened 2019-12-17
Fixed an issue where frame count was reset incorrectly every frame for the game view, resulting in temporal processes breaking.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue where frame count was reset incorrectly every frame for the game view, resulting in temporal processes breaking.

Repro steps for the issue:
- Create a new scene
- If created with the proper default scene it should already have volumetric fog in it. If not add a fog component to setup volumetric fog.
- Create a spot and enable volumetric on it.
- Have only the game view and enter playmode. Notice that reprojection works (noise in the volumetric should go away)
- Open both Scene View and Game view at the same time. Enter playmode and notice that reprojection does not work in the game view (volumetric fog stays noisy)

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1162,13 +1162,11 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
                     HDCamera.CleanUnused();
 
                     if (newTime > m_Time)
+                    {
```
sebastienlagarde: be careful, I remember that the reset to 0 was a requirement. I guess it was when you are changing from one scene to another and was reported by Robert. @EvgeniiG do you remind?

EvgeniiG: Their issue was using realtimeSinceStartup.
From the other PR:
 "If we switch to other scene Time.realtimeSinceStartup is reset, so we need to SRP.Render() can be called several times per frame."
Since we no longer use Time.realtimeSinceStartup, this may be safe. Needs to be thoroughly tested.

sebastienlagarde: Found the slack discussion: https://unity.slack.com/archives/C6Y79CZM0/p1566575224497500

sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp be sure we covert the case describe in this slack discussion: https://unity.slack.com/archives/C6Y79CZM0/p1566575224497500

PR #5372, opened 2019-12-17
Fix issue with Motion Blur sample weighting 

FrancescoC-unity:
There was a typo that caused the whole weighting that should reject samples. 

This fixes issues that the demo team had and should generally lead to higher quality to everyone as a whole big class of artefacts will be gone. 

For QA: 
 - A way to repro the problem is a bit tricky, but the demo team gave me a repro project. Whoever picks this PR up, please contact me and I'll send you the project. 

- I'd also like a small pass in general to see if it does something unwanted elsewhere (it really shouldn't, but hey, better safe than sorry)


FrancescoC-unity:
> PR fix the issue with motion blur default and "reasonable" values
> Could still repro some artifacts with higher up values (intensity ~x10)
> https://gyazo.com/19eebf9b441dc625d42f613dbeab137d

Yeah, with unreasonably large velocities stuff is bound to break :-) At that point users should clamp them anyway (via the volume settings)

PR #5374, opened 2019-12-18
Added support for physical parameters in the scene view camera

Chman:
### Purpose of this PR

Working with proper light units and using the physical camera to control exposure is currently painful in the scene view, as physical parameters aren't editable in this view. The scene view camera now replicates the physical parameters from the camera tagged as "MainCamera", the same way it does for volume triggers & layer masks.

This PR also removes some GC pressure in the scene view (it used to call `GetComponent<>()` on every frame).


TomasKiniulis:
Did a quick check through multiple scene views, no issues found

NmduMvvJq2

PR #5375, opened 2019-12-18
Fix issues of decals duplicating and baking into the scene when editing

FrancescoC-unity:
This fixes: https://fogbugz.unity3d.com/f/cases/1204967/ 

The issue was that under some editor operation both OnValidate and OnEnable could be called and both of those functions could add the decal to the list. This fixes should avoid the double add.

There was a separate issue, that didn't remove a decal if the material would be set to null.


iM0ve:
Before:
Different undo opperations or even gameobject name change would sometimes cause Decal duplication. In the video you can see an example of moving undo.
bug

This PR fixes the issue. I have tested:
* Moving decal + undo
* Removing the material from Decal Projector
* Toggled visibility of the Decal on/off + undo
* Renamed the gameobject + undo
* Tested the combination of all above. Each of them could previously cause the issue alone.
* Real usage test. Opened Fontainebleau 19.3.0f3. Tried to reproduce the bug. Checked for unwanted graphical changes.
* Real usage test. Opened Amalienborg 20.1.0a15. Tried to reproduce the bug. Checked for unwanted graphical changes.

PR #5377, opened 2019-12-18
Fix ray traced screen space shadows

anisunity:
Change the ray tracing condition to change from screen space shadow index to screen space shadow slot index.
Allocate two channels for ray traced area light shadows
Fix an issue with ray traced area light shadows index (uses shadow even if not required)
Added a shadow history validation based on the light's instanceID
Store the filtering sample count in a different texture for shadows and use the slot for the analytic accumulation for area lights
and various other fixes to make ray traced shadow work properly


anisunity:
This PR also fixes this issues:
https://fogbugz.unity3d.com/f/cases/1208489/
https://fogbugz.unity3d.com/f/cases/1208492/
https://fogbugz.unity3d.com/f/cases/1207501/
https://fogbugz.unity3d.com/f/cases/1207942/

PR #5379, opened 2019-12-18
Finally activate the GC.Alloc check in HDRP graphics tests (for most scenes)

alelievr:
### Purpose of this PR
Remove most of the remaining GC alloc calls in HDRP and enable the GC alloc test in our graphic tests.

The GC alloc check in our graphics tests was enabled with a threshold of 2 GC calls allowed, because currently calling Camera.Render allocate memory ... This issue have been reported here: https://fogbugz.unity3d.com/f/cases/1206364/ and the threshold should be set to 0 once the fix lands.

Note that there is now a `Check Memory Allocation` boolean in the HDRP test settings to toggle the allocation test per-scene:
image
For every scene that contains a planar / reflection probe, this boolean is disabled because there are still a ton of allocation when you have one of them.


com.unity.testing.hdrp/TestRunner/HDRP_TestSettings.cs
```diff
@@ -20,6 +20,7 @@ public class HDRP_TestSettings : GraphicsTestSettings
 	public int captureFramerate = 0;
 	public int waitFrames = 0;
     public bool xrCompatible = true;
+    public bool checkMemoryAllocation = true;
```
fabien-unity: Can you move this either up or down ?
Right now, it is in between the 2 XR settings.

alelievr: ha yes, sorry

alelievr: done

TestProjects/UniversalGraphicsTest/Assets/Test/Runtime/UniversalGraphicsTests.cs
```diff
@@ -85,7 +85,7 @@ public IEnumerator Run(GraphicsTestCase testCase)
         var mainCamera = GameObject.FindGameObjectWithTag("MainCamera").GetComponent<Camera>();
         try
         {
-            ImageAssert.AllocatesMemory(mainCamera, 512, 512); // 512 used for height and width to render
+            ImageAssert.AllocatesMemory(mainCamera, settings?.ImageComparisonSettings); // 512 used for height and width to render
```
JulienIgnace-Unity: Can probably remove the comment :)

com.unity.testing.hdrp/TestRunner/HDRP_GraphicTestRunner.cs
```diff
@@ -65,22 +65,23 @@ public IEnumerator Run(GraphicsTestCase testCase)
...
-            bool allocatesMemory = false;
-            try
+            if (settings.checkMemoryAllocation)
             {
-                ImageAssert.AllocatesMemory(camera, 512, 512); // 512 used for height and width to render
+                // Does it allocate memory when it renders what's on camera?
+                bool allocatesMemory = false;
+                try
+                {
+                    ImageAssert.AllocatesMemory(camera, settings?.ImageComparisonSettings); // 512 used for height and width to render
```
JulienIgnace-Unity: same comment to remove here.

com.unity.testframework.graphics/Runtime/ImageAssert.cs
```diff
@@ -205,20 +205,31 @@ public static void AreEqual(Texture2D expected, Texture2D actual, ImageCompariso
...
             if (camera == null)
                 throw new ArgumentNullException(nameof(camera));
+            
+            if (settings == null)
+                settings = new ImageComparisonSettings();
+
+            int width = settings.TargetWidth;
+            int height = settings.TargetHeight;
+
+            var defaultFormat = (settings.UseHDR) ? SystemInfo.GetGraphicsFormat(DefaultFormat.HDR) : SystemInfo.GetGraphicsFormat(DefaultFormat.LDR);
```
JulienIgnace-Unity: We can compare HDR images now? Does the comparison function actually supports HDR correctly?

alelievr: I copied this from the actual ImageAssert that we use so ib hope so ^^

sebastienlagarde:
Force merging this PR as people in SDETs are in vacation and we need to have this PR in to detect futher GC alloc issue before there is more PR merge.

PR #5383, opened 2019-12-19
HDRP Volume System update refactor [skip CI]

JulienIgnace-Unity:
### Purpose of this PR
Currently HDRP is using the global Volume System stack for all its cameras. It means that each camera updating the volume system will overwrite the result of the previous camera. To go around that for rendering purpose, Volume System update is done late at the beginning of ExecuteRenderRequest. Because of that, any piece of code accessing the volume system in the first update phase (where we go through all camera to setup culling and reflection probes) will get the wrong result.
To fix that, now each camera has its own volume stack and the global one should not be used anymore by the HDRP render loop.

This fixes the shadow resolution issue raised here: https://fogbugz.unity3d.com/f/cases/1198559/

---
### Testing status

- [ ] Built a player


Chman:
Aside from the commented part in `CleanupCustomPostProcess` (I'll talk to @alelievr) it looks alright to me, we're gonna need a QA pass to make sure nothing broke as it's touching quite a lot of code.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1803,6 +1795,14 @@ AOVRequestData aovRequest
             // Updates RTHandle
             hdCamera.BeginRender(cmd);
 
+            if (m_RayTracingSupported)
+            {
+                // This call need to happen once per camera
+                // TODO: This can be wasteful for "compatible" cameras.
+                // We need to determine the minimum set of feature used by all the camera and build the minimum number of acceleration structures.
+                BuildRayTracingAccelerationStructure(hdCamera);
```
anisunity: @sebastienlagarde as discussed with @JulienIgnace-Unity, In the current state, having one build per camera is the only way this can work even if both cameras build exactly the same RAS. Code needs to be produced to avoid the build when RAS can be shared

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1576,7 +1565,7 @@ private int GetResolutionFromSettings(ShadowMapType shadowMapType, HDShadowInitP
             }
         }
 
-        internal void ReserveShadowMap(Camera camera, HDShadowManager shadowManager, HDShadowInitParameters initParameters, Rect screenRect)
+        internal void ReserveShadowMap(Camera camera, HDShadowManager shadowManager, HDShadowSettings shadowSetings, HDShadowInitParameters initParameters, Rect screenRect)
```
FrancescoC-unity: Typo: shadowSetings

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: so this change is ok?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1849,19 +1849,15 @@ AOVRequestData aovRequest
                 }
             }
 
-            using (new ProfilingScope(cmd, ProfilingSampler.Get(HDProfileId.VolumeUpdate)))
+            using (new ProfilingScope(null, ProfilingSampler.Get(HDProfileId.CustomPassVolumeUpdate)))
             {
-                VolumeManager.instance.Update(hdCamera.volumeAnchor, hdCamera.volumeLayerMask);
-                // We need to cache the max shadow distance to make sure the culling uses the correct value. Refer to the comment
-                // above m_ShadowMaxDistance in hdCamera.
```
sebastienlagarde: expected to have remove 
                hdCamera.shadowMaxDistance

sebastienlagarde: ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2827,6 +2824,7 @@ struct DepthPrepassParameters
             public bool                shouldRenderMotionVectorAfterGBuffer;
             public RendererListDesc    rayTracingOpaqueRLDesc;
             public RendererListDesc    rayTracingTransparentRLDesc;
+            public bool                renderRayTracingPrepass;
```
sebastienlagarde: expected?

sebastienlagarde:
Note: yamato is green on this PR (failure aren't related to this PR)

PR #5385, opened 2019-12-19
Lighting explorer fixes

chikuba:
### Purpose of this PR
Fixing more issues with the Lighting Explorer:
* Fixed multi-edit and undo on Light and Reflection probes
* Disables the prefab field since it's not editable
* Fixed null-ref issue on Reflection probes
* Fixed HDLightType for Lights

### Testing status
I have extensively tested and made sure all fields work (sorting, multi-edit, undo etc) 
I would love to have someone else make sure all this works:
* Sorting of each column for Lights, Volumes and Reflection probes etc.
* Select a few row, change a value, and then hit undo (this should multi-edit, and then revert back properly) 

### When backporting
Please be aware that the name change "camera" -> "cameraSettings" might not work on older versions of Unity. This was a regression in newer versions, and might need to remain how it was. Not sure when this name change happened so this will need to be tested by whoever is doing the backport :) The rest of the changes should be simple to merge in.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
@@ -192,6 +192,7 @@ namespace UnityEditor.Rendering.HighDefinition
 
                     HDLightType lightType = lightData.type;
 
+                    EditorGUI.BeginProperty(r, GUIContent.none, prop);
                     EditorGUI.BeginChangeCheck();
                     lightType = (HDLightType)EditorGUI.IntPopup(r, (int)lightType, HDStyles.LightTypeTitles, HDStyles.LightTypeValues);
 
@@ -200,6 +201,7 @@ namespace UnityEditor.Rendering.HighDefinition
                         Undo.RecordObjects(new Object[] { prop.serializedObject.targetObject, lightData }, "Changed light type");
                         lightData.type = lightType;
                     }
+                    EditorGUI.EndProperty();
                 }, (lprop, rprop) =>
                 {
                     TryGetAdditionalLightData(lprop, out var lLightData);
--- a/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
+++ b/com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
@@ -211,10 +211,7 @@ namespace UnityEditor.Rendering.HighDefinition
                     return ((int)lLightData.type).CompareTo((int)rLightData.type);
                 }, (target, source) => 
                 {
-                    TryGetAdditionalLightData(target, out var tLightData);
-                    TryGetAdditionalLightData(source, out var sLightData);
-
-                    if (!tLightData || !sLightData)
+                    if (!TryGetAdditionalLightData(target, out var tLightData) || !TryGetAdditionalLightData(source, out var sLightData))
                         return; 
 
                     Undo.RecordObjects(new Object[] { target.serializedObject.targetObject, tLightData }, "Changed light type");
@@ -265,10 +262,7 @@ namespace UnityEditor.Rendering.HighDefinition
                     return ((float)lLightData.intensity).CompareTo((float)rLightData.intensity);
                 }, (target, source) => 
                 {
-                    TryGetAdditionalLightData(target, out var tLightData);
-                    TryGetAdditionalLightData(source, out var sLightData);
-
-                    if (!tLightData || !sLightData)
+                    if (!TryGetAdditionalLightData(target, out var tLightData) || !TryGetAdditionalLightData(source, out var sLightData))
                         return; 
 
                     Undo.RecordObject(tLightData, "Changed light intensity");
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -726,7 +889,7 @@ protected virtual LightingExplorerTableColumn[] GetHDReflectionProbeColumns()
                     }
 
                     reflectionData.Update();
-                    EditorGUI.PropertyField(r, reflectionData.FindProperty("m_ProbeSettings.camera.frustum.farClipPlane"), GUIContent.none);
+                    EditorGUI.PropertyField(r, reflectionData.FindProperty("m_ProbeSettings.cameraSettings.frustum.farClipPlane"), GUIContent.none);
```
chikuba: This was the issue with the null-ref

sebastienlagarde:
can you add a changelog please? thanks

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -217,6 +253,22 @@ protected virtual LightingExplorerTableColumn[] GetHDLightColumns()
...
+                    if (IsNullComparison(lLightData, rLightData, out var order))
+                        return order; 
+
+                    return ((float)lLightData.intensity).CompareTo((float)rLightData.intensity);
+                }, (target, source) => 
+                {
+                    if (!TryGetAdditionalLightData(target, out var tLightData) || !TryGetAdditionalLightData(source, out var sLightData))
+                        return; 
+
+                    Undo.RecordObject(tLightData, "Changed light intensity");
```
RSlysz: Can you test it with the debug inspector? I think that the intensity also modify the built-in light intensity. So you must certainly register both component in the Undo

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -244,6 +296,13 @@ protected virtual LightingExplorerTableColumn[] GetHDLightColumns()
                         return order; 
 
                     return ((int)lLightData.lightUnit).CompareTo((int)rLightData.lightUnit);
+                }, (target, source) => 
+                {
+                    if (!TryGetAdditionalLightData(target, out var tLightData) || !TryGetAdditionalLightData(source, out var sLightData))
+                        return; 
+
+                    Undo.RecordObject(tLightData, "Changed light unit");
```
RSlysz: Same here: changing the light intensity unit change the intensity value. So check if the built-in light intensity is changed and if its the case, also register it in the Undo.

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
RSlysz: Do you know if there is a way to make the renaming more safe regarding serialization path, in general?

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
```diff
@@ -736,7 +899,15 @@ protected virtual LightingExplorerTableColumn[] GetHDReflectionProbeColumns()
                     if (IsNullComparison(lReflectionData, rReflectionData, out var order))
                         return order; 
 
-                    return lReflectionData.FindProperty("m_ProbeSettings.camera.frustum.farClipPlane").floatValue.CompareTo(rReflectionData.FindProperty("m_ProbeSettings.camera.frustum.farClipPlane").floatValue);
+                    return lReflectionData.FindProperty("m_ProbeSettings.cameraSettings.frustum.farClipPlane").floatValue.CompareTo(rReflectionData.FindProperty("m_ProbeSettings.camera.frustum.farClipPlane").floatValue);
```
RSlysz: If I follow well, this should be "m_ProbeSettings.cameraSettings.frustum.farClipPlane" for the rReflectionData too

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
chikuba: not really no

com.unity.render-pipelines.high-definition/Editor/Lighting/HDLightExplorerExtension.cs
chikuba: oh thanks

laurenth-unity:
Light type works as expected as well as multi selection and undo (on the tested values).
I spotted some issues that could be fixed in another PR :

- The fact that there are 2 columns called Contact shadows and 2 columns called Shadow resolution makes it impossible to hide them all from the UI. If I do right click on the header and disable shadow resolution for instance, only one of the columns disappears.
- We could rename one of the column "custom contact shadow" and "custom shadow resolution".

- Also there is now inconsitency between Color temperature and scalability settings. When color temperature is disabled the value is hidden, whereas for scalability settings when custom is not selected the custom value is disabled and not hideen. 
- I would suggest we also disable the color temperature instead of hiding.

sebastienlagarde:
@iM0ve  / @laurenth-unity I will merge this PR, could you log into a single fogbugzz or somewhere else for tracking the remaining issue? Thanks!

RSlysz:
@iM0ve It is strange but this is standard behavior amongst all Unity inspector using IMGUI to not update always. They refresh while the mouse over them.

Hopefully, one day we will be able to write this in UIElement with Updatable callbacks.

PR #5389, opened 2019-12-19
Make motion vectors work properly in MSAA

FrancescoC-unity:
They were in a broken state before: 
 - Camera motion vectors where never rendered in the MSAA target
 - Object motion vectors where rendered in the MSAA target
 - MSAA target was never resolved. 

I originally wanted to _not_ have them MSAA at all, but since we have a partial depth prepass (completed by MV pass) and we need to populate an MSAA Depth, so MV target needs to be MSAA as well... 
Of course this is suboptimal for various reasons: 

 - We render MV and Normals in MSAA, while it really shouldn't be needed for what we have now. 
 - We do a dumb resolve for both normals and MV, this is a cost that shouldn't be there. 

However, I guess re-render the geometry twice (i.e. forcing full depth pre pass when MSAA) would probably be as bad or worse in other cases, that's why for now I went with this solution.

Note that the resolve of MV I don't think can be well defined, for now I pick the closest sample to camera as it is the most likely to be relevant.


FrancescoC-unity:
They were both calculated, is just that one was _MSAA_ (the per object) one was not (the per camera), this would have resulted in one of the two being used depending on how the user code requested the MV target

PR #5390, opened 2019-12-19
Fixed cleanup of custom post process

alelievr:
### Purpose of this PR
Fixed a bug where the custom post process Cleanup only called on volume used for interpolation of the main volume stack.

This PR resolves this issues and the `Cleanup` function is called on all custom post processes instances (default ones + interpolated + custom stack).

For QA: this can be tested by ensuring that the number of call to Setup is matching the number of Cleanup (modulo the current running post processes). Can be test on different scenario with domain reload: play mode, script reload, etc.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/CustomPostProcessing/CustomPostProcessVolumeComponent.cs
```diff
@@ -33,6 +33,13 @@ public abstract class CustomPostProcessVolumeComponent : VolumeComponent
         /// </summary>
         public virtual void Cleanup() {}
 
+        /// <inheritdoc />
```
Chman: Unfortunately this tag isn't supported, so it won't pass the doc validation process...

alelievr: yep, fixed

PR #5391, opened 2019-12-19
Avoid UINT gathers in AO code

FrancescoC-unity:
Gather4 on uint texture is effectively working on most GPUs, but is not guaranteed by standard and is producing D3D warnings.

This goes around the problem. Note that because of the asfloat(asuint()) shenanigans, depth now occupies 23 bits instead of 24. This should still be more than enough since it is used only for bilateral weights. 

Adding QA to check to see if some regressions are caused by this. 

Note to QA: Please check across all vendors GPUs (Intel, NVIDIA and AMD)


FrancescoC-unity:
Apparently the float sampler might screw things up a bit when reading. Need to check with vendors.

FrancescoC-unity:
I remember that specific card having problem with AO without this change, could you verify please that it is this PR that makes it not working? 

Also, before going forward with the PR, is really important to have the Intel GPU tested.

PR #5392, opened 2019-12-19
Hdrp/fix material baked emission

RSlysz:
### Purpose of this PR
Hide non supported option for emission in Materials

image

---
### Testing status

---
### Comments to reviewers
Unfortunately, we need to use Reflection once more.


sebastienlagarde:
Question: Does this take into account the compatibility mode with Enlighten in case we come from 19.2 ? 
Otherwise it will be a 2020.1 only PR

For QA: Think to test with our various shader (layered, lit, tessellation, unlit and with shader graph)

laurenth-unity:
Tested this today :
- works on lit shader
- I see nothing on layered lit, was there ever something actually ?
- works on unlit 
- doesn't work on shader graph materials, they still show the old UI

Ideally i't be great if we could detect if enlighten is supported or not so we can backport this to 19.3 because it'll be weird if enlighten is hidden but we leave realtime emission options

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
```diff
@@ -145,12 +159,46 @@ void DrawEmissionGUI()
...
+        }
+
+
+        public bool BakedEmissionEnabledProperty()
+        {
+            Material[] materials = Array.ConvertAll(materialEditor.targets, (UnityEngine.Object o) => { return (Material)o; });
+
+            var settings = GetLightingSettingsOrDefaultsFallback();
+
+            MaterialGlobalIlluminationFlags defaultEnabled = settings.realtimeGI ? MaterialGlobalIlluminationFlags.RealtimeEmissive
```
alelievr: Is defaultEnabled  used somewhere else ?

RSlysz:
@JMargevics 
Ok so the initial request was to simplify emission mode into only the supported "Baked Emission"
Before this PR:
image
After:
image
(only the baked mode is supported amongst emission mode in HDRP)

AXF and LayeredLit do not have those option currently (it is a feature request?).
Though LayeredLitTesselation have it and it was changed accordingly.

> Material assigned to static objects still gets baked. Is this intended? 
For this I am not sure. The scope of this PR is only to hide non usable mode. If this is an issue, please fill a FB.

For all ShaderGraph based, I'll take a look.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/EmissionUIBlock.cs
RSlysz: Seams to be an unused remains of when I copy pasted a part of this functionality in built-in before tweeking it. Need to be removed

PR #5396, opened 2019-12-20
Support for alpha channel in some post-processing passes (DoF, TAA, Uber)

pmavridis:
This PR adds support for alpha channel in some post processing passes. Depending on the pass, the alpha channel is either propagated (pass-through) to the next pass, which is the case in the Uber shader for example, or it is processed (for example in DoF and TAA). 

I have tested this on the compositor prototype but I think QA should also have a careful look.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -380,15 +389,6 @@ void PoolSource(ref RTHandle src, RTHandle dst)
 
             using (new ProfilingSample(cmd, "Post-processing", CustomSamplerId.PostProcessing.GetSampler()))
             {
-                // Save the alpha and apply it back into the final pass if working in fp16
-                if (m_KeepAlpha)
-                {
-                    using (new ProfilingSample(cmd, "Alpha Copy", CustomSamplerId.AlphaCopy.GetSampler()))
```
sebastienlagarde: Let's discuss but we should keep this initial behavior. Maybe it is the case already, haven't read all the PR

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/DepthOfFieldGather.compute
```diff
@@ -118,32 +126,49 @@ void MAIN(uint3 dispatchThreadId : SV_DispatchThreadID)
         sampTap = sampTap * sampDist * clipping + uv;
         sampTap = ClampAndScaleUV(sampTap, _TexelSize.zw, texelsToClamp);
 
-        float3 sampColor = SAMPLE_TEXTURE2D_X_LOD(_InputTexture, SamplerTap, sampTap, mip).xyz;
+        CTYPE sampColor = SAMPLE_CTYPE_LOD(_InputTexture, SamplerTap, sampTap, mip);
```
sebastienlagarde: Maybe keep the name _X_ in the macro so we know it is VR compatible

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/PostProcessDefines.hlsl
```diff
@@ -0,0 +1,48 @@
+#ifndef UNITY_PP_DEFINES_INCLUDED
+#define UNITY_PP_DEFINES_INCLUDED
+
+#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl"
+#include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Filtering.hlsl"
+
+#if !defined(ENABLE_ALPHA)
+    #define CTYPE float3
+
+    //Note: we use functions instead of defines for type safaty. The compiler should inline everything.
```
sebastienlagarde: I think it is overkill to redifine all the type only to change the extension from .xyz; to .xyzw;

I propose that you add a MACRO only for the extension
i.e
float3 tapColor = LOAD_TEXTURE2D_X(_InputTexture, posInputs.positionSS).xyz;

=>
CTYPE tapColor = LOAD_TEXTURE2D_X(_InputTexture, posInputs.positionSS).CTYPE_SWIZZLE;

with CTYPE_SWIZZLE (or any good name)
#define CTYPE_SWIZZLE xyz
#else
CTYPE_SWIZZLE xyzw

sebastienlagarde: and it will work with the function later:
   return SampleTexture2DBicubic(TEXTURE2D_X_ARGS(_InputTexture, s_linear_clamp_sampler),
                                    UV * _RTHandleScale.xy,
                                    bicubicWnd,
                                    (1.0f - 0.5f * _ScreenSize.zw) * _RTHandleScale.xy,
                                    unity_StereoEyeIndex);

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/RTUpscale.hlsl
```diff
@@ -112,4 +124,24 @@ float3 Lanczos(TEXTURE2D_X(_InputTexture), float2 inUV)
     accumulation += float4(Bilinear(_InputTexture, float2(UV3.x, UV0.y)), 1) * weight5.x * weight23.y;
 
     return accumulation.xyz /= accumulation.w;
+#else
+    CTYPE color_accumulation = 0;
+    float weight_accumulation = 0;
+
+    WeightedAcc(Bilinear(_InputTexture, float2(UV_2.x, UV0.y)), weight0.x * weight23.y, color_accumulation, weight_accumulation);
```
sebastienlagarde: Please, add comment about what is happening here :)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/PostProcessDefines.hlsl
Chman: 👍

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
pmavridis: Oh yes, you are right, I must revert this change. Initially I left the initial behavior in, but then decided to remove it to clean-up/simplify the code. But now I see that removing it also causes a test to fail, so I will revert back (and also merge the last changes).

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/PostProcessDefines.hlsl
pmavridis: Indeed!

sebastienlagarde:
Please merge latest HDRP/staging then launch Yamato, should be good (then we wait QApass :) )

PR #5406, opened 2019-12-23
Universal/public renderpass

yugo-unity:
### Purpose of this PR
Customize FowardRenderer(or any extended ScriptableRenderer) to optimize for each projects.

---
### Comments to reviewers
I'd glad if you give me a information, if there are any reasons to be "internal".


phi-lira:
We are not ready to make these classes public yet.

kaychang-unity:
Is there an ETA to make these classes public? This means that currently, users cannot copy-paste the code of ForwardRenderer.cs and make their own renderer code without importing the package locally and modifying the original source code.

yugo-unity:
>We are not ready to make these classes public yet.
I got it. Thanks for your reply, Felipe. And thanks for help, Kay!

I want to know schedule to make public like Kay said. And I would be glad if you tell me the reason for internal yet.
We are really looking forward to the potential of Universal RP by custormized ForwardRenderer or any extended ScriptableRenderer. (Actually this is necessary for Unity-chan Toon Shader).

PR #4940, opened 2019-10-31
[Skip CI] Shader Graph Targets

Kink3d:
### Purpose of this PR
This PR is a near total overhaul of the API for shader generation. Per-pipeline, per-master SubShaders are replaced with static definitions for SubShader, Pass, Struct, RenderState etc. All stateful generation switch logic inside SubShaders is now based on static definitions of conditions on active Fields. Generation logic is handled by a single, internal, Generator class instance. This work is the next step on the path to Stack masters.

Changes are too vast to explain in detail here. Join #project-shadergraphtargets for discussion/progress. See pinned items in that channel for full documentation of the changes.

---
### Testing status

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally
- [x] Checked new UI names with UX convention
- [x] Tested UI multi-edition + Undo/Redo
- [x] C# and shader warnings (supress shader cache to see them)

- Other: 

**Automated Tests**: None yet...

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?: N/A

---
### Comments to reviewers
Join #project-shadergraphtargets

PR #5407, opened 2019-12-23
Universal/deferred

kaychang-unity:
---
### Purpose of this PR
This adds deferred shading support for UniversalRP.

---
### Testing status

- [ X ] Built a player

---
### Comments to reviewers
Opening this PR now to get feedback.


com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalPBRSubShader.cs
```diff
@@ -191,6 +191,100 @@ class UniversalPBRSubShader : IPBRSubShader
                 "multi_compile_instancing",
             },
         };
+
+        ShaderPass m_GBufferPass = new ShaderPass
```
phi-lira: ShaderGraph work will conflict with https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4940

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            m_DepthCopyTexture.Init("_CameraDepthCopyTexture");
+            m_OpaqueColor.Init("_CameraOpaqueTexture");
+            m_AfterPostProcessColor.Init("_AfterPostProcessTexture");
+            m_ColorGradingLut.Init("_InternalGradingLut");
+            m_DepthInfoTexture.Init("_DepthInfoTexture");
+            m_TileDepthInfoTexture.Init("_TileDepthInfoTexture");
+
+            supportedRenderingFeatures = new RenderingFeatures()
+            {
+                cameraStacking = true,
```
phi-lira: Is camera stacking already working in deferred? maybe safer to disable it for now. I can look at it, I think we need to solve how to stack deferred + other cameras, maybe we only allow to stack deferred with forward? but not deferred + deferred?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+        {
+            Camera camera = renderingData.cameraData.camera;
+            ref CameraData cameraData = ref renderingData.cameraData;
+            RenderTextureDescriptor cameraTargetDescriptor = renderingData.cameraData.cameraTargetDescriptor;
+
+            // Depth prepass is generated in the following cases:
+            // - Scene view camera always requires a depth texture. We do a depth pre-pass to simplify it and it shouldn't matter much for editor.
+            // - If game or offscreen camera requires it we check if we can copy the depth from the rendering opaques pass and use that instead.
+            bool requiresDepthPrepass = cameraData.isSceneViewCamera;
+            requiresDepthPrepass |= (cameraData.requiresDepthTexture && !CanCopyDepth(ref renderingData.cameraData)) || m_DeferredLights.tiledDeferredShading;
```
phi-lira: in deferred we always have a depth buffer this logic should go away no?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            // TODO: There's an issue in multiview and depth copy pass. Atm forcing a depth prepass on XR until we have a proper fix.
+            if (cameraData.isStereoEnabled && cameraData.requiresDepthTexture)
+                requiresDepthPrepass = true;
+
+            // Special path for depth only offscreen cameras. Only write opaques + transparents.
+            bool isOffscreenDepthTexture = cameraData.targetTexture != null && cameraData.targetTexture.format == RenderTextureFormat.Depth;
```
phi-lira: TODO: We should probably force offscreen depth only cameras to use forward renderer and avoid this condition here.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+
+            // The copying of depth should normally happen after rendering opaques only.
+            // But if we only require it for post processing or the scene camera then we do it after rendering transparent objects
+            m_CopyDepthPass.renderPassEvent = (!cameraData.requiresDepthTexture && (applyPostProcessing || cameraData.isSceneViewCamera)) ? RenderPassEvent.AfterRenderingTransparents : RenderPassEvent.AfterRenderingOpaques;
+
+            //bool createColorTexture = RequiresIntermediateColorTexture(ref renderingData, cameraTargetDescriptor) || camera.forceIntoRenderTexture;
+            bool createColorTexture = true; // we always create a new colorTexture in this pass: it is used as the 3rd GBuffer slice used in the GBuffer pass, and as lightingBuffer RenderTarget in the Deferred pass
+
+            // If camera requires depth and there's no depth pre-pass we create a depth texture that can be read later by effect requiring it.
+            bool createDepthTexture = cameraData.requiresDepthTexture && !requiresDepthPrepass;
```
phi-lira: should always be true?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            createDepthTexture |= (renderingData.cameraData.renderType == CameraRenderType.Base && !renderingData.resolveFinalTarget);
+
+            // Configure all settings require to start a new camera stack (base camera only)
+            if (cameraData.renderType == CameraRenderType.Base)
+            {
+                m_ActiveCameraColorAttachment = (createColorTexture) ? m_CameraColorAttachment : RenderTargetHandle.CameraTarget;
+                m_ActiveCameraDepthAttachment = (createDepthTexture) ? m_CameraDepthAttachment : RenderTargetHandle.CameraTarget;
+
+                bool intermediateRenderTexture = createColorTexture || createDepthTexture;
```
phi-lira: Note, unless we have ontile deferred we should always have an intermediate render texture?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            }
+            bool hasAfterRendering = activeRenderPassQueue.Find(x => x.renderPassEvent == RenderPassEvent.AfterRendering) != null;
+
+            if (mainLightShadows)
+                EnqueuePass(m_MainLightShadowCasterPass);
+
+            if (additionalLightShadows)
+                EnqueuePass(m_AdditionalLightsShadowCasterPass);
+
+            if (requiresDepthPrepass)
```
phi-lira: Do we actually benefit from a depth prepass in deferred in any case? (considering we had a proper depth prepass)

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+                m_SceneViewDepthCopyPass.Setup(m_DepthTexture);
+                EnqueuePass(m_SceneViewDepthCopyPass);
+            }
+#endif
+        }
+
+        /// <inheritdoc />
+        public override void SetupLights(ScriptableRenderContext context, ref RenderingData renderingData)
+        {
+            m_ForwardLights.Setup(context, ref renderingData);
```
phi-lira: we should probably move some code from forward lights to helper functions somewhere, maybe RenderingUtils

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+
+        /// <inheritdoc />
+        public override void SetupCullingParameters(ref ScriptableCullingParameters cullingParameters,
+            ref CameraData cameraData)
+        {
+            // TODO: PerObjectCulling also affect reflection probes. Enabling it for now.
+            // if (asset.additionalLightsRenderingMode == LightRenderingMode.Disabled ||
+            //     asset.maxAdditionalLightsCount == 0)
+            // {
+            //     cullingParameters.cullingOptions |= CullingOptions.DisablePerObjectCulling;
```
phi-lira: we need to disable per-object light culling

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            if (m_ActiveCameraDepthAttachment != RenderTargetHandle.CameraTarget)
+            {
+                cmd.ReleaseTemporaryRT(m_ActiveCameraDepthAttachment.id);
+                m_ActiveCameraDepthAttachment = RenderTargetHandle.CameraTarget;
+            }
+        }
+
+        void EnqueueDeferred(ref RenderingData renderingData, bool requiresDepthPrepass)
+        {
+            if (requiresDepthPrepass)
```
phi-lira: are you enqueuing depth prepass twice?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+                // provides better performance that doing it in a single above pass.
+                if (m_DeferredLights.HasTileDepthRangeExtraPass())
+                    EnqueuePass(m_TileDepthRangeExtraPass);
+            }
+
+            EnqueuePass(m_DeferredPass);
+
+            // Must explicitely set correct depth target to the transparent pass (it will bind a different depth target otherwise).
+            m_RenderOpaqueForwardOnlyPass.ConfigureTarget(m_CameraColorAttachment.Identifier(), m_DepthTexture.Identifier());
+            EnqueuePass(m_RenderOpaqueForwardOnlyPass);
```
phi-lira: I'm not sure if I follow, seems like you have this:

 depth prepass
 enqueue deferred (includes depth prepass (again), gbuffer, lighting and transparent forward)
 enqueue skybox (after transparent?)
 enqueue transparent (again?)

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+                depthDescriptor.depthBufferBits = k_DepthStencilBufferBits;
+                depthDescriptor.bindMS = msaaSamples > 1 && !SystemInfo.supportsMultisampleAutoResolve && (SystemInfo.supportsMultisampledTextures != 0);
+                cmd.GetTemporaryRT(m_ActiveCameraDepthAttachment.id, depthDescriptor, FilterMode.Point);
+            }
+
+            context.ExecuteCommandBuffer(cmd);
+            CommandBufferPool.Release(cmd);
+        }
+
+        void SetupBackbufferFormat(int msaaSamples, bool stereo)
```
phi-lira: we should move this logic to scriptable renderer and avoid code duplication.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
phi-lira: ah this is required because transparents?

phi-lira: we should probably discuss what optimizations we could do for lights in transparent objects for deferred. Currently per-object culling will test all objects against all lights.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+            QualitySettings.antiAliasing = msaaSamples;
+
+            if (stereo && msaaSampleCountHasChanged)
+                XR.XRDevice.UpdateEyeTextureMSAASetting();
+#else
+            QualitySettings.antiAliasing = msaaSamples;
+#endif
+        }
+        
+        bool RequiresIntermediateColorTexture(ref RenderingData renderingData, RenderTextureDescriptor baseDescriptor)
```
phi-lira: this should be much simplified for deferred I guess, without render pass code it should just return yes as we need to create gbuffer + depth. with render pass we have a scenario we stay on tile and don't need to create gbuffer (only vulkan I suppose)

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
```diff
@@ -0,0 +1,571 @@
...
+
+            bool requiresBlitForOffscreenCamera = cameraData.postProcessEnabled || cameraData.requiresOpaqueTexture || requiresExplicitMsaaResolve;
+            if (isOffscreenRender)
+                return requiresBlitForOffscreenCamera;
+
+            return requiresBlitForOffscreenCamera || cameraData.isSceneViewCamera || isScaledRender || cameraData.isHdrEnabled ||
+                   !isCompatibleBackbufferTextureDimension || !cameraData.isDefaultViewport || isCapturing || Display.main.requiresBlitToBackbuffer;
+        }
+
+        bool CanCopyDepth(ref CameraData cameraData)
```
phi-lira: I think this should always be true in deferred, we don't support MSAA and a platform that require gbuffer should support copy depth (to be double checked)

com.unity.render-pipelines.universal/Runtime/ForwardLights.cs
```diff
@@ -10,8 +10,8 @@ public class ForwardLights
     {
         static class LightConstantBuffer
         {
-            public static int _MainLightPosition;
-            public static int _MainLightColor;
+            public static int _MainLightPosition;   // DeferredLights.LightConstantBuffer also refers to the same ShaderPropertyID - TODO: move this definition to a common location shared by other UniversalRP classes
```
phi-lira: Agreed.

com.unity.render-pipelines.universal/Runtime/ForwardLights.cs
```diff
@@ -163,21 +144,56 @@ void InitializeLightConstants(NativeArray<VisibleLight> lights, int lightIndex,
...
                 float smoothAngleRange = Mathf.Max(0.001f, cosInnerAngle - cosOuterAngle);
                 float invAngleRange = 1.0f / smoothAngleRange;
                 float add = -cosOuterAngle * invAngleRange;
                 lightAttenuation.z = invAngleRange;
                 lightAttenuation.w = add;
             }
+        }
+
+        // called from DeferredLights.cs too
+        public static void InitializeLightConstants_Common(NativeArray<VisibleLight> lights, int lightIndex, out Vector4 lightPos, out Vector4 lightColor, out Vector4 lightAttenuation, out Vector4 lightSpotDir, out Vector4 lightOcclusionProbeChannel)
```
phi-lira: Should probably move to some utils helper class, maybe RenderingUtils or UniversalRenderPipelineCore

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
```diff
@@ -0,0 +1,44 @@
...
+        public DeferredPass(RenderPassEvent evt, DeferredLights deferredLights)
+        {
+            base.renderPassEvent = evt;
+            m_DeferredLights = deferredLights;
+        }
+
+        // ScriptableRenderPass
+        public override void Configure(CommandBuffer cmd, RenderTextureDescriptor cameraTextureDescripor)
+        {
+            // TODO: Cannot currently bind depth texture as read-only!
```
phi-lira: What is required to do this, should we extend ScriptableRenderPass?

phi-lira: RenderPass api supports it, I'm not sure about SetRenderTarget

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -323,42 +338,55 @@ half3 EnvironmentBRDF(BRDFData brdfData, half3 indirectDiffuse, half3 indirectSp
 // * NDF [Modified] GGX
 // * Modified Kelemen and Szirmay-Kalos for Visibility term
 // * Fresnel approximated with 1/LdotH
-half3 DirectBDRF(BRDFData brdfData, half3 normalWS, half3 lightDirectionWS, half3 viewDirectionWS)
+half3 DirectBDRF(BRDFData brdfData, half3 normalWS, half3 lightDirectionWS, half3 viewDirectionWS, bool specularHighlightsOff)
```
phi-lira: you need to add an override that takes additional parameter, otherwise it's a breaking change as any function in shader is public and user can be relying on it.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
```diff
@@ -531,16 +559,16 @@ half3 LightingSpecular(half3 lightColor, half3 lightDir, half3 normal, half3 vie
     return lightColor * specularReflection;
 }
 
-half3 LightingPhysicallyBased(BRDFData brdfData, half3 lightColor, half3 lightDirectionWS, half lightAttenuation, half3 normalWS, half3 viewDirectionWS)
+half3 LightingPhysicallyBased(BRDFData brdfData, half3 lightColor, half3 lightDirectionWS, half lightAttenuation, half3 normalWS, half3 viewDirectionWS, bool specularHighlightsOff)
```
phi-lira: Same here, we need override

com.unity.render-pipelines.universal/Runtime/DeferredRendererData.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        internal LayerMask opaqueLayerMask => m_OpaqueLayerMask;
+
+        public LayerMask transparentLayerMask => m_TransparentLayerMask;
+
+        public StencilStateData defaultStencilState => m_DefaultStencilState;
+
+        public bool shadowTransparentReceive => m_ShadowTransparentReceive;
+
+        public bool tiledDeferredShading => m_TiledDeferredShading;
```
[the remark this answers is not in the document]
kaychang-unity: Where do you add public api docs? Do you mean doxygen styles comments?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: It is mostly working but the viewport is on the 2nd camera render in scene 015_CameraMulti_FPSCam has a slightly incorrect viewport.
I'll disable cameraStacking for now.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: Fixed.

com.unity.render-pipelines.universal/Runtime/Passes/DeferredPass.cs
kaychang-unity: It's not possible to do this using SetRenderTarget(), so I am waiting for RenderPass api.

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
kaychang-unity: Already exist (see function under).

com.unity.render-pipelines.universal/ShaderLibrary/Lighting.hlsl
kaychang-unity: Ok, I added an override.

com.unity.render-pipelines.universal/Runtime/DeferredTiler.cs
```diff
@@ -0,0 +1,650 @@
...
+            m_TileHeaderSize = tilerLevel == 0 ? 4 : 2;
+            m_AvgLightPerTile = avgLightPerTile;
+            m_MaxLightPerTile = 0;
+            m_TileDataSize = 0;
+            m_TileDataCapacity = 0;
+            m_TilerLevel = tilerLevel;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public int GetTileXCount()
```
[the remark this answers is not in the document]
kaychang-unity: Ok, done.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: So I am not entirely sure how it will work with the new RenderPass api. In theory we can make it memoryless on TBDR hardware. Does it still mean we need depth buffer for all case?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: Deferred can benefit from depth-prepass if it can speed up the gbuffer passs.

The logic that dictates if we want a depth-prepass is copied from the Forward renderer. I think it would be better if both Forward and Deferred renderer had an UI option to force-enable depth-prepass, as performance gain depends on scene complexity, so the user should decide.
Then the existing code also implicitly force a depth-prepass in specific cases (msaa texture, cannot copy depth texture on platform, etc...).

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: ForwardLights needs to be setup for the standard transparent pass, but also for a new forward-only pass.
Forward-only pass is a new pass which is a mechanism copied from HDRP. Some "rare" materials cannot be rendered in deferred mode because they use an exotic BRDF, so they are rendered in forward-only (car paint, hair shader, ...).

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: I see. It should be done only for objects which are forward-only and transparent.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: Fixed:
(optional) depth prepass
enqueue deferred (gbuffer, lighting, opaque forward-only)
enqueue skybox
enqueue transparent

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: I moved `SetupBackbufferFormat()` up to ScriptableRenderer.

com.unity.render-pipelines.universal/Runtime/DeferredRendererData.cs
```diff
@@ -0,0 +1,113 @@
...
+using UnityEditor;
+using UnityEditor.ProjectWindowCallback;
+#endif
+using System;
+using UnityEngine.Scripting.APIUpdating;
+
+namespace UnityEngine.Rendering.Universal
+{
+    [Serializable, ReloadGroup]
+    [MovedFrom("UnityEngine.Rendering.LWRP")]
```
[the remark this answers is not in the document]
kaychang-unity: OK.

com.unity.render-pipelines.universal/Runtime/DeferredRendererData.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            // Upon asset creation, OnEnable is called and `shaders` reference is not yet initialized
+            // We need to call the OnEnable for data migration when updating from old versions of LWRP that
+            // serialized resources in a different format. Early returning here when OnEnable is called
+            // upon asset creation is fine because we guarantee new assets get created with all resources initialized.
+            if (shaders == null)
+                return;
+
+#if UNITY_EDITOR
+            try
```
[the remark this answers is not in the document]
kaychang-unity: Ok.

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: I removed the logic related to CanCopyDepth(), as we assume it always returns true.

com.unity.render-pipelines.universal/Runtime/ForwardLights.cs
kaychang-unity: Not sure about this. It's because DeferredRenderer uses ForwardLights class to render forward-only and transparent meshes. Maybe remove the comment "// called from DeferredLights.cs too"?

com.unity.render-pipelines.universal/Runtime/DeferredRenderer.cs
kaychang-unity: I removed `RequiresIntermediateColorTexture()`.
@phi-lira, can you check the logic related with
`m_ActiveCameraColorAttachment`/`m_ActiveCameraDepthAttachment` and `m_CameraColorTexture`/`m_CameraDepthTexture` ? Should I just delete `m_ActiveCameraColorAttachment`/`m_ActiveCameraDepthAttachment` and use `m_CameraColorTexture`/`m_CameraDepthTexture` all the time?

com.unity.render-pipelines.universal/Runtime/ForwardLights.cs
kaychang-unity: If moved InitializeLightConstants_Common() into UniversalRenderPipelineCore.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -29,10 +30,12 @@ public CopyDepthPass(RenderPassEvent evt, Material copyDepthMaterial)
         /// </summary>
         /// <param name="source">Source Render Target</param>
         /// <param name="destination">Destination Render Targt</param>
-        public void Setup(RenderTargetHandle source, RenderTargetHandle destination)
+        /// <param name="allocateRT">The destination must be temporarily allocated</param>
+        public void Setup(RenderTargetHandle source, RenderTargetHandle destination, bool allocateRT = true)
```
phi-lira: Suggestion: we want to change this creation behaviour, so render pass will declare input/output on the _CameraDepthTexture and the pipeline itself will manage it.

 For now, instead of changing public API to add parameter, can we make this AllocateRT internal property?

com.unity.render-pipelines.universal/Runtime/Passes/AdditionalLightsShadowCasterPass.cs
```diff
@@ -248,11 +255,19 @@ public override void FrameCleanup(CommandBuffer cmd)
             }
         }
 
+        public int GetShadowLightIndexForLightIndex(int visibleLightIndex)
```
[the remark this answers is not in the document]
kaychang-unity: Ok, I renamed the method.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
kaychang-unity: I made AllocateRT an internal parameter.

com.unity.render-pipelines.universal/Editor/Unity.RenderPipelines.Universal.Editor.asmdef
```diff
@@ -2,11 +2,11 @@
     "name": "Unity.RenderPipelines.Universal.Editor",
     "references": [
         "GUID:15fc0a57446b3144c949da3e2b9737a9",
-        "GUID:d60799ab2a985554ea1a39cd38695018",
```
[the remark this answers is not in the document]
kaychang-unity: To fix this, I reverted to the original file in the master branch, and re-added Unity.Mathematics.Editor and Unity.Burst.Editor. Burst and Mathematics are both needed and used by the tile deferred renderer version.
Note that even in master branch, the references d60799ab2a985554ea1a39cd38695018 and a35efad8797223d499f8c68b1f545dbc are not recognised.
image

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRGBufferPass.hlsl
```diff
@@ -0,0 +1,80 @@
...
+    // in Deferred rendering we store the sum of these values (and of emission as well) in the GBuffer
+    BRDFData brdfData;
+    InitializeBRDFData(surfaceDescription.Albedo, metallic, specular, surfaceDescription.Smoothness, surfaceDescription.Alpha, brdfData);
+    
+    Light mainLight = GetMainLight(inputData.shadowCoord);                                      // TODO move this to a separate full-screen single gbuffer pass?
+    MixRealtimeAndBakedGI(mainLight, inputData.normalWS, inputData.bakedGI, half4(0, 0, 0, 0)); // TODO move this to a separate full-screen single gbuffer pass?
+
+    half3 color = GlobalIllumination(brdfData, inputData.bakedGI, surfaceDescription.Occlusion, inputData.normalWS, inputData.viewDirectionWS);
+
+    color += LightingPhysicallyBased(brdfData, mainLight, inputData.normalWS, inputData.viewDirectionWS, false); // TODO move this to a separate full-screen single gbuffer pass?
```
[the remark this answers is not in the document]
kaychang-unity: I profiled it before (on Switch) and computing main light during gbuffer is faster than doing an extra pass during the light accumulation phase. Comparisons with built-in deferred also support this conclusion.
For SSAO, it is indeed a problem. Maybe enforce depth-prepass in that case, or explicitely move the main light as a separate pass?

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -1,695 +1,696 @@
...
- - Added Material Upgrader for stock Unity Mobile and Legacy Shaders.
- - UI improvements
-
-### Changed
-- Realtime shadow filtering was improved. 
-
-### Fixed
- - Fixed an issue that was including unreferenced shaders in the build.
- - Fixed a null reference caused by Particle System component lights.
+# Changelog
```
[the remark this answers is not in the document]
kaychang-unity: I think incorrect line endings. I reverted to original file and re-added my change (deferred renderer).

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -888,6 +888,12 @@ float3 ComputeWorldSpacePosition(float2 positionNDC, float deviceDepth, float4x4
     return hpositionWS.xyz / hpositionWS.w;
 }
 
+float3 ComputeWorldSpacePosition(float4 positionCS, float4x4 invViewProjMatrix)
```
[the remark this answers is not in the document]
kaychang-unity: I am not sure what you mean. The new version does slighly less computation by not calling ComputeClipSpacePosition().

PR #5410, opened 2019-12-28
[HDRP] Fix iridescence color space

tomix1024:
The iridescence implementation in the HDRP uses the Gaussian fits of the sensitivity curves in Fourier space from the original publication, which works in CIE XYZ color space.
However, Unity works in sRGB color space.
In the existing implementation, the result of `EvalSensitivity()` is an CIE XYZ color, and is interpreted as an sRGB color.
This pull request converts the result of `EvalSensitivity()` to sRGB.
Since out of gamut colors might be produced, negative color values are clamped to 0.
The commented out conversion from CIE XYZ to sRGB in the end of `EvalIridescence()` is not sufficient, since the `Cm`'s are sRGB colors, and would have to be converted to CIE XYZ first.

**As a consequence of this pull request, the iridescence material option in the HDRP/Lit material produces more vibrant colors than the previous implementation.**


sebastienlagarde:
Thanks for the PR. Appreciated.

sebastienlagarde:
Recreated the PR here: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5691 to be able to manipualte it and trigger yamato test.

PR #5414, opened 2020-01-03
Improve play mode change speed

alelievr:
### Purpose of this PR
Improved the speed of HDRP when entering / exiting play mode with the experimental play mode option enabled by a factor of ~25.

- Changed the texture caches to use a RenderTexture instead of a standard Texture which avoid to pay the cost of a unnecessary texture upload (that was taking almost one second).
- Also changed some TextureFormat ot GraphicsFormat because it can be used on both Texture and RenderTextures.

**Side note:** HDRP is still the slowest part of the enter play mode thing, but now it takes something like 50ms instead 1200 which is more reasonable. The rest ofd the time is spread over a lot of things:
image

### QA
Focus on testing Cookies, reflection probes and planars, the behavior must not have changed. You can also ensure that there is a performance gain when we change a setting in the HDRP asset, or when we change the play mode.


com.unity.render-pipelines.high-definition/Runtime/Core/Textures/TextureCacheCubemap.cs
```diff
@@ -97,7 +98,7 @@ public bool AllocTextureArray(int numCubeMaps, int width, TextureFormat format,
                     wrapModeV = TextureWrapMode.Clamp,
                     filterMode = FilterMode.Trilinear,
                     anisoLevel = 0,
-                    name = CoreUtils.GetTextureAutoName(panoWidthTop, panoHeightTop, format, TextureDimension.Tex2DArray, depth: numCubeMaps, name: m_CacheName)
+                    name = CoreUtils.GetTextureAutoName(panoWidthTop, panoHeightTop, TextureFormat.RGBAHalf, TextureDimension.Tex2DArray, depth: numCubeMaps, name: m_CacheName)
```
sebastienlagarde: Why is extureFormat.RGBAHalf hard coded here?

alelievr: It's because the `TextureFormat` is also hardcoded in the texture creation few line above. This was causing a mismatch between the format in the name of the texture and the actual format.

But is it intended that the `m_CacheNoCubeArray ` format is hardcoded ?

sebastienlagarde:
There is currently 4 test failing on Yamato with this PR:
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/798f743d-d6ab-45fb-984d-c8a5d942d347%2Flogs%2FTestProjects%2FHDRP_Tests%2Fupm-ci~%2Ftest-results/TestReportV1.html
(SSR fail is from HDRP/Staging)

PR #5416, opened 2020-01-03
Fix issue with AO and Dynamic Resolution

FrancescoC-unity:
For a combination of issues, AO ended up sampling the wrong region of the depth pyramid when dynamic resolution was involved. 

This should fix bug: https://fogbugz.unity3d.com/f/cases/1207408/.


FrancescoC-unity:
I added to this commit also a change that makes changes in resolution handle the temporal accumulation gracefully (that helps also when resizing the window)

PR #5418, opened 2020-01-06
Hdrp/remove HDAdditionalCameraData [2020] [need C++] [Do not merge]

RSlysz:
# THIS PR IS NOT FINISHED !!!

### Checklist for PR maker

Need C++ https://ono.unity3d.com/unity/unity/pull-request/99920/_/graphics/remove-additional-data-pattern-for-camera

---
### Purpose of this PR
This will remove the HDAdditionalCameraData component. Instead HDRP data will be now stored inside a HDCamera that inherits from the new ScriptableCamera (see C++ PR)

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
To achieve this:
- Former HDCamera class is renamed HDCameraInfo.
- HDCamera will know and behave like former HDAdditionalCameraData except that it will simplify the co-component handling
- HDAdditionalCameraData will still exist but purely for data migration purpose. It will have no logic, no accessible data and will allow to migrate smoothly the co-component Camera to a HDCamera

Important informations:
- this do not prevent the cases where the user have a Camera and not a Camera+HDAdditionalCameraData (case in user data from non HDRP, SceneCamera and PreviewCamera). So we cannot simplify the render loop co-component gestion and still use default value to handle these special cases.
- There is a Migration function to transform a Camera into a HDCamera and vice versa. But to do this we must destroy the former Camera and component relying on it will lose their references. This migration function is accessible in contextual menu of the Camera and the HDCamera. It can is also call if the Camera have still the HDAdditionalCameraData co-component while being loaded.


RSlysz:
Not continued

PR #5419, opened 2020-01-06
Only assign the reflection system when an HDRP asset is assigned

fredericv-unity3d:
### Purpose of this PR
Only assign the reflection system when an HDRP asset is assigned

Fixes 1205109 - custom reflection probe not working when having HDRP code but not using an HDRP asset

---
### Testing status

**Manual Tests**: Manual local testing
Tested to add/remove the HDRP asset in Editor/Play mode to check the reflection probe system is still active in all cases.

**Automated Tests**: -

**Yamato**:-

---
### Comments to reviewers
QA: A more thorough manual testing can be done with different probe types from both legacy and HDRP, and during play mode or editor mode.
In all cases reflection from reflection probes must work


sebastienlagarde:
Hey, I have merged this PR but it was currently not compiling :(, 
I don't understand how this haven't been catch, the type should be HDRenderPipelineAsset, not HDRenderPipeline

image

I have fixed this directly on HDRP/staging, but how have you been able to test the PR ?

fredericv-unity3d:
This seems weird but the qa did not reproduced either the bug.
Before shipping next package I think this should be included in the manual test to ensure there is no regression for the next release (@iM0ve)

PR #5426, opened 2020-01-06
Surface Gradient Bug Fixes [Skip CI]

zadi0648:
Creating this PR for Alex as she can't access her GitHub atm

---
### Summary:

**New Feature:**
- Added drop down to Lit and PBR master nodes to use Fragment Normals in different coordinate spaces.

**Fixed:**
- Object to Tangent transform https://fogbugz.unity3d.com/f/cases/1162203/
- Keywords on HD master nodes don't work https://fogbugz.unity3d.com/f/cases/1204827/
- Object space normals scale with object https://fogbugz.unity3d.com/f/cases/1176581/
- Triplanar node on objects with no UVs https://fogbugz.unity3d.com/f/cases/1152044/ (when used with surface gradients)

_Fixed Unreported Bugs:_
- TransformWorldToTangent delivered incorrect results
- Tangent WS in vert stage is 4d normalized
- TangentToWorld now properly normalizes when mode is set to directional (commit: normalize tangent to world in direction transform)

**Backport PR:** #5427

---
**Notes to QA:**
- Please ask Alex Lindman and/or Morten Mikkelsen for questions, I (Zack) only created the PR and probably won't be able to answer questions.
- Known issue: if you have an edge connected to the normal port on a master node then change the output space, the edge will disconnect. Being investigated but hopefully not a ship stopper.

---

### Testing:

Halo Effect: 2/4
Technical Risk: 2/4

Halo is probably a 2 at least, it changes a space translation function in core. It fixes a bug but does slightly change the behavior.
Tech risk likely the same, for the same reason.

**Automation:** Added automated test for all of the fixed bugs, as well as the new feature.


LandonTownsendUnity:
Note: the NaN component of this bug is not fixed:
https://fogbugz.unity3d.com/f/cases/1152044/
I've talked to Alex and she says this is known, and the PR does not fix that bug.

mmikk:
> Note: the NaN component of this bug is not fixed:
> https://fogbugz.unity3d.com/f/cases/1152044/
> I've talked to Alex and she says this is known, and the PR does not fix that bug.

That bug can't be fixed as long as that node is required to deliver the result in tangent space since tangent space is broken when there are no UVs. This is part of the reason we are making it possible to specify normal drop-off in world/object space as opposed to tangent space. However, this requires a triplanar subgraph/node that delivers in world space (such as the one I made for the surface gradient based framework). You could say what we are doing provides most of the important changes to support bump mapping without UVs but it'll always require a drop-off that is not in uv0 tangent space.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -138,22 +149,38 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
 real3 TransformWorldToTangent(real3 dirWS, real3x3 tangentToWorld)
```
EvgeniiG: This function should probably have a "Normal" suffix since you perform an inverse-transpose.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -138,22 +149,38 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
...
+	return SafeNormalize( sgn * mul(matTBN_I_T, dirWS) );
 }
 
 real3 TransformTangentToObject(real3 dirTS, real3x3 tangentToWorld)
 {
     // Note matrix is in row major convention with left multiplication as it is build on the fly
-    real3 normalWS = mul(dirTS, tangentToWorld);
-    return mul((real3x3)GetWorldToObjectMatrix(), normalWS);
+	real3 normalWS = TransformTangentToWorld(dirTS, tangentToWorld);
+	return TransformWorldToObjectNormal(normalWS);
```
EvgeniiG: Maybe it's cheaper to have a matrix multiplication and inverse-transpose the result?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
```diff
@@ -810,11 +827,24 @@ public sealed override void UpdateNodeAfterDeserialization()
...
+                RemoveSlot(NormalSlotId);
+                AddSlot(new NormalMaterialSlot(NormalSlotId, NormalSlotName, NormalSlotName, CoordinateSpace.Object, ShaderStageCapability.Fragment));
+                validSlots.Add(NormalSlotId);
+            }
+            if (MaterialTypeUsesSlotMask(SlotMask.Normal) && m_NormalDropOffSpace == NormalDropOffSpace.World)
+            {
+                RemoveSlot(NormalSlotId);
+                AddSlot(new NormalMaterialSlot(NormalSlotId, NormalSlotName, NormalSlotName, CoordinateSpace.World, ShaderStageCapability.Fragment));
+                validSlots.Add(NormalSlotId);
+            }
```
EvgeniiG: Some redundant code above.
Should probably write:

if (MaterialTypeUsesSlotMask(SlotMask.Normal)
{
    RemoveSlot(NormalSlotId);

    var coordSpace;

    switch (m_NormalDropOffSpace)
    {
    	case NormalDropOffSpace.Tangent:
    		coordSpace = CoordinateSpace.Tangent;
    		break;
    	/* etc */
    }

    AddSlot(new NormalMaterialSlot(NormalSlotId, NormalSlotName, NormalSlotName, coordSpace, ShaderStageCapability.Fragment));
    validSlots.Add(NormalSlotId);
}

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -56,6 +65,9 @@ Pass
     $TransparentWritesMotionVec:         #define _WRITE_TRANSPARENT_MOTION_VECTOR 1
     $DepthOffset:                        #define _DEPTHOFFSET_ON 1
     $BlendMode.PreserveSpecular:        #define _BLENDMODE_PRESERVE_SPECULAR_LIGHTING 1
+    $NormalDropOffTS:			#define _NORMAL_DROPOFF_SPACE	0
+    $NormalDropOffOS:			#define _NORMAL_DROPOFF_SPACE	1
+    $NormalDropOffWS:			#define _NORMAL_DROPOFF_SPACE	2
```
EvgeniiG: So this is only done for the Lit shader?

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -242,12 +253,18 @@ $include("SharedCode.template.hlsl")
...
 
         // compute world space normal
-        GetNormalWS(fragInputs, normalTS, surfaceData.normalWS, doubleSidedConstants);
+#if _NORMAL_DROPOFF_SPACE==0
+        GetNormalWS(fragInputs, normalSrc, surfaceData.normalWS, doubleSidedConstants);
+#elif _NORMAL_DROPOFF_SPACE==1
+		surfaceData.normalWS = TransformObjectToWorldNormal(normalSrc);
+#elif _NORMAL_DROPOFF_SPACE==2
+		surfaceData.normalWS = normalSrc;
+#endif
```
EvgeniiG: I feel like readability suffers with 0 1 2 instead of defines with proper names, but not a critical issue.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
```diff
@@ -157,12 +168,18 @@ $include("SharedCode.template.hlsl")
...
 
         // compute world space normal
-        GetNormalWS(fragInputs, normalTS, surfaceData.normalWS, doubleSidedConstants);
+#if _NORMAL_DROPOFF_SPACE==0
+        GetNormalWS(fragInputs, normalSrc, surfaceData.normalWS, doubleSidedConstants);
+#elif _NORMAL_DROPOFF_SPACE==1
+		surfaceData.normalWS = TransformObjectToWorldNormal(normalSrc);
+#elif _NORMAL_DROPOFF_SPACE==2
+		surfaceData.normalWS = normalSrc;
+#endif
```
EvgeniiG: So I see we need to copy-paste this code every single time.
Since the final normal is in the world space, why not modify GetNormalWS?

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
```diff
@@ -157,12 +168,18 @@ $include("SharedCode.template.hlsl")
...
 
         surfaceData.geomNormalWS = fragInputs.tangentToWorld[2];
         surfaceData.tangentWS = normalize(fragInputs.tangentToWorld[0].xyz);    // The tangent is not normalize in tangentToWorld for mikkt. TODO: Check if it expected that we normalize with Morten. Tag: SURFACE_GRADIENT
```
EvgeniiG: Check with Morten. :P

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRSubShader.cs
```diff
@@ -508,6 +509,22 @@ private static ActiveFields GetActiveFieldsFromMasterNode(AbstractMaterialNode i
...
+                case NormalDropOffSpace.Object:
+                    baseActiveFields.AddAll("features.NormalDropOffOS");
+                    break;
+                case NormalDropOffSpace.World:
+                    baseActiveFields.AddAll("features.NormalDropOffWS");
+                    break;
+                default:
+                    UnityEngine.Debug.LogError("Unknown normal drop off space: " + masterNode.normalDropOffSpace);
+                    break;
+            }
```
EvgeniiG: Make a function to avoid copy-pasting?

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/SharedCode.template.hlsl
```diff
@@ -30,8 +30,8 @@
         SurfaceDescriptionInputs output;
         ZERO_INITIALIZE(SurfaceDescriptionInputs, output);
 
-        $SurfaceDescriptionInputs.WorldSpaceNormal:          output.WorldSpaceNormal =            normalize(input.tangentToWorld[2].xyz);
-        $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           mul(output.WorldSpaceNormal, (float3x3) UNITY_MATRIX_M);           // transposed multiplication by inverse matrix to handle normal scale
+        $SurfaceDescriptionInputs.WorldSpaceNormal:          output.WorldSpaceNormal =            input.tangentToWorld[2].xyz;	// normal was already normalized in BuildTangentToWorld()
+        $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           normalize(mul(output.WorldSpaceNormal, (float3x3) UNITY_MATRIX_M));           // transposed multiplication by inverse matrix to handle normal scale
```
EvgeniiG: Why normalize here and SafeNormalize elsewhere? Can probably use the regular normalize everywhere...

com.unity.render-pipelines.high-definition/Runtime/Material/MaterialUtilities.hlsl
```diff
@@ -28,10 +28,8 @@ void GetNormalWS(FragInputs input, float3 normalTS, out float3 normalWS, float3
 #else // SURFACE_GRADIENT
 
 #ifdef _DOUBLESIDED_ON
-    // Just flip the TB in the 'flip normal' mode. Conceptually wrong, but it works.
     float flipSign = input.isFrontFace ? 1.0 : doubleSidedConstants.x;
-    input.tangentToWorld[0] = flipSign * input.tangentToWorld[0]; // tangent
-    input.tangentToWorld[1] = flipSign * input.tangentToWorld[1]; // bitangent
+    normalTS.xy *= flipSign;
```
EvgeniiG: I trust you that this works but it's not immediately clear why.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/SpriteNormalPass.hlsl
```diff
@@ -15,5 +15,6 @@ half4 frag(PackedVaryings packedInput) : SV_TARGET
     SurfaceDescriptionInputs surfaceDescriptionInputs = BuildSurfaceDescriptionInputs(unpacked);
     SurfaceDescription surfaceDescription = SurfaceDescriptionFunction(surfaceDescriptionInputs);
 
-    return NormalsRenderingShared(surfaceDescription.Color, surfaceDescription.Normal, unpacked.tangentWS.xyz, unpacked.bitangentWS, unpacked.normalWS);
+    float3 bitangent = (unpacked.tangentWS.w > 0.0 ? 1.0 : -1.0) * cross(unpacked.normalWS.xyz, unpacked.tangentWS.xyz);
```
EvgeniiG: What's in tangentWS.w?

com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/BuildSurfaceDescriptionInputs.template.hlsl
```diff
@@ -3,15 +3,27 @@ SurfaceDescriptionInputs BuildSurfaceDescriptionInputs(Varyings input)
...
 
-    $SurfaceDescriptionInputs.WorldSpaceNormal:          output.WorldSpaceNormal =            input.normalWS;
-    $SurfaceDescriptionInputs.ObjectSpaceNormal:         output.ObjectSpaceNormal =           mul(output.WorldSpaceNormal, (float3x3) UNITY_MATRIX_M);           // transposed multiplication by inverse matrix to handle normal scale
+	$SurfaceDescriptionInputs.WorldSpaceNormal: // must use interpolated tangent, bitangent and normal before they are normalized in the pixel shader.
+	$SurfaceDescriptionInputs.WorldSpaceNormal: float3 unnormalizedNormalWS = input.normalWS;
+    $SurfaceDescriptionInputs.WorldSpaceNormal: const float renormFactor = 1.0 / length(unnormalizedNormalWS);
+
+	$SurfaceDescriptionInputs.WorldSpaceBiTangent: // use bitangent on the fly like in hdrp
+	$SurfaceDescriptionInputs.WorldSpaceBiTangent: // IMPORTANT! If we ever support Flip on double sided materials ensure bitangent and tangent are NOT flipped.
+	$SurfaceDescriptionInputs.WorldSpaceBiTangent: float3 bitang = (input.tangentWS.w>0.0 ? 1.0 : -1.0) * cross(input.normalWS.xyz, input.tangentWS.xyz);
```
EvgeniiG: I am confused.
First, we support double-sided with flip, don't we?
Secondly, doesn't $SurfaceDescriptionInputs.WorldSpaceBiTangent mean that the user provides a bitangent as an input? Yet we ignore it?

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
mmikk: I think perhaps just rename dirWS to normWS is a better idea because otherwise it opens a bit of a can of worms. The way TBN normal mapping is done since it was introduced into realtime cg (incl. when using mikkts) is to transform a normal using M = TBN in the columns which corresponds to T*m.x + B*m.y + N*m.z where m is the tangent space normal. Technically the correct thing would have been to use M_I_T as in inverse transposed to transform the normal but again this is not going to change since using M directly to transform a normal out of tspace has been the standard for 2 decades and all normal maps are baked accordingly. It's done this way because it's the fastest way to construct the transformation on the fly from interpolated inputs.

In the other examples where we have special transform functions for normal vs. direction we are explicitly talking about using the inverse transposed of the matrix which is not the case here. We use M directly to transform the normal. The WorldToTangent() function is just designed to be the exact inverse of TangentToWorld() (for transforming normals). Since TBN (as in for normal mapping) tangent space is skewed and squeezed a bit it's not really useful for anything other than storing normals either statically or temporarily. The main need we have for the function TransformWorldToTangent() is so we can support object/world space normal maps  and transform them into "mikktspace" in a compliant way so we don't screw up the normal when it's transformed back out by the template into world space.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
mmikk: I don't think so because one of the 2 matrices is from constants (the world to object one) and the other is from interpolated inputs which are delivered to the pixel shader in world space.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/SharedCode.template.hlsl
mmikk: We've just followed what people were already doing in the various sections of the shader code. I am guessing in some cases people are anticipating input vectors "could be" too short and in others they are known not to be.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/SpriteNormalPass.hlsl
mmikk: That's the sign bit which dictates if the tangent frame is a left or right-hand coordinate system. This is dictated by whether or not the texture coordinates are defined clockwise or counterclockwise.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/BuildSurfaceDescriptionInputs.template.hlsl
mmikk: No that means the user is using the bitangent vector node as an input to their shader graph. So it means we are providing it on demand for the graph/user.

com.unity.render-pipelines.high-definition/Runtime/Material/MaterialUtilities.hlsl
mmikk: The tangent and bitangent conceptually represent the derivative of the surface position wrt. s and t. These do not flip on the back-side of the face. Remember as I wrote above that since the normal is transformed as T * m.x + B * m.y + N * m.z and since N is already negated this implies that the way it was before where T and B are being negated is identical to negating m.x and m.y instead. However negating m.x and m.y is both faster and conceptually more correct because it implies the height function is inverted (relative to the negated direction of the normal).
Another way to think of it is scenario Flip negates the final perturbed normal but since N is already negated we have to negate (T * m.x + B * m.y) separately to achieve the same.
Ultimately we're not changing results here. Just making it faster and conceptually more correct.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
mmikk: GetNormalWS() is used in other contexts (bent normal for instance) and is external to template/shader graph specifics. Plus it's HDRP specific.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
EvgeniiG: I am not sure I follow. You changed the implementation of this function to perform inverse-transpose. It's not just doing a matrix multiplication anymore.

EvgeniiG: Okay, I think I see what you mean. So we use the inverse-transpose only for object-space normal maps?

EvgeniiG: In any case, there is a problem here, because going from our convention, we usually have one variant of the function to transform positions, and one to transform normals, so the name here is misleading.

EvgeniiG: Generally speaking, I am not sure we need the TransformBlaDir() variant. We need TransformBla() for positions and TransformBlaNormal() for normals. TransformBlaDir() has a bit of a misleading name for less technical people, and this use case can be covered by TransformBla().

EvgeniiG: My point is that we need to be more specific with comments and naming. Otherwise, you need tribal knowledge to use these functions. :-)

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
EvgeniiG: That's a good point that it's HDRP specific. Still, since we have dozens of shaders, I would like us to avoid copy-pasting too much code if possible.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
mmikk: it's not really the inverse transpose no. TransformWorldToTangent() is doing the exact inverse of what TangentToWorld() does. The _T part is removed as you see in the comment by swapping args because we only want the inverse of what TangentToWorld() does. Not inverse transposed of what TangentToWorld() does.
The transpose part happens as a result of dropping the columns into a float3x3() syntax. We then remove that transpose by swapping args cuz we only want the exact inverse of what TangentToWorld() does.

EvgeniiG: I see, my bad. It's OK, then. Thanks.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/SharedCode.template.hlsl
EvgeniiG: I feel like it's janky and leaves perf on the table.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
marctem: I agree, but these dozens of shaders are largely being combined soon (in another PR) via some refactoring so we didn't spend too much time on it in this PR.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/SpriteNormalPass.hlsl
EvgeniiG: OK, this comment should probably be in the code above, that's not obvious IMO.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/BuildSurfaceDescriptionInputs.template.hlsl
EvgeniiG: Ah, my bad.

com.unity.render-pipelines.high-definition/Runtime/Material/MaterialUtilities.hlsl
EvgeniiG: I see, thanks. This comment should be in the code.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
EvgeniiG: Maybe add a high-level comment describing what function does.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
russoh: This looks like it should be OK.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
```diff
@@ -28,15 +28,27 @@ Pass
...
+
     $Material.SpecularColor:             #define _MATERIAL_FEATURE_SPECULAR_COLOR 1
     $SurfaceType.Transparent:            #define _SURFACE_TYPE_TRANSPARENT 1
     $BlendMode.Alpha:                    #define _BLENDMODE_ALPHA 1
     $BlendMode.Add:                      #define _BLENDMODE_ADD 1
     $BlendMode.Premultiply:              #define _BLENDMODE_PRE_MULTIPLY 1
     $DoubleSided:                        #define _DOUBLESIDED_ON 1
+    $features.NormalDropOffTS:			#define _NORMAL_DROPOFF_SPACE	0
+    $features.NormalDropOffOS:			#define _NORMAL_DROPOFF_SPACE	1
+    $features.NormalDropOffWS:			#define _NORMAL_DROPOFF_SPACE	2
```
russoh: Any reason these aren't just unique names? Just to enforce mutual exclusion?
_NORMAL_DROPOFF_SPACE_TS 1
_NORMAL_DROPOFF_SPACE_OS 1
_NORMAL_DROPOFF_SPACE_WS 1

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRForwardPass.hlsl
```diff
@@ -2,8 +2,17 @@
 {
     inputData.positionWS = input.positionWS;
 #ifdef _NORMALMAP
-    inputData.normalWS = TransformTangentToWorld(normal,
-        half3x3(input.tangentWS.xyz, input.bitangentWS.xyz, input.normalWS.xyz));
+
+#if _NORMAL_DROPOFF_SPACE==0
+	// IMPORTANT! If we ever support Flip on double sided materials ensure bitangent and tangent are NOT flipped.
+    float3 bitangent = (input.tangentWS.w>0.0 ? 1.0 : -1.0) * cross(input.normalWS.xyz, input.tangentWS.xyz);
```
phi-lira: Seems like you are now always computing bitangent in fragment. However you did not remove the bitangent interpolator in Varyings.hlsl is that expected?

phi-lira: Also seems like ShaderGraph doesn't depend on unity_WorldTransformParams.w which URP Standard shaders uses to compute the cross sign as `unity_WorldTransformParams.w * tangentOS.w`. Is this expected? Why is `unity_WorldTransformParams.w` not used here?

alindmanUnity: Bitangent was an oversight -- I removed it from the generation process just missed the include it seems. Will fix that

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
alindmanUnity: Correct this is only for Lit and PBR

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitMasterNode.cs
alindmanUnity: noted and changed

com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRForwardPass.hlsl
alindmanUnity: fixed the `unity_WorldTransformParams.w` issue, the cross sign now uses `GetOddNegativeScale()` which returns `unity_WorldTransformParams.w`

mmikk: > Also seems like ShaderGraph doesn't depend on unity_WorldTransformParams.w which URP Standard shaders uses to compute the cross sign as `unity_WorldTransformParams.w * tangentOS.w`. Is this expected? Why is `unity_WorldTransformParams.w` not used here?

As alex wrote we've fixed this by using GetOddNegativeScale(), however, what about the URP shader SpriteNormalPass.hlsl? Should we use it there too?

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
sebastienlagarde: Note: Be very careful about function renaming now that we are out of preview. We can't change this easily. Just something to have in mind.

sebastienlagarde: >We then remove that transpose by swapping args cuz we only want the exact inverse of what TangentToWorld() does. >Not the inverse transposed.
Guess we should add a comment in the code to explain exactly this, as in current state it is not obvious.

sebastienlagarde: This is obviously more costly but I think the cost is ok as the only place (outside of this PR) where we used TransformWorldToTangent is for Parallax occlusion mapping to transform the view vector in tangent space. BTW this will require that QA take care of impact on POM.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
sebastienlagarde: Note that as far as I can tell we don't use this function currently, so guess it is ok :)

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
sebastienlagarde: This will need to be added to stacklit as well. it is ok for fabric, eye, hair, axf as object or world space is not used, but stacklit is a superest of lit.

sebastienlagarde: (can be done in another PR)

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
mmikk: > > We then remove that transpose by swapping args cuz we only want the exact inverse of what TangentToWorld() does. >Not the inverse transposed.
> 
> Guess we should add a comment in the code to explain exactly this, as in current state it is not obvious.

But the comment is already there. It says:

// remove transpose part by using matrix as the first arg in mul()
// this makes it the exact inverse of what TransformTangentToWorld() does.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
sebastienlagarde: + 1
I don't understand why we define both

 $features.NormalDropOffTS:			#define _NORMAL_DROPOFF_SPACE	0
    $features.NormalDropOffOS:			#define _NORMAL_DROPOFF_SPACE	1
    $features.NormalDropOffWS:			#define _NORMAL_DROPOFF_SPACE	2

and 

 $features.NormalDropOffTS:			#define _NORMAL_DROPOFF_TS 1
    $features.NormalDropOffOS:			#define _NORMAL_DROPOFF_OS 1
    $features.NormalDropOffWS:			#define _NORMAL_DROPOFF_WS 1

We can simply always define 
#define _NORMAL_DROPOFF_TS 0
#define _NORMAL_DROPOFF_OS 1
#define _NORMAL_DROPOFF_WS 2

then  test 

#if _NORMAL_DROPOFF_SPACE== _NORMAL_DROPOFF_TS
        GetNormalWS(fragInputs, normalSrc, surfaceData.normalWS, doubleSidedConstants);
#elif _NORMAL_DROPOFF_SPACE== _NORMAL_DROPOFF_OS
		surfaceData.normalWS = TransformObjectToWorldNormal(normalSrc);
#else //_NORMAL_DROPOFF_SPACE== _NORMAL_DROPOFF_WS
		surfaceData.normalWS = normalSrc;
#endif

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
sebastienlagarde: As Marc mention, shader target refactor will change a lot of thing, so better to copy/paste for now

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
sebastienlagarde: So @mmikk can you answer this comment I have written 2 years ago :P ?

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
alindmanUnity: Sorry that looks like an error on my part. it should only be declaring NORMAL_DROPOFF_TS and _OS etc, not both. I will fix that

com.unity.render-pipelines.high-definition/Runtime/Material/MaterialUtilities.hlsl
sebastienlagarde: +1 Agree, put this comment in the code

sebastienlagarde:
Merged master and triggered Yamato to see the current state of this change: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fsurfgrad-bug-fixes/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-trunk/1040753/job/pipeline

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
mmikk: > So @mmikk can you answer this comment I have written 2 years ago :P ?

The answer is it depends on what you are using surfaceData.tangentWS for :) If you are using it to transform normals or build TBN style surface gradients then the answer is no you cannot normalize the tangent. But in HDRP we use fragInputs.tangentToWorld[] for that so whatever we use surfaceData.tangentWS for it must be something unrelated.

com.unity.render-pipelines.universal/Editor/ShaderGraph/Templates/BuildSurfaceDescriptionInputs.template.hlsl
mmikk: > First, we support double-sided with flip, don't we?

Not in the PBR shader (used in URP and supported in HDRP). It's always a scenario None.

sebastienlagarde:
So there is various failure on Yamato test:
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/5991fa18-3c24-4e03-bfc8-a6cfab0ff498%2Flogs%2FTestProjects%2FHDRP_Tests%2Fupm-ci~%2Ftest-results/TestReportV1.html

1202_Lit_DoubleSideNormalMode don't work anymore. This need to be investigate:
image

Diffusion profile are broekn: 1218_Lit_DiffusionProfiles - I not sure to understand why?

Shader graph decal are broken: 1708_DecalsSG_on_SG
image

Material quality don't work anymore: 1801_MaterialQuality

SSR have issue

There is issue with GI , bent normal etc...

There is something going wrong in the PR. Not sure why it have such impact on the graphic test, there is certainly one change that is wrong somewhere that may fix a lot of test

mmikk:
> So there is various failure on Yamato test:

Odd. The 1202_Lit_DoubleSideNormalMode test isn't using Shader Graph is it? It's using HDRP built-in lit shader if I recall correctly? The change we did in GetNormalWS() is for when #ifdef SURFACE_GRADIENT is not enabled. For Lit built-in shader it's enabled.

LandonTownsendUnity:
Tomorrow I'm going to test this branch with my Universal Shadergraph automated tests and see if we would have any failures on those.

LandonTownsendUnity:
Experiencing major regressions on tests with baked lighting.
image
[Diffs.zip](https://github.com/Unity-Technologies/ScriptableRenderPipeline/files/4072918/Diffs.zip)
Attached diffs.
Going to do some checking to make sure this isn't a mistake on my end.

alindmanUnity:
tests on yamato are all green

sebastienlagarde:
Test green, all reviewver green, merging :)

PR #5429, opened 2020-01-07
Improve custom pass

alelievr:
### Purpose of this PR

- Fix a nullref in the Custom passes due to the profiling API changes
- Fix the black/grey screen issue on after post process Custom Passes in non dev builds.
- Add a boolean in the custom pass to allow the custom pass to be rendered in scene view or not 

### QA:
I tested a to build the windows standalone in dev / non-dev + editor with playmode. The behavior should be the same on other platforms.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPass.cs
```diff
@@ -32,7 +32,7 @@ internal ProfilingSampler   profilingSampler
             get
             {
                 if (m_ProfilingSampler == null)
-                    m_ProfilingSampler = new ProfilingSampler("Custom Pass");
+                    m_ProfilingSampler = new ProfilingSampler(m_Name ?? "Custom Pass");
```
sebastienlagarde: it is not expected to use a string here? Right @JulienIgnace-Unity ?

JulienIgnace-Unity: You can use a string. The enumeration version is just a helper to not have to manage the lifetime of the ProfilingSampler objects (it allows us to have an automatic array of samplers based on the enum size).

PR #5431, opened 2020-01-07
Various improvements to volume gizmos

Chman:
### Purpose of this PR

- Added a menu item to toggle the display of Volume gizmos in `Edit > Render Pipeline > Display Selected Volume Gizmos`. Its default shortcut is `SHIFT+ALT+V` and can be remapped in the shortcut manager.
- Improved the look of the gizmos themselves a bit.
- Fixed the sphere & mesh gizmos that wouldn't show their "outer skin" (from the volume's blend distance).

---
### Testing status

**Manual Tests**: I opened an existing project with a few local volumes (of types box, sphere and mesh), selected them and made sure the gizmos were being correctly displayed and hidden when using the new menu item. I also checked that URP compiled fine as some code got moved from URP to CoreRP.


phi-lira:
Adding @martint-unity

JMargevics:
WRLFhdFfIq
Bonus issue: holding shift for uniform scale on collider gizmo has more scaling issues when volume gizmo is displayed

PR #5438, opened 2020-01-08
Set maximum on number of visible lights & fix how we calculate the maximum in URP [skip ci]

ellioman:
### Purpose of this PR
The purpose of this PR is:
1. To be able to set the maximum amount of visible lights.
2.  Fix an issue with the maximum number of visible lights on Metal (32 to 256)

This PR also increases the timeout, for Yamata Standalone Metal test, from 1200s to 1400s because of the increased number of lights.

The maximum amount of lights is by default set to -1 which means no limit. 
This PR sets the amount to the maximum allowed on URP (32 for mobile otherwise 256)

When the maximum gets set, then the code will only return at most the number of light assigned to that variable. Directional lights have the highest importance and the remaining lights will then be sorted by the distance to the camera.

**Before**
Before

**After**
After

**Performance**
LightSortingProfiler

**Example 1**
A scene contains 30 lights: 10 directional lights, 10 spotlights and 10 point lights.
The maximumVisibleLights is set to 20.
The code will return all 10 directional lights and then the 10 spot/point lights that are closest to the camera.

**Example 2**
A scene contains 40 lights: 20 directional lights, 10 spotlights and 10 point lights.
The maximumVisibleLights is set to 20.
The code will return all 20 directional lights.

---
### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 

**Yamato**:

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Flight-sorting

---
### Comments to reviewers
I have created a test scene, 010_AdditionalLightsSorted.unity, which contains 5 directional and 4096 spot/point lights.

In the SetupCullingParameters() function inside ForwardRenderer.cs, you can try out different values for the maximumVisibleLights.


phi-lira:
What's missing for this PR?

ellioman:
> What's missing for this PR?

I just need to fix the conflicts, most likely bump the Unity version inside the package.json and make a backport PR.

Will get it done on Monday (03.02).

com.unity.render-pipelines.universal/ShaderLibrary/Input.hlsl
```diff
@@ -1,27 +1,19 @@
 #ifndef UNIVERSAL_INPUT_INCLUDED
 #define UNIVERSAL_INPUT_INCLUDED
 
-#define MAX_VISIBLE_LIGHTS_SSBO 256
-#define MAX_VISIBLE_LIGHTS_UBO  32
+#define MAX_VISIBLE_LIGHTS_MOBILE  32
```
phi-lira: thinking about this we don't actually need to introduce this new define. we only use MAX_VISIBLE_LIGHTS in other scripts we can define 32 / 256 there and avoid this.

TestProjects/UniversalGraphicsTest/Assets/Test/Runtime/CameraCallbackTests.cs
```diff
@@ -35,8 +35,10 @@ public override void Create()
 		    data = UniversalRenderPipeline.asset.m_RendererDataList[0] as ForwardRendererData;
 		if (data == null)
 			return;
-		var shader = data.shaders.samplingPS;
-		m_SamplingMaterial = CoreUtils.CreateEngineMaterial(shader);
+
+        // Setting this to null because it's giving errors on our tests and the test for this is currently disabled.
+		//var shader = data.shaders.samplingPS;
+        m_SamplingMaterial = null;// CoreUtils.CreateEngineMaterial(shader);
```
phi-lira: we should probably use a try / catch block here instead if the problem is a null reference due to asset database.

PR #5439, opened 2020-01-08
Fix test framework baking issue [Skip CI]

phi-lira:
### Purpose of this PR

Fix baking issue with the graphics test framework. This changes framework to not bake lights additively.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/gfxframework%252Ffix-baking-issue


sophiaaar:
just to confirm, does this fix the issue that VFX was having too? cc @PaulDemeulenaere

PaulDemeulenaere:
> just to confirm, does this fix the issue that VFX was having too? cc @PaulDemeulenaere

To check this, we have to merge `vfx/staging`, I tried to graft this change here : https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/e4d0b649ac1c4c1e57928d7a8923650c282a3431
But it also fails with the same value, do we have something to do on this scene ?

There is only the label `TestRunnerBake` in meta file for this scene : https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/vfx/test/fix-baking-issue-graft/TestProjects/VisualEffectGraph_LWRP/Assets/GraphicsTests/Shadergraph/SampleScene/101_ShadergraphSampleScene.unity.meta

PaulDemeulenaere:
You can launch VFX_LWRP on this branch, the difference should be only in this red square :
image

sophiaaar:
> You can launch VFX_LWRP on this branch, the difference should be only in this red square

btw the result images will still show up incorrectly until the utr fix is in

PaulDemeulenaere:
PlayMode VFX URP:
image

Standalone VFX URP :
image

I confirm the GI is still missing on this test...

sophiaaar:
should this be closed?

phi-lira:
Closing. Issue was caused by running test in batchmode.

PR #5446, opened 2020-01-08
Remove some possible NaNs sources in DoF

FrancescoC-unity:
Remove some possible sources of NaNs in Depth of Field. This should fix https://fogbugz.unity3d.com/f/cases/1206585/ on Xbox


FrancescoC-unity:
Will have a look asap on Monday.

PR #5447, opened 2020-01-08
Lighting/probevolumes [2020]

thefranke:
This is a test PR to get an overview of the code fallout.


com.unity.render-pipelines.high-definition/Runtime/Core/Utilities/GeometryUtils.cs
```diff
@@ -48,7 +48,7 @@ public static void Create(Frustum frustum, Matrix4x4 viewProjMatrix, bool depth_
     } // struct Frustum
 
     [GenerateHLSL]
-    struct OrientedBBox
+    public struct OrientedBBox
```
sebastienlagarde: We need to be careful, everything public need to be documented :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -861,6 +870,7 @@ void CleanupLightLoop()
 
             CoreUtils.Destroy(m_DebugViewTilesMaterial);
             CoreUtils.Destroy(m_DebugHDShadowMapMaterial);
+			CoreUtils.Destroy(m_DebugDisplayProbeVolumeMaterial);
```
sebastienlagarde: indent

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2672,6 +2706,11 @@ static void BuildPerTileLightList(in BuildGPULightListParameters parameters, in
                     {
                         baseFeatureFlags |= LightDefinitions.s_MaterialFeatureMaskFlags;
                     }
+                    if (parameters.probeVolumeEnabled)
+                    {
+                        // TODO: Verify that we should be globally enabling ProbeVolume feature for all tiles here, or if we should be using per-tile culling.
```
sebastienlagarde: you shouldn't have a if (parameters.probeVolumeEnabled) here. proveVolume should work like reflection probe,

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2743,6 +2782,11 @@ static void BuildDispatchIndirectArguments(in BuildGPULightListParameters parame
                     {
                         baseFeatureFlags |= LightDefinitions.s_LightFeatureMaskFlags;
                     }
+                    if (parameters.probeVolumeEnabled)
+                    {
+                        // TODO: Verify that we should be globally enabling ProbeVolume feature for all tiles here, or if we should be using per-tile culling.
```
sebastienlagarde: same remark as above

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -1,13 +1,30 @@
...
+// We perform scalarization only for forward rendering as for deferred loads will already be scalar since tiles will match waves and therefore all threads will read from the same tile.
 // More info on scalarization: https://flashypixels.wordpress.com/2018/11/10/intro-to-gpu-scalarization-part-2-scalarize-all-the-lights/
 #define SCALARIZE_LIGHT_LOOP (defined(PLATFORM_SUPPORTS_WAVE_INTRINSICS) && !defined(LIGHTLOOP_DISABLE_TILE_AND_CLUSTER) && SHADERPASS == SHADERPASS_FORWARD)
 
 //-----------------------------------------------------------------------------
 // LightLoop
 // ----------------------------------------------------------------------------
 
+// Copied from VolumeVoxelization.compute
+float ProbeVolumeComputeFadeFactor(
```
sebastienlagarde: This function shouldn't be here, we can discuss about where to put it.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -409,6 +431,329 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+                        float3x3 obbFrame = float3x3(s_probeVolumeBounds.right, s_probeVolumeBounds.up, cross(s_probeVolumeBounds.right, s_probeVolumeBounds.up));
+                        float3 obbExtents = float3(s_probeVolumeBounds.extentX, s_probeVolumeBounds.extentY, s_probeVolumeBounds.extentZ);
+
+                        // Note: When normal bias is > 0, bounds using in tile / cluster assignment are conservatively dilated CPU side to handle worst case normal bias.
+                        float3 samplePositionWS = bsdfData.normalWS * _ProbeVolumeNormalBiasWS + posInput.positionWS;
+                        float3 samplePositionBS = mul(obbFrame, samplePositionWS - s_probeVolumeBounds.center);
+                        float3 samplePositionBCS = samplePositionBS * rcp(obbExtents);
+
+                        // TODO: Verify if this early out is actually improving performance.
+                        bool isInsideProbeVolume = max(abs(samplePositionBCS.x), max(abs(samplePositionBCS.y), abs(samplePositionBCS.z))) < 1.0;
```
sebastienlagarde: indeed, not sure it save anything here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -409,6 +431,329 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+                                        (int)(floor(probeVolumeTexel3D.y - 0.5) + 0.5)
+                                    );
+                                    int2 probeVolumeTexel2DFrontSW = int2(probeVolumeTexel2DBackSW.x + (int)s_probeVolumeData.resolution.x, probeVolumeTexel2DBackSW.y);
+
+                                    // TODO: Rather than sampling validity data from a slice in our texture array, we could place it in a different texture resource entirely.
+                                    // This would allow us to use a single channel format, rather than wasting memory with float4(validity, unused, unused, unused).
+                                    // It would also allow us to use a different texture format (i.e: 1x8bpp rather than 4x16bpp).
+                                    // Currently just using a texture slice for convenience, and with the idea that MAYBE we will end up using the remaining 3 channels.
+                                    probeWeightBSW = max(_ProbeVolumeBilateralFilterWeightMin, LOAD_TEXTURE2D_ARRAY_LOD(_ProbeVolumeAtlasSH, int2(probeVolumeTexel2DBackSW.x + 0, probeVolumeTexel2DBackSW.y + 0),  3, 0).x);
```
sebastienlagarde: We should really use a texture3D atlas here, the code seems very heavy :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -409,6 +431,329 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+                        // Note: When normal bias is > 0, bounds using in tile / cluster assignment are conservatively dilated CPU side to handle worst case normal bias.
+                        float3 samplePositionWS = bsdfData.normalWS * _ProbeVolumeNormalBiasWS + posInput.positionWS;
+                        float3 samplePositionBS = mul(obbFrame, samplePositionWS - s_probeVolumeBounds.center);
+                        float3 samplePositionBCS = samplePositionBS * rcp(obbExtents);
+
+                        // TODO: Verify if this early out is actually improving performance.
+                        bool isInsideProbeVolume = max(abs(samplePositionBCS.x), max(abs(samplePositionBCS.y), abs(samplePositionBCS.z))) < 1.0;
+                        if (!isInsideProbeVolume) { continue; }
+
+                        float3 samplePositionBNDC = samplePositionBCS * 0.5 + 0.5;
```
sebastienlagarde: Please move all the evaluation of the volume probe inside a separate file, something like: VolumeLightProbeUitilties.hlsl
and move ProbeVolumeComputeFadeFactor() and evaluation code to it. It pollute too much the light loop which should be restrict to the loop and data retrieval itself, not the evaluation :).

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -409,6 +431,329 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+            builtinData.backBakeDiffuseLighting = 0.0;
+        }
+        else
+#endif
+        {
+            probeVolumeDiffuseLighting *= bsdfData.diffuseColor;
+        }
+
+        // When probe volumes are enabled, builtinData.bakeDiffuseLighting only contains emissiveColor contribution.
+        builtinData.bakeDiffuseLighting += probeVolumeDiffuseLighting;
```
sebastienlagarde: We will need to talk about this one, right now you handle only Lambert lighting model, we want also to support Disney diffuse (and others). For this we need to call similar function that what we have in PostInitBuiltinData()
i.e builtinData.bakeDiffuseLighting *= _IndirectLightingMultiplier.x; (that you call above)
and  ModifyBakedDiffuseLighting(V, posInput, surfaceData, builtinData);

Note that we also must handle backface lighting, mean we want to evaluate the probe in an opposite direction.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -409,6 +431,329 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
...
+                                sampleShAb = lerp(
+                                    SAMPLE_TEXTURE2D_ARRAY_LOD(_ProbeVolumeAtlasSH, s_linear_clamp_sampler, probeVolumeAtlasUV2DBack,  2, 0),
+                                    SAMPLE_TEXTURE2D_ARRAY_LOD(_ProbeVolumeAtlasSH, s_linear_clamp_sampler, probeVolumeAtlasUV2DFront, 2, 0),
+                                    lerpZ
+                                );
+                            }
+                        }
+                    }
+
+                    float3 sampleOutgoingRadiance = SHEvalLinearL0L1(bsdfData.normalWS, sampleShAr, sampleShAg, sampleShAb);
```
sebastienlagarde: In lightloop we should never access to bsdfData, otherwise we break the contract of decoupling the lighting loop and material (and force all material to have the accessed data). This evaluation must be on material side and also handle transmission (backface lighting) when required. We can have a quick chat so I explain you the design. This is important for thing like hair and other material.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeController.cs
```diff
@@ -0,0 +1,31 @@
+using UnityEngine.Rendering;
+using UnityEngine.Serialization;
+using System;
+
+namespace UnityEngine.Rendering.HighDefinition
+{
+    [VolumeComponentMenu("Lighting/Probe Volume Global Illumination quality")]
```
sebastienlagarde: I don't think we need to add "Quality", let's just call it "Probe Volume Global Illumination"

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
```diff
@@ -113,6 +113,7 @@ class RenderDebugOverlayPassData
                     RenderRayCountOverlay(debugParams, ctx.cmd, ref x, ref y, overlaySize);
                     RenderLightLoopDebugOverlay(debugParams, ctx.cmd, ref x, ref y, overlaySize, ctx.resources.GetTexture(data.depthPyramidTexture));
                     RenderShadowsDebugOverlay(debugParams, shadowAtlases, ctx.cmd, ref x, ref y, overlaySize, ctx.renderGraphPool.GetTempMaterialPropertyBlock());
+                    // RenderProbeVolumeDebugOverlay(debugParams, ctx.cmd, ref x, ref y, overlaySize, m_DebugDisplayProbeVolumeMaterial); // TODO(Nicholas): not sure when is this code path used, please see the other usage of RenderProbeVolumeDebugOverlay.
```
sebastienlagarde: Is it expected to have this code commented?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -834,6 +837,7 @@ protected override void Dispose(bool disposing)
             m_XRSystem.Cleanup();
             m_SkyManager.Cleanup();
             CleanupVolumetricLighting();
+            m_ProbeVolumeSystem.Cleanup(); // TODO(Kuba)
```
sebastienlagarde: // TODO(Kuba) ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -933,6 +937,7 @@ void PushGlobalParams(HDCamera hdCamera, CommandBuffer cmd)
                 Fog.PushFogShaderParameters(hdCamera, cmd);
 
                 PushVolumetricLightingGlobalParams(hdCamera, cmd, m_FrameCount);
+                m_ProbeVolumeSystem.PushGlobalParams(hdCamera, cmd, m_FrameCount); // TODO(Nicholas): make symmetrical with the other volumes
```
sebastienlagarde: // TODO(Nicholas): make symmetrical with the other volumes ?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2515,7 +2523,7 @@ void InitializeGlobalResources(ScriptableRenderContext renderContext)
             hdCamera = HDCamera.GetOrCreate(camera, xrPass.multipassId);
 
             // From this point, we should only use frame settings from the camera
-            hdCamera.Update(currentFrameSettings, this, m_MSAASamples, xrPass);
+            hdCamera.Update(currentFrameSettings, this, m_ProbeVolumeSystem, m_MSAASamples, xrPass); // TODO(Kuba) this probably doesn't actually want the  m_ProbeVolumeSystem
```
sebastienlagarde: I need to check why you require the probevolumesystem for the camera here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
thefranke: Eventually, yes. The reason we use a 2D atlas is because the 3D version didn't allow to combine multiple volumes of different sizes. Also, managing fragmentation in 3D becomes much more complex than in 2D.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
thefranke: Wanted to do that anyhow for my Hackweek project, seems like the best solution.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -0,0 +1,570 @@
...
+                return;
+
+            int numProbes = parameters.resolutionX * parameters.resolutionY * parameters.resolutionZ;
+            SphericalHarmonicsL1[] data = new SphericalHarmonicsL1[numProbes];
+            float[] dataValidity = new float[numProbes];
+
+            var sh = new NativeArray<SphericalHarmonicsL2>(numProbes, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
+            var validity = new NativeArray<float>(numProbes, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
+
+            UnityEditor.Experimental.Lightmapping.GetAdditionalBakedProbes(GetID(), sh, validity);
```
thefranke: @sebastienlagarde 2020 API call.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -0,0 +1,570 @@
...
+
+            UnityEditor.EditorUtility.SetDirty(probeVolumeAsset);
+            UnityEditor.AssetDatabase.Refresh();
+
+            dataUpdated = true;
+        }
+
+        public void DisableBaking()
+        {
+            UnityEditor.Experimental.Lightmapping.additionalBakedProbesCompleted -= OnBakeCompleted;
```
thefranke: @sebastienlagarde 2020 API call.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -0,0 +1,570 @@
...
+
+            UnityEditor.Lightmapping.lightingDataCleared -= OnLightingDataCleared;
+            UnityEditor.Lightmapping.lightingDataAssetCleared -= OnLightingDataAssetCleared;
+
+            if (GetID() != -1)
+                UnityEditor.Experimental.Lightmapping.SetAdditionalBakedProbes(GetID(), null);
```
thefranke: @sebastienlagarde 2020 API call.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -0,0 +1,570 @@
...
+        }
+
+        public void EnableBaking()
+        {
+            UnityEditor.Experimental.Lightmapping.additionalBakedProbesCompleted += OnBakeCompleted;
```
thefranke: @sebastienlagarde 2020 API call.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolume.cs
```diff
@@ -0,0 +1,570 @@
...
+                                indexInCurrentBatch = 0;
+                            }
+                        }
+                    }
+                }
+            }
+
+            m_DebugProbeInputHash = debugProbeInputHash;
+
+            UnityEditor.Experimental.Lightmapping.SetAdditionalBakedProbes(GetID(), positions);
```
thefranke: @sebastienlagarde 2020 API call.

thefranke:
The only calls to the new API are made from ProbeVolume.cs. These are hooks to the baking callbacks to push probe positions and, when the bake finishes, pull and extract SH data.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
thefranke: So this code should go into `PostEvaluateBSDF`? How do we deal with all the normals we access above?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
pastasfuture: Yep, I'm overdue to create a ProbeVolume.hlsl include that contains these utility functions - does that seem appropriate?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
pastasfuture: Yeah. Given that I already wrote a custom allocator for the 2D case, I feel more confident I could put together the 3D case.

That said, this section of code could be boiled down to less scale bias terms. Its just the brute force version for now to get it up and running.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.Debug.cs
pastasfuture: We should remove it. It's some old cruft from a previous upgrade.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
pastasfuture: If I understand the architecture correctly, in 2019.3, HDRP changed from having systems be members of the HDRenderPipeline class, into implementing systems as partial class HDRenderPipeline.

ProbeVolumeSystem is still in the 2019.2 mode where is is a separate class, that HDRenderPipeline owns. Should be easy enough to switch over, just some typing. Before doing so, I just want to confirm, is my understanding correct?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
pastasfuture: Thanks Seb, that makes sense. If possible, I'd love to be on that call as well to understand how to fully decouple the two.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: yes,we have stop to use "plugin" system as it is not a plugin at all, there is too much dependency between system and nobody will replace this code anyway

PR #5457, opened 2020-01-09
Disable Async for SSR, SSAO and Contact shadow when aggregated ray tracing frame setting is on

anisunity:
In the player, when ray tracing is running, there was a possibility that ray tracing effects were executed in the async command queue (which should not happen). This PR solves this BUG -> https://fogbugz.unity3d.com/f/cases/1208354/

Also added documentation for a method that is public.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -615,9 +615,9 @@ internal static void Sanitize(ref FrameSettings sanitizedFrameSettings, Camera c
 
             bool async = sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.AsyncCompute] &= SystemInfo.supportsAsyncCompute;
             sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.LightListAsync] &= async;
-            sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.SSRAsync] &= async;
-            sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.SSAOAsync] &= async;
-            sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.ContactShadowsAsync] &= async;
+            sanitizedFrameSettings.bitDatas[(int)FrameSettingsField.SSRAsync] &= (async && !rayTracingActive);
```
sebastienlagarde: I guess it should be more subtle than that. You should not run async only if the effect is not using raytracing. 
So for example we want to keep SSR async if it use rasterization.
Or is there a general issue with async code ?

PR #5461, opened 2020-01-09
Switch to global ray tracing resources for a few HDRP constant buffer [2020]

IonutNedelcuUnity:
- Bind a few global resources and constant buffers (e.g. UnityGlobal) through the global root signature in ray tracing shaders. More details here: https://favro.com/organization/c564ede4ed3337f7b17986b6/503f833c7b8ff3e58d179086?card=Uni-108369

- Global resource binding can be enabled/disabled for debugging using DEBUG_RAY_TRACING_USE_GLOBAL_RESOURCES define.

- Ray Traced Reflection performance in a scene with ~900 Renderers using Lit.shader (Editor).

Before: https://i.imgur.com/LyCp1wN.png
After: https://i.imgur.com/McX8XR1.png

- Standalone build https://drive.google.com/file/d/1SGuiIRMStxZSlzY2cRwkqqDD_8AY5hb-/view?usp=sharing

- Manually tested using DXR projects from TestProjects\HDRP_DXR_Tests


com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/RayTracingGlobalShaderVariables.hlsl
```diff
@@ -0,0 +1,31 @@
...
+#if defined(RAY_TRACING_USE_GLOBAL_RESOURCES)
+	#define RAY_TRACING_GLOBAL_RESOURCE(type, name, reg) type name : register(reg, space1);
+	#define RAY_TRACING_GLOBAL_CBUFFER_START(name, reg) cbuffer name : register(reg, space1) {
+#else
+	#define RAY_TRACING_GLOBAL_RESOURCE(type, name, reg) type name;
+	#define RAY_TRACING_GLOBAL_CBUFFER_START(name, reg) CBUFFER_START(name)
+#endif
+
+// Global Constant Buffers - b registers. Unity supports a maximum of 16 global constant buffers.
+#define UNITY_GLOBAL_CBUFFER_REGISTER                           b0
```
sebastienlagarde: @IonutNedelcuUnity Is there a way to have automatic assignment of the register index like we have for texture, constant buffer and others stuff?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/RayTracingGlobalShaderVariables.hlsl
```diff
@@ -0,0 +1,31 @@
+#ifndef UNITY_RAY_TRACING_GLOBAL_RESOURCES_INCLUDED
+#define UNITY_RAY_TRACING_GLOBAL_RESOURCES_INCLUDED
+
+#define DEBUG_RAY_TRACING_USE_GLOBAL_RESOURCES 1
```
sebastienlagarde: Why is this macro named DEBUG_RAY_TRACING_USE_GLOBAL_RESOURCES  ? shouldn't be call ENABLE_RAY_TRACING_USE_GLOBAL_RESOURCES  ?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -132,7 +133,11 @@ TEXTURE2D(_PrevExposureTexture);
 
 // Important: please use macros or functions to access the CBuffer data.
 // The member names and data layout can (and will) change!
+#if defined(RAY_TRACING_USE_GLOBAL_RESOURCES)
+RAY_TRACING_GLOBAL_CBUFFER_START(UnityGlobal, UNITY_GLOBAL_CBUFFER_REGISTER)
```
sebastienlagarde: given that RAY_TRACING_GLOBAL_CBUFFER_START is already define as CBUFFER_START if RAY_TRACING_USE_GLOBAL_RESOURCES isn't define, we could simply let RAY_TRACING_GLOBAL_CBUFFER_START 

I understand that the problem here is the name of the Macro itself which could be misleading. So maybe we should call the macro: GLOBAL_CBUFFER_START instead. And it have only effect if RAY_TRACING_USE_GLOBAL_RESOURCES  is define, otherwise it fallback to CBUFFER_START. This is cleaner I guess and less misleading. Though?

sebastienlagarde: i.E I suggest to replace 
#if defined(RAY_TRACING_USE_GLOBAL_RESOURCES)
RAY_TRACING_GLOBAL_CBUFFER_START(UnityGlobal, UNITY_GLOBAL_CBUFFER_REGISTER)
#else
CBUFFER_START(UnityGlobal)
#endif

by 

GLOBAL_CBUFFER_START(UnityGlobal)

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/RayTracingGlobalShaderVariables.hlsl
sebastienlagarde: Note: Just to keep compatibility with 19.3 we can do

#if VERSION_GREATER_EQUAL(8, 0)
DEBUG_RAY_TRACING_USE_GLOBAL_RESOURCES 1
#endif

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/RayTracingGlobalShaderVariables.hlsl
IonutNedelcuUnity: @sebastienlagarde 
For register space0 where we don't specify the register, yes. Space0 is used for local root signature and the register is assigned by dxcompiler automatically. For global resources is not possible because a global resource can be automatically assigned to different registers in different compilation units (like hit groups / raytrace shaders) or a register can be used by different resources. To recap the idea with global resources is to setup them only once per DispatchRays and not for each hit group table record.
Say you have 2 shaders that use UnityGlobal and other cbuffers. Since they are different code with different features, is possible that these resources can be assigned automatically to different registers.

Here's just an example of global cbuffers usage and the register assigned by dxcompiler assuming we don't specify it manually and let the compiler assign it.

Hit Group 1:
cbuffer UnityGlobal : b0
cbuffer SomeGlobalShaderParams : b1  - this is because the shader uses an uniform from this cbuffer and the cbuffer is declared before UnityRayTracingLightLoop in the code.
cbuffer UnityRayTracingLightLoop : b2

Hit Group 2:
cbuffer UnityGlobal : b0
cbuffer UnityRayTracingLightLoop : b1

Hit Group 1 and Hit Group 2 being two different shaders, they are compiler separately and in this case dxcompiler can assign same register b1 to SomeGlobalShaderParams and UnityRayTracingLightLoop. Also UnityRayTracingLightLoop occupies different registers in different hit groups.

Same thing can happen with other resources. dxcompiler can assign t0 to a RaytracingAccelerationStructure in a hit group because is doing a TraceRay but in another hit group t0 is occupied by an environment texture and the RaytracingAccelerationStructure is not used. If we plan to make the RaytracingAccelerationStructure or the env texture global resources then in this case is not possible since they use the same register.

How registers are assigned by the compiler is directly linked to how the global root signature is created and this is unique per DispatchRays invocation. If the registers are specified in the code then we guarantee that there will be no conflicts since a resource will occupy the same register in different hit groups. For this reason I created RayTracingGlobalShaderVariables.hlsl where the register indices used for global resources in HDRP are centralized. 

I added some information in the favro card https://favro.com/organization/c564ede4ed3337f7b17986b6/503f833c7b8ff3e58d179086?card=Uni-108369

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/RayTracingGlobalShaderVariables.hlsl
IonutNedelcuUnity: I can delete #define DEBUG_RAY_TRACING_USE_GLOBAL_RESOURCES 1. To turn on/off the support for global resources we can comment out #define RAY_TRACING_USE_GLOBAL_RESOURCES.

IonutNedelcuUnity: For compatibility I created UNITY_RAY_TRACING_GLOBAL_RESOURCES which I defined when I added the support for global resources in the engine code. UNITY_RAY_TRACING_GLOBAL_RESOURCES is always define starting with the PR for global resources in the engine code. If we use a newer HDRP version with an older engine version if this is possible then UNITY_RAY_TRACING_GLOBAL_RESOURCES won't be there and it will turn off the support.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
IonutNedelcuUnity: > given that RAY_TRACING_GLOBAL_CBUFFER_START is already define as CBUFFER_START if RAY_TRACING_USE_GLOBAL_RESOURCES isn't define, we could simply let RAY_TRACING_GLOBAL_CBUFFER_START
> 
> I understand that the problem here is the name of the Macro itself which could be misleading. So maybe we should call the macro: GLOBAL_CBUFFER_START instead. And it have only effect if RAY_TRACING_USE_GLOBAL_RESOURCES is define, otherwise it fallback to CBUFFER_START. This is cleaner I guess and less misleading. Though?

I'm using the RAY_TRACING_USE_GLOBAL_RESOURCES and RAY_TRACING_GLOBAL_CBUFFER_START to not confuse the reader. 
The register index has to be specified anyway. 

I can change the cbuffer definition to something like:

GLOBAL_CBUFFER_START(UnityGlobal, UNITY_GLOBAL_CBUFFER_REGISTER)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/ShaderVariablesRaytracing.hlsl
```diff
@@ -1,23 +1,28 @@
...
 #include "Packages/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/HDRayTracingManager.cs.hlsl"
 
 // The target acceleration acceleration structure should only be defined for non compute shaders
 #ifndef SHADER_STAGE_COMPUTE
-RaytracingAccelerationStructure         _RaytracingAccelerationStructure;
+RAY_TRACING_GLOBAL_RESOURCE(RaytracingAccelerationStructure, _RaytracingAccelerationStructure, RAY_TRACING_ACCELERATION_STRUCTURE_REGISTER);
 #endif
+
+RAY_TRACING_GLOBAL_CBUFFER_START(UnityRayTracingGlobals, UNITY_RAY_TRACING_GLOBAL_CBUFFER_REGISTER)
+float                                   _RaytracingCameraNearPlane;
```
anisunity: @IonutNedelcuUnity maybe its a display thing why only these 3 variables are int the constant buffer? not the rest?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Raytracing/Shaders/ShaderVariablesRaytracing.hlsl
```diff
@@ -1,23 +1,28 @@
...
+int                                     _RaytracingMinRecursion;
+int                                     _RaytracingMaxRecursion;
+CBUFFER_END
+
```
anisunity: @IonutNedelcuUnity maybe its a display issue, but why only these 3 variables are in the constant buffer?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
```diff
@@ -31,7 +31,7 @@
 #endif
 
 // Workaround for lack of multi compile in compute/ray shaders
-#if defined(UNITY_TEXTURE2D_X_ARRAY_SUPPORTED) && (defined(SHADER_STAGE_COMPUTE) || defined(SHADER_STAGE_RAYTRACING))
+#if defined(UNITY_TEXTURE2D_X_ARRAY_SUPPORTED) && (defined(SHADER_STAGE_COMPUTE) || defined(SHADER_STAGE_RAY_TRACING))
```
anisunity: This has already been changed

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -913,7 +913,7 @@ PositionInputs GetPositionInput(float2 positionSS, float2 invScreenSize, uint2 t
     ZERO_INITIALIZE(PositionInputs, posInput);
 
     posInput.positionNDC = positionSS;
-#if SHADER_STAGE_COMPUTE || SHADER_STAGE_RAYTRACING
+#if SHADER_STAGE_COMPUTE || SHADER_STAGE_RAY_TRACING
```
anisunity: This has already been changed

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -330,6 +330,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Search field selects the appropriate fields in both project settings panels 'HDRP Default Settings' and 'Quality/HDRP'
 - Disabled the refraction and transmission map keywords if the material is opaque.
 - Keep celestial bodies outside the atmosphere.
+- Changed a few resources used by ray tracing shaders to be global resources (using register space1) for improved CPU performance.
```
anisunity: This is in the wrong location  now.

sebastienlagarde:
@IonutNedelcuUnity No worry, we will handle the PR and redo it. thanks for the insight.

sebastienlagarde:
close as replaced by this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6300

PR #5463, opened 2020-01-09
 Added a menu option that checks scene issues with ray tracing. 

anisunity:
Also removed the previously existing warning at runtime. The new code reports if a material is null, if a mesh is null and if a mesh has both transparent and opaque submeshes.

image


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/HDRenderPipelineMenuItems.cs
```diff
@@ -517,5 +517,94 @@ public void LogException(Exception exception, UnityObject context)
...
+                    // Initially we consider the potential mesh as invalid
+                    if (materialArray.Count > meshIdx)
+                    {
+                        // Grab the material for the current sub-mesh
+                        Material currentMaterial = materialArray[meshIdx];
+
+                        // The material is transparent if either it has the requested keyword or is in the transparent queue range
+                        if (currentMaterial != null)
+                        {
+                            bool materialIsTransparent = currentMaterial.IsKeywordEnabled("_SURFACE_TYPE_TRANSPARENT")
```
sebastienlagarde: there is no _SURFACE_TYPE_TRANSPARENT for shader graph master node.

sebastienlagarde: only renderqueue should be tested for this

anisunity:
Reflection probes are now ignored, it is not posisble to detect shaders that do not compile or that are not impltement in ray tracing yet (because of SG)

PR #5465, opened 2020-01-09
Fixing unnecessary memory allocations in the ray tracing cluster build and Updating the screen shots for the ray tracing tests

anisunity:
I also disabled the memory check due to the necessity of Calling FindAllOfType for ray tracing.


RemyUnity:
Pair-tested with Anis, approved ✅

PR #5468, opened 2020-01-10
Added API to update probe settings at runtime

fredericv-unity3d:
### Purpose of this PR
Add an api to update probe settings at runtime.

Users don't have access to all settings used internally be the reflection probe and this additional api enable them to use all of them from C# scripts.

---
### Testing status

**Manual Tests**:
- Tested locally with a script that settings can be modified at runtime

**Automated Tests**: not appropriate

**Yamato**: -

---
### Comments to reviewers
Doc: Can you review the xmldoc of the new api?
QA: I am not sure that a lot of testing is relevant, it is only exposing existing API to public.
The impact is one ReflectionProbe and PlanarProbe settings, so a small smoke test on those can be appropriate.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -143,16 +143,35 @@ internal bool requiresRealtimeUpdate
         // Texture asset
         /// <summary>
         /// The baked texture. Can be null if the probe was never baked.
+        ///
+        /// Use the setter only if you know what you are doing.
```
reallyfancy: Change to:

Most of the time, you do not need to set this value yourself. You can set this property in situations where you want to manually assign data that differs from what Unity would generate.

reallyfancy: And can you suggest a use case for when a user would want to do this?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -143,16 +143,35 @@ internal bool requiresRealtimeUpdate
...
+        public Texture customTexture
+        {
+            get => m_CustomTexture;
+            set => m_CustomTexture = value;
+        }
+
         /// <summary>
         /// The allocated realtime texture. Can be null if the probe never rendered with the realtime mode.
+        ///
+        /// Use the setter only if you know what you are doing.
```
reallyfancy: As above, please change to:

Most of the time, you do not need to set this value yourself. You can set this property in situations where you want to manually assign data that differs from what Unity would generate.

reallyfancy: And suggest a use case if possible!

reallyfancy: Change to:

Most of the time, you do not need to set this value yourself. You can set this property in situations where you want to manually assign data that differs from what Unity generates.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
reallyfancy: Change to:

Most of the time, you do not need to set this value yourself. You can set this property in situations where you want to manually assign data that differs from what Unity generates.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -325,7 +343,19 @@ public Vector3 proxyExtents
         /// </summary>
         public Bounds bounds => influenceVolume.GetBoundsAt(transform.position);
 
-        internal ProbeSettings settings
+        /// <summary>
+        /// Use this property to edit the settings of the probe.
+        ///
+        /// If you intend to read the settings back, you may need the sanitized version <see cref="settings"/>.
```
reallyfancy: To read the settings of this probe, most of the time you should use the sanitized version of this property: `<see cref="settings"/>`. Use this property to read the settings of the probe only when it is important that you read the raw data.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -325,7 +343,19 @@ public Vector3 proxyExtents
...
+        /// </summary>
+        public ref ProbeSettings settingsRaw => ref m_ProbeSettings;
+
+        /// <summary>
+        /// Use this property to get the settings used for calculations.
+        ///
+        /// This is a sanitized version of the property <see cref="settingsRaw"/>.
```
reallyfancy: Change to:

To edit the settings of the probe, use the unsanitized version of this property: `<see cref="settingsRaw"/>`.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
fredericv-unity3d: Fixed

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
fredericv-unity3d: Fixed

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
fredericv-unity3d: Fixed

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
fredericv-unity3d: Fixed

PR #5480, opened 2020-01-10
Fix various issues with time and animated materials

EvgeniiG:
### Purpose of this PR
Simplify time management in the HDRP.
Fix various issues with time and animated materials, such as:
https://fogbugz.unity3d.com/f/cases/1210068/
https://fogbugz.unity3d.com/f/cases/1210064/

---
### Testing status

**Manual Tests**: I only tested 2020.1 (need to test 2019.3!)
**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-time/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_trunk/1040174/job

---
### Comments to reviewers
Need to manually test 2019.3 (and maybe double-check 2020.1).

Expected behavior for all animated materials (including density volumes) now is:
1. game view - animate if playing;
2. scene view - animate if "animate materials" is checked;
3. previews - always animate.


sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp In addition to what Evgenii Said. 

Expected behavior for all animated materials (including density volumes) now is:
game view - animate if playing;
scene view - animate if "animate materials" is checked;
previews - always animate.

Here is a list of additional thing to check:
Check animated material and density volume in this context:

- Check with multiple game view and scene view at the same time + preview (material preview for example). 
- Check the preview of shader graph (maybe with multiple window of shader graph)
- Check with planar reflection and realtime reflection probe inside the scene

Do all at the same time ^

Then:
- Check that antialiasing work correctly in the sceneview and multiple gameview, same for volumetric fog with reprojection.

Check that we haven't break previous PR fix (if denoising after reload of a scene and other stuff):
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5369
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5423

Need to manually test 2019.3 (and maybe double-check 2020.1).

sebastienlagarde:
@iM0ve 
>Realtime Reflection probe is breaking rendering (one scene view is enough to cause the error)
Is it from this PR? or this was failing before this PR ?

TomasKiniulis:
> @TomasKiniulis
> 
> > Realtime Reflection probe is breaking rendering (one scene view is enough to cause the error)
> 
> Is it from this PR? or this was failing before this PR ?

Yes it seems like this is failing from this PR. I've checked staging from Jan 10 and it works there. Bisected now a bit and it turns out that the latest commit is breaking the probe.
Fix animation in reflection a1403e4 commit - breaks pr
Only update once per frame a550aec commit - works fine

sebastienlagarde:
> > @TomasKiniulis
> > > Realtime Reflection probe is breaking rendering (one scene view is enough to cause the error)
> > 
> > 
> > Is it from this PR? or this was failing before this PR ?
> 
> @sebastienlagarde
> Yes it seems like this is failing from this PR. I've checked staging from Jan 10 and it works there. Bisected now a bit and it turns out that the latest commit is breaking the probe.
> Fix animation in reflection [a1403e4](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/a1403e439f26b849d9ed44f7948548a1b6bf1495) commit - breaks pr
> Only update once per frame [a550aec](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/a550aec1c3a0ce404008a8840ba2aaf943a7b23f) commit - works fine

@EvgeniiG if you can check, thanks

EvgeniiG:
I tested on the same scene and I don't see errors like this. I will keep trying to repro.

EvgeniiG:
I fixed the issue with null reference exception.
However, the problem remains: since I cannot access the camera which requested a spherical reflection probe capture (it is "view-independent" whatever that is supposed to mean), we have no parent camera and therefore cannot determine whether we should animate anything or not.
So animation in spherical reflection probes is broken.
And I cannot figure out the code.

EvgeniiG:
Note that I am pretty sure that it has never worked.

EvgeniiG:
Textured POM seems broken in test 1708.

sebastienlagarde:
> Textured POM seems broken in test 1708.
> 
> image

This is random. We have this sometime locally, sometime on yamato. Related to depth offset. Sadly we only get this on the screenshots, if you open the scene all is fine. Not related to your PR.

sebastienlagarde:
@TomasKiniulis 
"previews - always animate - this is working, but the result is choppy or requires mouse movement to refresh the view. Perhaps, it would be possible to trigger the Preview window to be in "Playing" by default?"

was it already like this or not working at all before this PR? 
and if it was working like this already, can you check if URP behave the same? 
Thanks

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbeSystem.cs
```diff
@@ -138,22 +138,22 @@ static Texture CreateAndSetRenderTargetIfRequired(HDProbe probe, ProbeSettings.M
...
-                {
-                    case ProbeSettings.RealtimeMode.EveryFrame:
-                        return probe.lastRenderedFrame != Time.frameCount;
-                    case ProbeSettings.RealtimeMode.OnEnable:
-                        return !probe.wasRenderedAfterOnEnable;
-                }
-            }
-            return true;
-        }
+        //static bool DoesRealtimeProbeNeedToBeUpdated(HDProbe probe)
```
JulienIgnace-Unity: Was unused? Why not remove the code? @fredpointzero  for awarness.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1422,26 +1396,35 @@ float ComputeVisibility(int visibleInIndex, HDProbe visibleProbe)
...
                                 HDUtils.GetSceneCullingMaskFromCamera(visibleInRenderRequest.hdCamera.camera),
+                                parentCamera,
                                 visibleInRenderRequest.hdCamera.camera.fieldOfView,
                                 visibleInRenderRequest.hdCamera.camera.aspect
                             );
                         }
                     }
                     else
                     {
+                        // No parent camera.
```
JulienIgnace-Unity: This is expected as a "view independent" probes might be used in several different cameras so as such they don't really have a single parent.
Not sure what's best in this case for animated materials. Can we just enable it if ApplicationIsPlaying or the animated material checkbox is enabled? Not sure how consistent that would be with the rest though.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1138,51 +1142,18 @@ protected override void Render(ScriptableRenderContext renderContext, Camera[] c
...
-                    // Therefore, outside of the Play Mode we update the time at 60 fps,
-                    // and in the Play Mode we can rely on 'Time.frameCount'.
-                    newTime  = Time.realtimeSinceStartup;
-                    newFrame = (newTime - m_Time) > 0.0166f;
-
-                    // If we switch to other scene 'Time.realtimeSinceStartup' is reset, so we need to
-                    // reset also m_Time. Here we simply detect ill case to trigger the reset.
-                    newFrame = newFrame || (newTime <= m_Time);
-                }
+            int  newCount = Time.frameCount;
```
JulienIgnace-Unity: Afaik, the existing code made a different case for Application.isPlaying and the rest because time management behaved differently in both cases? Did something change that allows us to change the code like this? If so will this work for 2019.3 when backported?

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbeSystem.cs
fredericv-unity3d: This is an issue, If we render multiple cameras in a single frame realtime probe probably will be rendered for each camera.

And it is best to remove code instead of commenting it. (Also, more general topic: we have a lot of dead code we still maintain, it can help to delete the dead code)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
fredericv-unity3d: "View independent" camera means "We can use the result of a single rendering for several camera requests"
If we have in a single frame multiple cameras looking at a reflection from a realtime reflection probe and they should look the same, that probe should go in the "view independent" list.

If the probe needs to be rendered differently for each viewing cameras, it is no more "view-independent" and needs to go in the "view-dependent" list and depends on the said viewer camera.

Maybe call it "viewer-dependent" and "viewer-independent".

But in all case assigning a "parentCamera" to the "view-independent" probe list is incorrect.

TomasKiniulis:
> @TomasKiniulis
> "previews - always animate - this is working, but the result is choppy or requires mouse movement to refresh the view. Perhaps, it would be possible to trigger the Preview window to be in "Playing" by default?"
> 
> was it already like this or not working at all before this PR?
> and if it was working like this already, can you check if URP behave the same?
> Thanks

Before the PR it was the same, and URP also has the same choppy behavior.

> Check that we haven't break previous PR fix (if denoising after reload of a scene and other stuff):
> #5369
> #5423

#5423 works checking in 2020.1, but can't approve that it works in 2019.3. Switching to DX12 in 2019.3 produces the following crash:
========== OUTPUTTING STACK TRACE ==================
0x00007FFFFDD2F44E (nvwgf2umx) OpenAdapter12
0x00007FF809B18EAE (d3d12) D3D12PIXReportCounter
0x00007FF612B04B5E (Unity) FillTableUNIAndBarrier
0x00007FF612B028D8 (Unity) CmdListFillTableUNI
0x00007FF612B0A2AD (Unity) D3D12DeviceState::Transition
0x00007FF612B016A8 (Unity) D3D12CommandListWithState::ApplyStateDrawcall
0x00007FF612B28513 (Unity) GfxDeviceD3D12Base::GenerateMips
0x00007FF612B30AFE (Unity) GfxDeviceD3D12Base::SetRenderTargetsImpl
0x00007FF612B78ABF (Unity) GfxDeviceWorker::RunCommand
0x00007FF612B82A2B (Unity) GfxDeviceWorker::RunExt
0x00007FF612B82DE8 (Unity) GfxDeviceWorker::RunGfxDeviceWorker
0x00007FF613A0B093 (Unity) Thread::RunThreadWrapper
0x00007FF83FD77974 (KERNEL32) BaseThreadInitThunk
0x00007FF8409DA271 (ntdll) RtlUserThreadStart
========== END OF STACKTRACE ===========

It doesn't seem to be caused by this PR. Will investigate the crash more.

sebastienlagarde:
@TomasKiniulis you can't test on 19.3, we need to backport the PR to check this.

TomasKiniulis:
> @TomasKiniulis you can't test on 19.3, we need to backport the PR to check this.

Tha

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: Hope so. QA will find out.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: I suppose in this case the criteria of view-independence need to be adjusted to account for animation settings.

EvgeniiG: > Can we just enable it if ApplicationIsPlaying or the animated material checkbox is enabled?

No, because this behavior is valid only for game and VR views.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
EvgeniiG: But yes, indeed, something was fixed. This used to not work.

TomasKiniulis: I did check in 2019.3, and the changes are indeed working

TomasKiniulis:
> @TomasKiniulis you can't test on 19.3, we need to backport the PR to check this.

I thought this is necessary to check in 19.3 too, so I've locally modified the release/2019.3 package with the changes. The PR appears to be good. The only unfortunate part is that realtime reflection probes cannot update animated materials.

TomasKiniulis:
Did sanity after the merges, seems to be working fine. 

@EvgeniiG
Noticed that Realtime Reflection Probes are actually working in game view playmode, but it is also being animated even in Scene views and is not affected by the scene view animated material toggle. It just depends if you are in play mode or not.
FbLGyqCFqA

PR #5487, opened 2020-01-13
Make the shadow behavior for transparent and transmissive objects more intuitive.

anisunity:
Previously, to have a colored or semi transparent shadow. We needed to set the shadow mode to off on transparent/transmissive objects (otherwise they would cast opaque shadows) which doesnt make sense on the user side. Now transparent objects cast transparent and tranmissive object cast color shadows are soon as the light supports it. I also added two  test scenes to validate the behavior.

This PR also adds a proper management of submeshes with both transparent and opaque meshes (which was not the case before).

105_RTColorShadow
106_RTShadow


sebastienlagarde:
For QA:, there is this PR to check: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5490 that will cover both PR, so fine to only test 5490

sebastienlagarde:
part of https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5490

PR #5491, opened 2020-01-13
Fixed the first frame flash with auto exposure

Chman:
### Purpose of this PR

In scenes with very high or very low exposure, the first frame when entering play mode would be either fully white or fully black when using Auto Exposure. This PR fixes it by manually applying the exposure in the same frame instead of N+1 on the first frame (or on `ResetHistory()`).

Note that because the first frame is not pre-exposed it can result in visible compression artifacts and clipped white values (a color buffer with a high exposure will clip as we're not running full precision and are using R11G11B10 or FP16 targets). But at least it doesn't flash anymore.

---
### Testing status

Manual testing. Create a new HDRP scene, put an HDRI with about 16ev for the exposure. The screen will be white. Add an Auto Exposure effect to a volume and press play. The screen won't flash in white on the first frame.


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -420,6 +420,28 @@ void PoolSource(ref RTHandle src, RTHandle dst)
                     using (new ProfilingScope(cmd, ProfilingSampler.Get(HDProfileId.DynamicExposure)))
                     {
                         DoDynamicExposure(cmd, camera, source, lightingBuffer);
+
+                        // On reset history we need to apply dynamic exposure immediately to avoid
+                        // white or black screen flashes when the current exposure isn't anywhere
+                        // near 0
+                        if (m_ResetHistory)
```
sebastienlagarde: minor optimization, we should only do that when we are in auto-exposure mode. But agree that this doesn't have any impact for a single frame, so ok :)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -420,6 +420,28 @@ void PoolSource(ref RTHandle src, RTHandle dst)
...
+                        {
+                            var destination = m_Pool.Get(Vector2.one, k_ColorFormat);
```
sebastienlagarde: You should merge HDRP/staging and update k_ColorFormat to m_ColorFormat (due to Pavlos PR merge).
As this will not be detected until we merge in HDRP/staging

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Chman: It's already done only when Auto Exposure is enabled, see the if just above it ;)
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/f3846d598f5ca20da39f7bdf3fa3a33d1834db1c/com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs#L418

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
Chman: Ah right, Pavlos' PR wasn't merged when I pushed mine. Will do!

Chman: Fixed

sebastienlagarde:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Fexposure-flash-fix/.yamato%252Fupm-ci-hdrp.yml%2523All_HDRP_fast-trunk

TomasKiniulis:
There's a certain case where a flash is still present.

To reproduce:
Scene must contain two cameras
Neither of them should have TAA enabled
At this point going to playmode you get the flash

sebastienlagarde:
@TomasKiniulis can you recreate a fogbugz for this case:
"There's a certain case where a flash is still present."

I will merge this PR as it is a high user pain for now (and RM poke me all the time about it so we can close original fogbugz).

sebastienlagarde:
@TomasKiniulis if you can recheck the PR with last change, thanks.

TomasKiniulis:
> @TomasKiniulis if you can recheck the PR with last change, thanks.

The issue with two cameras is fixed, but there's another issue introduced with this PR -  the flash is introduced at first frame in Scene view now. Filed a bug report here: https://fogbugz.unity3d.com/f/cases/1211923/

sebastienlagarde:
ok. Note: I have merged the PR to be compliant with RM, Thomas will work on another PR for the new issue you raised. Thanks

PR #5499, opened 2020-01-14
Disable "Copy Shader" for VFX master node

pbbastian:
### Purpose of this PR
Fixes UNITY-1206090 by removing the "Copy Shader" option for the VFX master node.

---
### Testing status

**Manual Tests**: Verified locally

**Automated Tests**: Shader Graph UI changes only, which doesn't have automated tests.

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
+++ b/com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
@@ -311,7 +311,7 @@ namespace UnityEditor.ShaderGraph.Drawing
                         _ => isActive ? DropdownMenuAction.Status.Checked : DropdownMenuAction.Status.Normal);
                 }
 
-                var canViewShader = node.hasPreview || node is IMasterNode || node is SubGraphOutputNode;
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Drawing/Views/MaterialNodeView.cs
```diff
@@ -311,7 +311,7 @@ public override void BuildContextualMenu(ContextualMenuPopulateEvent evt)
                         _ => isActive ? DropdownMenuAction.Status.Checked : DropdownMenuAction.Status.Normal);
                 }
 
-                var canViewShader = node.hasPreview || node is IMasterNode || node is SubGraphOutputNode;
+                var canViewShader = !(node is VfxMasterNode) && (node.hasPreview || node is IMasterNode || node is SubGraphOutputNode);
```
zadi0648: Two things:
1) Usually it's nice to have a small comment before these special exceptions to explain why it's there (especially for us new comers 😛)
2) Is there a plan to remove GetShader's System.NotImplementedException someday? Atm I'm unfamiliar with how our VFX integration works. Obviously I can see the use case of a user wanting to do that, although getting the preview working is a higher priority (and I would image the two go together). Alternatively we may want to disable the Master Preview too for when it's VFX master node...

PR #5503, opened 2020-01-14
Fix for case 1200161 Save as subgraph issue

martint-unity:
---
### Purpose of this PR
Fixes for case https://fogbugz.unity3d.com/f/cases/1200161/
In where using save as on a subgraph would result in a null ref exception.

---
### Testing status

Running Yamato


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -376,7 +376,7 @@ public void SaveAs()
                         if (success)
                         {
                             ShaderGraphImporterEditor.ShowGraphEditWindow(newPath);
-                            if (GraphData.onSaveGraph != null)
+                            if (GraphData.onSaveGraph != null && extension != ShaderSubGraphImporter.Extension)
```
zadi0648: Simple change. Two things:

Just wanted to run by: Is this future proof? Will onSaveGraph possibly have other roles in the future that do involve Sub Graphs? Since saveContext is master node specific I believe this to be fine.

At first glance it's confusing what this conditional statement is about and why Sub Graphs can safely ignore it (namely to the SG dev newbies such as myself: GraphData.onSaveGraph sounds important; based on its name it sound like something that should happen to all graphs, putting aside that saveContext is master node specific). Looking into it it looks like this calls code only for HDRP Master Nodes that updates material dependencies/keywords (why nothing for URP atm?).
Not that this is new with this PR (although feel free to add this comment in the PR!); just stating that for future reference & maintainability reasons having a comment before these special cases / pipeline specific things explaining them would be great.

martint-unity: Nothing is future proof :) But I don't think it will touch subgraphs. But you are right a comment for the material update could probably be in order. Will do that after next bug I am working on.

PR #5505, opened 2020-01-14
Rename post formats

alelievr:
### Purpose of this PR
Renamed the the formats of post processing in the HDRP ui to be GraphicsFormat complient: 
image

Serialization should be fine since enums are saved as ints


sebastienlagarde:
Being a very small change that don't affect the code (only the UI), I will merge it, no need for QA.

sebastienlagarde:
I let Remy for awareness

PR #5506, opened 2020-01-14
Fixed resource lifecyle issue.

phi-lira:
### Purpose of this PR
 - Fixed resource lifetime issue that caused references to be lost in case of multiple renders assigned in the pipeline asset.
 - Fixed a compiler issue in VR when having URP without VR module.

---
### Testing status

- [x ] Built a player
- [x ] Checked new UI names with UX convention
- [x ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [x ] C# and shader warnings (supress shader cache to see them)
- [x ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

---
### Comments to reviewers
 Changelog is not required as this is fixing a regression in a package not released yet. (only in 2020.1)
 This will allow to backport another PR that fixed resource leak in post-processing to 2019.3


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -105,8 +105,9 @@ public ForwardRenderer(ForwardRendererData data) : base(data)
         }
 
         /// <inheritdoc />
```
Chman: Note: `<inheritdoc />` isn't accepted by our doc validation tool... So you need to copy/paste the base field doc :/

PR #5508, opened 2020-01-14
Testing/disable scene005

phi-lira:
Disabling test 005 as it's failing yamato. we opened a bug about it.
 Fixing VFX Tests.


PaulDemeulenaere:
Feel free to also disable 101_ShadergraphSampleScene : https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/testing/disable-scene005/TestProjects/VisualEffectGraph_LWRP/ProjectSettings/EditorBuildSettings.asset#L35

PR #5509, opened 2020-01-14
Hdrp/staging thin

eturquin:
Added a full REFRACTION_THIN mode, instead of using REFRACTION_BOX and overwriting Absorption Distance and Thickness values.

For QA: This could affect the refractive materials (using Box, Sphere or Thin models), although if working properly, it is not supposed to change anything. The only visible change user-wise, should be that setting the "Refraction Model" to Thin won't overwrite the "Thickness" and "Transmittance Absorption Distance" anymore.


sebastienlagarde:
Please add a changelog.
give better PR description for QA.

QA you will need to check that Refraction thin mode work correctly with shader graph Lit and Fabric (the PR change fabric but not sure if fabric support refraction?), and the insector version of lit and littesselation

com.unity.render-pipelines.high-definition/Editor/Material/Fabric/ShaderGraph/FabricRaytracingPass.template
```diff
@@ -38,6 +38,7 @@ Pass
     $Refraction:                         #define _HAS_REFRACTION 1
     $RefractionBox:                      #define _REFRACTION_PLANE 1
     $RefractionSphere:                   #define _REFRACTION_SPHERE 1
+    $RefractionThin:                     #define _REFRACTION_THIN 1
```
sebastienlagarde: Can you check if Refraction appear in the UI? we should remove refraction from Fabric

eturquin: No it did not, so I removed these.

anisunity: This does not exist anymore in the refactored version of the shaders

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -447,8 +447,13 @@ BSDFData ConvertSurfaceDataToBSDFData(uint2 positionSS, SurfaceData surfaceData)
 
 #if HAS_REFRACTION
     // Note: Reuse thickness of transmission's property set
-    FillMaterialTransparencyData(surfaceData.baseColor, surfaceData.metallic, surfaceData.ior, surfaceData.transmittanceColor, surfaceData.atDistance,
-        surfaceData.thickness, surfaceData.transmittanceMask, bsdfData);
+    FillMaterialTransparencyData(surfaceData.baseColor, surfaceData.metallic, surfaceData.ior, surfaceData.transmittanceColor,
+    #ifdef _REFRACTION_THIN
+                                 0.005, 0.005, // We set both atDistance and thickness to the same, small value
```
anisunity: Maybe a define somewhere up like the clear coat values would be a better more maintanable way?

eturquin: Done.

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -376,8 +376,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Now each camera has its own Volume Stack. This allows Volume Parameters to be updated as early as possible and be ready for the whole frame without conflicts between cameras.
 - Disable Async for SSR, SSAO and Contact shadow when aggregated ray tracing frame setting is on.
 - Improved performance when entering play mode without domain reload by a factor of ~25
-- Renamened the camera profiling sample to include the camera name
+- Renamed the camera profiling sample to include the camera name
```
sebastienlagarde: Generally speaking try to avoid to change anything in the log in a PR other that your own line. To fix the changelog we must do a separate PR (otherwise the merge with other PR start to conflict a lot on this file). ok for this PR for now, but be careful with next one.

sebastienlagarde:
> HDRP/LitTesselation shader is missing the thin mode, 
fixed in latest commit.

PR #5512, opened 2020-01-14
Alpha post-processing bugfixes and improvements

pmavridis:
This PR fixes the following issues:
- The flag to enable alpha processing was not set properly (due to an untested last-minute change)
- The alpha channel was not copied properly when chromatic aberration was enabled
- Alpha/color was blended two times when DOF was enabled, making the falloff at the alpha/non-alpha boundaries more steep than it should.


sebastienlagarde:
(note: no need for changelog as we haven't created a version with the issue).

PR #5513, opened 2020-01-14
Fixed for case 1201208.

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1201208/

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue that caused errors if you disabled the VR Module when building a project.
 - Fixed an issue where the default TerrainLit Material was outdated, which caused the default Terrain to use per-vertex normals instead of per-pixel normals.
 - Fixed shader errors and warnings in the default Universal RP Terrain Shader. [case 1185948](https://issuetracker.unity3d.com/issues/urp-terrain-slash-lit-base-pass-shader-does-not-compile)
+- Fixed an issue where partical shaders failed to handle Single-Pass Stereo VR rendering with Double-Wide Textures.
```
phi-lira: do we have an issue tracker/case id for the bug? If so let's add here. (see line above for example)

thomas-zeng: Addressed in 2822ec1

phi-lira:
Ok. LGTM.

com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl
```diff
@@ -73,7 +73,8 @@ half3 AlphaModulate(half3 albedo, half alpha)
 half3 Distortion(float4 baseColor, float3 normal, half strength, half blend, float4 projection)
 {
     float2 screenUV = (projection.xy / projection.w) + normal.xy * strength * baseColor.a;
-    float4 Distortion = SAMPLE_TEXTURE2D(_CameraOpaqueTexture, sampler_CameraOpaqueTexture, screenUV);
+    screenUV = UnityStereoTransformScreenSpaceTex(screenUV);
+    float4 Distortion = SAMPLE_TEXTURE2D_X(_CameraOpaqueTexture, sampler_CameraOpaqueTexture, screenUV);
     return lerp(Distortion.rgb, baseColor.rgb, saturate(baseColor.a - blend));
 }
```
mikesfbay: While on this particle fix, we should also update Particle's BlendTexture() which is broken by not sampling correctly from VR texture:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/2822ec18eeb88b2fa4bd5d894cee5ccce9ea83e4/com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl#L84

This is used by ParticlesLit and ParticlesUnlit.

mikesfbay: You also need to fix Particle's BlendTexture() which is also wrong and not sampling correctly from VR texture:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/2822ec18eeb88b2fa4bd5d894cee5ccce9ea83e4/com.unity.render-pipelines.universal/ShaderLibrary/Particles.hlsl#L84-L86

BlendTexture() is used by ParticlesLit and ParticlesUnlit.

thomas-zeng: Mike, BlendTexture is correct, what fix you are referring to here? I don't think it needs any fix. Could you let's me know what needs to done here?

mikesfbay: BlendTexture is a utility function, and if it's called with a camera texture as input then sampling would be wrong in VR.

thomas-zeng: BlendTexture is being used against normal/albedo/gloss/metallic maps. They are not XR textures. Your change request will break particle effects in a pretty bad way by introducing inconsistency between gfx cmd stream and shader asset.

mikesfbay: The current uses are those textures but BlendTexture() is a utility function that can be called by users with any texture including a camera texture. There should be a comment limiting input texture to non-camera textures.

DennisDeRykeUnity:
The following Yamato pipelines became green after I re-ran them: 
Universal on OSX_Metal_playmode on version fast-trunk
VFXMain on Win_DX11_editmode on version fast-trunk
Universal on OSX_Metal_Standalone on version fast-trunk

DennisDeRykeUnity:
@thomas-zeng Your dev branch crashes the Unity Editor in Play mode on Windows MR.  See the following bug report for logs and the repro case.  
https://fogbugz.unity3d.com/f/cases/1214194/

@phi-lira This is similar to behaviour I saw in one of your recent PRs.  Perhaps you and Thomas can compare notes on this while I am away.

thomas-zeng:
@DennisDeRykeUnity I was able to repro the case with this PR branch and the current URP master branch. So far I could only repro crash for Windows MR. Investigation suggests this is related to the culling system. Could you confirm this issue is caused by trunk regression and not specific to this PR? 
I have also submitted trunk PR that fixed the crash issue: https://ono.unity3d.com/unity/unity/pull-request/99916/_/fix/xr/wmr-editor-crash

@phi-lira This crash regression is introduced by trunk changes because current URP XR relies on trunk legacy XR logic. Would be nice to move to Pure XRSDK(2020.1+) asap to avoid the situation where we need trunk changes to fix URP XR

phi-lira:
Ok. Once everyone approves and validates no regression please ping me to merge.

thomas-zeng:
@DennisDeRykeUnity Trunk PR landed, could you verify this PR again?

DennisDeRykeUnity:
Per discussion with Ryan, he is going to review this PR for XR QA.

thomas-zeng:
Ryan is on vacation. 
From last week:
The distortion shader the 1201208 test project is using does not seem compatible with multiview, on Quest its just grey.  Is this something that should be fixed along with this PR?

thomas-zeng:
Ryan is on vacation this week. I will talk to XR QA lead to get some help for this PR.

Here is the Q&A between Ryan and I from last week(might be helpful for QA when testing):
Q: The distortion shader the 1201208 test project is using does not seem compatible with multiview, on Quest its just grey.  Is this something that should be fixed along with this PR?
A: In that example, android is using medium URP quality setting. And medium URP asset pipeline setting doesn't have opaque texture checked on -- without opaque texture, distortion pass won't work. Please enabled opaque texture on medium URP pipeline asset and try Quest.

Q: Your fix still has some lingering issues with this shader. There is some tearing of what kind of looks like the occlusion mesh or something still in the render.
A: This is the expected behavior. It is caused by the distortion effect sampling from the occluded area. We provide https://docs.unity3d.com/ScriptReference/XR.XRSettings-useOcclusionMesh.html to enable/disable occlusion mesh using script. so users can make the effect to work by adding the script.

thomas-zeng:
@phi-lira This PR is ready to land.

PR #5517, opened 2020-01-15
Fix colum headers

RSlysz:
### Purpose of this PR
Rename the column of the Light Explorer for HDRP lights and make color temperature use same concept (show but disabled) than the actual value of scalable settings.

---
### Testing status

**Manual Tests**: 

---
### Comments to reviewers
It is not possible to hide the value column in case of scalable settings as it display what correspond to the selected level when not editable. So here I prefer to display the color temperature but in a disabled state for coherency>


sebastienlagarde:
small change, so merging.

PR #5518, opened 2020-01-15
Cookie and planar atlas

alelievr:
### Purpose of this PR
Replace the current system of texture array by atlases for:
- 2D Cookies (Spot, Directional, Rectangle) and Surface Texture (Celestial body in the directional light).
- Realtime Planar reflections

Added more debug modes to display the cookie atlas, the point light cookie array and the planar reflection atlas:
image

Added a resolution setting in planar reflection component:
image

In the HDRP asset, there is the cookie atlas resolution field that replace the cookie cube array resolution:
image
There is also the format of the cookie atlas (default have changed from R8G8B8A SRGB to R11G11B10) and can support R16G16B16A16 for HDR cookies.  
Lastly the 2D Atlas Last Valid Mip field is used to adjust the padding between the textures in the atlas, by default it is set to 0 which means that mip 0 will be valid (1 pixel of border). The number of pixels in the border is computed with `(1 << LastValidMipLevel)`. This is only useful for area light cookies because the other lights are not using mip maps.

For planars, there is the atlas size as well:
image
And the max planar reflection on screen that remains because we need it to allocate compute buffer to their max size.

**Warning:** this branch is more than 1 year old and I have merge too many times big refactors so there might be some merge issues.


com.unity.render-pipelines.high-definition/Runtime/Lighting/LightCookieManager.cs
```diff
@@ -0,0 +1,276 @@
...
+            {
+                Debug.LogError( "FilterAreaLightTexture has an invalid shader. Can't filter area light cookie." );
+                return null;
+            }
+
+            int sourceWidth = m_CookieAtlas.AtlasTexture.rt.width;
+            int sourceHeight = m_CookieAtlas.AtlasTexture.rt.height;
+            int mipMapCount = 1 + Mathf.FloorToInt(Mathf.Log(Mathf.Max(source.width, source.height), 2));
+
+            if (m_TempRenderTexture0 == null )
```
sebastienlagarde: try to have consistent if ()  indent :)
you have if(tutu),  if (tutu ) and if ( tutu ) inside the same file, please do if (tutu)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightCookieManager.cs
```diff
@@ -0,0 +1,276 @@
...
+            {
+                string cacheName = m_CookieAtlas.AtlasTexture.name;
+                m_TempRenderTexture0 = new RenderTexture(sourceWidth, sourceHeight, 1, RenderTextureFormat.ARGB32, RenderTextureReadWrite.sRGB )
```
sebastienlagarde: as we use float format now without sRGB we shouldn't use RenderTextureFormat.ARGB32, RenderTextureReadWrite.sRGB ). We must allocated same texture format than selected by users

sebastienlagarde: same for m_TempRenderTexture1

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightEvaluation.hlsl
```diff
@@ -53,11 +53,10 @@ float3 SampleAreaLightCookie(int cookieIndex, float4x3 L, float3 F)
     //          mix of up into right that needs to be subtracted from simple projection on right vector
     //
     float   u = (dot(hitPosition, right) - upRightMixing * v) * recSqLengthRight;
-    // Currently the texture happens to be reversed when comparing it to the area light emissive mesh itself. This needs
-    // Further investigation to solve the problem. So for the moment we simply decided to inverse the x coordinate of hitUV
-    // as a temporary solution
-    // TODO: Invesigate more!
-    float2  hitUV = float2(1.0 - u, v);
+    // We need to flip the UV in x because we're emitting light from the area light and not projecting it.
```
sebastienlagarde: Better comment:
We create automatic quad emissive mesh for area light. For those to be dispplay in the direction of the light when they are single sided, we need to reverse the winding order. Because of this reverse of windows order, to get a matching area light reflection, we need to flip the x axis.
float2  hitUV = float2(1.0 - u, v);

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -3724,19 +3779,22 @@ LightLoopDebugOverlayParameters PrepareLightLoopDebugOverlayParameters()
...
             parameters.debugShadowMapMaterial = m_DebugHDShadowMapMaterial;
+            parameters.debugBlitMaterial = m_DebugBlitMaterial;
+            parameters.cookieManager = m_TextureCaches.lightCookieManager;
+            parameters.planarProbeCache = m_TextureCaches.reflectionPlanarProbeCache;
 
             return parameters;
         }
 
-        static void RenderLightLoopDebugOverlay(in DebugParameters debugParameters, CommandBuffer cmd, ref float x, ref float y, float overlaySize, RTHandle depthTexture)
+        static void RenderLightLoopDebugOverlay(in DebugParameters debugParameters, CommandBuffer cmd, ref float x, ref float y, float overlaySize, RTHandle depthTexture, MaterialPropertyBlock mpb)
```
sebastienlagarde: Do we really need a MaterialPropertyBlock mpb here?

TomasKiniulis:
Looks pretty stable!

- Changing "2D Atlas Valid Mip" values also introduces "zoom in" effect to a cookie
- Debug Exposure does not show up when only Display Point Light Cookie Array is enabled, point light cube texture should be either not affected by exposure or the option should be present

Attaching test doc: https://confluence.unity3d.com/display/HDRP/Atlas+for+cookie+2D+and+Planar+reflection

I've ran DXR tests locally on my machine. And the test results are identical to HDRP/staging
One thing to note, the default cookie atlas size must be increased or 4001, 4002, 4003 tests will fail
image (5)
101_RTReflections and 3002_AreaShadows_Denoiser are failing on HDRP/staging too

anisunity:
Something must have broken them in the mean time... i'll look into it.

PR #5523, opened 2020-01-15
Fix of warnings in Shader Graphs and Sub Graphs being treated as errors

elizabeth-legros:
### Checklist for PR maker
- [x ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
There was a bug in the Shader Graph import process which would treat warnings and errors as both being shader-breaking errors and thus a warning in the graph would cause the fallback error shader to display.
---
### Testing status

- Other: 

Added automated tests to the test project to verify warnings will not break shaders while errors still will
Added graphic test to verify material on left with error is broken and material or right with warning is okay

ShaderWarningsAndErrors


elizabeth-legros:
This is addressing https://fogbugz.unity3d.com/f/cases/1186867/

PR #5526, opened 2020-01-16
Bugfix 168283: URP shaders output incorrect alpha values

ellioman:
### Purpose of this PR
URP Shaders are currently outputting incorrect alpha values not taking into account whether the object is opaque or transparent. This causes issues when where rendering to a render texture. (Example case [#168283](https://issuetracker.unity3d.com/issues/lwrp-alpha-clipping-material-makes-other-materials-look-like-alpha-clipping-when-gameobject-is-shown-in-render-texture)).

This PR adds an OutputAlpha() function to Core.hlsl which URP shaders now call in the fragment functions.
We now also send the DrawObjectPassData to the shaders with information whether the object is opaque or alpha blended.

---
### Comments to reviewers
Test Scene: 109_URPShadersAlphaOutput
Contains a two cameras, one rendering to a render texture and another main camera showing the plane that has a material displaying the contents of that render texture. Then I have 12 planes with opaque and transparent Lit/SimpleLit/Unlit normal and particle shaders.

109_URPShadersAlphaOutput

### Checklist for PR maker

---
### Testing status

- [ ] Built a player
- Other: 

**Yamato**
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fbugfix-1168283-rendering-to-rendertexture


com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -117,6 +117,11 @@ void AlphaDiscard(real alpha, real cutoff, real offset = 0.0h)
 #endif
 }
 
+half Alpha(half outputAlpha)
```
phi-lira: Looking at code seems like we already have an Alpha function in SurfaceInput. Should we add the code there instead of adding a new function? 

Or maybe name this OutputAlpha()?

ellioman: OutputAlpha() sounds good. I have renamed the function and modified the shaders calling it.

PR #5531, opened 2020-01-16
Make APIs internal when they should be - 1 - [Skip CI]

JulienIgnace-Unity:
### Purpose of this PR
First pass of making APIs internal when they should be.

Note to reviewers:
Make sure that public APIs that I made internal should indeed not be public.


com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace UnityEditor.Rendering.HighDefinition
 {
-    public abstract class BaseShaderPreprocessor
+    abstract class BaseShaderPreprocessor
```
RSlysz: Doesn't Users need this to write their own?
If not, Why do we need an abstract base class?

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/FullScreenCustomPassDrawer.cs
```diff
@@ -13,7 +13,7 @@ namespace UnityEditor.Rendering.HighDefinition
 	/// FullScreen custom pass drawer
 	/// </summary>
 	[CustomPassDrawerAttribute(typeof(FullScreenCustomPass))]
-    public class FullScreenCustomPassDrawer : CustomPassDrawer
+    class FullScreenCustomPassDrawer : CustomPassDrawer
```
RSlysz: alignment issue?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightUtils.cs
```diff
@@ -5,7 +5,7 @@ namespace UnityEngine.Rendering.HighDefinition
     /// <summary>
     /// Light Utils contains function to convert light intensities between units
     /// </summary>
-    public class LightUtils
+    internal class LightUtils
```
RSlysz: Note: Internal is the default visibility for class directly within namespace (not nested). (The default is private if nested). So you can remove the keyword.

com.unity.render-pipelines.high-definition/Runtime/Debug/LightingDebug.cs
```diff
@@ -33,7 +33,7 @@ public enum DebugLightFilterMode
         IndirectPlanarProbe = 1 << 8,
     }
 
-    public static class DebugLightHierarchyExtensions
+    internal static class DebugLightHierarchyExtensions
```
RSlysz: Note: Internal is the default visibility for class directly within namespace (not nested). (The default is private if nested). So you can remove the keyword.

com.unity.render-pipelines.high-definition/Runtime/Debug/MaterialDebug.cs
```diff
@@ -66,7 +66,7 @@ public enum MaterialSharedProperty
             Alpha,
         }
 
-        public class MaterialSharedPropertyMappingAttribute : Attribute
+        internal class MaterialSharedPropertyMappingAttribute : Attribute
```
RSlysz: Note: Internal is the default visibility for class directly within namespace (not nested). (The default is private if nested). So you can remove the keyword.

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
JulienIgnace-Unity: We have several different preprocessors for all our shaders, that's why there is a base abstract class. But we don't expect users to use the same basis a us.

com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/FullScreenCustomPassDrawer.cs
JulienIgnace-Unity: Nope, just removed the public specifier.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightUtils.cs
JulienIgnace-Unity: Yeah I usually do it right :p
Must have been caught in a "internalization" spree of members and did it mechanically. Will fix other occurrences.

com.unity.render-pipelines.high-definition/Editor/Material/BaseShaderPreprocessor.cs
sebastienlagarde: hmm... currently we want, but as all this is not super well define let's not expose it

PR #5538, opened 2020-01-17
Shader Graph: Color Properties to have IsGammaSpace() check

zadi0648:
**Summary:**

Fix for https://fogbugz.unity3d.com/f/cases/1184248/
Presumably Zack's last fix for the January Quality Drive

* Was caused by the fact that ColorNodes had a 'IsGammaSpace() ?' check and the Color Properties did not.
* The bug thinks that the Inline node was incorrect. That is incorrect; Inline had it right: they were using a very intense color node (may want to explain that to them when we close the bug).

---
**To Optimize, or To Not Optimize..?:**

Feels like we should change both this and the Color node to use https://docs.unity3d.com/ScriptReference/QualitySettings-activeColorSpace.html
If we find a way to cause the Shaders to rebuild if the setting is changed.
_Discussion:_ https://unity.slack.com/archives/G7HQCPV71/p1579226248050000

My current thought is that we should perhaps do that for 2020, but backport this change to 19.3

---
**Manually Tested:**

* That the bug no longer reproduces
* My other SGs still seem to work fine
* Checked this in sub graphs real quick

---
**Halo Effect:** 2/4: This does affect color properties everywhere...
**Technical Risk:** 0/4: Simple single line change that other nodes are already using

---
**Notes to QA:**

* I was going to make a build of this, I didn't get around to it tonight (whether or not that's important, I have no reason to think so... but who knows)

---
**Yamato:**
...

### DIFF
--- a/com.unity.shadergraph/Editor/Data/Nodes/Input/PropertyNode.cs
+++ b/com.unity.shadergraph/Editor/Data/Nodes/Input/PropertyNode.cs
@@ -139,7 +139,7 @@ namespace UnityEditor.ShaderGraph
                     sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = {property.referenceName};");
                     break;
                 case PropertyType.Color:
-                    sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = {property.referenceName};");
+                    sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = IsGammaSpace() ? {property.referenceName} : float4(SRGBToLinear({property.referenceName}));");
                     break;
                 case PropertyType.Matrix2:
                     sb.AppendLine($"$precision2x2 {GetVariableNameForSlot(OutputSlotId)} = {property.referenceName};");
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Data/Nodes/Input/PropertyNode.cs
```diff
@@ -140,7 +140,7 @@ public void GenerateNodeCode(ShaderStringBuilder sb, GenerationMode generationMo
                     sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = {property.referenceName};");
                     break;
                 case PropertyType.Color:
-                    sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = {property.referenceName};");
```
zadi0648: Here

com.unity.shadergraph/Editor/Data/Nodes/Input/PropertyNode.cs
```diff
@@ -140,7 +140,7 @@ public void GenerateNodeCode(ShaderStringBuilder sb, GenerationMode generationMo
...
+                    sb.AppendLine($"$precision4 {GetVariableNameForSlot(OutputSlotId)} = IsGammaSpace() ? {property.referenceName} : SRGBToLinear({property.referenceName});");
```
zadi0648: Here

zadi0648:
With removals: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fcolor-property-to-match-inline/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1746013/job/pipeline

PR #5545, opened 2020-01-17
Fix custom pass disable stencil

alelievr:
### Purpose of this PR
+ Custom passes now disable the stencil when overwriting the depth and not writing into it.
+ Update the documentation to add a table of allowed materials per injection point

This fixes a potential issues where a material that writes to the stencil (SSS for example) when rendered via a custom pass with an override depth and no depth write would mark the stencil bits affecting objects that shouldn't have the effect. This can happen when rendering to the custom buffer and binding the camera depth buffer (to have depth test with the scene).


com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass.md
```diff
@@ -109,7 +109,18 @@ Filters allow you to select which objects will be rendered, you have the queue w
...
 By default, the objects are displayed with their material, you can override the material of everything in this custom pass by assigning a material in the `Material` slot. There are a bunch of choices here, both unlit ShaderGraph and unlit HDRP shader works and additionally there is a custom unlit shader that you can create using **Create/Shader/HDRP/Custom Renderers Pass**.
 
-> **Note that Lit Shaders aren't supported on every injection point as they require the lighting data to be ready.**
+**⚠️ Note that not all kind of materials are supported by every injection point. Here is the compatibility table:**
+
+Injection Point               | Material Type
+----------------------------- | -------------------------------------------
+Before Rendering              | Unlit forward but without writing to the camera color
+After Opaque Depth And Normal | Unlit forward
+Before PreRefraction          | Unlit + Lit forward only
```
sebastienlagarde: it is not Lit material, it is "all forward material with lighting". (i.e stacklit, lit, fabric etc...)

sebastienlagarde: guess we should first define:
Set of material forward with lighting: ...

then in the doc use that as reference

sebastienlagarde: tagging @JordanL8

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/DrawRenderersCustomPass.cs
```diff
@@ -110,9 +110,13 @@ protected override void Execute(ScriptableRenderContext renderContext, CommandBu
                 return;
             }
 
-            var stateBlock = new RenderStateBlock(overrideDepthState ? RenderStateMask.Depth : 0)
+            var mask = overrideDepthState ? RenderStateMask.Depth : 0;
+            mask |= overrideDepthState && !depthWrite ? RenderStateMask.Stencil : 0;
+            var stateBlock = new RenderStateBlock(mask)
             {
                 depthState = new DepthState(depthWrite, depthCompareFunction),
+                // We disable the stencil when the depth is overwritten but we don't write to it, to prevent writing to the stencil.
```
sebastienlagarde: hmm... we should maybe wait for the PR of Francesco then update all custom pass code to allow to write in bit 6 and 7 of the stencil? Let's talk

PR #5548, opened 2020-01-17
Refactor of HDCamera (accesibility mostly) [Skip CI]

JulienIgnace-Unity:
### Purpose of this PR
The goal of this PR is to make everything that should not have been public, either internal or private and thus define a clearer public API for HDCamera
- Made internal/private APIs that should not have been public
- Fixed naming convention for private members
- Moved everything by accessibility for clarity (public, internal, then private)

Note to reviewer:
The goal is to assess whether or not APIs I made internal should stay public.
The question we need to ask is "Does a user writing code *outside* of HDRP need APIs XXX in a relevant way?".
Don't forget that for now, HDRP extensibility is almost non existent beside Custom Passes, Custom Post processes and Custom Render so most APIs are indeed internal (for example, Volumetric parameters can only be used inside HDRP and don' t make sense exposed to the users).

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -15,6 +15,7 @@ namespace UnityEngine.Rendering.HighDefinition
     // not used during a frame.
     public class HDCamera
     {
+        #region Public API
```
sebastienlagarde: as a reminder, Region aren't in our code convention chart. (I know there is some in our package, still want to note it)

glabute:
Cinemachine is using HDCamera like this, please don't make it break:

#if CINEMACHINE_HDRP_7_0_0
            // Reset temporal effects
            var cam = brain.OutputCamera;
            if (cam != null)
            {
    #if CINEMACHINE_HDRP_7_1_0
                HDCamera hdCam = HDCamera.GetOrCreate(cam);
    #else
                HDCamera hdCam = HDCamera.GetOrCreate(cam, new XRPass());
    #endif
                hdCam.volumetricHistoryIsValid = false;
                hdCam.colorPyramidHistoryIsValid = false;
                hdCam.Reset();
            }
#endif

Chman:
Hmm both `volumetricHistoryIsValid` and `colorPyramidHistoryIsValid` have been moved to the internal namespace. @JulienIgnace-Unity these two fields should probably be part of `Reset()` with all the other history resets.

glabute:
I agree that Reset() should handle that, but if you change the public API you will break Cinemachine, so please leave it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
JulienIgnace-Unity: Yes but honestly without that it's really hard to navigate in there.

PR #5552, opened 2020-01-17
Disable the new input system for runtime debug menu

alelievr:
### Purpose of this PR
Put the debug menu input code between `ENABLE_LEGACY_INPUT_MANAGER` define.

Added an error message in the debug menu window to warn about debug menu not being available in build when using the new input system
image

#### QA
There should be no more errors when enabling the new input system package in the player settings and in the build we should not be able to open it since it does not work yet.  
I tested with the input mode both and the debug menu works fine in both build and editor.


com.unity.render-pipelines.core/Editor/Debugging/DebugWindow.cs
```diff
@@ -377,6 +381,10 @@ void OnGUI()
                 DebugManager.instance.Reset();
             GUILayout.EndHorizontal();
 
+#if !ENABLE_LEGACY_INPUT_MANAGER
```
sebastienlagarde: we shouldn't test for the non prescence of legacy input system here, but more for the prescence of new input system. ISn't there a define for it? (guess if you haven't done it is because it is not define?

sebastienlagarde: OR maybe we can rephrase the sentence and say that debug menu only work with LEGACY_INPUT_MANAGER. It make more sense with current test.

alelievr: There is the `ENABLE_INPUT_SYSTEM` define which is only here if the new debug menu is enabled but because we can have both new and old system at the same time just checking if the new one is here doesn't work (debug menu works correctly in `both` input settings).

sebastienlagarde: ok, can you add that as a comment, so we know why :)

PR #5554, opened 2020-01-17
Fix dispatcher to only perform remove Camera if no component depends on it

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1206823/
(It occures when you first remove the camera and then a component that depend on it with the [RequireComponent] attribute)
This was productible in:
- HDRP
- Universal
- built-in if CoreRP package was installed

---
### Testing status

**Manual Tests**: 
- Other: 
    - Checked that the returned dependency collection match what is actually on the GameObject with attribute targeting the Camera.
    - Checked in HDRP that the same windows open if a FlareLayer is on the GameObject and trying to remove the Camera than in built-in (without CoreRP package)
    - Checked built-in with CoreRP package that the same windows open if a FlareLayer is on the GameObject and trying to remove the Camera than in built-in (without CoreRP package)

---
### Comments to reviewers
I have not checked Universal but the code is similar than HDRP


RSlysz:
> 
> 
> Approve subject to docs change - let me know if I incorrectly summarised the bug.

@reallyfancy It miss an information.
How about:
Fixed a bug where Unity would crash if you tried to remove a Camera component from a GameObject using the Inspector window, while other components dependended on the Camera component and CoreRP package is installed.

By the way, does changelog must be user oriented or technically oriented?
Should I change the 2 other changelog (HDRP and Universal ones) to match this change?

PR #5557, opened 2020-01-17
Hdrp/smallfixes wizard dxr

RSlysz:
### Purpose of this PR
Fix https://fogbugz.unity3d.com/f/cases/1211987/
The issue was due to wrong reference in new default DXR scene. I dunno exactly how it happened but GUID registered here are not the right ones.

---
### Testing status

**Manual Tests**: 
- Other: Run the wizard again and checked default prefab scene

---


iM0ve:
Also while not a bug, I am unsure why we store separate Sky files for DXR and non DXR. Since sky does not have raytracing specific features, this profile should be shared. And because it's not shared now, we already have some inconsistencies. DXR template does not have fog while regular does. Upgrading them separately might cause more issues like this in the future.

DXR sky
image

Non DXR sky
image

RSlysz:
PR linking this one will redo these changes and add the fix on bug spited by @iM0ve

PR #5561, opened 2020-01-17
Merge bugfix for case 1181518

parashar-unity:
### Purpose of this PR
Fix for issue #1181518 -- https://fogbugz.unity3d.com/f/cases/1181518/
The issue appears to indicate that the X-component of the normal-mapped result on terrain vs meshes using the same normal mapping results in lighting that appears flipped on the X-axis.  More accurately, it is that the tangent direction is flipped.

---
### Testing status
Test scene result.

**Buggy State** 
Broken1181518

**Fixed State**
Fixed1181518

Also verified via another test case that this seems to affect Universal, but not HDRP.

---
### Comments to reviewers
Also needs backport to 2019.3


ellioman:
Approved but we need to run Yamato again and make sure it goes green.

PR #5553, opened 2020-01-17
Shader Graph memory leak plumbing

pbbastian:
### Checklist for PR maker
- [x] Have you added a backport label (if needed)? For example, the `need-backport-2019.3` label. After you backport the PR, the label changes to `backported-2019.3`.
- [x] Have you updated the changelog? Each package has a `CHANGELOG.md` file.

---
### Purpose of this PR
Fix a significant memory leak when closing Shader Graph windows.

---
### Testing status

**Manual Tests**: What did you do?
- [x] Opened test project + Run graphic tests locally

- Other: 
Extensive QA by Gfx QA and close review with MemoryProfiler package.

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

PR #5563, opened 2020-01-18
[Backport] Shader Graph memory leak plumbing

marctem:
7.1.8 backport for https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5553

Removes one change from that PR which was about using less memory but did not impact the leak. Good stuff but not critical.


LandonTownsendUnity:
Confirmed to be fixed and not cause additional bugs when used with the corresponding fix branch on the engine.

PR #5569, opened 2020-01-20
Move platform specific files to separate package

sebastienlagarde:
Move platform specific files (Xbox and PS4) to separate packages.


phi-lira:
This PR is removing the whole texture abstraction / common instrisics from Common.hlsl for those platforms. 
What happens in URP case? URP doesn't use in stock shaders wave instrinsics. It compiles fine without platform specific code, yet this change not only removes platform specific code but also the entire texture / common intrinsics abstraction and if I understand correctly will cause URP to fail compilation with PS4 and XboxOne.

Could you confirm this @sebastienlagarde @stramit ?

PR #5579, opened 2020-01-21
Redesign the recursive rendering queue system to make it work when the effect is disabled

anisunity:
This PR is a major refactor of how the queue system behaves so that transparent objects behave correctly (meaning they appear) when the recursive rendering effect is disabled and they are flagged with ray tracing on. First, the option is not displayed in the rendering pass but on top. Also it only appears if support ray tracing is enabled on the current asset.

image
image

This PR depends on a migration feature that @FrancescoC-unity is working on due to render queue upgrade breaking.

It clearly needs extensive testing.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
```diff
@@ -329,6 +343,24 @@ void PrepareForwardPassData(RenderGraph renderGraph, RenderGraphBuilder builder,
                 ReadDBuffer(dbuffer.Value, builder);
         }
 
+        void PrepareTransparentForwardPassData(RenderGraph renderGraph, RenderGraphBuilder builder, TransparentForwardPassData data, FrameSettings frameSettings, RendererListDesc rendererpreRefractionListDesc, RendererListDesc rendererListDesc, RendererListDesc rendererLowTransparentListDesc, RenderGraphMutableResource depthBuffer, ShadowResult shadowResult, DBufferOutput? dbuffer = null)
```
JulienIgnace-Unity: Can be changed in a subsequent PR but the way it's implemented is problematic:
This is called several times so we create and use multiple time renderer list that might not be used depending on the frame settings.
A better way will be to prepare all renderer list once and pass them to the various required function.
Please add this as a comment.

alelievr:
Just to be sure, is the only master node concerned by these changes is the Lit one ? no Unlit / PBR / another HDRP master node ?

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
```diff
@@ -357,6 +357,9 @@ Shader "HDRP/LayeredLit"
         [ToggleUI] _LinkDetailsWithBase2("LinkDetailsWithBase2", Float) = 1.0
         [ToggleUI] _LinkDetailsWithBase3("LinkDetailsWithBase3", Float) = 1.0
 
+        // Ray Tracing
+        [ToggleUI] _RayTracing("Ray Tracing", Float) = 0
```
sebastienlagarde: Please add ""Ray Tracing (Preview)"

everywhere :), thanks

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderQueue.cs
```diff
@@ -20,30 +20,33 @@ public enum Priority
...
             LowTransparent = 3400,
             LowTransparentLast = 3400 + k_TransparentPriorityQueueRange,
+            LowTransparentRayTracing = LowTransparentLast + 10,
 
             AfterPostprocessTransparentFirst = 3700 - k_TransparentPriorityQueueRange,
             AfterPostprocessTransparent = 3700,
             AfterPostprocessTransparentLast = 3700 + k_TransparentPriorityQueueRange,
 
-            RaytracingTransparent = 3900,
+            OldRaytracingTransparent = 3900,
```
sebastienlagarde: Add a comment that it is obsolete and can be remove in 2020.2

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderQueue.cs
```diff
@@ -71,21 +79,27 @@ public enum RenderQueueType
 
         public static readonly RenderQueueRange k_RenderQueue_OpaqueNoAlphaTest = new RenderQueueRange { lowerBound = (int)Priority.Background, upperBound = (int)Priority.OpaqueAlphaTest - 1 };
         public static readonly RenderQueueRange k_RenderQueue_OpaqueAlphaTest = new RenderQueueRange { lowerBound = (int)Priority.OpaqueAlphaTest, upperBound = (int)Priority.OpaqueLast };
-        public static readonly RenderQueueRange k_RenderQueue_AllOpaqueRaytracing = new RenderQueueRange { lowerBound = (int)Priority.RaytracingOpaque, upperBound = (int)Priority.RaytracingOpaque };
-        public static readonly RenderQueueRange k_RenderQueue_AllOpaque = new RenderQueueRange { lowerBound = (int)Priority.Background, upperBound = (int)Priority.OpaqueLast };
+        public static readonly RenderQueueRange k_RenderQueue_OpaqueRayTracing = new RenderQueueRange { lowerBound = (int)Priority.OpaqueRayTracing, upperBound = (int)Priority.OpaqueRayTracing };
+        public static readonly RenderQueueRange k_RenderQueue_AllOpaque = new RenderQueueRange { lowerBound = (int)Priority.Background, upperBound = (int)Priority.OpaqueRayTracing };
```
sebastienlagarde: I need to check what are the consequence of this one. k_RenderQueue_AllOpaque

sebastienlagarde: maybe we need also a k_RenderQueue_AllOpaqueNoRayTracing

anisunity: where would that be needed? @sebastienlagarde

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderQueue.cs
anisunity: Not sure to understand, also we need to discuss how to do it with the migration

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
anisunity: done

anisunity:
> Just to be sure, is the only master node concerned by these changes is the Lit one ? no Unlit / PBR / another HDRP master node ?

Yes!

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.RenderGraph.cs
anisunity: Will do

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderQueue.cs
sebastienlagarde: I will perform change, there is something wrong here, I will explain Monday

anisunity:
@sebastienlagarde  closing this PR given the discussion we had

PR #5592, opened 2020-01-22
Fix decal orientation when created from script.

JulienIgnace-Unity:
### Purpose of this PR
Fix decal orientation when created from script.

The issue was caused by migration code being called when it should not have been because the "current" version for decals was not the latest one.
Had to update the contextual creation of decals to include the 90° rotation to match the old convention.
Fixed a lot of other migration classes to the right latest version as well.

Fixes: https://fogbugz.unity3d.com/f/cases/1211479/

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Editor/DefaultScene/HDProjectSettings.cs
```diff
@@ -18,7 +19,7 @@ enum Version
         }
 #pragma warning disable 414 // never used
         [SerializeField]
-        Version version = Version.First;
+        Version version = MigrationDescription.LastVersion<Version>();
```
sebastienlagarde: is it expected here ?  is it overwrite by the serialized value from the file?

PR #5593, opened 2020-01-22
Shader Graph: Inline Color Node Fix for Switching between Modes (Default & HDR)

zadi0648:
**Summary:**

* Fix for https://fogbugz.unity3d.com/f/cases/1185260/

---
**Implementation Notes:**

* Removed code that was trying to convert from one value type to another unnecessarily: Unity already handles this.
* Without this code, it is consistent with how Color Blackboard Properties work as well as switching from HDR to Default in Unity outside of Shader Graph.

---
**Manually Tested:**

* The bug no longer repros.
* What Color Blackboard Properties do: it's now consistent.
* Switching between HDR & Default in a non-SG Shader.

---
**Halo Effect:** 1/4: only affects this node.
**Technical Risk:** 3/4: I don't have a reason to think so, but changing how things are converted could cause problems.

---
**Notes to QA:**

* Since the removed code only triggered when a Mode is switched, there shouldn't be a problem with opening up pre-existing Shaders that used the Color Node before this change. May be good to double check.

---
**Yamato:**
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Finline-color-node-hdr-to-default-fix


zadi0648:
Going to pick this back up during next bug bash week, if not sooner.

zadi0648:
I guess I didn't need to close this

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -84,7 +84,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a bug where adding a HDRP Master Node to a Shader Graph would softlock the Shader Graph.
 - Fixed a bug where shaders fail to compile due to `#pragma target` generation when your system locale uses commas instead of periods.
 - Fixed a compilation error when using Hybrid Renderer due to incorrect positioning of macros.
-- Fixed a bug with the `Transform` node where converting from `Absolute World` space in a sub graph causes invalid subscript errors. [1190813](https://issuetracker.unity3d.com/issues/shadergraph-invalid-subscript-errors-are-thrown-when-connecting-a-subgraph-with-transform-node-with-unlit-master-node)
+- Fixed a bug with the `Transform` node where converting from `Absolute World` space in a sub graph causes invalid subscript errors. [1190813](https://issuetracker.unity3d.com/issues/
+- Fixed a bug where switching from Color Nodes' Mode from Default to HDR or vice-versa would cause the Color to be altered incorrectly.
```
alindmanUnity: merge error? looks like you deleted the hyperlink in the entry above yours.

zadi0648: Good catch!

zadi0648:
📗 Green: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Finline-color-node-hdr-to-default-fix/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1708425/job/pipeline

LandonTownsendUnity:
Added needs backport tags. If there is any disagreement let me know, but I made some errors in the bug case's description and "package found version" that I just corrected, which may have lead to Zack not thinking backports were needed.

PR #5594, opened 2020-01-23
Fixed an issue where creating decals at runtime could cause a null ref exception.

JulienIgnace-Unity:
### Purpose of this PR
Fixed an issue where creating decals at runtime could cause a null ref exception.
This was caused by a different code path for object creation happening at runtime compared to the editor. In this case, a decal with null material was first registered but then never removed from internal structures.

This should fix two issues:
https://fogbugz.unity3d.com/f/cases/1211760/
https://fogbugz.unity3d.com/f/cases/1201697/

---
### Testing status

- [ ] Built a player


TomasKiniulis:
Checked both projects from reports. Not encountering any null ref or crashing issues anymore

PR #5596, opened 2020-01-23
Fixed a null reference exception in the HDRI Sky editor when no HDRP asset is present

JulienIgnace-Unity:
### Purpose of this PR
Fixed a null reference exception in the HDRI Sky editor when no HDRP asset is present

---
### Testing status

- [ ] Built a player


com.unity.render-pipelines.high-definition/Editor/Sky/HDRISky/HDRISkyEditor.cs
```diff
@@ -62,7 +62,8 @@ public override void OnEnable()
 
             m_IntensityTexture = RTHandles.Alloc(1, 1, colorFormat: GraphicsFormat.R32G32B32A32_SFloat);
             var hdrp = HDRenderPipeline.defaultAsset;
-            m_IntegrateHDRISkyMaterial = CoreUtils.CreateEngineMaterial(hdrp.renderPipelineResources.shaders.integrateHdriSkyPS);
+            if (hdrp != null)
+                m_IntegrateHDRISkyMaterial = CoreUtils.CreateEngineMaterial(hdrp.renderPipelineResources.shaders.integrateHdriSkyPS);
```
sebastienlagarde: we should do this for all sky type rights?

PR #5601, opened 2020-01-23
Fix disk change check

pbbastian:
### Purpose of this PR
Fixes FB-1198885. The issue is that when the user has an open Shader Graph that is modified, external changes to Sub Graphs in use involving input/output will cause the main Shader Graph to think that the asset changed on disk, and thus ask if the user wants to reload the graph. 

I have fixed this by keeping around the original JSON string from last disk access -- i.e. opening the window, or saving the asset -- and then using that to check for disk changes.

---
### Testing status

- Repro the issue without the fix
- Verify that the issue does not repro with the fix
- Actual file changes still yields the dialog

**Automated Tests**: UI only changes, no tests

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
+++ b/com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
@@ -25,6 +25,9 @@ namespace UnityEditor.ShaderGraph.Drawing
         string m_Selected;
 
         [SerializeField]
+        string m_DiskJson;
+
+        [SerializeField]
         GraphObject m_GraphObject;
 
         [NonSerialized]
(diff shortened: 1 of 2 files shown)


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -262,10 +270,17 @@ bool IsDirty()
 
         public void CheckForChanges()
         {
-            var isDirty = IsDirty();
-            if (isDirty)
+            var fileJson = File.ReadAllText(AssetDatabase.GUIDToAssetPath(selectedGuid));
+            bool isDirty;
+            if (string.CompareOrdinal(fileJson, m_DiskJson) != 0)
```
zadi0648: For maintainability could be good to have a comment here explaining why there are two different kinds of comparisons being done

PR #5604, opened 2020-01-23
Fix ray tracing tests and add a test scene for lights and cookies in ray tracing

anisunity:
Apparently the HDNoise texture file was deleted. Also adding a test scene to try to capture problems with lights and cookies in ray tracing


sebastienlagarde:
Already tested by @RemyUnity . Need for Antoine for testing

PR #5609, opened 2020-01-23
Clear shader error cache every time we save the graph

marctem:
This is a partial workaround for an AssetDb behavior change. It keeps old error messages from re-appearing on the material inspector after you fix an issue with a Custom Function Node. See https://fogbugz.unity3d.com/f/cases/1196304/.

There are still instances where this issue will reproduce, specifically with CFNs in subgraphs or when importing a graph from the cache server. Fixing those will require a more significant change but this is low-risk and eliminates the most common cause of this issue.

*NO CHANGELOG* - since this doesn't fully fix the bug, I don't want to highlight it.

### Checklist for PR maker

---
### Purpose of this PR
Partial fix for https://fogbugz.unity3d.com/f/cases/1196304

---
### Testing status

- Other: 
Created and cleared errors on custom nodes in several scenarios.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fclear-errors-on-save

---
### Comments to reviewers
As discussed above, this does not solve the specific repro steps in the bug (with subgraphs) but it does fix the following repro:
1. Create Shader Graph
2. Add Custom Function Node, add an output port named "Out", and wire it to the master node
3. In "String" mode, add a non-compiling piece of code (I like to use `Out = 0.6; a` so I can easily clear the error)
4. Save the graph, then select the graph asset in the Project Browser
5. Observe the error appears in the inspector
6. Back in the graph, fix the error in the CFN so the error badge in the graph goes away
7. Save the graph, then select the graph in the Project Browser
8. Observe the error is gone
9. Select another asset, then re-select the graph
10. Observe the error is back.

### DIFF
--- a/com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
+++ b/com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
@@ -341,6 +341,10 @@ namespace UnityEditor.ShaderGraph.Drawing
 
                 ShaderGraphAnalytics.SendShaderGraphEvent(selectedGuid, graphObject.graph);
 
+                var oldShader = AssetDatabase.LoadAssetAtPath<Shader>(path);
+                if (oldShader != null)
+                    ShaderUtil.ClearShaderMessages(oldShader);
+
                 UpdateShaderGraphOnDisk(path);
 
                 if (GraphData.onSaveGraph != null)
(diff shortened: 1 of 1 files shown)


com.unity.shadergraph/Editor/Drawing/MaterialGraphEditWindow.cs
```diff
@@ -349,7 +353,7 @@ public void UpdateAsset()
                     if (shader != null)
                     {
                         GraphData.onSaveGraph(shader, (graphObject.graph.outputNode as MasterNode).saveContext);
-                    }                    
+                    }
```
alindmanUnity: do we care about the random whitespace here?

marctem: Dang Rider. I deleted that before committing and it added it back. 🙄 I've been not caring about a little bit in most PRs.

PR #5612, opened 2020-01-24
Merge Hdrp/fix/case 1210058 to master

parashar-unity:
### Purpose of this PR
Fix for case 1210058 --  The issue was presumed to be related to SpeedTree, but is really a generic problem caused by INFs/NaNs in the derivatives on sub-pixel geometry.  The quick fix for this is to put some safeguards to catch INF/NaN when calculating basis spaces.

---
### Testing status
Using test cases and assets provided in the bug case, verified that the bug was fixed and did not hit on any additional corner cases.

This issue shows up specifically on the provided SpeedTree assets in HDRP because SpeedTree can have thin geometry (e.g. branches) and has no specific HDRP shader at this time and so it is using the standard HDRP Lit shader.  In theory, though, it probably actually could affect anything.  While the NaN may only appear on one or two pixels at render time and be innocuous.  However, because of the post-processes (in this case, Bloom), it gets spread to the entire view.

RenderDoc captures of first 4 Bloom filter passes showing the NaN spreading with each pass --
BloomPass1NaN
BloomPass2NaN
BloomPass3NaN
BloomPass4NaN

After adding the SanitizeFinite calls and the SafeNormalize, we get the fixed result --
Fixed1210058

---
### Comments to reviewers
There is an ifdef guard for the SanitizeFinite calls because of the fact that its definition in Common.hlsl is similarly guarded.

See https://fogbugz.unity3d.com/f/cases/1210058/ for the asset used in testing.
The example project does not seem to function, but at least the tree asset located in the .unitypackage file does.  For me to test, I had to create an empty HDRP test project and import the tree asset from the package.  The bug would show up intermittently when zooming out and circling the camera around at both LOD1 and LOD2.


sebastienlagarde:
Adding @EvgeniiG and @mmikk for validation of this change. Seems overkill

parashar-unity:
Initially, I tried it with just the SafeNormalize alone, and that fixed 90% of the instances of the bug, but I couldn't get to 100% elimination until I added the extra SanitizeFinite lines.

parashar-unity:
Tried it just now.  As expected, it doesn't serve as a complete fix.  Fixes the vast majority of the cases, but I can still repro the bug in some cases so long as it is just the SafeNormalize.

SafeNormalize really only protects against the cases where you've got a near divide-by-zero issue, which is the most common form in which the bug appears because you can get some derivatives that are finite, but really enormous (e.g. 1.E+37 or something).  But it doesn't help when you've already got INFs or NaNs in vT before you get to the point of normalizing.

EvgeniiG:
The NaN is generated because the texture coordinates passed to the function are invalid.
It appears the shader is using UV1, even though I don't see for what purposes. These UVs are broken.
The bug does not appear to have anything to do with the Surface Gradient code.

image

parashar-unity:
Well, it's probably using texcoord1 because it's active on the asset.  The debug view inside the fragment shader might look that way, but when I look through all the texcoord1 channel data in VS_IN and VS_OUT from the vertex shaders, there's no instance of that sort of nonsense anywhere.  It's more likely that the register otherwise mapped to texcoord1 is just intermittently being overwritten during the course of execution of that code.

mmikk:
Well that basically just confirms what Evgenii is saying. Since you're getting bad results on the screen we know it's using a fragment shader that is using tangent and bitangent on uv1 (as confirmed by Evgenii). The other end which you confirmed on your end is the vertex shader is not trying to produce a uv1 so it is garbage which matches Evgenii's operation. So the point Evgenii is making is the real bug is why is it using a fragment shader that's trying to normal map with uv1 when there is no uv1.

parashar-unity:
What I mean is not that the vertex shader isn't _trying_ to produce a uv1 -- but that it isn't producing any _bad_ uv1 values.  Looking through all the VS_IN and VS_OUT shows that there isn't garbage anywhere in those results, but there appears to be when it gets to the fragment shader.

As for why it is trying to compute basis using uv1, Lit Shader _always_ does that 
If you look at litdata.hlsl and in GenerateLayerTexCoordBasisTB, there is always at least a call to SurfaceGradientGenBasisTB using uv1, (potentially uv2 and uv3 as well, but uv1 is certain as long as you are using surface gradients).

mmikk:
Yes but calling it is not the same as it being used to produce a result. That was to reduce the number of keywords I think. It can be called without the result being used in any subsequent calculation that leads to the value that is ultimately written to the frame buffer.

parashar-unity:
Actually, if you look through litdataindividuallayer, you'll see that, in a sense, it basically always uses them.  Lines like the following --

    ADD_IDX(layerTexCoord.base).bitangentWS =   uvMappingMask.x * layerTexCoord.vertexBitangentWS0 +
                                                uvMappingMask.y * layerTexCoord.vertexBitangentWS1 +
                                                uvMappingMask.z * layerTexCoord.vertexBitangentWS2 +
                                                uvMappingMask.w * layerTexCoord.vertexBitangentWS3;

Basically means that even if the mask weight is zero, NaN/INF in any of the vertexTangents or vertexBitangents vertex will still screw the pooch.

mmikk:
Okay, I haven't seen that before but it looks like that is the real problem then. I'd argue a better course of action would be to do an explicit if() statement so we don't accumulate the bitangentWS  for uv1 unless uvMappingMask.y>0.0 and then for the other two we could use _REQUIRE_UV2 and _REQUIRE_UV3 similar to how it's done in GenerateLayerTexCoordBasisTB().

mmikk:
Or alternatively inside GenerateLayerTexCoordBasisTB() nullify vertexTangentWS1 and vertexBitangentWS1 there if there's no uv1. Perhaps that's even better.

EvgeniiG:
Correctness aside, keeping all UVs alive at all times is really bad from the perf point of view. This needs to be fixed.

@sebastienlagarde I can take a look next week if you like.

mmikk:
Another observation here is it would have been more practical to pack uv1 together with uv0. That way we could have prevented the situation where it becomes garbage (it would have been zero).

mmikk:
Sebastien since we are not spending a keyword on whether or not there’s a uv1 this means the fragment shader (referring to the uvMappingMask stuff) is essentially assuming there is always a uv1. In this case it is a requirement we also ensure it cannot deliver garbage ever. It has to pass through (0,0) when there’s no uv1 for this to work.

sebastienlagarde:
Hey,
so yes current code is to deal with the crazyness of Keyword. We have already too much of them, and we can't add new one for this purpose (in particular as Layered Lit also belong to this path).

In addition, uv1 is the uv reseved for lightmap (lightmap static always use uv1 and lightmap dynamic (enlighten) use uv2). Using uv1 here can result in weird result if you are using static batching, as the static batcher will perform operation on it thinking it is a lightmap.

>all UVs alive at all times is really bad from the perf point of view
Not exactly, we already have
#pragma shader_feature_local _ _REQUIRE_UV2 _REQUIRE_UV3
We haven't done it for UV1 to avoid the extra complexity and save keyword as said above, as using UV1 is very frequent due to lightmap. Using UV2 and UV3 is less frequent.

Which do:
#ifdef SURFACE_GRADIENT
void GenerateLayerTexCoordBasisTB(FragInputs input, inout LayerTexCoord layerTexCoord)
{
    float3 vertexNormalWS = input.tangentToWorld[2];

    layerTexCoord.vertexTangentWS0 = input.tangentToWorld[0];
    layerTexCoord.vertexBitangentWS0 = input.tangentToWorld[1];

    float3 dPdx = ddx_fine(input.positionRWS);
    float3 dPdy = ddy_fine(input.positionRWS);

    float3 sigmaX = dPdx - dot(dPdx, vertexNormalWS) * vertexNormalWS;
    float3 sigmaY = dPdy - dot(dPdy, vertexNormalWS) * vertexNormalWS;
    //float flipSign = dot(sigmaY, cross(vertexNormalWS, sigmaX) ) ? -1.0 : 1.0;
    float flipSign = dot(dPdy, cross(vertexNormalWS, dPdx)) < 0.0 ? -1.0 : 1.0; // gives same as the commented out line above

    // TODO: Optimize! The compiler will not be able to remove the tangent space that are not use because it can't know due to our UVMapping constant we use for both base and details
    // To solve this we should track which UVSet is use for normal mapping... Maybe not as simple as it sounds
    SurfaceGradientGenBasisTB(vertexNormalWS, sigmaX, sigmaY, flipSign, input.texCoord1.xy, layerTexCoord.vertexTangentWS1, layerTexCoord.vertexBitangentWS1);
    #if defined(_REQUIRE_UV2) || defined(_REQUIRE_UV3)
    SurfaceGradientGenBasisTB(vertexNormalWS, sigmaX, sigmaY, flipSign, input.texCoord2.xy, layerTexCoord.vertexTangentWS2, layerTexCoord.vertexBitangentWS2);
    #endif
    #if defined(_REQUIRE_UV3)
    SurfaceGradientGenBasisTB(vertexNormalWS, sigmaX, sigmaY, flipSign, input.texCoord3.xy, layerTexCoord.vertexTangentWS3, layerTexCoord.vertexBitangentWS3);
    #endif
}
#endif

That's said. We have indeed this issue:
>Basically means that even if the mask weight is zero, NaN/INF in any of the vertexTangents or >vertexBitangents vertex will still screw the pooch.

Now  I have a question regarding this ^. We have been agree that SpeedTree will use Master node, and not a inspector shader version. So this shouldn't be a problem (regarding speedtree) :P

Still. Not sure how to deal the best with garbage in this case.
Maybe in the particular case of 
SurfaceGradientGenBasisTB(vertexNormalWS, sigmaX, sigmaY, flipSign, input.texCoord1.xy, layerTexCoord.vertexTangentWS1, layerTexCoord.vertexBitangentWS1);

do a isnan after it? I am not keen at all to add a another keyword here.

if (isnan(layerTexCoord.vertexTangentWS1.x) 
layerTexCoord.vertexTangentWS1 = 0;
of maybe just a saturate?

layerTexCoord.vertexTangentWS1 = saturate(layerTexCoord.vertexTangentWS1), as the verctor should be between 0 and 1 , right?

Thougth?

mmikk:
I think a significantly better and more lightweight approach is to make sure texCoord1.xy is cleared in UnpackVaryingsMeshToFragInputs() when the mask is 0 or whichever way we have to know that texCoord1 garbage.

sebastienlagarde:
|when the mask is 0 or whichever way we have to know that texCoord1 is garbage.
if we where having a way to say that, then we the fix will be easy :)
but I don't see way to know we have garbage in uv1

All is drive by LitSharePass.hlsl which always include UV1

// Attributes
#define ATTRIBUTES_NEED_NORMAL
#define ATTRIBUTES_NEED_TANGENT // Always present as we require it also in case of anisotropic lighting
#define ATTRIBUTES_NEED_TEXCOORD0
#define ATTRIBUTES_NEED_TEXCOORD1
#define ATTRIBUTES_NEED_COLOR

#if defined(_REQUIRE_UV2) || defined(_REQUIRE_UV3) || defined(DYNAMICLIGHTMAP_ON) || defined(DEBUG_DISPLAY) || (SHADERPASS == SHADERPASS_LIGHT_TRANSPORT)
#define ATTRIBUTES_NEED_TEXCOORD2
#endif
#if defined(_REQUIRE_UV3) || defined(DEBUG_DISPLAY)
#define ATTRIBUTES_NEED_TEXCOORD3
#endif

Adding a keyword is your way to know that UV1 is used.

mmikk:
Hacking GenerateLayerTexCoordBasisTB() is not the answer. I agree with Evgenii that's chasing symptom rather than fixing the problem. Plus it makes it globally more expensive in the the other 3 cases which don't need it. 
The proper way is to clear texcoord1.xy in FragInputs. If that can't be done then the second approach could be to change LitDataIndividualLayer.hlsl by giving it an if-statement there so it only accumulates contributions associated with texcoord1 if uvMappingMask.y > 0 or alternatively to clear  vertexTangentWS1 and vertexBitangentWS1 to zero somewhere outside of GenerateLayerTexCoordBasisTB().

EvgeniiG:
>do a isnan after it? I am not keen at all to add a another keyword here.

Passing garbage through the system is unacceptable. We need a proper fix,

sebastienlagarde:
> > do a isnan after it? I am not keen at all to add a another keyword here.
> 
> Passing garbage through the system is unacceptable. We need a proper fix,

So as stated above, I guess the proper fix is currently that the null stream for this UV1 should be passed by the C++ code when dealing with SpeedTree (this is a guess based on memory), but I know that in Fontainebleau we use details map on UV3, mean we pass UV0 to UV3 even if the mesh don't have it, and all the stream UV2 is correctly setup to 0 by Unity (we use UV1 for lightmap).

mmikk:
Sebastien I understand the concern with a keyword but I don't follow why it's so difficult to clear texCoord1 in FragInputs based on a global such as uvMappingMask.y or some form of similar uniform/global?

sebastienlagarde:
> Sebastien I understand the concern with a keyword but I don't follow why it's so difficult to clear texCoord1 in FragInputs based on a global such as uvMappingMask.y or some form of similar uniform/global?

ah right, ok, I wasn't seeing it this way, yes we can do this.

sebastienlagarde:
So I suppose we can do something like this:

void GetSurfaceAndBuiltinData(FragInputs input, float3 V, inout PositionInputs posInput, out SurfaceData surfaceData, out BuiltinData builtinData RAY_TRACING_OPTIONAL_PARAMETERS)
{
input.texCoord1 = _UVMappingMask.y != 1 ? input.texCoord0  : 0;

Note: _UVMappingMask only exist in the context of Lit.shader

mmikk:
> > Sebastien I understand the concern with a keyword but I don't follow why it's so difficult to clear texCoord1 in FragInputs based on a global such as uvMappingMask.y or some form of similar uniform/global?
> 
> ah right, ok, I wasn't seeing it this way, yes we can do this.

Btw null stream should also work. If that's an option that sounds great too.

EvgeniiG:
If this bug not reproducible with the Lit shader, then?
If it's not, we should indeed fix Speedtree to work the same way other geometry does.

This is also a good place to invest in another keyword, since there is a fair amount of instructions, and I can easily see UV1 not being necessary for distant LODs (for which perf really matters). Detail maps are more likely to be used for nearby LODs, and light maps do not work well for all types of objects, plus we are switching to light probes for GI, anyway.

We could remove another keyword that is less perf intensive and properly fix this case.

parashar-unity:
Theoretically, this may not be an issue in general with other Lit meshes.  Well, fixing SpeedTree is the target such that it has its own series of ways of handling things, and SpeedTree assets are anyway exported with all sorts of things packed inside its UVs like geometry tags and what not.  At the same time, due to the cycling back and forth over how best to handle SpeedTree, a proper SpeedTree material will be 2020.1+, so that means 2019.3 will at least need something like the mask-based UV zeroing mentioned or if a new keyword at the cost of a less-meaningful is preferred, we can do that, too.

mmikk:
How about the idea of ensuring a null stream like seb mentioned? I'm guessing we're already doing this when it's not SpeedTree.

EvgeniiG:
So I see that we always access both vertex colors and UV1:

image

I propose replacing these three variants with branches:

image

Then we can add two variants for vertex colors and UV1, and still compile faster.

sebastienlagarde:
>Then we can add two variants for vertex colors and UV1, and still compile faster.

At this step of the package, if it is for 2019.3, then we should use the solution
input.texCoord1 = _UVMappingMask.y > 0 ? input.texCoord1 : 0;

Other solution implying changing keyword will be postpone to 2020.1, it is too dangerous.

EvgeniiG:
If it fixes the problem, go for it.

parashar-unity:
Well, that along with the SafeNormalize (_just_ the SafeNormalize without the Sanitize calls) in the Surface Gradients code seems to do the trick.  Though, based on Seb's comment and the other PR for keyword changes, does that imply I should only target this change for release/19.3 or also make the change for master?

sebastienlagarde:
The change must target master (then I can backport).

mmikk:
I would really greatly prefer we do not change SurfaceGradientGenBasisTB(). The bug we're dealing with comes from garbage in the UV. Once we correctly pass through (0,0) to texcoord1 as planned then you will not need the SafeNormalize() anymore.

I know from experience it's very risky to put in something like a SafeNormalize() in this context because the determinant of the screen-space derivatives on the normalized texture coordinate gets quite small so using an epsilon based approach such as SafeNormalize() doesn't work in general. You'll just end up breaking it in the general case.

EvgeniiG:
Why do we still need SafeNormalize if we now compute the derivative of the tex coord which was set to 0? What else are we missing?

EvgeniiG:
`   texST = 0; 
    float2 dSTdx = ddx_fine(texST), dSTdy = ddy_fine(texST); // 0

    float det = dot(dSTdx, float2(dSTdy.y, -dSTdy.x)); // 0

    if (abs(det) > 0.0) // false
        vT = normalize(vT);`

parashar-unity:
Ack...  it's probably because I removed the determinant check on my local copy after trying out one of the earlier suggestions.  That would make sense considering that the SafeNormalize is meant to prevent those divide-by-zero issues.

Let me try and run a test with that back in place.

sebastienlagarde:
hmmm... several test fail now, which is rather unexpected. Need to investagate why

sebastienlagarde:
So Julien have updated the PR with working code with yamato green. 
@parashar-unity could you do a last check to be sure it is fine on your side? thanks

mmikk:
Nothing further on my end.

parashar-unity:
Looks clear for me.  Oddly enough, I was not getting test failures on OSX
even before, but at least Windows appears to be clear now.

PR #5616, opened 2020-01-24
Make user stencil bits public

FrancescoC-unity:
As in the title


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDStencilUsage.cs
```diff
@@ -34,4 +34,13 @@ internal enum StencilUsage
 
         HDRPReservedBits = 255 & ~(UserBit0 | UserBit1),
     }
+
+    /// <summary>
+    /// Stencil bit exposed to user and not reserved by HDRP.
+    /// </summary>
+    public enum UserStencilUsage
```
sebastienlagarde: Please, add also the users stencil  mask

PR #5621, opened 2020-01-24
Fix default material transparent created with ZWrite enabled

alelievr:
### Purpose of this PR
Depth write is now disabled when switching from opaque to transparent for new materials.

TransparentZWrite

Tested the creation of materials and switch opaque to transparent + migration code with standard shader and shadergraphs


sebastienlagarde:
QA are at quality week, please track it to check after

TomasKiniulis:
Depth write is disabled now when switching to Transparent. No issues found with Depth Write functionality.

PR #5625, opened 2020-01-24
Adding the initial implementation of sub-surface scattering

anisunity:
This PR adds a volume override that allows user to have ray traced sub-surface scattering instead of the rasterized variant. 

image

image

The only user facing parameter that affects it is the "Scattering Distance" parameter in the Diffusion profile.


anisunity:
Also added a new test scene for this that checks Fabric SG, Lit.Shader, LayeredLit.Shader and Lit SG

remi-chapelain:
General state of the volume is ✔️ 
Found two minor concerns ❗️ : 

Ghosting (only in scene view without PLAY) when you are close to the mesh 
faf2eae847193d567648d388db6306af

Also ghosting when reactivating volume after moving the camera (in scene view)
2b6e4bc0414720be0f86f702811d669c

PR #5630, opened 2020-01-25
19.3 Blackboard Visibility Issue Bug Fix

zadi0648:
**Summary:**

* Fix for https://fogbugz.unity3d.com/f/cases/1214332/
* Also fixes the issues mentioned in #5062

---
**Implementation Notes:**

* Not an exact backport(?) for, while this follows the same format as what is in master, it is not the same fix as that fix (#5062) caused https://fogbugz.unity3d.com/f/cases/1214177/.
* In order to avoid any more unforeseen regressions & since this is a new way of disabling sub windows in Shader Graph, instead of having the Master Preview do the same thing I left the Master Preview alone (besides moving it into a function). I'll leave that up to the team if that's the right approach.
**Update:** ^ This definitely looks like the right approach because I tried `m_MasterPreviewView.style.display = DisplayStyle.Flex;` ect. for the Master Preview and I'm still hitting https://fogbugz.unity3d.com/f/cases/1214177/

---
**Game Plan:**

QA will be gone for Quality week when we want to land (Which is why Bellevue QA was not added to this PR), which includes me, so:

* Marc T or Alex L can take over this branch if further changes are needed.
* Perhaps Wyatt S, if we can be taken off Environment for a short bit, or Alex L can QA. Or Zack's testing could be sufficient to merge; QA will double check the package.

---
**Manually Tested:**

* All related bug cases: did not reproduce (see Notes to QA)
* All four visibility combinations; closing and reopening the SGs several times.
* Disabling the Blackboard's and Master Preview's Visibility, then resizing the Shader Graph, then re-enabling the visibility. 

---
**Technical Risk:** 2.5/4, informed that this is the correct way to do this, but this is a new way to temporarily remove things from Graph View that has not had any direct nor indirect testing.
**Halo Effect:** 1/4, Only affects the Blackboard & Master Preview.

---
**Notes to QA:**

_Related Cases:_
https://fogbugz.unity3d.com/f/cases/1214177/
https://fogbugz.unity3d.com/f/cases/1211132/ (Duplicate)
https://fogbugz.unity3d.com/f/cases/1187828/

---
**Yamato:**
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252F19.3%252Fblackboard-visibility-fix


spaztronaut:
tested for #5062
also unable to reproduce https://fogbugz.unity3d.com/f/cases/1214177/ with universal rp and 2019.3.0f6
uielements debugger shows correct display state for the blackboard when toggling the blackboard visibility setting

PR #5485, opened 2020-01-13
Hdrp/update default scenes

RSlysz:
### Checklist for PR maker
- [x] Have you added a backport label (if needed)? 
- [x] Have you updated the changelog? 
- [ ] Have you updated or added the documentation for your PR? NOT NEEDED
- [ ] Have you added a graphic test for your PR (if needed)? NOT NEEDED

---
### Purpose of this PR
This PR update the default scene and default DXR scene and DefaultVolumeProfile to simplify the new scene created in HDRP

Note: as usual until we will be able to use Scene-Template package, this will work only if the user recreate the HDRPAsset and the DefaultScene prefab in its project.

---
### Testing status

**Manual Tests**: What did you do?
- [x] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)

---

PR #5633, opened 2020-01-27
Hdrp/small wizard fixes for dxr

RSlysz:
### Purpose of this PR
This redo the Wizard update from PRs: 
- https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5485
- https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5557

The default FrameSettings for ScreenSpaceShadows is set to true for Camera in order to give a better workflow for DXR.

Added back disappeared Shadow test (I'm pretty sure to have done it but it still not here so redoing it)

Added an Hardware and OS test so the user can understand better why this is not working. 
image

---
### Testing status

**Manual Tests**: 

---
### Comments to reviewers
Note Killed test as the PR ticket was send on master instead of HDRP/staging due to the infamous Ctrl + Enter on github

@JordanL8 the Material-Upgrade.html page will be added by PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5613


remi-chapelain:
Got the "hardware & OS" error in the wizard. 

It appears that you need to be in DX12 to have the hardware test **TRUE**. 

Or the DX12 test is **AFTER** the hardware test so that prevent you from going further in the DXR settings.

The workaround we decided is to put the info about the Hardware & OS inside the DXR resources for a better workflow. 

If the DXR resources is false it's either the DXR resources or the hardware is not supported

iM0ve:
image

PR #5635, opened 2020-01-27
More code share between inspector materials 

FrancescoC-unity:
Similarly to the pattern used in shader graph.


com.unity.render-pipelines.high-definition/Editor/Material/Unlit/UnlitGUI.cs
```diff
@@ -41,36 +41,9 @@ public static void SetupUnlitMaterialKeywordsAndPass(Material material)
...
-            // ObjectVelocity need to be tagged in velocity pass.
-            // As velocity pass can be use as a replacement of depth prepass it also need to have StencilUsage
-            // As GBuffer pass can have no depth prepass, it also need to have StencilUsage
-            // Object velocity is always render after a full depth buffer (if there is no depth prepass for GBuffer all object motion vectors are render after GBuffer)
-            // so we have a guarantee than when we write object velocity no other object will be draw on top (and so would have require to overwrite velocity).
-            // Final combination is:
-            // Prepass: StencilUsage
-            // Motion vectors: StencilUsage, ObjectVelocity
-            // Forward: LightingMask
+            BaseLitGUI.SetupStencil(material, receivesSSR: false, useSplitLighting: false);
```
sebastienlagarde: is it possible to call BaseLitGUI from the unlit? 
if it is the case, please add a comment to explain why we use BaseLitGUI currently (say it is to share code and that the lit part will be ignore anyway)

FrancescoC-unity: Yup, good point, will do.

FrancescoC-unity: Actually, noticed a mistake even though test pass. Will fix soon as soon as I reopen this branch

PR #5646, opened 2020-01-27
Changing the tier system for a preset mode system and removing unnecessary options

anisunity:
Tier1 was changed to Performance and it implies deferred and ray bining.
Tier2 was changed to Quality.

Also changed most of the tests and increased coverage signficantly

image
image
image
image


anisunity:
@sebastienlagarde I did not forget the changelog, it is in the 3rd commit!

remi-chapelain:
✔️ 
_Side note : not related to this PR but I think it would help to disabled the denoiser options (radius, pass... etc) when the denoise checkbox is not enabled because you can still move the sliders but there's no repercussion on screen._

PR #5652, opened 2020-01-27
Add missing hdlight api (useVolumetrics)

alelievr:
### Purpose of this PR
Added a affectsVolumetric field to the HDAdditionalLightData to enabled / disable volumetric for lights.


EvgeniiG:
There's already a volumetric dimmer. Why do we need this?

alelievr:
It's needed because users can't know if the volumetric dimmer was set to 0 or just disabled. It's needed for tools that filter lights (Light Explorer for example) so they can only display lights that affect volumetric

PR #5662, opened 2020-01-28
Shadow Atlas is no longer allocated for area lights when they are disabled in the shader config file.

JulienIgnace-Unity:
### Purpose of this PR
Shadow Atlas is no longer allocated for area lights when they are disabled in the shader config file.

---
### Testing status

- [ ] Built a player


TomasKiniulis:
Not getting Shadow Atlas allocated for area lights anymore. A couple of notes and findings:

When Shadow Atlas are no longer allocated, VisualizeAreaLights in Shadow Debug Mode does not show any atlas anymore - in 7.1.8 black atlas is displayed

If only ShaderConfig.cs file is changed (works fine changing both .cs and .cs.hlsl), the following errors occur:
```
Compute shader (Deferred): Property (_AreaShadowmapMomentAtlas) at kernel index (8) is not set
```
modifying only ShaderConfig.cs file in 7.1.8 does not introduce these errors.

PR #5663, opened 2020-01-28
Support semi-transparent shadows for unlit shader and unlit shader graph

anisunity:
This PR is done on top of https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5490. 

It adds support for unlit and unlit SG to solve this bug https://fogbugz.unity3d.com/f/cases/1209779/

Tests still pass locally and unlit shader have been added to the test scenes. Example: 
701_DirectionalColorShadow


remi-chapelain:
All good ✔️

PR #5669, opened 2020-01-28
Hdrp/backplate ao fix

skhiat:
Change log added

### Checklist for PR maker
- [X] Have you added a backport label (if needed)?
- [X] Have you updated the changelog?
- [X] Have you updated or added the documentation for your PR?

---
### Purpose of this PR
Add new node on the ShaderGraph for being able to project a cubemap on the geometry.

---
### Testing status

- [ ] Built a player
- Other: 
Test the features

---
### Comments to reviewers


marctem:
Thanks for the PR! For 2019.3 this can be done as a Custom Function Node in a Subgraph for delivery to the automotive team (so I removed the backport tag). For 20.1 we're happy to take it but please add a test (see https://github.com/Unity-Technologies/ScriptableRenderPipeline/tree/master/TestProjects/ShaderGraph/Assets/Testing/IntegrationTests/Graphs/Input/Texture for examples), and we need to wait for our QA to return next week.

reallyfancy:
Hi! Further to Marc's comment concerning QA, the documentation review will also need to wait until next week when the writer who works with Shader Graph returns from vacation.

skhiat:
Added/Update Gfx test

LandonTownsendUnity:
Why does origin offset default to (1, 1, 1)? Doesn't it make more sense for it to default to (0, 0, 0)?

LandonTownsendUnity:
Can I get some more information as to how this would be used? I can understand why you would want a sample cube map mode that does not function as a reflection, but I don't understand why all of these inputs are needed. Isn't it enough to just take in a cubemap, samplerstate, vector3 (sample direction) and vector1 (LOD), and then leave it up to the user to decide how they want to calculate the sample direction? All of these extra variables feel like they make the node unnecessarily complicated and add extra performance cost.

skhiat:
Here an illustration of a use case:
https://youtu.be/RhCdOXvZTMA
When we have a "Backplate" or just an HDRI to visualize a product, automotive, ... We want to have the scene matching with the HDRI. This node allow that without modeling, just by projecting the HDRI on a given geometry. Which allows us to have a more consistent rendering with SSR etc.

skhiat:
@LandonTownsendUnity: For the default value, you're right > Updated

com.unity.shadergraph/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Add SampleCubemapProjectionNode to be able to project Cubemap to Geometry
```
sharlenet: Suggestion:
Added `SampleCubemapProjectionNode` that lets you project cubemaps to Geometry.

PR #5671, opened 2020-01-28
Fixed fogbugz case 1197958

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Fixed fogbugz case 1197958
---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
- The issue is that XR inits before URP overwrites MSAA settings.
- The solution is to update XR MSAA settings once it is changed by URP
- Thoughts: if XR could init after URP constructor we can save this UpdateEyeTextureMSAASetting. Will need to decouple texture creation logic from XR start logic to achieve this.


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -134,7 +134,12 @@ public UniversalRenderPipeline(UniversalRenderPipelineAsset asset)
 
             // Let engine know we have MSAA on for cases where we support MSAA backbuffer
             if (QualitySettings.antiAliasing != asset.msaaSampleCount)
+            {
                 QualitySettings.antiAliasing = asset.msaaSampleCount;
+#if ENABLE_VR && ENABLE_VR_MODULE
+                XR.XRDevice.UpdateEyeTextureMSAASetting();
```
mikesfbay: Why is this call added here? When I wrote the UpdateEyeTextureMSAASetting() function, it checks for AA changes each frame in post-render. So it should've been called each frame already on cpp side. 

URP MSAA was working fine until this fogbugz case in early December. We need to find the real reason that caused the breakage rather than adding this hack here. 

Adding this also won't help when user changes AA in script each frame since this in URP init code.

thomas-zeng: This is a correct fix. There is no post-render check for AA in 19.3

mikesfbay: AA was working fine in 19.3 URP and legacy XR, and AA checks in UpdateEyeTextureMSAASetting() is called in post-render: https://github.cds.internal.unity3d.com/unity/unity/blob/trunk/Modules/VR/VRDevice.cpp#L1346

For XRSDK, William added proper engine side AA checks in his MSAA PR. The right solution is to backport that PR to 19.3 rather than hacking this in.

thomas-zeng: Having to make trunk changes to fix URP is a hack. This PR fix is better.
AA is all broken according to fogbugz and oculus partners. So I don't know where you get the info AA was working fine. (Its broken for legacy XR too! which you just linked above).

mikesfbay: No, the fogbugz case said MSAA being broken is a recent change. Also, the line you are adding is already handled in ForwardRenderer.cs: 
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs#L419

thomas-zeng: Hey Mike, MSAA is broken in URP with legacy XR/XRSDK. This is known. I don't understand why you disagree with me about the above statement... 
Fogbug case: "Is Regression: No"
You can try it out yourself if you don't believe our QA

You pointed me to URP master branch code base, I don't know how that can help.. Are you suggesting MSAA is working in the current URP master? Oculus partner can repro the bug on 01/23/2020 though.. 

I will try the master branch tmrw and see if its already fixed.

mikesfbay: Like I said, MSAA has been working URP/LWRP + legacy XR since 19.1, so this breakage is a recent thing. The same code I pointed out has been around for a while and in 19.3:
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/release/2019.3/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs#L477

As for XRSDK shim, MSAA was listed as missing in my "Missing Features in XR SDK' doc, line 2 : https://docs.google.com/document/d/1K1epsA9NRVV5kjEjFv5-QYj-dSooPmTYoj0E6ydX24M

thomas-zeng: So legacy XR + URP is broken, even with 
https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/release/2019.3/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs#L477
and
https://github.cds.internal.unity3d.com/unity/unity/blob/trunk/Modules/VR/VRDevice.cpp#L1346

This PR fixed legacy XR. could you green light?

mikesfbay: ForwardRenderer.cs#L477 is the right place to call UpdateEyeTextureMSAASetting() since it done per frame. Your change in init code won't work when AA is changed per frame in script. So we need to figure out why ForwardRenderer.cs update isn't working.

thomas-zeng: Hey Mike, The bug is not about ForwardRenderer.cs update isn't working. Its about XR init sequence in URP. See *Comments to reviewers*.
Let's sync up tmrw so I can explain to you how XR works in URP.

phi-lira: What's the decision on this?

thomas-zeng: I will get back to this PR next week. (all-hands on 7.2.0 release issues right now)

thomas-zeng:
@ryanhy-unity Hey ryan, do you think you have time this week to verify this PR? This PR is a high priority one to unblock further MSAA investigation/fix.

mikesfbay:
I looked into the update AA call in ForwardRenderer. There are multiple places where the logic prevents UpdateEyeTextureMSAASetting() from getting called in ForwardRender. I'm looking into simplifying the logic.

Meanwhile, QA can go ahead and verify the patch fix.

ryanhy-unity:
I'll have a look this afternoon, hope to have feedback by eod today.

thomas-zeng:
@mikesfbay This PR is ready to land. Could you greenlight? We should get this fix into next 7.2.x/7.3.x release.

thomas-zeng:
@phi-lira this PR is good to land.

PR #5672, opened 2020-01-28
[Draft] Pure URP XRSDK feature parity with XR shim

thomas-zeng:

### Checklist for PR maker

---
### Purpose of this PR

Pure XRSDK in URP to replace legacy XR in one go.
---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -52,6 +52,8 @@ public sealed class ForwardRenderer : ScriptableRenderer
 
         public ForwardRenderer(ForwardRendererData data) : base(data)
         {
+            UniversalRenderPipeline.m_XRSystem.RegisterXRShaders(data.shaders.xrOcclusionMeshPS, data.shaders.xrMirrorViewPS);
```
phi-lira: Can this be moved to ScriptableRenderer or UniversalRenderPipeline? Reason: make as much possible XR logic out of the renderer, otherwise you will have to maintain this for Deferred render and it will also not work for users custom renderers.

com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
```diff
@@ -164,7 +164,13 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             }
 
             // Render the lut
-            Blit(cmd, m_InternalLut.id, m_InternalLut.id, material);
+            Vector4 scaleBias = new Vector4(1, 1, 0, 0);
```
phi-lira: if this is constant why not define the constant in shader?

com.unity.render-pipelines.universal/Runtime/Passes/ColorGradingLutPass.cs
```diff
@@ -164,7 +164,13 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+            Vector4 scaleBiasRT = new Vector4(1, 1, 0, 0);
+            cmd.SetGlobalVector(ShaderConstants._BlitScaleBias, scaleBias);
+            cmd.SetGlobalVector(ShaderConstants._BlitScaleBiasRt, scaleBiasRT);
+
+            CoreUtils.SetRenderTarget(cmd, m_InternalLut.id);
+            cmd.DrawProcedural(Matrix4x4.identity, material, 0, MeshTopology.Quads, 4, 1, null);
```
phi-lira: does not work on all platforms

com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -73,22 +75,31 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
             RenderTargetIdentifier opaqueColorRT = destination.Identifier();
 
+            Vector4 scaleBias = new Vector4(1, 1, 0, 0);
+            Vector4 scaleBiasRT = new Vector4(1, 1, 0, 0);
+            cmd.SetGlobalVector(ShaderConstants._BlitScaleBias, scaleBias);
```
phi-lira: I'm not sure if I understand but it seems these are compile time constants? so why not declare it in shader?

com.unity.render-pipelines.universal/Runtime/Passes/CopyColorPass.cs
```diff
@@ -73,22 +75,31 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+            cmd.SetGlobalVector(ShaderConstants._BlitScaleBiasRt, scaleBiasRT);
+            cmd.SetGlobalTexture("_BlitTex", source);
+
+            ScriptableRenderer.SetRenderTarget(cmd, new RenderTargetIdentifier(opaqueColorRT, 0, CubemapFace.Unknown, -1),
+                                                       BuiltinRenderTextureType.CameraTarget, clearFlag, clearColor);
             switch (m_DownsamplingMethod)
             {
                 case Downsampling.None:
-                    Blit(cmd, source, opaqueColorRT);
+                    cmd.DrawProcedural(Matrix4x4.identity, m_CopyColorMaterial, 0, MeshTopology.Quads, 4, 1, null);
```
phi-lira: Does not work in all platforms.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -57,35 +57,44 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-                cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa4);
-                CopyTexture(cmd, depthSurface, copyDepthSurface, m_CopyDepthMaterial);
+                    break;
+                case 4:
+                    cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthNoMsaa);
+                    cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa2);
+                    cmd.EnableShaderKeyword(ShaderKeywordStrings.DepthMsaa4);
+                    break;
+                default:
+                    // XRTODO: Add err msg. This case shouldn't really happend. Could be undefined behavior
```
phi-lira: hm, actually user can choose 8 samples. (I think you just exposed an existing bug in URP)

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -57,35 +57,44 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+                    cmd.EnableShaderKeyword(ShaderKeywordStrings.DepthNoMsaa);
+                    cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa2);
+                    cmd.DisableShaderKeyword(ShaderKeywordStrings.DepthMsaa4);
+                    break;
             }
+
+            ScriptableRenderer.SetRenderTarget(cmd, new RenderTargetIdentifier(copyDepthSurface, 0, CubemapFace.Unknown, -1),
+                                               BuiltinRenderTextureType.CameraTarget, clearFlag, clearColor);
+            
+            cmd.DrawProcedural(Matrix4x4.identity, m_CopyDepthMaterial, 0, MeshTopology.Quads, 4, 1, null);
```
phi-lira: does not work in all platforms.

com.unity.render-pipelines.universal/Runtime/Passes/CopyDepthPass.cs
```diff
@@ -108,5 +117,12 @@ public override void FrameCleanup(CommandBuffer cmd)
             cmd.ReleaseTemporaryRT(destination.id);
             destination = RenderTargetHandle.CameraTarget;
         }
+
+        // Precomputed shader ids to save some CPU cycles (mostly affects mobile)
+        static class ShaderConstants
```
phi-lira: why do you define these in multiple places?

com.unity.render-pipelines.universal/Runtime/Passes/DrawSkyboxPass.cs
```diff
@@ -15,7 +15,47 @@ public DrawSkyboxPass(RenderPassEvent evt)
...
+                context.DrawSkybox(renderingData.cameraData.camera);
+
+                // Disable Legacy XR path
+                cmd.SetSinglePassStereo(SinglePassStereoMode.None);
+                context.ExecuteCommandBuffer(cmd);
+            }
+            else
+            {
+                // Setup legacy XR before calling into skybox. In non-XR case, this function just returns false.
+                UniversalRenderPipeline.m_XRSystem.MountShimLayer();
```
phi-lira: ifguard all logic if in stereo mode

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -53,44 +53,45 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-                    ClearFlag.None,
-                    Color.black,
-                    m_TargetDimension);
-
-                Camera camera = cameraData.camera;
-                cmd.SetViewProjectionMatrices(Matrix4x4.identity, Matrix4x4.identity);
-                cmd.SetViewport(cameraData.pixelRect);
-                cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, blitMaterial);
-                cmd.SetViewProjectionMatrices(camera.worldToCameraMatrix, camera.projectionMatrix);
+                blitTarget = new RenderTargetIdentifier(cameraData.xrPass.renderTarget, 0, CubemapFace.Unknown, -1);
```
phi-lira: you killed the entire LOAD/STORE action segment.

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -53,44 +53,45 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+            cmd.SetGlobalTexture("_BlitTex", m_Source.Identifier());
+
+            cmd.DrawProcedural(Matrix4x4.identity, m_BlitMaterial, 0, MeshTopology.Quads, 4, 1, null);
+
             context.ExecuteCommandBuffer(cmd);
             CommandBufferPool.Release(cmd);
         }
+
+        // Precomputed shader ids to save some CPU cycles (mostly affects mobile)
+        static class ShaderConstants
```
phi-lira: You don't have to declare this in multiple places.

com.unity.render-pipelines.universal/Runtime/Passes/FinalBlitPass.cs
```diff
@@ -53,44 +53,45 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
+            // We need to keep in the pipeline of first render pass to each render target to propertly set load/store actions.
+            // meanwhile we set to load so split screen case works.
+            CoreUtils.SetRenderTarget(cmd, blitTarget, RenderBufferLoadAction.Load, RenderBufferStoreAction.Store, ClearFlag.None, Color.black);
+            cmd.SetViewport(cameraData.xrPass.GetViewport(0));
+
+            // We f-flip if
+            // 1) we are bliting from render texture to back buffer(UV starts at bottom) and
+            // 2) renderTexture starts UV at top
+            bool yflip =  !cameraData.xrPass.renderTargetIsRenderTexture && SystemInfo.graphicsUVStartsAtTop;
+            Vector4 scaleBias = yflip ? new Vector4(1, -1, 0, 1) : new Vector4(1, 1, 0, 0); ;
```
phi-lira: Seems like we should setup _ProjectionParams.x instead to know if it's flipped or not? https://docs.unity3d.com/Manual/SL-UnityShaderVariables.html

com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -338,32 +342,36 @@ int GetDestination()
...
+                                    colorLoadAction, RenderBufferStoreAction.Store, RenderBufferLoadAction.DontCare, RenderBufferStoreAction.DontCare);
 
-                    cmd.DrawMesh(RenderingUtils.fullscreenMesh, Matrix4x4.identity, m_Materials.uber);
-                    cmd.SetViewProjectionMatrices(cameraData.camera.worldToCameraMatrix, cameraData.camera.projectionMatrix);
-                }
+
+                // We f-flip if
+                // 1) we are bliting from render texture to back buffer and
+                // 2) renderTexture starts UV at top
+                bool yflip = isRenderToBackBufferTarget && SystemInfo.graphicsUVStartsAtTop;
```
phi-lira: same thing we should update/set _ProjectionParams and use it instead?

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -81,39 +81,101 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-                        camera.nearClipPlane, camera.farClipPlane);
-
-                    Matrix4x4 viewMatrix = camera.worldToCameraMatrix;
-                    Vector4 cameraTranslation = viewMatrix.GetColumn(3);
-                    viewMatrix.SetColumn(3, cameraTranslation + m_CameraSettings.offset);
-
-                    cmd.SetViewProjectionMatrices(viewMatrix, projectionMatrix);
-                    context.ExecuteCommandBuffer(cmd);
+                    // XR Pass viewport will handle camera stack too
+                    Rect pixelRect = cameraData.xrPass.GetViewport(0);
```
phi-lira: This is possibly breaking camera stacking as we stack pixelRect

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -49,6 +48,8 @@ static internal class PerCameraBuffer
         const string k_RenderCameraTag = "Render Camera";
         static ProfilingSampler _CameraProfilingSampler = new ProfilingSampler(k_RenderCameraTag);
 
+        public static XRSystem m_XRSystem = new XRSystem();
```
phi-lira: wrap around XR_ENABLED XR_MODULE_ENABLED?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -234,20 +249,25 @@ static void RenderSingleCamera(ScriptableRenderContext context, CameraData camer
...
 
-            if (!camera.TryGetCullingParameters(IsStereoEnabled(camera), out var cullingParameters))
-                return;
-
-            SetupPerCameraShaderConstants(cameraData);
-
             string tag = (asset.debugLevel >= PipelineDebugLevel.Profiling) ? camera.name: k_RenderCameraTag;
             CommandBuffer cmd = CommandBufferPool.Get(tag);
-            using (new ProfilingScope(cmd, _CameraProfilingSampler))
+            var xrPasses = m_XRSystem.SetupFrame(cameraData, /*XRTODO XR single pass settings in urp asset pipeline*/ true, /*XRTODO: test mode*/ false);
```
phi-lira: wrap around any XR specific code to check if XR is enabled

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -234,20 +249,25 @@ static void RenderSingleCamera(ScriptableRenderContext context, CameraData camer
...
+            foreach (XRPass xrPass in xrPasses)
```
phi-lira: If I understand correctly, now multipass will execute all passes (including monoscopic ones like shadow maps and color grading LUT). This most likely will be a perf regression in multi pass.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -251,25 +251,8 @@ half3 MixFog(real3 fragColor, real fogFactor)
 
 #endif
 
-#if defined(UNITY_SINGLE_PASS_STEREO)
```
phi-lira: you cannot remove shader api function as users might be relying on it, move to a deprecated file instead. Then we can remove in a major version update of URP.

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
```diff
@@ -121,42 +126,23 @@ real4 unity_SHBb;
 real4 unity_SHC;
 CBUFFER_END
 
-#if defined(UNITY_STEREO_MULTIVIEW_ENABLED) || ((defined(UNITY_SINGLE_PASS_STEREO) || defined(UNITY_STEREO_INSTANCING_ENABLED)) && (defined(SHADER_API_GLCORE) || defined(SHADER_API_GLES3) || defined(SHADER_API_METAL) || defined(SHADER_API_VULKAN)))
-    #define GLOBAL_CBUFFER_START(name)    cbuffer name {
```
phi-lira: we cannot remove this macros, users might be relying on them, move to deprecate file.

com.unity.render-pipelines.universal/ShaderLibrary/UnityInput.hlsl
```diff
@@ -190,7 +172,6 @@ float4x4 unity_MatrixV;
 float4x4 unity_MatrixInvV;
 float4x4 unity_MatrixVP;
 float4x4 unity_MatrixInvVP;
-float4 unity_StereoScaleOffset;
```
phi-lira: don't remove, deprecate instead

com.unity.render-pipelines.universal/Shaders/PostProcessing/Common.hlsl
```diff
@@ -13,6 +13,12 @@ struct Attributes
     UNITY_VERTEX_INPUT_INSTANCE_ID
 };
 
+struct VertQuadAttributes
```
phi-lira: Our shader convetions says to use just Attributes

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -63,9 +63,7 @@ namespace UnityEngine.Rendering.Universal
 
         public SortingCriteria defaultOpaqueSortFlags;
 
-        public bool isStereoEnabled;
```
phi-lira: This is breaking API change, you deprecate but not remove this now.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -181,15 +178,39 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
...
+                SetupBackbufferFormat(backbufferMsaaSamples);
+            }
+
+            if (createColorTexture)
+            {
+                ConfigureCameraTarget(m_ActiveCameraColorAttachment.Identifier(), m_ActiveCameraDepthAttachment.Identifier(), cameraTargetDescriptor.dimension);
+            }
+            else
+            {
+                ConfigureCameraTarget(cameraData.xrPass.renderTarget, cameraData.xrPass.renderTarget, cameraTargetDescriptor.dimension,
```
phi-lira: I'm worried about xrPass logic being the only way to access camera render target info (or some other properties). Consider the case people don't use XR? Now from the API stand point this doesn't make sense.

I wonder if we should encapsulate the camera matrices and render target into wrapper properties / methods that work in a more generic way. We should have a talk about it to brainstorm some possibilities.

thomas-zeng: Yes, let's sync up on this in out 2/19/2020 meeting. We could generalize xrPass logic here for URP usage, eg camera stacking. We should chat about the opportunity here.
Also, I would like to integrate xrPass logic to the future URP render graphs (provide XR info for RG to build gfx work dependencies tree). On that part, I would need your expertise on how to architect things robust enough to be compatible with RG.

thomas-zeng:
Next iteration: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6346

PR #5674, opened 2020-01-29
Remove 3 variants from the Lit shader

EvgeniiG:
### Purpose of this PR
I need to remove several variants to solve the SpeedTree case.
So, for Lit/LitTessellation/LayeredLit/LayeredLitTessellation shaders, I replaced _VERTEX_DISPLACEMENT_LOCK_OBJECT_SCALE, _DISPLACEMENT_LOCK_TILING_SCALE, and _PIXEL_DISPLACEMENT_LOCK_OBJECT_SCALE with _MaterialInstanceFlags. I will reclaim two variants in the next PR (for a bug fix and a perf uplift).

---
### Testing status

**Manual Tests**: Tested on SRP Data repository.

---
### Comments to reviewers
I have NOT upgraded the materials. I HAVE written the upgrader. It appears to work. We can upgrade materials after we are done reviewing the code.

### Comments to QA
Must test all combinations of "Lock Object Scale" and "Lock Tiling Scale" for Vertex Displacement, Tessellation Displacement, and Pixel Displacement before and after this change. You can start with some assets in the SRP Data repository (there is a setup with tessellation and pixel displacement side by side in which they look nearly identical).

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/ImageBasedLighting.hlsl
@@ -8,7 +8,7 @@
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/Sampling/Sampling.hlsl"
 
 #ifndef UNITY_SPECCUBE_LOD_STEPS
-    #define UNITY_SPECCUBE_LOD_STEPS 6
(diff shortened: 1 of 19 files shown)


EvgeniiG:
OK, so the test 1207 fails because the upgrader does not upgrade certain materials (if you click on the mesh and expand the material, the inspector runs and the material is fixed). There is some kind of weird materials nesting (clone) and I am not sure how that works. I suppose it's not upgrader-friendly.

PR #5682, opened 2020-01-29
Various Fixes and Adding Contact Shadow tests

anisunity:
- Fixed the radius value used for ray traced directional light (diameter was used as radius, which caused a mismatch between ray tracing and path tracing)
- Fixed compilation issues with the layered lit in ray tracing shaders (error was in the sponza scene)
- Changed the way the length parameter is used for ray traced contact shadows (was divided by 2 for no reason)
- Added contact shadow test scenes for directional, point and spot lights


remi-chapelain:
All tests pass independently, **BUT** I got a crash 90% of the time when I try to run them all 😑

PR #5683, opened 2020-01-29
Fix debug menu bitfield

alelievr:
### Purpose of this PR
Fixed runtime debug menu bitfield controls
DebugMenu


iM0ve:
The None checkbox still breaks the functionality. In the video:
1) All lights are shown
2) Selecting checkbox to only see Punctual spot light
3) Selecting None
4) All lights shown again, but the Punctual spot light is still checked.



Also the debug mode is unusable with real world units

PR #5688, opened 2020-01-30
Better directional PCSS blur handling

FrancescoC-unity:
This PR: 

 - Simplifies a bit the softness calculation for Directional Lights (better than before and still checked vs reference)
 - Now the visual softness is as much as possible matching between cascades. 
 The only issue is when an occluder is 1.0 in the shadow map. In that case the distance information is lost due to clamping, so we cannot get that info out of there cheaply enough. In those cases you'll still have a mismatch in effective blur. 

With this change for example this fairly blurred shadow spans 3 cascades and the softness is somewhat coherent between them. 

image

Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-pcss-for-directional/.yamato%252Fupm-ci-hdrp.yml%2523HDRP_Win_DX11_playmode%2520XR_fast-trunk/1174105/job


TomasKiniulis:
Checked PCSS shadows comparing with 7.1.8, the result looks a lot more accurate than it was before.

7.2.0
image
7.1.8
image

PR #5689, opened 2020-01-30
Fix mip slider reflectionprobe

alelievr:
### Purpose of this PR
image
Remove the slider knob when the reflection probe preview has no mipmap


TomasKiniulis:
The slider knob gets disabled accordingly. No issues found

PR #5702, opened 2020-01-30
Hdrp xr public layout api cluster display

etienne-p:

### Checklist for PR maker

---
### Purpose of this PR

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UberPost.compute
```diff
@@ -14,6 +15,15 @@
 #pragma kernel KMain_Variant5 MAIN=KMain_Variant5 VARIANT=5
 #pragma kernel KMain_Variant6 MAIN=KMain_Variant6 VARIANT=6
 #pragma kernel KMain_Variant7 MAIN=KMain_Variant7 VARIANT=7
+// Cluster Display variants
+#pragma kernel KMain_Variant8  MAIN=KMain_Variant8  VARIANT=0 USING_CLUSTER_DISPLAY
```
sebastienlagarde: as it is for 2020.1 and we support multi compile in compute, we should use multicompile for USING_CLUSTER_DISPLAY (we will soon do this for all our compute shaders)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UberPost.compute
```diff
@@ -110,27 +119,28 @@ void MAIN(uint3 dispatchThreadId : SV_DispatchThreadID)
 {
     UNITY_XR_ASSIGN_VIEW_INDEX(dispatchThreadId.z);
     PositionInputs posInputs = GetPositionInput(float2(dispatchThreadId.xy), _ScreenSize.zw, uint2(GROUP_SIZE, GROUP_SIZE));
-    float2 uv = posInputs.positionNDC;
+    float2 uv = DEVICE_TO_CLUSTER_FULLSCREEN_UV(posInputs.positionNDC.xy); // posInputs.positionNDC.xy is in fullscreen uv not ndc space
```
sebastienlagarde: >// posInputs.positionNDC.xy is in fullscreen uv not ndc space
What do you mean? 
positionNDC is in 0..1 range

sebastienlagarde:
From my point of view the PR is good and we can merge it when it is approve.
@fabien-unity waiting for your review for approval.
Also an improvement to do is to add a multi-compile for the cluster display as multicompile is now supported in 2020.1, this is something we need to do on our side for all compute too.

Sadly currently we can't strip compute shader varaint in the sahder stripper at build time, which mean that for now, all extra variant will be embed even on console like PS4 :( . 

@joravainen is working on stripping of compute shader variant, guess it is for 2020.2 ? which will be ok for us, if it is for a later version, then build time will increase which will not be nice.

Please add a changelog. This PR will land for the package 9.0.0-preview

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
```diff
@@ -95,6 +95,7 @@ private enum SMAAStage
         bool m_FilmGrainFS;
         bool m_DitheringFS;
         bool m_AntialiasingFS;
+        bool m_ClusterDisplayFS;
```
clusty: This might be a trivial comment, but does it make sense to try to "sell" the cluster display features basically as tiling support. What in mean by this, I would not mention clusterDisplay anywhere in the code.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRPass.cs
```diff
@@ -44,16 +46,18 @@ internal XRView(Camera camera, Camera.StereoscopicEye eye, int dstSlice)
             projMatrix = camera.GetStereoProjectionMatrix(eye);
             viewMatrix = camera.GetStereoViewMatrix(eye);
             viewport = camera.pixelRect;
+            clusterDisplayParams = Matrix4x4.zero;
```
clusty: Matrix4x4.identity or it makes no difference ?

etienne-p: this matrix is used as a storage, it will not be multiplied, zero initialization reflects that :)

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/PostProcessSystem.cs
etienne-p: Sounds like a relevant point, however if we're going to update terminology, let's all agree on something, @sebastienlagarde  @fabien-unity what do yout think?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UberPost.compute
etienne-p: ndcxy

maybe i'm missing something, if i assign ndc.xy to color output, such as `    _OutputTexture[COORD_TEXTURE2D_X(posInputs.positionSS)] = float4(posInputs.positionNDC.xy, 0, 1);
`, i see the image above wich led me to think those were in the 0 .. 1 range?

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UberPost.compute
etienne-p: cool! lets me reduce the footprint of this PR, fix in https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/42046cdac6d90d06f1be6ad3714aaf1e98d65feb

etienne-p:
> 
> 
> Overall, I think this is going in the right direction.
> On top of @seblagarde feedback, I'll add:
> 
>     * we need to fully document the initial changes in this branch (moving the XRLayout API from internal to public)
> 
>     * we should revisit the naming of some code to make it a bit more generic, esp the bits in the XR module. What about using _**Tile**_ ? For example in cases like CLUSTER_SCREEN_SIZE or DEVICE_TO_CLUSTER_FULLSCREEN_UV. Thoughts?
> 
>     * and we'd need to add a test to the automation suite to make sure this feature will stay alive
> 
>     * is CLUSTER_GRID_SIZE used at all ?

Let's all agree on proper terminology, i also think cluster display is a bit of a narrow term :) Regarding the XR API this PR rely on, who'll document/test it? CLUSTER_GRID_SIZE in not used yet, but it though it'd have value to add macros to save users the headache of unpacking data from the cluster display parameters matrix. I wanted to provide intelligible macros for anything currently stored in the parameters matrix.

fabien-unity:
I can work on the documenting on the API but this will have to wait until I'm back (end of March). And as discussed before, we have to properly review it first also.

etienne-p:
> 
> 
> I can work on the documenting on the API but this will have to wait until I'm back (end of March). And as discussed before, we have to properly review it first also.

Sounds good, thank you!

com.unity.render-pipelines.high-definition/Runtime/PostProcessing/Shaders/UberPost.compute
etienne-p: Also, taking a look at the template fullscreen custom pass shader i see:
`    //     float2 positionNDC; // Normalized screen coordinates within the viewport    : [0, 1) (with the half-pixel offset)
`

PR #5717, opened 2020-01-31
Fix some causes of NaNs on consoles when using probes

FrancescoC-unity:
NOTE: This PR is against release/2019.3 and need to be ported to staging

This contains several things that could have caused NaNs: 

 - We make sure that the RenderTextures are created whenever we need them. 
 - Clear to black the texture caches whenever they are created (otherwise the first frame, garbage will be read if probes need to look into probes) 
 - Make sure that OnDemand Probes are rendered only when requested. 
 - Make sure that we consider a probe in the lightloop only if it has valid rendered data (has ever rendered if realtime or has a valid texture if not) 
 - Make sure that we set the vbuffer history validity even when not using reprojection to make sure it stays in the correct state.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -119,7 +119,8 @@ float aspect
         // Runtime Data
         RenderTexture m_RealtimeTexture;
         RenderData m_RealtimeRenderData;
-        bool m_WasRenderedSinceLastOnDemandRequest;
+        bool m_WasRenderedSinceLastOnDemandRequest = true;
```
JulienIgnace-Unity: Seems like this is unused?

TomasKiniulis:
Checked template Xbox and PS4 - works correctly, but adding additional planar probes on Xbox template scene still leaves NaNs which is already fixed by Francesco, unfortunately, that does not cover a fix for HDRP_RuntimeTests template scene which is still displaying black screen on Xbox

PR #5718, opened 2020-01-31
Virtual texturing [2020.1] [Skip CI]

GraphineCharles:

### Checklist for PR maker
- No backport needed 2020.1 only.
- [x] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
Add virtual texturing support to HDRP.
Unity trunk now has support for VT but without the SRP's driving this it is not possible for user to use this. This PR adds the SRP side of VT for HDRP. This includes two major parts of code
- Feedback analysis: Add an extra rendertarget written during the relevant phases (GBuffer, forward) that allows the VT system to collect information on which textures were accessed by the shaders.
- Shader graph support: VT enabled materials can be created in shadergraph through a new texture stack node. If VT is enabled in the project this will be using VT if VT is disabled this will fall back to using regular 2D texture sampling.

For more information on virtual texturing please see:
- A presentation on Virtual Texturing at the graphics summit (November) here https://drive.google.com/open?id=1wLsvSlUoy9rqcoQ9xEUiRO-tn5n_yzCW
- A draft of the Streaming Virtual Texturing User Guide here https://docs.google.com/document/d/1pzKz7bCPTjAvMcpUlxOs83KXwKZKC7rdTlBqwGIbOfU/edit
- A presentation about the state of Procedural Virtual Texturing here https://drive.google.com/file/d/1EmnYDMAVwMzuoDE8Mg6KKwCVWMJBLBSQ/view
- A sample project demonstrating Procedural VT here https://github.cds.internal.unity3d.com/unity/VTTestProjects/tree/master/PVTExampleProject
- A sample project demonstrating Streaming VT (a bit rough, we use it to develop) here https://github.cds.internal.unity3d.com/unity/VTTestProjects/tree/master/HDRP

---
### Testing status

- [x] Built a player
- Other: 

- Graphics test 9713

---
### Comments to reviewers


zadi0648:
Doesn't this need HDRP qa?

LandonTownsendUnity:
Test document: https://docs.google.com/document/d/1t9dp1c00bUztdH760RjdkZeHmPamts0heghYXVOSvoM/edit?usp=sharing

LandonTownsendUnity:
What does the "no resolve" checkbox do?

LandonTownsendUnity:
Also, if there's anything major in this PR that I should be testing (as shadergraph QA) besides the new shadergraph node, please let me know.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+        float4 color = SampleStack(info, TextureSlot1);
+        float4 color2 = SampleStack(info, TextureSlot2);
+        ...
+
+    The above steps can be repeated for multiple stacks. But be sure that when using the SampleStack you always
+    pass in the result of the PrepareStack for the correct stack the texture belongs to.
+
+*/
+
+#ifdef UNITY_VIRTUAL_TEXTURING
```
EvgeniiG: #else and #endif parts are pretty hard to locate. Please add a comment // UNITY_VIRTUAL_TEXTURING at the appropriate locations.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+#define TEXTURESTACK_include
+
+#define GRA_HLSL_5 1
+#define GRA_ROW_MAJOR 1
+#define GRA_TEXTURE_ARRAY_SUPPORT 1
+#define GRA_PACK_RESOLVE_OUTPUT 0
+#if SHADER_API_PSSL
+#define GRA_NO_UNORM 1
+#endif
+#include "GraniteShaderLib3.cginc"
```
EvgeniiG: What is the reason this is not a part of SRP? Do you plan to support VT in Legacy?

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+    The above steps can be repeated for multiple stacks. But be sure that when using the SampleStack you always
+    pass in the result of the PrepareStack for the correct stack the texture belongs to.
+
+*/
+
+#ifdef UNITY_VIRTUAL_TEXTURING
+
+struct StackInfo
+{
+    GraniteLookupData lookupData;
```
EvgeniiG: For anyone else who may be wondering:

`struct GraniteLookupData
{
    gra_Float4 translationTableData;
    gra_Float2 textureCoordinates;
    gra_Float2 dX;
    gra_Float2 dY;
};`

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+    GraniteLODLookupData lookupDataLod;
```
EvgeniiG:
```
struct GraniteLODLookupData
{
    gra_Float4 translationTableData;
    gra_Float2 textureCoordinates;
    float cacheLevel;
};
```

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+        ...
+
+    The above steps can be repeated for multiple stacks. But be sure that when using the SampleStack you always
+    pass in the result of the PrepareStack for the correct stack the texture belongs to.
+
+*/
+
+#ifdef UNITY_VIRTUAL_TEXTURING
+
+struct StackInfo
```
EvgeniiG: 17x floats.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+    #define GR_LOOKUP Granite_Lookup_Anisotropic
+    #define GR_LOOKUP_LOD Granite_Lookup
+#endif
+
+// This can be used by certain resolver implementations to override screen space derivatives
+#ifndef RESOLVE_SCALE_OVERRIDE
+#define RESOLVE_SCALE_OVERRIDE float2(1,1)
+#endif
+
+StructuredBuffer<GraniteTilesetConstantBuffer> _VTTilesetBuffer;
```
EvgeniiG:
```
struct GraniteTilesetConstantBuffer
{
    #define _grTilesetCBSize 2
    gra_Float4x4 data[_grTilesetCBSize];
};
```

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+
+#define DECLARE_STACK_CB(stackName) \
+    float4 stackName##_atlasparams[2]
+
+#define DECLARE_STACK_BASE(stackName) \
+TEXTURE2D(stackName##_transtab);\
+SAMPLER(sampler##stackName##_transtab);\
+\
+GraniteTilesetConstantBuffer GetConstantBuffer_##stackName() \
```
EvgeniiG: Seems like it would be cleaner to define the function once (instead of pasting it multiple ones as a macro) that takes the stack as a parameter. Not a major issue.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+#ifdef UNITY_VIRTUAL_TEXTURING
+
+struct StackInfo
+{
+    GraniteLookupData lookupData;
+    GraniteLODLookupData lookupDataLod;
+	float4 resolveOutput;
+};
+
+#ifdef TEXTURESTACK_CLAMP
```
EvgeniiG: Is this not configurable?

com.unity.render-pipelines.high-definition/Editor/Material/Fabric/ShaderGraph/FabricPass.template
```diff
@@ -369,6 +370,10 @@ Pass
 
         $DepthOffset: builtinData.depthOffset = surfaceDescription.DepthOffset;
 
+#if !defined(_SURFACE_TYPE_TRANSPARENT)
+        $SurfaceDescription.VTFeedback: builtinData.vtFeedback = surfaceDescription.VTFeedback;
```
EvgeniiG: Couldn't find a single comment describing what's stored in there. It's float4, by the way.

TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/9x_Other/9713_VirtualTexturing/LowResolutionRequester.cs
```diff
@@ -0,0 +1,51 @@
...
+    public int firstMipToRequest;
+
+    // Start is called before the first frame update
+    void Start()
+    {
+        if (materialToRequest == null) return;
+        Shader shad = materialToRequest.shader;
+
+        // Hack, we simply request all stacks in the shader ideally we would have some way
+        // to identify a specific stack within a shader.
```
EvgeniiG: Do we ship this hack? What are the implications?

EvgeniiG:
I'd like to estimate the perf impact of this PR. Say we have a full-screen quad with a base color texture and a normal map. What's the register count delta? Instruction count delta? How much slower is the G-Buffer pass? Preferably on the PS4.

EvgeniiG:
Maybe test Forward as well.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+	cache.TextureArray = stackName##_c##layerIndex;\
+	cache.Sampler = sampler##stackName##_c##layerIndex;\
+\
+	float4 output;\
+	Granite_Sample_HQ(grCB, info.lookupData, cache, layerIndex, output);\
+	return output;\
+} \
+float3 SampleVT_Normal_##layerSamplerName(StackInfo info, float scale)\
+{\
+	return Granite_UnpackNormal( jj(SampleVT_,layerSamplerName)( info ), scale ); \
```
EvgeniiG: This seems rather restrictive.
It needs to at least support both 2 and 3 channel normal maps (see SampleUVMappingInternal.hlsl).

EvgeniiG:
```
gra_Float3 Granite_UnpackNormal(in gra_Float4 PackedNormal, float scale)
{
    gra_Float2 reconstructed = gra_Float2(PackedNormal.x * PackedNormal.a, PackedNormal.y) * 2.0f - 1.0f;
    reconstructed *= scale;
    float z = sqrt(1.0f - GranitePrivate_Saturate(dot(reconstructed, reconstructed)));
    return gra_Float3(reconstructed, z);
}
```

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+	DECLARE_STACK_LAYER(stackName, layer1SamplerName,1)\
+	DECLARE_STACK_LAYER(stackName, layer2SamplerName,2)
+
+#define DECLARE_STACK4(stackName, layer0SamplerName, layer1SamplerName, layer2SamplerName, layer3SamplerName)\
+	DECLARE_STACK_BASE(stackName)\
+    DECLARE_STACK_RESOLVE(stackName)\
+	DECLARE_STACK_LAYER(stackName, layer0SamplerName,0)\
+	DECLARE_STACK_LAYER(stackName, layer1SamplerName,1)\
+	DECLARE_STACK_LAYER(stackName, layer2SamplerName,2)\
+	DECLARE_STACK_LAYER(stackName, layer3SamplerName,3)
```
EvgeniiG: So 4 texture layers is a limitation? Is the environment team OK with it?

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
EvgeniiG: In fact, is there a good reason you handle normal maps differently at all? I would personally prefer that VT acts like a complicated sampler, and the user then interprets the data in whichever way she wants.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
```diff
@@ -0,0 +1,318 @@
...
+    NOTE: The Stack shaderlab property and DECLARE_STACKn define need to match i.e. the same name and same texture slots.
+
+    Then in the pixel shader function (likely somewhere at the beginning) do a call:
+
+        StackInfo info = PrepareStack(uvs, MyFancyStack);
+
+    Then later on when you want to sample the actual texture do a call(s):
+
+        float4 color = SampleStack(info, TextureSlot1);
+        float4 color2 = SampleStack(info, TextureSlot2);
```
EvgeniiG: What's the purpose of Resolve? It has no comments or documentation.

com.unity.render-pipelines.high-definition/Editor/Material/Lit/ShaderGraph/HDLitPass.template
```diff
@@ -403,6 +406,11 @@ Pass
 
         $SurfaceDescription.Emission: builtinData.emissiveColor = surfaceDescription.Emission;
 
+        //Note this will not fully work on transparent surfaces (can check with _SURFACE_TYPE_TRANSPARENT define)
+        //We will always overwrite vt feeback with the nearest. So behind transparent surfaces vt will not be resolved
+        //This is a limitation of the current MRT approach.
+        $SurfaceDescription.VTFeedback: builtinData.vtFeedback = surfaceDescription.VTFeedback;
```
EvgeniiG: So, should we perhaps automatically NOT support VR on transparent objects?
Seems like this could potentially adversely affect the quality of opaque surfaces hiding behind?
I feel like we should either offer a solution that is robust, or not offer it at all (and explain why).
Perhaps there could be an expensive option for transparent objects where you use atomics?

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
```diff
@@ -238,6 +239,10 @@ $include("SharedCode.template.hlsl")
 
         $SurfaceDescription.Emission: builtinData.emissiveColor = surfaceDescription.Emission;
 
+#if !defined(_SURFACE_TYPE_TRANSPARENT)
+        $SurfaceDescription.VTFeedback: builtinData.vtFeedback = surfaceDescription.VTFeedback;
```
EvgeniiG: The comment
> //Note this will not fully work on transparent surfaces (can check with _SURFACE_TYPE_TRANSPARENT define)
> //We will always overwrite vt feeback with the nearest. So behind transparent surfaces vt will not be resolved
> //This is a limitation of the current MRT approach.
should be replicated here.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
```diff
@@ -238,6 +239,10 @@ $include("SharedCode.template.hlsl")
...
+#endif
```
EvgeniiG: Should we perhaps disallow VT on transparent objects?
As is, it will look awful (I suppose?), without any clear indication as to why.
Or can we perhaps add an expensive option that uses atomics (opt-in)?

EvgeniiG:
Is there a small (but not trivial) sample project one can use to test this?

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
EvgeniiG: I suppose it performs blending of the texture layers? If that's the case, I would prefer to have programmable blending (e.i. we handle it ourselves).

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
EvgeniiG: Is there only a single sampler per stack? Is this an acceptable limitation?

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaderUtilities.cs
```diff
@@ -200,6 +203,7 @@ internal struct SurfaceDescriptionInputs
                 new Dependency("SurfaceDescriptionInputs.uv3",                       "FragInputs.texCoord3"),
                 new Dependency("SurfaceDescriptionInputs.VertexColor",               "FragInputs.color"),
                 new Dependency("SurfaceDescriptionInputs.FaceSign",                  "FragInputs.isFrontFace"),
+                new Dependency("SurfaceDescriptionInputs.PixelCoordinate",           "FragInputs.positionSS"),
```
EvgeniiG: Can't you use "SurfaceDescriptionInputs.ScreenPosition" (a couple of lines above)? What's the context?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -42,8 +46,10 @@ TEXTURE2D_X(_GBufferTexture0);
 TEXTURE2D_X(_GBufferTexture1);
 TEXTURE2D_X(_GBufferTexture2);
 TEXTURE2D_X(_GBufferTexture3); // Bake lighting and/or emissive
-TEXTURE2D_X(_GBufferTexture4); // Light layer or shadow mask
-TEXTURE2D_X(_GBufferTexture5); // shadow mask
+TEXTURE2D_X(_GBufferTexture4); // VTFeedbakc or Light layer or shadow mask
+TEXTURE2D_X(_GBufferTexture5); // Light layer or shadow mask
+TEXTURE2D_X(_GBufferTexture6); // shadow mask
```
EvgeniiG: Not related to the current PR, but we should really somehow unify (put it in one place) G-buffer declaration since we also need certain G-buffers in forward (albedo for SSS, normals for decals, now VT).
@seblagarde

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs
```diff
@@ -248,6 +248,10 @@ public sealed class ShaderResources
             public Shader temporalAntialiasingPS;
             [Reload("Runtime/PostProcessing/Shaders/ContrastAdaptiveSharpen.compute")]
             public ComputeShader contrastAdaptiveSharpenCS;
+#if ENABLE_VIRTUALTEXTURES
+            [Reload("Runtime/ShaderLibrary/DownsampleVTFeedback.compute")]
+            public ComputeShader VTFeedbackDownsample;
```
EvgeniiG: We don't put shaders into the Shader Library. Only HLSL headers.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassForward.hlsl
```diff
@@ -60,20 +60,37 @@ PackedVaryingsToPS VertTesselation(VaryingsToDS input)
...
-#ifdef _DEPTHOFFSET_ON
-    , out float outputDepth : SV_Depth
+#define EXTRA_BUFFER_TARGET SV_Target1
 #endif
+
+void Frag(PackedVaryingsToPS packedInput,
+        #ifdef OUTPUT_SPLIT_LIGHTING
+            out float4 outColor : SV_Target0,  // outSpecularLighting
+            #ifdef UNITY_VIRTUAL_TEXTURING
+                out float4 outVTFeedback : VT_BUFFER_TARGET,
```
EvgeniiG: Please don't slot your RT between diffuse and specular. It's annoying to debug. Just put it last.

com.unity.render-pipelines.core/Runtime/Textures/RTHandle.cs
```diff
@@ -50,6 +50,8 @@ public class RTHandle
         /// </summary>
         public string name { get { return m_Name; } }
 
+        public bool enableMSAA { get { return m_EnableMSAA; } }
```
FrancescoC-unity: Aside from missing doc, I know this replicates the var name, but being a function and given that is part of the public API I would name this something isMSAAEnabled or MSAAEnabled or something that could not be confused with something that enables MSAA.

LandonTownsendUnity:
I somehow forgot to link this test doc, but here:
https://docs.google.com/document/d/1t9dp1c00bUztdH760RjdkZeHmPamts0heghYXVOSvoM/edit?usp=sharing

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
GraphineCharles: - Copy paste: Yes, it wouln't make much difference though this is all inlined and define-generated code so the user will never see this code.
- Single samples: This is very much by desin, the whole purpose of a "stack" is that everything in the stack uses the same UV's , filtering , ... this means we can optimize the VT sampling and streaming of things being together in the same stack. Without this it's probably easier to use just individual stack nodes and have additional freedom (but also increased performance cost).

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
GraphineCharles: This was placeholder in the most recent update of the branch this is now a parameter.

com.unity.render-pipelines.high-definition/Editor/Material/Fabric/ShaderGraph/FabricPass.template
GraphineCharles: It's virtual texturing feedback data, information the VT system uses to determine what tiles to load. It is indeed a float4 that is stored as RGBA8 in the rendertarget.

TestProjects/HDRP_Tests/Assets/GraphicTests/Scenes/9x_Other/9713_VirtualTexturing/LowResolutionRequester.cs
GraphineCharles: This is not shipped it's a script that's only part of the test. It's works for what we want to test but wouldn't be applicable in a general case. (So it's more some sort of warning for someone looking to copy paste this elsewhere...)

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
GraphineCharles: This is actually legacy code from when we did our initial integration. The Shadergraph node actually uses the standard unpack functions using the same logic at the ordinary SampleTexture2D node. I'll remove this old code so users don't get confused about this.

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
GraphineCharles: 4 textures is a limitation due to the maximum number of channels the translation table texture supports. The Environment team is well aware of this limit. If you need more layers note that you can always do multiple VT samples (at an additional cost).

com.unity.render-pipelines.core/ShaderLibrary/TextureStack.hlsl
GraphineCharles: This is an old function and should be removed. It gets the virtual texturing feedback info and is not related to doing any blending. Note that layers in VT speak are not like layers to be blended but different "channels" for example think of it like a pbr stack texture like "Diffuse", "Normal", "Ambient Occlusion" how these layers are interpreted/blended is up to the shadergraph authors.

com.unity.render-pipelines.high-definition/Editor/Material/PBR/ShaderGraph/HDPBRPass.template
GraphineCharles: It is disabled the shadergraph/material inspector will error/warn the user it's not working. An option using Atomics is out of scope for now but we might consider it in the future.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaderUtilities.cs
GraphineCharles: We've updated to code to use ScreenPosition in the latest iteration. If I remember correctly ScreenPosition (sometimes? this is a define/function spagetti) uses an extra vertex interpolation instead of SV_Position so it might actually be slower. But just to keep things clean it's indeed better to just share it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs
GraphineCharles: Will move it.

com.unity.render-pipelines.core/Runtime/Textures/RTHandle.cs
GraphineCharles: Done

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPipelineResources.cs
GraphineCharles: Moved it to Runtime/VirtualTexturing/Shaders/DownsampleVTFeedback.compute . Note we just put it there as there theare are some other shaders in there wich have similar functionality : DownsampleDepth.shader , CopyDepthBuffer.shader , ClearStencilBuffer.shader these might need to be moved too at some point.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/ShaderPass/ShaderPassForward.hlsl
GraphineCharles: This is necessary as unlit uses the same forward pass and the same render target layout. And we need to write to the feedback render target from Unlit.

GraphineCharles:
> I'd like to estimate the perf impact of this PR. Say we have a full-screen quad with a base color texture and a normal map. What's the register count delta? Instruction count delta? How much slower is the G-Buffer pass? Preferably on the PS4.

A full performance analysis on different platforms is out of scope for this PR but it is important to note the following when considering the wider performance impact:
- If VT is disabled in the project settings the ENABLE_VIRTUALTEXTURES c# define will be false, thus the HDRP will essentially be the same as the no-vt HDRP without any performance impact at all.
- If vt is enabled but a no-vt material is used, the only performance overhead is the extra render target bound (and outputting 0 to that rendertarget). Needed by the VT feedback. Avoiding this would mean splitting shader passes or something which we felt was not desirable.
- Only when using the Texture Stack node, and VT is enabled will there be any real overhead cost this is usually a few ALU instructions (15~30) and one dependent texture read.

GraphineCharles:
> So I'm not the one QAing this (at least at the moment), but since you're new to making Unity PRs (welcome! 😃) I wanted to say that PR descriptions like this, especially PRs this large, could use more information, especially for QA to test efficiently and effectively.
> 
> For example:
> 
> * How do you set up Virtual Texturing in a project?
> * How do you set up Virtual Textures? + Additional ones you used for testing during development, to help us as a starting point.
> * How do you use the new Shader Graph node?
> * (^ first three points: basically, how do we begin testing this?)
> * Perhaps a link to a doc or something explaining the behind the scenes of this implementation of Virtual Texturing
> * In your mind, what is the checklist of the things QA should hit? Settings, types, integration, ect.
> * Things you ran into during development, for us to keep an eye out for?
> * Anything else that we may not be aware of, as people who are new to Virtual Texturing.
> * An image/further explanation of the automated test could be nice.
> 
> Example of other PRs: https://ono.unity3d.com/unity/unity/pull-request/82477/_/video/vfr-encoding, https://ono.unity3d.com/unity/unity/pull-request/85273/_/lighting/shadowmaskcolor, #3334

I've updated the PR description with links to relevant documentation & tutorials on how to use the feature and how it works. As for the QA focus I think it's best to talk to @hdb-unity he's been mapping this for a wile supporting us with the foundation side of QA.

PR #5719, opened 2020-01-31
Add support to run post-processing on each camera in the stack

phi-lira:
### Checklist for PR maker
- [x ] Have you added a backport label (if needed)? For example, the `need-backport-2019.3` label. After you backport the PR, the label changes to `backported-2019.3`.
- [ ] Have you updated the changelog? Each package has a `CHANGELOG.md` file.
- [ ] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR.
- [ ] Have you added a graphic test for your PR (if needed)? When you add a new feature, or discover a bug that tests don't cover, please add a graphic test.

---
### Purpose of this PR
 This PR adds support to run post-processing in each camera in the stack, as opposed to run once at the end of the stack. 
 
  This fixes the issue with the previous solution that camera stacking would always apply post-processing on UI screen space and world space UI. 

  Now it's possible to run post-processing in the middle of the stack while still compositing results on top.

---
### Testing status

**Manual Tests**: What did you do?
- [x ] Opened test project + Run graphic tests locally
- [x ] Built a player
- [x ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [x ] C# and shader warnings (supress shader cache to see them)
- [x ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 
 Tested several camera stacking demos (cockpit, 3D UI, shooter, shop menu) from Universal Rendering Examples
  Tested BoatAttack for regression.

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #5720, opened 2020-01-31
[2019.3 Backport] Add support to run post-processing on each camera in the stack (#5719)

phi-lira:
Backport #5719


com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -310,15 +310,32 @@ void DrawElementCallback(Rect rect, int index, bool isActive, bool isFocused)
...
+                m_NameContent.text = "MISSING CAMERA";
+                string warningInfo = "Camera is missing";
+                EditorGUI.LabelField(rect, m_NameContent, TempContent("", warningInfo, m_ErrorIcon), errorStyle);
+
+                // We shouldn't do this!
                 // Automagicaly deletes the entry if a user has removed a camera from the scene
-                m_AdditionalCameraDataCameras.DeleteArrayElementAtIndex(index);
-                m_AdditionalCameraDataSO.ApplyModifiedProperties();
+                //m_AdditionalCameraDataCameras.DeleteArrayElementAtIndex(index);
+                //m_AdditionalCameraDataSO.ApplyModifiedProperties();
```
ellioman: Same comment here, maybe better to remove these four lines?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -310,15 +310,32 @@ void DrawElementCallback(Rect rect, int index, bool isActive, bool isFocused)
...
```
ellioman: Same comment here, maybe best to remove these 4 lines?

phi-lira: Fix submitted.

phi-lira:
Not much testing with PPv2 yet, I've pulled the package to check it compiles. PPv2 doesn't support camera stacking so testing should be easier.

phi-lira:
Looking at the code today I think some of the issues with fixed with integrated post-processing and render pass are likely to be in PPv2, we should take a look at it while doing package validation next week.

PR #5726, opened 2020-02-03
Updated missing API doc.

JulienIgnace-Unity:
### Purpose of this PR
Updated missing API doc.


com.unity.render-pipelines.high-definition/Runtime/VFXGraph/Utility/PropertyBinders/HDRPCameraBinder.cs
```diff
@@ -87,9 +104,14 @@ private void OnValidate()
                 AdditionalData.requestGraphicsBuffer += RequestHDRPBuffersAccess;
         }
 
+        /// <summary>
+        /// Returns true if the VisualEffect is valid for this binder.
```
peeweek: It's more precisely, return if the Visual Effect *and the configuration of the binder* are valid to perform the binding.

PR #5727, opened 2020-02-03
Fix AO depth issues

FrancescoC-unity:
The previous fix for the issue still has some precision issues with huge distances and A2 reported back the issue. With this PR I don't have the same issues on my machine, though the previous issue was vendor specific. 

So for QA, pls can you check the following scenario: 

A very long plane or something (long as that it span between near and far plane) and have big near/far plane difference, in my test scene and what A2 reported it was 0.01 near and 5000 far plane. 
On fail, the end of the range (farthest points) are black when AO is on. Depending on your test scene you might want to set the Direct Lighting strength to non 0 to see the AO. 

On AMD, NVIDIA and Intel please? 

Thank you!


sebastienlagarde:
I will merge this PR for the package of wednesday, so you can test it inside the package, thanks

PR #5759, opened 2020-02-05
Bugfix 1209255: Issue with reflections when using an orthographic camera

ellioman:
### Purpose of this PR
When using orthographic cameras, the reflection would be incorrect because the view direction was using the same calculations for perspective & ortho.

---

### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 

**Yamato**: 


ellioman:
**Yamato:** Everything is green except VFXMain, which is due to a known issue that's about to be fixed and not related to this PR.

com.unity.render-pipelines.universal/Shaders/LitForwardPass.hlsl
```diff
@@ -91,7 +91,8 @@ Varyings LitPassVertex(Attributes input)
 
     VertexPositionInputs vertexInput = GetVertexPositionInputs(input.positionOS.xyz);
     VertexNormalInputs normalInput = GetVertexNormalInputs(input.normalOS, input.tangentOS);
-    half3 viewDirWS = GetCameraPositionWS() - vertexInput.positionWS;
+
+    half3 viewDirWS = lerp(GetCameraPositionWS() - vertexInput.positionWS, UNITY_MATRIX_I_V[2].xyz, round(saturate(unity_OrthoParams.w))); // Perspective vs OrthoGraphic
```
phi-lira: using `UNITY_MATRIX_I_V[2].xyz` here will cost extra register pressure for additional matrix. we should use instead `-UNITY_MATRIX_V[2].xyz` 

Actually, we could define the following functions that HDRP define in the file below:
`float3 GetViewForwardDir()`
`bool IsPerspectiveProjection()`
`float3 GetWorldSpaceViewDir()`
`float3 GetWorldSpaceNormalizedViewDir()`

https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesFunctions.hlsl

phi-lira: we currently don't have a `ShaderVariablesFunctions.hlsl`, it seems we could add to `Core.hlsl` instead of adding a new include file.

ellioman: @phi-lira Good points. I've now added the functions to Core.hlsl and am now calling those.

ellioman: Should I be calling GetWorldSpaceNormalizedViewDir() rather than GetWorldSpaceViewDir()?

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -122,6 +122,39 @@ half OutputAlpha(half outputAlpha)
...
+}
+
+// Returns 'true' if the current view performs a perspective projection.
+bool IsPerspectiveProjection()
+{
+    return (unity_OrthoParams.w == 0);
+}
+
+// Computes the world space view direction (pointing towards the viewer).
+float3 GetWorldSpaceViewDir(float3 positionRWS)
```
phi-lira: positionWS, we don't have positionRWS (relative world space)

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -122,6 +122,39 @@ half OutputAlpha(half outputAlpha)
...
+{
+    if (IsPerspectiveProjection())
```
phi-lira: I'm not sure if we should use [flatten] or let compiler decide. I've asked in slack.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -122,6 +122,39 @@ half OutputAlpha(half outputAlpha)
...
+        return GetCameraPositionWS() - positionRWS;
+    }
+    else
+    {
+        // Orthographic
+        return -GetViewForwardDir();
+    }
+}
+
+float3 GetWorldSpaceNormalizeViewDir(float3 positionRWS)
```
phi-lira: positionWS

ellioman: Fixed!

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
ellioman: Fixed!

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
phi-lira: To be safe let's use ternary operator or [flatten] here.

PR #5773, opened 2020-02-05
Fixed an exposure issue with ray traced sub-surface scattering.

anisunity:
Exposure is now supported for ray traced sub-surface. It is applied afterwards to avoid a ghosting behavior when changing exposure.


anisunity:
Also updated the test scene to cover this.

PR #5774, opened 2020-02-05
Fix an issue where hair shader could write garbage in the diffuse lighting buffer.

JulienIgnace-Unity:
### Purpose of this PR
Fix an issue where hair shader could write garbage in the diffuse lighting buffer.
This was caused buy the "transparent write motion" define being active even for opaque hair.
Also updated the shader graph samples provided with HDRP to fix the shader graphs there and update materials to avoid errors at import.

---
### Testing status

- [ ] Built a player


iM0ve:
APPROVED
PR fixes:
* Blinking
* Scene exposure
* Import errors

New issue with the sample not related to the PR, planar reflection probes don't work and an error message due to too low default HDRP asset reflection probe resolution when created by Wizard.
image

PR #5775, opened 2020-02-05
Fix material properties not up to date when updating the preview

alelievr:
### Purpose of this PR
When the material of the main preview in ShaderGraph is sent to the SRP via `ProcessPreviewMaterial` it does not contains all the properties that it's own shader have. This is a workaround for this issue.

Note that there is still an issue when we edit a script with a ShaderGraph window opened, it breaks the preview but we can't fix it properly in C# and require more investigation.

### Story:
To give you a bit of context in HDRP we have a function which resets the material keywords and states by reading properties in the material (for example it will ask for the value of the _Transparent property and set a keyword to enable/disable transparency in the shader).
We call this function to 'fix' the material of the preview in ShaderGraph when the compilation ends but when i was debugging this function I saw some material.HasProperty returning false while the shader have these properties. As stated in the doc of the HasProperty function: https://docs.unity3d.com/ScriptReference/Material.HasProperty.html it says that it checks if the shader have the property so it should probably have returned true.
Now if we look at how the ShaderGraph compiles the preview, we can see these steps:
- Creates a dummy pink shader from sources stored in C# constant with ShaderUtil.CreateShaderAsset
- Creates a material from this shader
ShaderUtil.ClearCachedData is called on this shader
- Then we have the ShaderUtil.UpdateShaderAsset which replaces the pink shader by the actual code of the graph: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.shadergraph/Editor/Drawing/PreviewManager.cs#L469
- For debug purpose I added a log of this to check if the shader code was actually replaced: ShaderUtil.GetShaderData(shaderData.mat.shader).ActiveSubshader.GetPass(i).SourceCode  and it outputs the original pink shader, and same for material.passcount which is 1 in this case (that's one of the reason why the preview does not compile correctly)
- After that it calls ShaderUtil.CompilePass but i think here it will compile the pink shader instead of the graph one
- Then we have an update function that is called every frame which checks if the compilation have ended: https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.shadergraph/Editor/Drawing/PreviewManager.cs#L406
- If we debug at this point we can see that at the first frame the matreial.passcount is 1 (still the pink shader) but for next frames we have 3 (which is the number of pass of the preview shader)
- Here we enter a state where the ShaderUtil.IsPassCompiled will always return false, what i think is happening is that it compiled the pink shader but is actually asking if the graph shader is compiled (which we never requested the compilation) so it stays in this loop forever

To confirm this i added a call to ShaderUtil.CompilePass just after the IsPassCompiled check and it fixed the issue 🙂 
But I still have the problem that the HDRP reset keyword function have some HasProperty that fails, I think at this point since the update of the shader code have been delayed that the new properties of the shader haven't been reflected to the material which would explain why the function fails.

Finally, I managed to fix this issue by re-assigning the shader to the preview material after it have finished compiling to force all the properties to be in the material.


sebastienlagarde:
Yamato have issue not related to this PR that come from trunk, with the ALLOC_TEMP. There is however shader graph yamato coverage with the other PR

PR #5783, opened 2020-02-05
Fix case 1158661 hardware dynamic resolution for scene view

NetteVI:
### Purpose of this PR
Fixes case 1158661: https://fogbugz.unity3d.com/f/cases/1158661/ where enabled hardware dynamic resolution would cause the Scene View to be cropped/scaled incorrectly due to a Post-Processing render target having dynamic resolution enabled during Scene View rendering. This occurred on Metal and DX12.

The fix adjusts the condition for cleaning the render target pool so that it also cleans up the pool if render targets are expected to not have dynamic resolution enabled, but some do. 

---
### Testing status

- [x] Built a player
- Other: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Ffix-case-1158661-dynamic-resolution

---
### Comments to reviewers


FrancescoC-unity:
The change looks fine to me, adding QA for some validation

NetteVI:
Latest commit fixes issue on MacOS Metal.

(Before, when the pool was cleared, m_HasHWDynamicResolution would get initialized to 'false' and it would not correctly update to 'true' since the render targets already had dynamic scaling as requested.  This change has m_HasHWDynamicResolution being updated every time the state is set/checked, as long as the pool is not empty.)

PR #5786, opened 2020-02-05
[Workaround] Fixed WMR crashes editor when entering play mode regression for 7.2.0

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR
- Fixed WMR crashes editor issue introduced in 7.2.0. 

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
- 7.2.0 has WMR crashes editor issue caused by culling system
   - This is a regression since we didn't see this in 7.1.8
- The issue is WMR returns zero'd projection/view matrix data for the first couple frames.
   - culling system used to be able to handle this, but 7.2.0 can't handle this anymore(editor crashes)
- Fix is disable URP rendering until WMR is ready to return valid data
- Trunk proper fix is here: https://ono.unity3d.com/unity/unity/pull-request/94762/_/xr/display/noop-frame
    - QA: please verify this workaround is compatible with the trunk fix above

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
@@ -167,6 +167,7 @@ namespace UnityEngine.Rendering.Universal
 
 #if ENABLE_VR && ENABLE_XR_MODULE
         static List<XRDisplaySubsystem> xrDisplayList = new List<XRDisplaySubsystem>();
+        static bool xrSkipRender = false;
         internal void SetupXRStates()
         {
             SubsystemManager.GetInstances(xrDisplayList);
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -232,7 +275,7 @@ static void RenderSingleCamera(ScriptableRenderContext context, CameraData camer
                 return;
             }
 
-            if (!camera.TryGetCullingParameters(IsStereoEnabled(camera), out var cullingParameters))
```
joejo-unity: Probably can revert this.

thomas-zeng: Fixed in 
432da7a

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -162,13 +165,53 @@ protected override void Dispose(bool disposing)
             CameraCaptureBridge.enabled = false;
         }
 
+#if ENABLE_VR && ENABLE_XR_MODULE
+        static List<XRDisplaySubsystem> xrDisplayList = new List<XRDisplaySubsystem>();
```
phi-lira: @mikesfbay had already added a static display list so this is not necessary. I'll merge PR because we will remove this anyway in future.

PR #5817, opened 2020-02-06
Hdrp/fix wizard persistency

RSlysz:
### Purpose of this PR
- Fixed Wizard persistency of Direct3D12 change on domain reload. ( https://fogbugz.unity3d.com/f/cases/1208462/ )
- Fixed Wizard persistency of FixAll on domain reload.
- Fixed Wizard behaviour on domain reload.

---
### Testing status
Manually tested:
- persistency of Direct3D12 change (When press "Fix"):
  - check remains red after domain reload
  - check become green after next restart
  - checked state in file ProjectSettings > HDRPProjectSettings.asset (m_WizardNeedRestartAfterChangingToDX12)
- persistency of FixAll (in HDRP+VR that install several package which cause domaine reload after each):
  - FixAll do not stop anymore after first package install
  - checked state in file ProjectSettings > HDRPProjectSettings.asset (m_WizardNeedToRunFixAllAgainAfterDomainReload)
- Wizard behaviour:
  - opens 10 editor frames after Editor start only if "Show on start" checkbox is checked
  - do not show up on domain reload (either checkbox is checked or not)
  - checked state in file ProjectSettings > HDRPProjectSettings.asset (m_WizardPopupAtStart and m_WizardPopupAlreadyShownOnce)

---


com.unity.render-pipelines.high-definition/Editor/DefaultScene/HDProjectSettings.cs
```diff
@@ -29,10 +29,16 @@ enum Version
         [SerializeField]
         string m_ProjectSettingFolderPath = "HDRPDefaultResources";
         [SerializeField]
-        bool m_WizardPopupAtStart = false;
+        bool m_WizardPopupAtStart = true;
```
fredericv-unity3d: Why don't we use `EditorPrefs` here? It should be easier to handle persistency of the settings.

RSlysz: It should not be cross project

PR #5819, opened 2020-02-06
Hdrp/fix emissive mesh on lights

RSlysz:
### Purpose of this PR
Fix issue where MeshRenderer and MeshFilter were removed when adding Light component.
( https://fogbugz.unity3d.com/f/cases/1207148/ )

Note:
Now they are in a sub GameObject that is not editable in editor. This GameObject is hiden while light type is not Area.

---
### Testing status
Manualy tested all light type and the Display Emissive Mesh checkbox.
Tested also the coherency while moving and rotating and scaling the area light.

---


RSlysz:
Closed and rebased. More tests and fix in new version https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5882

PR #5823, opened 2020-02-06
[XR] Fix for XRSDK single-pass stereo mode incorrectly switch to multipass mode (case 1217120)

mikesfbay:
### Purpose of this PR 
Add query for XR SDK single-pass availability using XR SDK renderpass descriptors. This is a short-term fix for XR SDK single-pass stereo mode incorrectly switch to multipass mode ( case 1217120).

---
### Testing status
Tested locally on Oculus Quest and Rift with GLES/Vulkan with and without XRSDK.

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/7.2.x%252Furp%252Fxrsdk-single-pass-query

---
### Comments to reviewers
Proper XR SDK shim fix will land in 2019.3.2


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -478,7 +515,8 @@ static void InitializeStackedCameraData(Camera baseCamera, UniversalAdditionalCa
             cameraData.isXRMultipass = false;
 
 #if ENABLE_VR && ENABLE_VR_MODULE
-            if (cameraData.isStereoEnabled && !cameraData.isSceneViewCamera && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass)
+            if (cameraData.isStereoEnabled && !cameraData.isSceneViewCamera &&
+                !CanXRSDKUseSinglePass(baseCamera) && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass)
```
thomas-zeng: Please make sure XR.XRSettings.stereoRenderingMode is not being used for XRSDK. This function is not shimmed for XRSDK.
 if (displaySubsystemList.Count > 0)
 {
   XRSDK logic
 }
 else
 {
   XR.XRSettings.stereoRenderingMode logic
 }

mikesfbay: XRSDK does shim XRSettings.stereoRenderingMode. See VRBinding.gen.cpp line 1459 and IVRDeviceShimGeneration.h line 417. 
This has been tested to work with XRSDK and with legacy XR in 19.3.

thomas-zeng:
> Are we okay with the references to "XR SDK"? I'm not sure if we've been avoiding using that term in changelogs/method names.
> 
> This also prevents SPDW, but I believe that is unsupported in XR SDK to begin with.

I agree, we should avoid using XRSDK in method name if possible. XRSDK and legacyXR are legacy XR in URP. XRSDK is just a legacyXR shim device for URP. XRSDK is going through the same legacyXR code path.

thomas-zeng:
static bool IsMultiPassStereoEnabled(Camera camera)
        {
            if (camera == null)
                throw new ArgumentNullException("camera");

#if ENABLE_VR && ENABLE_VR_MODULE
            return IsStereoEnabled(camera) && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass;
#else
            return false;
#endif
        }

^^
This logic is broken too.

jackpr-unity:
I was able to reproduce https://fogbugz.unity3d.com/f/cases/1217120 with 2019.3/staging commit 9c473a0c8ba4

Then I tried this PR on Quest with 2019.3/staging commit 9c473a0c8ba4 and https://ono.unity3d.com/unity/unity/pull-request/100539/_/2019.3/xr/sdk/stereo-rendering-mode

So far so good. I still need to test with WindowsStandalonePlayer with the Oculus Rift and a WMR HMD on both editors.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
thomas-zeng: Mike, stereoRenderingMode is not shimmed. XRSDK always return multi-pass.

mikesfbay: The default shim function returns multipass (IVRDeviceShimGeneration.h ln 417). That's why there's a 19.3 backport PR to properly define GetStereoRenderingMode().

thomas-zeng: Yes, that is exactly my point. We should make sure to not use GetStereoRenderingMode() for XRSDK until your 19.3 backport PR land

thomas-zeng: And put that in else block will do exactly that

mikesfbay:
> > static bool IsMultiPassStereoEnabled(Camera camera)
> > return IsStereoEnabled(camera) && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass;
> ^^
> This logic is broken too.

This is fixed in changeset https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5823/commits/3bc9d57846e6e89b07030892d8350a9eb50142bf

thomas-zeng:
> > > static bool IsMultiPassStereoEnabled(Camera camera)
> > > return IsStereoEnabled(camera) && XR.XRSettings.stereoRenderingMode == XR.XRSettings.StereoRenderingMode.MultiPass;
> > > ^^
> > > This logic is broken too.
> 
> This is fixed in changeset [3bc9d57](https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/3bc9d57846e6e89b07030892d8350a9eb50142bf)

Thanks Mike. Fix looks good to me!

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
mikesfbay: Currently, in XR SDK XR.XRSettings.stereoRenderingMode just returns multipass, which is fine since CanXRSDKUseSinglePass() handles XR SDK single-pass query. The logic is tested to work for both XR SDK and legacy XR.

thomas-zeng: Ok, I won't block this PR because of the bad coding pattern due to its priority. On top of it, this PR is a workaround and code change won't be part of the master. Mark this as resolved.

jackpr-unity:
Ran into a problem with WMR. It is a known issue that's being fixed elsewhere. https://unity.slack.com/archives/CT9FB3FEW/p1581009181005700

Oculus ran fine in both editor versions, and in both editor and playmode.

jackpr-unity:
Magic Leap looks good

phi-lira:
Forward ported to release/2019.3 in #5826

PR #5830, opened 2020-02-07
 Hotfix for Hololens Remote Rendering

phi-lira:
### Purpose of this PR
 Hotfix to improve performance when running Hololens Remote Rendering

---
### Comments to reviewers
 This is a very blind fix as I don't have much experience in XR and I don't have a Hololens device to test the fix.


joejo-unity:
You should probably loop in our QA as well. Talk to @mario or @JasonCostanza .

mario:
What did I do wrong this time? :D

joejo-unity:
@mario Sorry, wrong Mario! ;-)
 Ignore that.

@phi-lira I meant Mario Baumann.

JasonCostanza:
Is this for Legacy VR or XR SDK?

thomas-zeng:
I know this fix is to work around a driver bug for hololens. Could you provide more info on what driver bug we are trying to workaround? From looking at the code, seems like driver bug is related to intermediate color buffer?

phi-lira:
I'm not sure about current driver bug. To be honest this is a workaround for a HUP issue that degradates performance when taking the intermediate render texture path. 

We need to have a follow up meeting after package is released to track real root cause of issue.

thomas-zeng:
If this is not to work around a driver bug, then I think the issue we are fixing here is more generic. Intermediate render texture path is costly on mobile platforms so in XR that has an impact on oculus quest/go and wmr hololens.

I recall @grayson-unity  is working on fixing oculus extra blit which is also related to this intermediate texture.
Adding grayson to reviewer list.

PR #5836, opened 2020-02-10
Fix some missing init of stereo parameters in a Blit variant 

FrancescoC-unity:


fabien-unity:
is this new variant also present in master? If so, let's make sure the same PR is integrated.

PR #5839, opened 2020-02-10
Fix a source of Inf in Area lights LTC tables

FrancescoC-unity:
The LTC tables have values above the max allowed by half precision, that leads to Inf values being stored in the texture and consequently bad stuff happening on consoles. 

Not sure if we want this for the release or not, please let me know if I need to target anther branch.


TomasKiniulis:
HDRP_RuntimeTests scene works fine on Xbox. However, checking PS4 with @JMargevics together, produces a black screen, but the black screen occurs in 7.2.x/release branch when we checked before the merge of the PR
10 37 33 195_Image_6

sebastienlagarde:
I guess it is because an out of memory on PS4?

JMargevics:
Hey, i've checked the console output on PS4 when running that scene.

'(Filename: .\Runtime/Allocator/MemoryManager.cpp Line: 1448)
00:03:01	
00:03:02	[Syscore App] App Crash : PID=0x3d, reason=0xb
00:03:02	[Syscore App] Syscore Event Queue Push : SCE_SYSCORE_EVENT_APP_CRASH
00:03:02	[Syscore App] Syscore Event Queue Pop : SCE_SYSCORE_EVENT_APP_CRASH
00:03:02	[SceLncService] AppCrash: pid={0x0000003d} appId={0x60000b09} appLocalPid={0x10b090e0}'

So a memory issue after all. I can send memory coredumps via slack if needed or a full log. 

Weird thing it doesn't close the app and allows for a normal exit via ps menu
image

PR #5847, opened 2020-02-10
Add missing exposure and mipmap sliders to the planar reflection probe preview

alelievr:
### Purpose of this PR
Add missing exposure to the planar reflection probe preview
PlanarProbe


JMargevics:
Hey, Antoine, is it possible to have Mip levels sliders on Planar Reflection Probe? As in regular Reflection Probes.
image

alelievr:
I think the problem is more why the preview in the scene is not squared ? Because IIRC only square render texture can be used for planar probe

com.unity.render-pipelines.high-definition/Editor/RenderPipelineResources/Shaders/GUITextureBlit2SRGB.shader
```diff
@@ -51,9 +53,9 @@ Shader "Hidden/GUITextureBlit2SRGB" {
             fixed4 frag (v2f i) : SV_Target
             {
                 UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(i);
-                fixed4 colTex = UNITY_SAMPLE_SCREENSPACE_TEXTURE(_MainTex, i.texcoord);
+                fixed4 colTex = tex2Dlod(_MainTex, float4(i.texcoord, 0, _MipLevel));
```
sebastienlagarde: please use our macro :)

PR #5851, opened 2020-02-10
Diffuse Shadow Denoiser Improvement.

anisunity:
Added a new mode that improves the filtering of ray traced shadows (directional, point and spot) based on the distance to the occluder. This makes the contact lines for shadows sharper whe they are supposed to be (in opposition to previous over blurring)
ContentExample
ContentExample000


sebastienlagarde:
Do we have an update for this PR?

anisunity:
Fixed the bug related to the denoiser grabbing data outisde of the screen

anisunity:
Rebased, tests are green

PR #5852, opened 2020-02-10
Fix/switch

kaychang-unity:
---
### Purpose of this PR
This fixes NaN generated during 2 post-processing tonemapping algorithms (neutral and ACES).
The NaNs appeared when extremely bright pixels were tonemapped in scenes 091 and 093 of UniversalGraphicsTest.

In particular, there was a strange bug where sqrt(k) with k == 0 caused NaN to be generated. According to the Nintendo docs, this should not happen though.

The changes were guarded for Switch platform only, so it should not affect any other platforms.

---
### Testing status

- [ X ] Built a player


com.unity.render-pipelines.core/ShaderLibrary/ACES.hlsl
```diff
@@ -315,7 +315,12 @@ half rgb_2_yc(half3 rgb)
     half r = rgb.x;
     half g = rgb.y;
     half b = rgb.z;
-    half chroma = sqrt(b * (b - g) + g * (g - r) + r * (r - b));
+    half k = b * (b - g) + g * (g - r) + r * (r - b);
+#if defined(SHADER_API_SWITCH)
+    half chroma = k == 0.0 ? 0.0 : sqrt(k); // Fix NaN on Nintendo Switch (should not happen in theory).
```
phi-lira: Maybe we should add a SafeSqrt in Common.hlsl like we have SafeDiv and SafeNormalize? 
@EvgeniiG wdyt?

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
```diff
@@ -659,13 +663,23 @@ float3 AcesTonemap(float3 aces)
     // Luminance fitting of *RRT.a1.0.3 + ODT.Academy.RGBmonitor_100nits_dim.a1.0.3*.
     // https://github.com/colour-science/colour-unity/blob/master/Assets/Colour/Notebooks/CIECAM02_Unity.ipynb
     // RMSE: 0.0012846272106
+#if defined(SHADER_API_SWITCH) // Fix floating point overflow on extremely large values.
```
phi-lira: should we also add SHADER_API_MOBILE? If this is because of precision issues most likely android and iOS would benefit from this change.

EvgeniiG:
Regarding the sqrt.

Unless Switch is not IEEE 754 compliant, sqrt(0) is well-defined and should produce Inf. It may be a bit weird, but that's what the standard implementation does.

image

Now, math involving Inf is tricky, and it can easily produce a NaN:

image

I am not opposed to adding SafeSqrt, s.t.

`// Avoid (sqrt(x < 0) == NaN) and (sqrt(x = 0) == Inf).
real SafeSqrt (real x)
{
	return sqrt(max(x, REAL_MIN));
}`

Although I am not sure how necessary that is given that we haven't encountered this problem until now.

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
EvgeniiG: Again, if they use IEEE 754, then all standard-compliant FP16 implementations may exhibit this issue (of course, whether the issue shows us may depend on compiler optimizations). So perhaps we should either always run these in full precision or, alternatively, find the spot where we should clamp to HALF_MAX to avoid Inf.

com.unity.render-pipelines.core/ShaderLibrary/ACES.hlsl
EvgeniiG: See my comment below.

kaychang-unity: The thing is I think sqrt(0) should work fine, I'm not sure why it failed in that case on Switch.

com.unity.render-pipelines.core/ShaderLibrary/Color.hlsl
kaychang-unity: I tested the same test scene on iOS (091 and 093 of UniversalGraphicsTest project) and it was fine, so I left it as Switch only issue.

kaychang-unity:
I did some testing and sqrt(x) with x = 0 (through an uniform) does not generate NaN.
My guess so far is that adding the code "k == 0.0 ? 0.0 : sqrt(k)" may force the shader compiler(s) to skip some optimization leading to NaN generation. This is difficult to tell since the original shader code goes through so many translation translation steps (DXC compiler, then HLSLcc, then Switch shader compiler).

PR #5862, opened 2020-02-11
Render pipeline editor settings refactor [ For Review/Discussion only for now ]

Thomasch-unity3d:

### Checklist for PR maker

---
### Purpose of this PR
The current RenderPipelineEditorResources system causes problems with some asset importers, specifically when assets are imported when Unity starts ( when upgrading or after deleting the Library folder for instance).
AssetDatabase.LoadAssetAtPath() returns null in this case, so importers that need to fetch a resource from RenderPipelineEditorResources at that point can't find it.
The symptoms of this problem are :
- The modelImporter falls back to the Legacy Render Pipeline's default material When MaterialImportMode == None
- SpeedTree always fall back to the Legacy Render Pipeline's speedTree shaders.
- AutodeskInteractive materials are falling back to the legacy render pipeline default shader.
- Maybe more ? Terrain, particles maybe ?
Those issues affect both URP and HDRP and must have been present for quite a while..

Moreover, there are unresolved import dependencies on those resources. For instance when the 'Lit material' in UniversalRenderPipelineEditorResources is changed, Model importers that depend on it should be reimported automatically to make sure they remain deterministic.

Also, the current implementation does not seem ideal because it's unclear which RenderPipelineEditorResources asset is actually used and they can be moved or deleted, etc... Moving those to the ProjectSettings with the proposed implementation ensures that there's only one instance in a specific and immutable location.

Finally, from a UX standpoint (in my opinion, of course that's debatable), it seems more logical to find those settings in the Project Settings window.

This PR fixes the issues enumerated above ( including https://fogbugz.unity3d.com/f/cases/1141488/ for reference ) and I did not have any issue when adding new fields in RenderPipelineEditorResources and simulating a package upgrade.

One remaining issue is the migration of existing projects where users did change the values in RenderPipelineEditorResources, those changes do not transfer to the new version. This does not seem to be an issue in HDRP but can be a problem for URP users.

I tried several ways to fix those issues ( well, actually this one https://fogbugz.unity3d.com/f/cases/1141488/ ) and that is so far the only workable solution I found so please don't discard it too hastily :) Thanks for your feedback !

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
@@ -1,4 +1,9 @@
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEditor;
+using UnityEditor.Experimental;
+using UnityEditorInternal;
 using UnityEngine.Scripting.APIUpdating;
 
 namespace UnityEngine.Rendering.LWRP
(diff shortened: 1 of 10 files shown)


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
```diff
@@ -9,73 +14,159 @@ public class LightweightRenderPipelineEditorResources
...
+                {
+                    Instance = objects.First() as UniversalRenderPipelineEditorResources;
+                    SaveSettings();
+                }
+                else
+                {
+                    //Was removed from package or other error, generate one from scratch ??
+                }
+
+                UpdateImportDependency(RenderPipelineAsset.DefaultMaterialImportDependency, Instance.materials.lit);
```
phi-lira: what about `defaultParticleMaterial` and `defaultTerrainMaterial`?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
```diff
@@ -9,73 +14,159 @@ public class LightweightRenderPipelineEditorResources
...
+            EditorGUILayout.PropertyField(property);
+            if (EditorGUI.EndChangeCheck())
+                UniversalRenderPipelineEditorResources.UpdateImportDependency(dependencyKey, property.objectReferenceValue);
+        }
 
-            // Add a "Reload All" button in inspector when we are in developer's mode
-            if (UnityEditor.EditorPrefs.GetBool("DeveloperMode") && GUILayout.Button("Reload All"))
+
+        [SettingsProvider]
+        public static SettingsProvider CreateRenderPipelineEditorResourcesProvider()
```
phi-lira: If I understand correctly we are enabling users to change resources with this. If so, I'm a little bit worried about exposing to this to users. What's the motivation to do it?

phi-lira:
Thanks for looking at this issue! I'm adding extra reviewers for coverability:
 @Chman for ResourceReloaded
 @Verasl for URP resources coverage

phi-lira:
Adding @Unity-Technologies/gfx-qa-foundation for testing:
We need to validate no regressions:
   - Create new pipeline asset and assign it to graphics settings. Enter playmode. Should render with no errors and all resources initialized correct.
   - Creating new renderer. Assign the new renderer to pipeline asset. Enter playmode. Should render with no errors and all resources initialized correct.
   - Upgrade UniversalRenderingExamples project to this PR. Test all scenes are rendering correctly with all resources.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
Thomasch-unity3d: Yes, that has to be added. I didn't find any reference to that in any Native Importer but thinking of it, since any of those properties can be used in ScriptedImporters, we need to declare import dependencies for all of them.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
Thomasch-unity3d: Well it looked like it's already possible to do so by copying the UniversalRenderPipelineEditorResources.asset in the project ( see LoadResourceFile<T>() ) so I tried to keep that.
If RenderPipelineEditorResources is not meant to be editable, then I guess we don't need it to be a scriptableObject, it could be just a static list of GUIDs. In this case, we don't even need to add import dependencies and the issue with LoadAssetAtPath would not happen.
It's up to you really, I think there's some value in being able to change the 'default material' for instance but probably not for others like 'Speed tree shaders'. We could show only a subset of those properties in the UI.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineEditorResources.cs
Thomasch-unity3d: > Adding @Unity-Technologies/gfx-qa-foundation for testing:
> We need to validate no regressions:
> 
> * Create new pipeline asset and assign it to graphics settings. Enter playmode. Should render with no errors and all resources initialized correct.
> * Creating new renderer. Assign the new renderer to pipeline asset. Enter playmode. Should render with no errors and all resources initialized correct.
> * Upgrade UniversalRenderingExamples project to this PR. Test all scenes are rendering correctly with all resources.

I forgot that this needs a few changes in Unity for the import dependencies to work and I also referenced variables that I introduced in trunk. I'll push my Unity branch and post a link to a build asap so you guys can test.

RSlysz:
I just want to highlight that the ResourceReloader initially was added to fix an issue regarding serialization.
The issue was, if an asset exist, and if its structure change (this happens when you upgrade the package and the asset now should have new fields), the serialized data is not accurate (new fields are null). So in this case the idea was to check the resources asset and  reload the null fields.

Please be sure the change do not introduce regression on this case (package upgrade).

Thomasch-unity3d:
> I just want to highlight that the ResourceReloader initially was added to fix an issue regarding serialization.
> The issue was, if an asset exist, and if its structure change (this happens when you upgrade the package and the asset now should have new fields), the serialized data is not accurate (new fields are null). So in this case the idea was to check the resources asset and reload the null fields.
> 
> Please be sure the change do not introduce regression on this case (package upgrade).

Yes I understand that, I did test for the case where RenderPipelineEditorResouces' structure changes by introducing new serialized properties in that class , when the project is opened, closed and even when it's closed and the library folder was deleted. No problem spotted but of course I might have missed something and yes, extensive testing is required here.
If you have any hints about how to best reproduce the original issue, please shoot. Thanks!

RSlysz:
> If you have any hints about how to best reproduce the original issue, please shoot. Thanks!

It was when we upgraded from a HDRP version A to a later version B that include new fields in resources using the package manager, there was a lot of error displayed in the console and the fields remained to null.

First time we introduced the ResourceReloader, there was still the error in the console (due to AssetDatabase.LoadAssetAtPath() returning null) but at the end of package import, the fields were not null anymore.

I discussed with AssetDataBase team to have a boolean to know if we can import or not (in order to only construct the render pipeline when we can load the required resources). I dunno exactly why they do not want to do it like this. So after discussion it is the current state:
- ResourceReloader try to load the resource. it can detect if the AssetDatabase is ready by trying to get the GUID (if this not work anymore, this is a regression on AssetPipeline). While it detect it, it send a error message with additional data that is handled at Render Pipeline creation and the render pipeline creation and usage is just delayed to next frame. So finally there was no error in console anymore and the RenderPipeline actually got constructed and run well with all its resources.

Now I agree that this pattern do not work well with lazy init. All resources must be loaded at Render Pipeline creation time. It is the case in HDRP but I had trouble to implement the fix in Universal due to some case where they used lazy init. Up to my tests and the review I was thinking it was ok but apparently not.

So this is why I asked if the changes are required in HDRP too, and why the Editor resources need the fix but not the Runtime resources (there is two assets one editor and one runtime)

sebastienlagarde:
@Thomasch-unity3d 
Hi, I just see the change happening in this PR. I don't have the full context, but could you elaborate why do you want to change the system of HDRP for reload resources?

We have get two years of issues / bugfix / fight / users pain bug etc... To reach the current state of the system in HDRP that is finally robust for us. I really don't want to change something that works and have cost us so many man/hour to fix and workaround various low level unity problem.

So unless there is a strong motivation with: There is this fogbugz or this users project that don't work with it, then we shouldn't change it.
Local testing isn't robust testing, all our issues have come from users, it was always working for us when we were doing change on the system until we deliver the packages and many users was starting complaining that something wasn't working. So any test should be in "real condition". i.e we have various projects out: Spaceship, VR labs, Fontainbleau, VFX samples... Try to add or remove resources in the code, create a local package, then try to upgrade project from one package to another when the project is open, when it is not open, when you have delete your library or not delete, when you have database v1 or v2, when you upgrade from 19.2 to 19.3, when you upgadre from 19.2 to 2020.1 etc... Try to build player also.

Thomasch-unity3d:
The reloadResource system still has issues when upgrading projects apparently, I think the root cause is those calls to LoadAssetAtPath<RenderPipelineEditorResources> in the getter methods. Those getters end up being called during import of other native importers and although I'm not entirely sure why, the AssetDatabase is 'not allowed' to load the .meta and returns null.
Using InternalEditorUtility.LoadSerializedFileAndForget instead seems to solve the problem by kinda bypassing the ADB and reading the content of the file directly. That's why I removed the Reload attributes from there, it does not seem to be necessary anymore ( of course you're right, it needs much more testing to make sure it's fine in all cases). Also, to be honest, I'm not a fan of the 'if it does not work, try again later' strategy of the ResourceReloader but I understand that it's there for good reasons and that you guys don't want to reopen that can of worms :)
I 'll revert all changes related to ResourceReloader but I think it's worth exploring the solution with the InternalEditorUtility.LoadSerializedFileAndForget changes.

RSlysz:
Could it be simply to "fix" the ResourceReloader by using the InternalEditorUtility if we are in `#if UNITY_EDITOR`?

Thomasch-unity3d:
> Could it be simply to "fix" the ResourceReloader by using the InternalEditorUtility if we are in `#if UNITY_EDITOR`?

Yeah I guess that could do the trick, I'll give it a try.

PR #5863, opened 2020-02-12
Shader Graph quick fix for unreported Keyword bug

zadi0648:
**Summary:**

Fixed an unreported bug where dragging keywords in GV would create a duplicate of the keyword on the blackboard.

---
**Manually Tested:**

* Adding Keywords & Properties of various types/modes to the graph via dragging
* Also across graphs

---
**Technical Risk:** 0/4
**Halo Effect:** 0/4

---
**Notes to QA:**
* I have already tested and would say that this is one of those fixes that you don't need to test at all. Added you nonetheless for awareness/habit.

---
**Yamato:**
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fyour-branch-dude


LandonTownsendUnity:
This needs to be backported to 8.x.x

PR #5870, opened 2020-02-12
Changed settings icon with the text "Inspect" [Skip CI]

Nzollner:
### Purpose of this PR

The issue was the settings icon used for a button to inspect a UniversalRenderPipelineAsset_Renderer was blurry on lower resolution screens as seen in the image bellow:
CogWheelBlurred

Fixed the issue by making the button display text which says "Inspect" instead of using the settings icon as seen in the image bellow:
Capture

---
### Testing status

- [ ] Built a player
- Other:


martint-unity:
I dont think that text is the right way of doing this. If the cogwheel isn't ok then try the icon with three dots instead. To keep consistency.
_Menu.png from resources. Try that

phi-lira:
@Nzollner can you try the approach above suggested by Martin?

Nzollner:
Yes I will look into that :)

Nzollner:
I have tried to change it and this is what is looks like:
image

PR #5871, opened 2020-02-12
Fix cookie index usage

alelievr:
## Purpose of this PR
Fixed potential issues with cookie index usage instead of cookie mode.

Ran the DXR test locally, no issue found
image


anisunity:
i'll add a test scene in your PR and make sure that the behavior is good. @sebastienlagarde Would it be possible to wait for me to do that before merge?

PR #5873, opened 2020-02-12
Fix player build with dx12 and dxr

anisunity:
It was not possible to build a player with either DX12 or DXR due to multiple issues:
-  SubsurfaceDXR shaders nto stripped
- LODCross fade clip not surrounded by a ifndef SHADER_STAGE_RAY_TRACING


anisunity:
Fixed an issue with the multi compile instancing and ray tracing shaders

anisunity:
Retargeting this to 8.x.x

PR #5875, opened 2020-02-13
Fix Material Samples : Revert correct GUIDs

laurenth-personal:
Change the GUIDs :
- material ball prefab
- fern material
to fix the missing prefab and material in the material samples scene.


iM0ve:
Currently has a missing material
image

iM0ve:
This needs to be merged into master. Because on master 9.0.0 it looks like this without this PR

PR #4495, opened 2019-09-05
Bump version to 7.2.0

sebastienlagarde:
Bump version to 7.2.0

PR #5881, opened 2020-02-13
7.3.0 Version Bump

sophiaaar:
### Purpose of this PR
* Version bump to 7.3.0
* Reorganise package dependency tests to workaround a upm-ci issue

---
### Testing status

**Yamato**: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/ci%252F20193-dependency-tests/.yamato%252Fupm-ci-packages.yml%2523all_package_ci/1316211/job/pipeline

The HDRP and LWRP failures are expected for now. These will not block development.

---
### Comments to reviewers


com.unity.render-pipelines.core/Runtime/Documentation.cs
```diff
@@ -15,7 +15,7 @@ public class DocumentationInfo
         /// <summary>
         /// Current version of the documentation.
         /// </summary>
-        public const string version = "7.1";
```
GrantLamb-Unity: Was this change intentionally not updated when we bumped the version to 7.2.0? Or was this a bug? 

I didn't see discussion around this when I skimmed the old PR related to this: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/4495

sophiaaar: i think this was just a mistake that it got missed for 7.2.0

PR #5888, opened 2020-02-13
Fix metal manual dof clear

victor-unity:
Clear farCoc UAV in a compute shader.

The bug :
image

Why is it occurring :

1. When we request a clear on a RT, the GPU does not need to actually clear the pixels. When we want to clear an UAV, the GPU does not know which pixel will be touched or not. On Tile based architecture GPU, the Hardware clear does not work for UAV.
2. Here the clear is postponed until the next drawcall that occurs after the compute that use the UAV, so the work if lost, the UAV is black.

image

image
we can see that the actual order of clear id not the one intended by the source code.

**Solution** is to **clear manually in a compute** the pixels that won't be affected.  

---
### Purpose of this PR
fix for https://fogbugz.unity3d.com/f/cases/1216799/
Clear the UAV farPong that will be used to blur/bokeh the far background.
---
### Testing status

- [x] Built a player
- Tested iOS, MacOs and Win64:


victor-unity:
A fix in the gfx metal here : https://ono.unity3d.com/unity/unity/changeset/7b58780437677beaa04da9205d2af163031dec0e
has fixed the issue by forcing a clear in that specific case. So I close the PR as it is not needed anymore.

PR #5897, opened 2020-02-14
Fix issues with ray traced sub surface scattering

anisunity:
Fix Ray Traced SubSurface Scattering not working with ray traced area lights
Fix Ray Traced SubSurface Scattering not working in forward mode.


anisunity:
As discussed, this is fixed on staging by an other PR

PR #5900, opened 2020-02-15
[backport 7.x] prevent creation of intermediate color texture for any stereo use case

grayson-unity:

### Checklist for PR maker

---
### Purpose of this PR
This change avoids an unnecessary intermediate color texture and full-screen blit on the Oculus Quest (and hopefully all XR devices).  This also allows FFR to work with GLES on Quest.

---
### Testing status

- [X] Built a player
- Other: 

---
### Comments to reviewers
Could you please double-check that I haven't broken other stereo devices?  Thanks!

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
+++ b/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
@@ -513,7 +513,7 @@ namespace UnityEngine.Rendering.Universal
             ref CameraData cameraData = ref renderingData.cameraData;
             int msaaSamples = cameraData.cameraTargetDescriptor.msaaSamples;
             bool isStereoEnabled = renderingData.cameraData.isStereoEnabled;
-            bool isScaledRender = !Mathf.Approximately(cameraData.renderScale, 1.0f);
+            bool isScaledRender = !Mathf.Approximately(cameraData.renderScale, 1.0f) && !cameraData.isStereoEnabled;
             bool isCompatibleBackbufferTextureDimension = baseDescriptor.dimension == TextureDimension.Tex2D;
             bool requiresExplicitMsaaResolve = msaaSamples > 1 && !SystemInfo.supportsMultisampleAutoResolve;
             bool isOffscreenRender = cameraData.targetTexture != null && !cameraData.isSceneViewCamera;
@@ -529,7 +529,8 @@ namespace UnityEngine.Rendering.Universal
                 return requiresBlitForOffscreenCamera;
 
             return requiresBlitForOffscreenCamera || cameraData.isSceneViewCamera || isScaledRender || cameraData.isHdrEnabled ||
-                   !isCompatibleBackbufferTextureDimension || !cameraData.isDefaultViewport || isCapturing || Display.main.requiresBlitToBackbuffer;
+                   !isCompatibleBackbufferTextureDimension || !cameraData.isDefaultViewport || isCapturing ||
+                   (Display.main.requiresBlitToBackbuffer && !isStereoEnabled);
         }
 
         bool CanCopyDepth(ref CameraData cameraData)


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -175,6 +175,7 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
 #endif
             bool createColorTexture = RequiresIntermediateColorTexture(ref renderingData, cameraTargetDescriptor) ||
                 (rendererFeatures.Count != 0 && !isRunningHololens);
+            createColorTexture &= !isStereoEnabled;  // TODO: Could this replace isRunningHololens?
```
phi-lira: What if you have post-processing? or HDR (tonemapping happens on final blit pass).

phi-lira:
@thomas-zeng do all XR Android platforms support sRGB framebuffers? iirc that Display.main logic is because of this.
@ROBYER1 and @soraryu could you tell us what conditions are causing your game to create a intermediate render texture? If we just force disable intermediate render texture some features will stop working (HDR or post-processing f.ex), instead you could disable them for your game. 

One known issues causing the intermediate render texture is when you have any custom render pass. This is not the case anymore in upcoming 8.0.0 package.

ROBYER1:
@Phi-lira the example repro project is i ln Case 1215830 or you can find the base files here. It is a scene with a Speedtree, WorldSpace UI Canvas and a Shader graph material of some water using Depth. I believe the cause is one of those three items in the scene, however in an older LWRP and early URP version, the scene rendered just fine on the Quest and Go with MSAA and FFR on so there has been a regression somewhere.

Find the raw scene file here, should work with 2019 and 2020. I have been using it as a simple test scene for many bug reports.

https://drive.google.com/file/d/1nctZLAGQj_BqpVSRbJvOSv_vCYoNmZD8/view?usp=drivesdk

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
soraryu: I'd also say just globally disabling this if stereo is on will cause many pathes to fail. Good opportunity to create the necessary VR tests, guys!

ROBYER1:
@thomas-zeng @phi-lira So I investigated this issue a little more and I found the main issue is when using OpenGLES 3 with MSAA being on for URP on Quest prevents UI and anything using Depth pass from rendering!

See a video of it below. I believe Thomas mentioned before this is an Oculus Driver issue? I am still unable to test Vulkan Multiview on Quest in 2020.1.0a23 but OpenGLES3 has this issue.

https://drive.google.com/open?id=1Sby5MH4NjUXUjSfwSasI9q45V3rhl_em

soraryu:
@phi-lira the issue in the past was that Display.main.requiresBlitToBackbuffer returned true on Quest, which is incorrect. Also, SystemInfo.supportsMultisampleAutoResolve returned false for Quest, also wrong. I just tested this and at least currently this does not happen anymore, so something has been improved there.

Test on Quest just now on 7.2.0:
image

ROBYER1:
@thomas-zeng @phi-lira I have made a new bug report for you guys so that you can see even in a scene that just has a camera, no skybox and a worldspace UI, the Intermediate Color texture is produced and Fixed Foveated Rendering is broken.

Also as per the Oculus Driver issue, if you enable MSAA in this project, the UI and Shader graph in the scene will not render.

Please can you look into this Case to tell me what is producing the Color texture even in the almost empty 'Intro' scene I created for this bug report

(Case 1220858) [XR][URP][Oculus Quest] Intermediate Color Texture is produced even when HDR and Postprocessing are off

thomas-zeng:
@phi-lira 
Not all XR platform supports sRGB framebuffer. However, they all support sRGB eye texture. The logic check here for Display.main only makes sense for non-XR case. When XR is enabled, "back buffer"(the one final blit will blit to) becomes XR eye target and thus should not check Display.main.

Also to point out, the current sRGB handling in final blit pass needs a polish for XR.
`bool requiresSRGBConvertion = Display.main.requiresSrgbBlitToBackbuffer;`
This check doesn't work with XR because XR target is eye texture. We are seeing both WMR and Oculus not handling sRGB correctly and we are actively tracking the bugs. 
We should have PR soon fixing finalblit pass sRGB issues for XR.

thomas-zeng:
@ROBYER1 
I haven't take a look at case 1220858 yet but I plan to send it to QA to verify and look into it this week. 
Just a couple sanity check:
If opaque texture is used in URP then the extra blit is required(it is required to generate the opaque texture). Could you confirm you have opaque off in the test project and extra blit is gone? You can go to quality settings and figure out which URP pipeline asset is used for android and in that URP pipeline asset, toggle off opaque texture.

Per Oculus driver bug. 
This driver bug only occurs if MSAA and Multiview are being used and opaque texture is being used at the same time.` (opaque texture requirement injected CopyColorPass in between draw opaque and draw trans, causing MSAAMultiview frame buffer discarded implicit MSAA data which should not happen according to spec).`
So if you disable opaque texture, UI and water should show up too.

In short, disable opaque texture in XR, you should have 1)UI&other trans obj show up and 2)no extra blit.

ROBYER1:
@thomas-zeng The Opaque Texture is disabled in my tests, you can use the repro case 1220858 with Opaque Texture switched off (Check the default pipeline asset for Android has it off as my submission may have had it on still). The extra blit is still happening due to FFR not working at all still.

I can confirm in the same repro project that Opaque Texture being off allows MSAA to work, however the extra blit is *still* there as FFR is not working at all still.

thomas-zeng:
@ROBYER1 Thanks! I put case 1220986 into my backlog and will take a look after QA verified the bug.

ROBYER1:
@thomas-zeng thankyou, and if there is anything I missed in the settings or if something else entirely is causing the FFR to not work due to the blit, please do let me know as it's a blocker for a current project.

ROBYER1:
@thomas-zeng @phi-lira Another forum user had the same issue on URP 7.2.0 and investigated using RenderDoc, the final blit is there regardless of them having anything that requires it and the Final Blit also has no render time?

I can "Verify no temporary buffers are used" - I'm rendering to RTDeviceEyeTextureArray with R8G8B8A8_SRGB - never to _CameraColorTexture or anything else.

It always has:
PlayerEndOfFrame
-XR.Blit

https://forum.unity.com/threads/quest-lwrp-urp-msaa-not-working.786026/page-2#post-5496472

phi-lira:
@grayson-unity looking at comments here and at code it seems to me for **Stereo** we should remove the following conditions:
  - Display.main.requiresBlitToBackbuffer
  - requiresExplicitMsaaResolve

thomas-zeng:
@grayson-unity @phi-lira 
Regarding XR MSAA:
In XR MSAA, we have 3 cases to cover
1. Implicit MSAA resolve on platform that supports auto-resolve
2. Explicit MSAA resolve(vs-ps workflow) on platforms that don't support auto-resolve
3. No resolve, XR eye texture is MSAA eye texture

Case 1: URP check if it can do auto-resolve, don't blit in case it can auto-resolve 
Case 2: URP check if can do auto-resolve, do explicit resolve pass if it can't auto-resolve
Case 3: Not currently supported. All current XR eye textures(legacyXR and XRSDK) must have sample 1. We want to support this use case in the future. Require trunk side XRSDK work.

grayson-unity:
@thomas-zeng I like that more specific test, thanks.  I'll test that locally today.

thomas-zeng:
Thanks grayson! 
I think this PR also fixes 
Extra blit https://fogbugz.unity3d.com/f/cases/1220986/ 
Extra blit: https://fogbugz.unity3d.com/f/cases/1220858/
So would be nice to include them in CHANGELOG.md fixed session.

grayson-unity:
Updated the PR with stricter logic to avoid postfx/HDR/etc regression.

ROBYER1:
@Thomas-zeng I can confirm this fixes these 2 cases reported by me:

Extra blit: https://fogbugz.unity3d.com/f/cases/1220986/
Extra blit: https://fogbugz.unity3d.com/f/cases/1220858/

@Phi-Lira @grayson-unity what's the ETA on this hitting Master branch for URP? It's a really vital fix for the Oculus Quest, Oculus Go and presumably Hololens and now allows URP to be used in production with FFR working properly due to the blit and no unnecessary blit killing performance.

thomas-zeng:
@phi-lira @grayson-unity 
This PR fixes HUP issue for oculus quest/go. Let's make sure this is part of 7.3.0 SRP

grayson-unity:
> LGTM. Please make sure to update CHANGELOG.md

Does this belong under 7.2.1, or has that ship already sailed?

phi-lira:
@grayson-unity all PR must target master first. Please change PR target. 
@ROBYER1 once all reviewers approve we will schedule to be merged to master.
@thomas-zeng We will backport this all the way to 7.x.x. I will update the PR backport tags. 

I just updated

phi-lira:
@grayson-unity please rebase this to master

grayson-unity:
@phi-lira https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5976
thanks!

phi-lira:
We are good to merge once QA approves. #5976 is the PR that targets master. changed this to backport.

phi-lira:
Where's the master PR for this one, is it the staging one?

grendell:
The master PR is now part of https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6036

PR #5909, opened 2020-02-17
Expose Asset properties and add the option to turn off PP & volume updates globally, 

ellioman:
### Purpose of this PR
This PR does three things:

1. Exposes various settings to users. Such as
1.1. Main Light Rendering Mode
1.2. Main Light Shadow Casting
1.3. Main Light Shadow Resolution
1.4. Additional Light Rendering Mode
1.5. Additional Light Shadow Casting
1.6. Additional Light Shadow Resolution
1.7. Soft Shadows

2. Adds the option to turn on/off Post Processing globally
3. Adds the option to turn on/off Volume updating per-renderer

---

### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
@@ -14,6 +14,7 @@ namespace UnityEditor.Rendering.Universal
             public static readonly GUIContent OpaqueMask = new GUIContent("Default Layer Mask", "Controls which layers to globally include in the Custom Forward Renderer.");
             public static readonly GUIContent defaultStencilStateLabel = EditorGUIUtility.TrTextContent("Default Stencil State", "Configure stencil state for the opaque and transparent render passes.");
             public static readonly GUIContent shadowTransparentReceiveLabel = EditorGUIUtility.TrTextContent("Transparent Receive Shadows", "When disabled, none of the transparent objects will receive shadows.");
+            public static readonly GUIContent shouldUpdateVolumesLabel = EditorGUIUtility.TrTextContent("Should Update Volumes", "When disabled, none of the volumes will be updated for this renderer");
         }
 
         SerializedProperty m_OpaqueLayerMask;
@@ -22,6 +23,7 @@ namespace UnityEditor.Rendering.Universal
         SerializedProperty m_PostProcessData;
         SerializedProperty m_Shaders;
         SerializedProperty m_ShadowTransparentReceiveProp;
+        SerializedProperty m_ShouldUpdateVolumesProp;
 
         private void OnEnable()
         {
@@ -31,6 +33,7 @@ namespace UnityEditor.Rendering.Universal
             m_PostProcessData = serializedObject.FindProperty("postProcessData");
             m_Shaders = serializedObject.FindProperty("shaders");
             m_ShadowTransparentReceiveProp = serializedObject.FindProperty("m_ShadowTransparentReceive");
+            m_ShouldUpdateVolumesProp = serializedObject.FindProperty("m_ShouldUpdateVolumes");
         }
 
         public override void OnInspectorGUI()
@@ -48,6 +51,10 @@ namespace UnityEditor.Rendering.Universal
 
             EditorGUILayout.LabelField("Shadows", EditorStyles.boldLabel);
             EditorGUILayout.PropertyField(m_ShadowTransparentReceiveProp, Styles.shadowTransparentReceiveLabel);
+            EditorGUILayout.Space();
+
+            EditorGUILayout.LabelField("Volumes", EditorStyles.boldLabel);
+            EditorGUILayout.PropertyField(m_ShouldUpdateVolumesProp, Styles.shouldUpdateVolumesLabel);
             EditorGUILayout.Space();
 
             EditorGUILayout.LabelField("Overrides", EditorStyles.boldLabel);
--- a/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
@@ -55,6 +55,7 @@ namespace UnityEditor.Rendering.Universal
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
+            public static GUIContent postProcessEnabled = EditorGUIUtility.TrTextContent("Enabled", "Enables or disables postprocessing. This is a global setting which affects every camera and renderer.");
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
             public static string colorGradingModeWarning = "HDR rendering is required to use the high dynamic range color grading mode. The low dynamic range will be used instead.";
@@ -130,6 +131,7 @@ namespace UnityEditor.Rendering.Universal
         SerializedProperty m_ShaderVariantLogLevel;
 
         LightRenderingMode selectedLightRenderingMode;
+        SerializedProperty m_PostProcessEnabled;
         SerializedProperty m_ColorGradingMode;
         SerializedProperty m_ColorGradingLutSize;
 
@@ -195,6 +197,7 @@ namespace UnityEditor.Rendering.Universal
 
             m_ShaderVariantLogLevel = serializedObject.FindProperty("m_ShaderVariantLogLevel");
 
+            m_PostProcessEnabled = serializedObject.FindProperty("m_PostProcessEnabled");
             m_ColorGradingMode = serializedObject.FindProperty("m_ColorGradingMode");
             m_ColorGradingLutSize = serializedObject.FindProperty("m_ColorGradingLutSize");
 
@@ -345,6 +348,7 @@ namespace UnityEditor.Rendering.Universal
 
                 EditorGUI.indentLevel++;
 
+                EditorGUILayout.PropertyField(m_PostProcessEnabled, Styles.postProcessEnabled);
                 EditorGUILayout.PropertyField(m_ColorGradingMode, Styles.colorGradingMode);
                 if (!isHdrOn && m_ColorGradingMode.intValue == (int)ColorGradingMode.HighDynamicRange)
                     EditorGUILayout.HelpBox(Styles.colorGradingModeWarning, MessageType.Warning);
--- a/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
+++ b/com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
@@ -55,7 +55,7 @@ namespace UnityEditor.Rendering.Universal
             public static GUIContent supportsSoftShadows = EditorGUIUtility.TrTextContent("Soft Shadows", "If enabled pipeline will perform shadow filtering. Otherwise all lights that cast shadows will fallback to perform a single shadow sample.");
 
             // Post-processing
-            public static GUIContent postProcessEnabled = EditorGUIUtility.TrTextContent("Enabled", "Enables or disables postprocessing. This is a global setting which affects every camera and renderer.");
+            public static GUIContent postProcessEnabled = EditorGUIUtility.TrTextContent("Enabled", "Enables or disables postprocessing. This is a global setting which affects all cameras and renderers.");
             public static GUIContent colorGradingMode = EditorGUIUtility.TrTextContent("Grading Mode", "Defines how color grading will be applied. Operators will react differently depending on the mode.");
             public static GUIContent colorGradingLutSize = EditorGUIUtility.TrTextContent("LUT size", "Sets the size of the internal and external color grading lookup textures (LUTs).");
             public static string colorGradingModeWarning = "HDR rendering is required to use the high dynamic range color grading mode. The low dynamic range will be used instead.";
--- a/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
@@ -495,11 +495,17 @@ namespace UnityEngine.Rendering.Universal
         public float cascade2Split
         {
             get { return m_Cascade2Split; }
+            set { m_Cascade2Split = Mathf.Clamp01(value); }
         }
 
         public Vector3 cascade4Split
         {
             get { return m_Cascade4Split; }
+            set {
+                m_Cascade4Split.x = Mathf.Clamp01(value.x);
+                m_Cascade4Split.y = Mathf.Clamp01(value.y);
+                m_Cascade4Split.z = Mathf.Clamp01(value.z);
+            }
         }
 
         public float shadowDepthBias
@@ -737,7 +743,7 @@ namespace UnityEngine.Rendering.Universal
 
         int ValidatePerObjectLights(int value)
         {
-            return System.Math.Max(0, System.Math.Min(value, UniversalRenderPipeline.maxPerObjectLights));
+            return Math.Max(0, Math.Min(value, UniversalRenderPipeline.maxPerObjectLights));
         }
 
         float ValidateRenderScale(float value)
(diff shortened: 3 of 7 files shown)


com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -473,11 +485,20 @@ public int maxAdditionalLightsCount
         public bool supportsAdditionalLightShadows
         {
             get { return m_AdditionalLightShadowsSupported; }
+            set { m_AdditionalLightShadowsSupported = value; }
         }
 
         public int additionalLightsShadowmapResolution
         {
             get { return (int)m_AdditionalLightsShadowmapResolution; }
+            set {
```
phi-lira: Same as comment above for shadowmap resolution in main light.

com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
```diff
@@ -31,6 +33,7 @@ private void OnEnable()
             m_PostProcessData = serializedObject.FindProperty("postProcessData");
             m_Shaders = serializedObject.FindProperty("shaders");
             m_ShadowTransparentReceiveProp = serializedObject.FindProperty("m_ShadowTransparentReceive");
+            m_ShouldUpdateVolumesProp = serializedObject.FindProperty("m_ShouldUpdateVolumes");
```
phi-lira: should this be applied to all renderers? Maybe move to asset? @kaychang-unity and @unity-cchu what do you think?

phi-lira: Also @Chman for awareness. It seems the motivation for this was users having performance issues in low end platforms due to always calling update volume framework. 

Can we cache it instead somehow?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -52,8 +52,6 @@ internal class Styles
 
             public static readonly GUIContent targetTextureLabel = EditorGUIUtility.TrTextContent("Output Texture", "The texture to render this camera into, if none then this camera renders to screen.");
 
-            public static readonly GUIContent cameraStackNotSupportedMessage = EditorGUIUtility.TrTextContent("Camera Stacking not supported.", "The renderer used by this camera doesn't support camera stacking.");
```
phi-lira: Did this get removed by mistake?

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
```diff
@@ -805,15 +803,15 @@ void DrawRenderer()
         void DrawPostProcessing()
         {
             bool hasChanged = false;
-            bool selectedRenderPostProcessing;
+            CameraOverrideOption selectedRenderPostProcessing;
```
phi-lira: What is the problem this is solving?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -447,21 +448,32 @@ public float renderScale
...
         public bool supportsMainLightShadows
         {
             get { return m_MainLightShadowsSupported; }
+            set { m_MainLightShadowsSupported = value; }
         }
 
         public int mainLightShadowmapResolution
         {
             get { return (int)m_MainLightShadowmapResolution; }
+            set {
```
phi-lira: Hum, I think I remember why this wasn't exposed. We thought about instead of exposing setting a fixed resolution, user configure in the asset resolutions for low, mid or high in the pipeline asset then here he would choose at runtime between low, mid, high quality shadows.

This scales better as user can configure per-quality settings values for each platform and at runtime switch between quality levels?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -447,21 +448,32 @@ public float renderScale
...
+                    return;
+                }
+                m_MainLightShadowmapResolution = (ShadowResolution) value;
+            }
         }
 
         public LightRenderingMode additionalLightsRenderingMode
         {
             get { return m_AdditionalLightsRenderingMode; }
+            set { m_AdditionalLightsRenderingMode = value; }
```
phi-lira: We should not change this. Impacts shader stripping.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -473,11 +485,20 @@ public int maxAdditionalLightsCount
         public bool supportsAdditionalLightShadows
         {
             get { return m_AdditionalLightShadowsSupported; }
+            set { m_AdditionalLightShadowsSupported = value; }
```
phi-lira: This also impacts shader stripping.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
phi-lira: Isntead we for additional lights we should actually expose low, mid, high per light as well as options to how to pack in shadow atlas.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -495,11 +516,17 @@ public ShadowCascadesOption shadowCascadeOption
...
+            set { m_Cascade2Split = Mathf.Clamp01(value); }
         }
 
         public Vector3 cascade4Split
         {
             get { return m_Cascade4Split; }
+            set {
+                m_Cascade4Split.x = Mathf.Clamp01(value.x);
+                m_Cascade4Split.y = Mathf.Clamp01(value.y);
+                m_Cascade4Split.z = Mathf.Clamp01(value.z);
```
phi-lira: We should validate cascades don't extrapolate each other.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -495,11 +516,17 @@ public ShadowCascadesOption shadowCascadeOption
         public float cascade2Split
         {
             get { return m_Cascade2Split; }
+            set { m_Cascade2Split = Mathf.Clamp01(value); }
```
phi-lira: Should validate value is between 0-1

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -517,6 +544,7 @@ public float shadowNormalBias
         public bool supportsSoftShadows
         {
             get { return m_SoftShadowsSupported; }
+            set { m_SoftShadowsSupported = value; }
```
phi-lira: cannot change this, affect shader stripping

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -547,6 +575,12 @@ public bool useSRPBatcher
             set { m_UseSRPBatcher = value; }
         }
 
+        public bool postProcessEnabled
```
phi-lira: What problem exposing this is solving? do we strip pp resources?

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -737,7 +771,7 @@ float ValidateShadowBias(float value)
 
         int ValidatePerObjectLights(int value)
         {
-            return System.Math.Max(0, System.Math.Min(value, UniversalRenderPipeline.maxPerObjectLights));
+            return Math.Max(0, Math.Min(value, UniversalRenderPipeline.maxPerObjectLights));
```
phi-lira: iirc there was some namespace conflict about this in some platform, I'd say to not change it.

com.unity.render-pipelines.universal/Runtime/UniversalAdditionalCameraData.cs
```diff
@@ -311,8 +311,19 @@ public Transform volumeTrigger
         /// </summary>
         public bool renderPostProcessing
         {
-            get => m_RenderPostProcessing;
-            set => m_RenderPostProcessing = value;
+            get {
```
phi-lira: I think in the end this is more confusing, users already have API to disable post-processing per camera. If we would use this to strip pp assets/shaders from build then it would be more useful, but in that case 1) should be best if a separate stripping settings like lightmap modes / fog stripping 2) no API to change it at runtime

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -447,21 +448,32 @@ public float renderScale
         public LightRenderingMode mainLightRenderingMode
         {
             get { return m_MainLightRenderingMode; }
+            set { m_MainLightRenderingMode = value; }
         }
 
         public bool supportsMainLightShadows
         {
             get { return m_MainLightShadowsSupported; }
+            set { m_MainLightShadowsSupported = value; }
```
phi-lira: cannot change touches shader stripping

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
```diff
@@ -447,21 +448,32 @@ public float renderScale
         public LightRenderingMode mainLightRenderingMode
         {
             get { return m_MainLightRenderingMode; }
+            set { m_MainLightRenderingMode = value; }
```
phi-lira: cannot change touches shader stripping

com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
Chman: We can't really cache it, that would defeat the purpose of having volumes in the first place (local volumes, blending etc). A global toggle to disable it sounds fine to me in the context of URP.

Also one thing to keep in mind when profiling volumes: they can get a bit sluggish in debug mode (build or editor) as it's doing quite a bit of validation etc but they are around 10x faster when building in release mode (for instance on PS4: from 1ms to 0.1ms on the CPU).

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
```diff
@@ -16,7 +16,7 @@ internal class Styles
             public static GUIContent qualitySettingsText = EditorGUIUtility.TrTextContent("Quality");
             public static GUIContent lightingSettingsText = EditorGUIUtility.TrTextContent("Lighting");
             public static GUIContent shadowSettingsText = EditorGUIUtility.TrTextContent("Shadows");
-            public static GUIContent postProcessingSettingsText = EditorGUIUtility.TrTextContent("Post-processing");
+            public static GUIContent postProcessingSettingsText = EditorGUIUtility.TrTextContent("Post Processing");
```
Chman: Why the renaming here? :)
The actual English term is `Post-processing`, it was correct.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
Chman: I kinda like having the ability to disable PP project-wide. We have it in HDRP and it's quite handy. As long as we hide the UI in cameras when it's disabled, I'm all for it.

com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
kaychang-unity: The deferred renderer should also expose the same parameters (m_ShouldUpdateVolumesProp).

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
ellioman: No, this was an unused variable.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
ellioman: I was just talking to @martint-unity and we thought it made more sense to have it as "Post Processing" to be consistent with how we write it in other places.

URP Asset: "Post-processing"
Camera inspector: "Post Processing"
Forward Renderer: "Post Process Data"
Package Manager: "Post Processing"

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Reverted.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Sounds good. I'll revert this and we'll do that in a separate PR.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Agreed. Reverting this one as well and we'll do that in a separate PR.

com.unity.render-pipelines.universal/Runtime/UniversalAdditionalCameraData.cs
ellioman: Yes, you're right. I'm going to revert this and we'll just have a toggle in the asset to disable PP globally for stripping.

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineCameraEditor.cs
ellioman: Reverting this. Will just have a toggle in the asset so we can strip.

com.unity.render-pipelines.universal/Runtime/Data/UniversalRenderPipelineAsset.cs
ellioman: Is clamping it not enough? Do you want an Assert there as well or...?

com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
Verasl: > We can't really cache it, that would defeat the purpose of having volumes in the first place (local volumes, blending etc). A global toggle to disable it sounds fine to me in the context of URP.
> 
> Also one thing to keep in mind when profiling volumes: they can get a bit sluggish in debug mode (build or editor) as it's doing quite a bit of validation etc but they are around 10x faster when building in release mode (for instance on PS4: from 1ms to 0.1ms on the CPU).

In this case should Volumes themselves have a debug mode? or maybe hook into our new debug level in URP? I wander now if most of the bugs we get about volumes being super slow are due to this...

com.unity.render-pipelines.universal/Editor/UniversalRenderPipelineAssetEditor.cs
Verasl: Probably means we need to change it everywhere else

com.unity.render-pipelines.universal/Editor/ForwardRendererDataEditor.cs
Chman: It goes further down that just the volume system doing some validation. Most of the speed gain is due to compiling for deployment instead of development, IL2CPP is just a lot faster overall when not in debug and this benefits the volume system greatly.
Just a side-note, but with the move to DOTS this system will be reworked eventually as we'll have access to more goodies compared to when this was initially written.

PR #5911, opened 2020-02-17
Fix inverted depth sampling

phi-lira:
### Purpose of this PR
- Fix y-flip issue when sampling depth texture and not rendering to render texture
- Fix issue that some shaders variables to not be defined for shadow caster pass (or any pass before opaques). Causing shadow pass to be one frame behind. 

---
### Testing status

**Manual Tests**: 
Added test scene to graphics project. Tested all combinations of (depth-prepass, depth copy) vs (render texture, screen).

- [ ] Built a player
- Other: 

**Automated Tests**: To be added

---
### Comments to reviewers
 I've re-purposed `_ProjectionParams` -> `unity_ProjectionParams` because Unity overwrites `_ProjectionParams.x` in several places. 
`_ProjectionParams` is still defined as `unity_ProjectionParams` so shaders don't break.


phi-lira:
Next iteration: #5913

PR #5913, opened 2020-02-18
 Fix y-flip issues in the pipeline

phi-lira:
### Purpose of this PR
- Handle camera matrices in URP side. Pre-work for Pure XR SDK.
- Fix y-flip issue when sampling depth texture and not rendering to render texture [case 1225362](https://issuetracker.unity3d.com/issues/game-object-is-rendered-incorrectly-in-the-game-view-when-sampling-depth-texture)
- Added support to 8 MSAA in Copy Depth

### Comments to reviewers
  All this weird y-flip logic handling is only required because we use CommandBuffer.Blit and CommandBuffer.SetViewProjectionMatrices. There is already another PR to remove this logic, but until then we fix it by adding a scale + bias on copy depth shader to unflip depth texture when rendering to RT.

---
### Testing status
**Manual Tests**: 
 - Tested all combinations of (depth-prepass, depth copy) vs (render texture, screen) in newly added test scene 126 that samples depth. 
 - Tested with all the examples of UniversalRenderingExamples
 - Tested with BoatAttack
 - Tested with DX11 and OpenGL (no y-flip)
 - Tested XR rendering using GfxStereoTest.

**Automated Tests**: 

- Added test scene 126 that tests sampling depth texture. The scene was added to both Graphics and GraphicsStereo Tests.

126_SampleDepth

126_SampleDepth

 - Added a script to GfxStereo tests to copy reference images to right folder (CopyImageToReferenceFolder.cs)

Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fbugfix%252Fflipdepth/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1475841/job

### Comment to Reviewers
 - Moved the `RenderingData.resolveFinalTarget` to `CameraData.resolveFinalTarget`. This setting should be per-camera at it actually makes logic simpler.
 - Removed unused ScreenSpaceShadowResolvePass
 - VFXManager.ProcessCameraCommand now reuses command buffer instead of creating a local one.


com.unity.render-pipelines.universal/Shaders/Utils/CopyDepthPass.hlsl
```diff
@@ -63,9 +67,33 @@ Varyings vert(Attributes input)
     #define DEPTH_OP max
 #endif
 
+half4 _ScaleBiasRT;
+
 float SampleDepth(float2 uv)
 {
-#ifdef _DEPTH_NO_MSAA
+    // Currently CopyDepthPass.cs uses cmd.Blit() that sets implicitly a projeciton matrix with y-flip
```
ellioman: typo => projeciton

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -242,6 +242,13 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             if (!requiresDepthPrepass && renderingData.cameraData.requiresDepthTexture && createDepthTexture)
             {
                 m_CopyDepthPass.Setup(m_ActiveCameraDepthAttachment, m_DepthTexture);
+
+                // TODO: This logic is only required because we use cmd.Blit() that handles y-flip in weird situations
+                // Remove it once we remove cmd.Blit();
+                // Check: Does stereo always flip as it's rendering to camera eye RT?
+                bool hasFlip = SystemInfo.graphicsUVStartsAtTop &&
+                    (cameraData.isStereoEnabled || cameraData.targetTexture != null || m_ActiveCameraColorAttachment.Identifier() != BuiltinRenderTextureType.CameraTarget);
```
thomas-zeng: XR eye target is a backbuffer type. so stereo doesn't flip for the most of the case. However, the provider could control this through VR flip flag that is not exposed in URP. 
For XRSDK, that flag is always y-flip off. For legacy XR that 19.3 supports, it is more complicated.

thomas-zeng: My question is:
CopyDepth is blitting from m_ActiveCameraDepthAttachment to m_DepthTexture. However, we are checking camera's final target and `m_ActiveCameraColorAttachment ` to figure out y-flip. These two don't really connect. Am I missing something?

phi-lira:
@thomas-zeng since now copy depth is not used in stereo and we plan to fix it with Pure XR SDK, is it ok to fix it in next PR?

phi-lira:
@thomas-zeng @DennisDeRykeUnity I added an automated test scene for both graphics and graphics stereo tests. I tested with double wide, copy depth works in XR correctly when using double wide (single pass?). I don't know how to test other XR modes in 2020.1 as the settings have moved so I'm not sure about MultiPass and Single Pass Instanced.

phi-lira:
@thomas-unity for a next PR, if you want to handle copy depth for XR using DrawProcedural, it's fine to do it and add a shader variant that handles blit with VertexID.

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -89,29 +89,31 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
...
-                cmd.Clear();
+                if (cameraData.isStereoEnabled)
+                    Debug.LogWarning("RenderObjects pass is configured to override camera matrices. While rendering in stereo camera matrices cannot be overriden.");
 
-                if (m_CameraSettings.overrideCamera)
+                if (m_CameraSettings.overrideCamera && !cameraData.isStereoEnabled)
                 {
                     Matrix4x4 projectionMatrix = Matrix4x4.Perspective(m_CameraSettings.cameraFieldOfView, cameraAspect,
                         camera.nearClipPlane, camera.farClipPlane);
+                    projectionMatrix = GL.GetGPUProjectionMatrix(projectionMatrix, cameraData.resolveFinalTarget);
```
thomas-zeng: We should add some comments here. We are making an assumption here that RenderObjectPass renders to RenderTexture type when resolveFinalTarget is true. For camera stacking case, only the last camera has resolveFinalTarget true, which means all previous cameras have resolveFinalTarget false (GetGPUProjectionMatrix won't do y-flip). If previous cameras are using intermediate textures, that could cause y-flip issues.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -421,6 +407,66 @@ void SetCameraRenderState(CommandBuffer cmd, ref CameraData cameraData)
...
+                zBufferParams.x = -zBufferParams.x;
+                zBufferParams.w += zBufferParams.z;
+                zBufferParams.z = -zBufferParams.z;
+            }
+
+            // -1.0 if flipped projection, 1.0 otherwise
+            // projection is flipped when rendering to a render texture in non OpenGL platforms
+            // this is because Unity uses bottom left coordinate system for uv (OpenGL convention)
+            // final blit pass will un-flip it when rendering to screen
+            float projectionFlipSign = RenderingUtils.GetProjectionFlipSign(cameraData.projectionMatrix);
```
thomas-zeng: Engine injects y-flip to projection matrix and VP matrix internally when SetRenderTarget.
Seems to me this flipsign could get out of sync with projection/VP matrix. Should we update flipSign in URP SetRenderTarget to keep SRP and core in sync? Otherwise, using projectionParams.x in shader could have y-flip issues.

phi-lira: this is `cameraData.projecitonMatrix` is URP only (maybe you are confusing with camera.projecitonMatrix that gets changed by SetRenderTarget?)

Either way, I'll change this because this is not a good method to check this. We could have cases the y is negative and projection is not flipped, f.ex, camera is upside down / 180deg rotation.

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
phi-lira: Indeed. Thanks for catching this.

phi-lira: Fix submitted in d085abc

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
phi-lira: Fix submitted in 
1e5b0fd

phi-lira:
Heads up. I've found issues by managing the camera matrices in URP, basically the same issues you had @thomas-zeng .

If we use same variable names unity_MatrixVP /  glstate_projection_matrix, Unity will override then and in some situations cause flip the already flipped matrix :tableflip:

If we use new name to avoid overriding then any cmd.Blit will stop working and we have no way to guarantee we are breaking user custom render passes.

phi-lira:
bors try+

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
```diff
@@ -89,29 +89,31 @@ public override void Execute(ScriptableRenderContext context, ref RenderingData
             CommandBuffer cmd = CommandBufferPool.Get(m_ProfilerTag);
             using (new ProfilingScope(cmd, m_ProfilingSampler))
             {
-                context.ExecuteCommandBuffer(cmd);
-                cmd.Clear();
+                if (cameraData.isStereoEnabled)
```
thomas-zeng: Should be if (m_CameraSettings.overrideCamera && cameraData.isStereoEnabled) check
We want to support RenderObjectPass in XR for cases like rendering different layers of the scene. The only case XR not supporting is overriding view projection matrix related data.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -585,10 +561,30 @@ static void InitializeAdditionalCameraData(Camera camera, UniversalAdditionalCam
             cameraData.postProcessEnabled &= SystemInfo.graphicsDeviceType != GraphicsDeviceType.OpenGLES2;
 
             cameraData.requiresDepthTexture |= cameraData.isSceneViewCamera || CheckPostProcessForDepth(cameraData);
+            cameraData.resolveFinalTarget = resolveFinalTarget;
+
+            cameraData.requiresIntermediateRenderTexture = RenderingUtils.RequiresIntermediateRenderTexture(ref cameraData);
+            cameraData.requiresFlip = SystemInfo.graphicsUVStartsAtTop && (cameraData.requiresIntermediateRenderTexture || camera.targetTexture != null);
```
thomas-zeng: There are corner cases for this requiresFlip.
For cases where requiresIntermediateRenderTexture is false, there are still render passes needs requiresFlip to be true.
Both ShadowPass and ColorLut pass are rendering to RenderTexture type and need to have requiresFlip true.
There are cases in the user's custom passes where renderpass target is a render texture type. Those passes require y-flip if they want to use projection matrix related data.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -44,14 +43,49 @@ namespace UnityEngine.Rendering.Universal
...
+        }
+
+        /// <summary>
+        /// Return the device projection matrix. This contains platform specific changes to handle y-flip and reverse z.
+        /// For more info on platform differences check: https://docs.unity3d.com/Manual/SL-PlatformDifferences.html
+        /// </summary>
+        /// <returns></returns>
+        public Matrix4x4 GetDeviceProjectionMatrix()
+        {
+            return GL.GetGPUProjectionMatrix(m_ProjectionMatrix, requiresFlip);
```
thomas-zeng: See comments above, 
When URP is configured to not using intermediate texture(drawing directly to the backbuffer), GetDeviceProjectionMatrix will return the wrongly y-flipped projection matrix for passes that renders to render texture(eg, shdaow, colorLut, etc..).

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
phi-lira: Hm, maybe I should rename this and make more comments?

`requiresFlip` basically means: What's the orientation of the camera textures (_CameraColorTexture and _CameraDepthTexture). This should only matter when I'm actually reading from those texture in a blit context (CopyDepth, CopyColor) as I need to guarantee that we keep the `same orientation` as source texture, otherwise all shaders will not work consistently. In other words. If we could use actual `CommandBuffer.CopyTexture`, we would not need this at all as we would just copy resource and have same orientation.

In short:
 - requiresFlip tells camera texture (source texture in a blit context) is flipped or not. It should only matter for render passes "copying" camera texture. Nothing else.
 - _ProjectionParam.x tells if current pass projection matrix is flipped or not. It matters if you are reading ANY texture as input (but not when copying as blit). `Color Grading` pass should use `_ProjectionParams.x` https://docs.unity3d.com/Manual/SL-PlatformDifferences.html

So maybe we should rename `requiresFlip` to `cameraProjectionIsFlipped`?

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
phi-lira: No. This is camera projection matrix, not per-pass. This projection matrix is set once per-frame and doesn't change. If it helps this is `unity_CameraProjection` not `unity_MatrixVP`. (well, technically it's also unity_MatrixVP `only` for the passes rendering to camera target).

phi-lira: From the API standpoint this is accessed as `cameraData.GetDeviceProjectionMatrix()` so it should be clear to client code this is camera projection matrix, not per-pass projection matrix.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
phi-lira: Another fun note. This is only required for copy depth because we don't create an intermediate render texture for copying depth. Then we end up in a situation that source depth might be generated in one orientation or the other. In theory, now we could `remove` the condition to require intermediate texture for `opaqueColorPass` if we apply the same logic to `CopyColor` pass. This will help VR performance IMO.

phi-lira:
bors try+

com.unity.render-pipelines.universal/Runtime/Passes/RenderObjectsPass.cs
phi-lira: Fix submitted in 
9924984

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -44,14 +43,49 @@ namespace UnityEngine.Rendering.Universal
...
+        {
+            return m_ProjectionMatrix;
+        }
+
+        /// <summary>
+        /// Return the camera device projection matrix. This contains platform specific changes to handle y-flip and reverse z.
+        /// For more info on platform differences check: https://docs.unity3d.com/Manual/SL-PlatformDifferences.html
+        /// </summary>
+        /// <returns></returns>
+        public Matrix4x4 GetDeviceProjectionMatrix()
```
thomas-zeng: I think we should make this private because we need to revisit this logic in Pure XRSDK. XR projection could be stereo projection.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -44,14 +43,49 @@ namespace UnityEngine.Rendering.Universal
...
+        public Matrix4x4 GetViewMatrix()
+        {
+            return m_ViewMatrix;
+        }
+
+        /// <summary>
+        /// Returns the camera projection matrix.
+        /// </summary>
+        /// <returns></returns>
+        public Matrix4x4 GetProjectionMatrix()
```
thomas-zeng: I think we should make this private because we need to revisit this logic in Pure XRSDK. XR projection could be stereo projection.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -44,14 +43,49 @@ namespace UnityEngine.Rendering.Universal
...
+        {
+            m_ViewMatrix = viewMatrix;
+            m_ProjectionMatrix = projectionMatrix;
+        }
+
+        /// <summary>
+        /// Returns the camera view matrix.
+        /// </summary>
+        /// <returns></returns>
+        public Matrix4x4 GetViewMatrix()
```
thomas-zeng: I think we should make this private because we need to revisit this logic in Pure XRSDK. XR view could be stereo view.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -64,6 +98,21 @@ namespace UnityEngine.Rendering.Universal
...
+        /// </summary>
+        public bool requiresIntermediateRenderTexture;
+
+        /// <summary>
+        /// True if the camera device projection matrix is flipped. This happens when the pipeline is rendering
+        /// to a render texture in non OpenGL platforms. If you are doing a custom Blit pass to copy camera textures
+        /// (_CameraColorTexture, _CameraDepthAttachment) you need to check this flag to know if you should flip the
+        /// matrix when rendering with for cmd.Draw* and reading from camera textures.
+        /// </summary>
+        public bool isDeviceProjectionMatrixFlipped;
```
thomas-zeng: If we move GetDeviceProjectionMatrix to private, this field should also be private.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
phi-lira: I could make this internal, but do we really need to?

We need to provide API for users that is not stereo aware and this should not be an issue to support stereo in the future. We can either add overload for stereo or make this method take a default parameter for eyeIndex like so: 
`public Matrix4x4 GetViewMatrix(int eyeIndex = 0)`

This is the reason I made these `functions` instead of `properties`. When we add stereo we just need to add overload and change the the data below to be an array of eye. Another good thing about this is that we can handle MultiPass stereo for users as the camera data knows the camera eyeIndex (I have a task to merge this PR for multi pass) https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5337/files

This is the same pattern we use for all functions that need stereo specific:
f.ex, `Camera.TryGetGetCullingParameters` has overload for stereo methods.

The matrix data is private to the CameraData, it's just the function that's public.
```
Matrix4x4 m_ViewMatrix;
Matrix4x4 m_ProjectionMatrix;
```

phi-lira:
I've moved the `RequiresIntermediateRendeTexture` back to ForwardRenderer (private). I had to make some changes to expose the ScriptableRenderer so we get the information if the pipeline is rendering to texture or not.

com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
thomas-zeng: Thanks for the details! I think this makes sense.

PR #5916, opened 2020-02-18
Fixed URP Terrain Depth pass not working in XR

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR

Fixed URP Terrain depth pass not working for XR
Related to https://fogbugz.unity3d.com/f/cases/assign/1219278

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
- Added XR Macros to URP Terrain DepthOnly Pass

### DIFF
--- a/com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitPasses.hlsl
+++ b/com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitPasses.hlsl
@@ -441,7 +441,7 @@ VaryingsLean DepthOnlyVertex(AttributesLean v)
 {
     VaryingsLean o = (VaryingsLean)0;
     UNITY_SETUP_INSTANCE_ID(v);
-	UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(o);
+    UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(o);
     TerrainInstancing(v.position, v.normalOS);
     o.clipPos = TransformObjectToHClip(v.position.xyz);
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.universal/Shaders/Terrain/TerrainLitPasses.hlsl
```diff
@@ -440,6 +441,7 @@ VaryingsLean DepthOnlyVertex(AttributesLean v)
 {
     VaryingsLean o = (VaryingsLean)0;
     UNITY_SETUP_INSTANCE_ID(v);
+	UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(o);
```
ramang-unity: Maybe you should add this too after line 407?

thomas-zeng: ShadowPass should be fine as it uses light proj view instead of XR proj view and it renders to shadow map which is not an XR aware texture (It is tex2D for both XR and non-XR afaik).

Maybe there is a case that shadow map need to be texture2DArray type for XR that I am not aware of?

ramang-unity: The only case I remember is in built-in renderer. K all good then, thanks!

phi-lira:
@thomas-zeng @DennisDeRykeUnity we should add automated tests for this in the GfxGraphicsTestStereo

thomas-zeng:
Moved to #6036

PR #5917, opened 2020-02-18
Sg/paste nodes at cursor

Nightmask3:
Yamato: https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252FPaste-Nodes-At-Cursor
---
### Purpose of this PR

This PR changes the behavior of copy-pasting nodes in the Shader Graph. 

Currently when a node is copied it is pasted slightly offset from its original location:

mGTgqJwxta

The convention followed by many node based editors (Blueprints, Houdini nodes etc.) is that copied nodes will be pasted at the cursor location, this behavior making it easier for user to recreate nodes at their desired location as opposed to having to drag them over manually.

New behavior (the gif compression is hiding the mouse cursor but I'm moving it around the graph and pasting the nodes):

jR63I1Nqas

---
### Testing status

- [x] Pasted single nodes at a time
- [x] Pasted multiple nodes at a time
- [x] Pasted multiple connected nodes at a time
- [x] Pasted subgraphs
- [x] Pasted nodes/subgraphs surrounded by a group selection box 
- [x] Pasted between graphs (after Landons QA pass mentioned this didnt work)
---
### Comments to reviewers

This is baby's first PR so please let me know if there's things I'm missing. 

Another note that I have is that currently sticky notes don't seem to get pasted at all, and their data isn't serialized out into the clipboard when they are copied even, so that's a separate bug that I reported. 

This was a pretty straightforward fix but one thing I'm wondering if there's a better way to get the cursor position in the graph. Currently I'm caching the mouse position in the MaterialGraphView using a delegate bound to the MouseMoveEvent.

I see that in cases of other mouse related input actions like dragging and dropping, there is a callback related to specifically those actions, and the event object usually provides the mouse position at the time of the event, but there isn't an equivalent callback for the copy and paste actions so caching the mouse position seemed like the only way to have it when needed.


LandonTownsendUnity:
Bug:
When copying from one graph to another, the nodes will not paste at the cursor's location
Concern:
Why is it when you paste one node, it pastes offset from the cursor down and right about half an inch? Shouldn't the offset go in the other direction, if there is one at all?

Nightmask3:
> W

Currently the behavior is just to copy the node at the mouse cursor position and offset by (30, 30) which is towards the bottom and right of the graph.

I could change that easily it just made sense to be offset a little from the cursor. There wasn't a conscious decision made towards which direction it goes. I'm interested to know why you think the offset should go in the other direction though!

LandonTownsendUnity:
If you offset it the other way then at least the node will always be visible no matter where the cursor is when you paste it. Right now if you drag the cursor off the right or bottom side of shadergraph and then paste the node will paste offscreen. I think for this reason it should be offset up and to the left.

Nightmask3:
> If you offset it the other way then at least the node will always be visible no matter where the cursor is when you paste it. Right now if you drag the cursor off the right or bottom side of shadergraph and then paste the node will paste offscreen. I think for this reason it should be offset up and to the left.

This is a good point! I didn't think at all about the cursor being offscreen or right at the edge of the screen. I guess this also raises the question though, what if the cursor is off to the left or top of the screen, wouldn't that be the same issue but from the opposite direction? I'm wondering how much work it would be to detect which way the cursor is relative to the center of the graph and then paste the nodes slightly offset to counter that so that you can always ensure they'll be within the view.

LandonTownsendUnity:
There would not be an issue for the upper left I don't think; it is the upper left corner of the node that is being offset, right? As in, with no offset, the cursor would be over the upper left corner of the node after pasting? As long as the cursor is overlapping the node after the paste operation it should be fine no matter where the cursor is. If you drag the cursor off the graph, the paste position is still locked to the graph, so there shouldn't be a chance of moving the cursor way off and then pasting somewhere far outside the bounds.

Nightmask3:
> There would not be an issue for the upper left I don't think; it is the upper left corner of the node that is being offset, right? As in, with no offset, the cursor would be over the upper left corner of the node after pasting? As long as the cursor is overlapping the node after the paste operation it should be fine no matter where the cursor is. If you drag the cursor off the graph, the paste position is still locked to the graph, so there shouldn't be a chance of moving the cursor way off and then pasting somewhere far outside the bounds.

Yes the offset applies from the top-left of each node, so with no offset the cursor would be at the upper left corner. 

I see your logic and I think it makes sense. Will change it to offset up and to the left in my next commit.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -41,6 +42,7 @@ public MaterialGraphView(GraphData graph) : this()
 
         public GraphData graph { get; private set; }
         public Action onConvertToSubgraphClick { get; set; }
+        public Vector2 cachedMousePosition { get; private set; }
```
marctem: Instead of caching this from the MouseMove you can use `UnityEngine.Event.current.mousePosition`. Simplify this whole thing.

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1220,8 +1231,10 @@ internal void PasteGraph(CopyPasteGraph graphToPaste, List<AbstractMaterialNode>
 
                 var drawState = node.drawState;
                 var position = drawState.position;
-                position.x += 30;
-                position.y += 30;
+                // Get the relative position to the calculated centroid and apply that relative offset
+                var relativeOffsetFromCentroid = position.position - centroid;
+                position.x = cursorGraphPosition.x + relativeOffsetFromCentroid.x + 30;
```
marctem: I saw your conversation with Landon about the offset - negative or positive. I'd say just don't offset for now and see what people think.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
Nightmask3: So I tried to do something similar prior to caching the position, and just to double check I tried to do it again and get the position from `UnityEngine.Event.current.mousePosition` but it's the weirdest thing, the value of `mousePosition` there is always (0,0) for some reason.

I theorize it's because we're still in the editor and the game isn't running so that value isn't updated?

com.unity.shadergraph/Editor/Data/Graphs/GraphData.cs
```diff
@@ -1173,6 +1174,16 @@ internal void PasteGraph(CopyPasteGraph graphToPaste, List<AbstractMaterialNode>
                 m_PastedStickyNotes.Add(pastedStickyNote);
             }
 
+            // Compute the mean of the copied nodes at their original position
```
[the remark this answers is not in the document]
Nightmask3: Will do, I'm currently keeping it in UIUtilities.cs with the signature `public static Vector2 CalculateCentroid(IEnumerable<Vector2> nodePositions)`.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
marctem: I saw some slack discussion a year about this being a bug but nobody every fixed it so we'll stick with the current way for now.

Nightmask3:
Status of master:
MasterStatus

Status of this branch:
BranchStatus

LandonTownsendUnity:
Here's a gif of my concern of pasting without an offset; if you drag your mouse off the bottom of shadergraph and then paste a single node, you can sometimes not see that you've pasted anything
pasteatcursorconcern
This is why I think nodes should be offset upwards and to the left.

Also, pasting on a separate graph still does not paste at the cursor, that needs to be fixed before I'll approve.

com.unity.shadergraph/Editor/Util/UIUtilities.cs
```diff
@@ -64,5 +65,18 @@ public static IEnumerable<Type> GetTypesOrNothing(this Assembly assembly)
                 return Enumerable.Empty<Type>();
             }
         }
+
+        public static Vector2 CalculateCentroid(IEnumerable<Vector2> nodePositions)
+        {
+            Vector2 centroid = Vector2.zero;
+            var count = 1;
```
zadi0648: This is definitely a developer style thing and of course not a blocker; and also another thing that I'm sure the team will have varying opinions on.

Anyhow I don't like using vars for simple types such as ints, floats, strings, ect.. Imo there is no reason to use a var when the type is simple (int, float, Vector2, string), known, and never going to change (as is the case here / is often the case).

The reason being, when ^ is the case, declaring the exact type makes stuff more readable at a glance imo (my eyes go to the left side of a variable to see what it is), doing otherwise feels lazy, I don't see any gain from keeping everything ambiguous. However this is super nit picky and we can agree to disagree 😛  (especially since I don't think the team will all agree with me).

Nightmask3:
> Here's a gif of my concern of pasting without an offset; if you drag your mouse off the bottom of shadergraph and then paste a single node, you can sometimes not see that you've pasted anything
> pasteatcursorconcern
> This is why I think nodes should be offset upwards and to the left. If not that, then it needs to paste at the center of the screen when your mouse is not on the shadergraph (this is a more complex solution that might not even be possible so I'd recommend just offsetting the pasting)
> Edit: Marc T's solution is to clamp the paste location so that the node is always pasted in a visible location (this requires clamping it a certain number of pixels up from the bottom as well as a certain number of pixels left from the right side).
> 
> Also, pasting on a separate graph still does not paste at the cursor, that needs to be fixed before I'll approve.

That is a good point, thanks for the visual demonstration of the issue. I think doing something where the pasted node gets clamped within the bounds of what is currently being viewed in the graph would be a good way to ensure that nodes don't ever get pasted out of view. I'll commit that in my next fix. 

The copy-paste between graphs is solved now, there's a comment in the shadergraph-megateam channel that goes into why that was an issue, **TL;DR** - It was explicit behavior someone coded a while ago and I just had to remove it.

com.unity.shadergraph/Editor/Util/UIUtilities.cs
Nightmask3: I replied to Landon's comment with the details but yes I fixed the copy-paste between graphs. So that'll be in the next commit. 

Yeah Marc suggested the Event.current.mousePosition route too but the issue is that the current event after pasting a node is the "Paste" input event, and mouse position is apparently not stored in that, so it always returns (0,0). 

Also yes I noted the convention of only lowercase letters in branch names, will be sure to stick to that going forward. 

I actually agree with less use of vars, I like concrete types insofar as possible, even though some places I worked at encouraged use of it in their style guides for some reason? 
Will change it up to be less ambiguous.

zadi0648:
I just thought: this is now a VFX parity problem: they should do this too. Would this be easy to put in VFX? (We probably just want them to be aware of this and do it themselves)

Nightmask3:
> I just thought: this is now a VFX parity problem: they should do this too. Would this be easy to put in VFX? (We probably just want them to be aware of this and do it themselves)

Yeah good point, VFX graph doesn't copy-paste at the cursor right now, but they should be able to integrate this fairly easily with my code as reference. 

We should maybe have a GitHub tag like #needs-VFX-Parity or something? that way we can let Tristan just have a bucket of PRs that he could integrate as well. Or maybe put it onto the shared Favro board or something. Some way of messaging stuff like this.

com.unity.shadergraph/Editor/Drawing/Views/MaterialGraphView.cs
```diff
@@ -993,5 +981,65 @@ internal static void InsertCopyPasteGraph(this MaterialGraphView graphView, Copy
...
+            // Get center of the current view
+            var center = graphView.contentViewContainer.WorldToLocal(graphView.layout.center);
+            // Get offset from center of view to mouse position
+            var mouseOffset = center - graphMousePosition;
+
+            var zoomAdjustedViewScale = 1.0f / graphView.scale;
+            var graphViewScaledHalfWidth = (graphView.layout.width * zoomAdjustedViewScale) / 2.0f;
+            var graphViewScaledHalfHeight = (graphView.layout.height * zoomAdjustedViewScale) / 2.0f;
+            const float widthThreshold = 40.0f;
+            const float heightThreshold = 20.0f;
```
zadi0648: Could be k_WidthThreshold for quick changing at the top as that's what most of the rest of the code base is, but this is fine

PR #5918, opened 2020-02-18
Fix issues with normals of super-unit length

EvgeniiG:
### Purpose of this PR
Compression can cause normal maps to store normals of length greater than 1. This causes a reconstruction error. We cannot fix the visuals (caused by the error in the texture compression), but we can at least minimize problems in our code.

---
### Testing status

**Manual Tests**: Yes.
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Fsafe-nmap-unpack/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_trunk/1366384/job

---
### Any test projects to go with this to help reviewers?

Use this normal map with the Lit shader, and reduce the scale of the normal to a small value.

sdprobs

---
### Comments to reviewers
None.


EvgeniiG:
Added Morten for awareness, since he reported the bug.

mmikk:
I just wanted to say I'm good with the proposed fix but let's get confirmation from Martin V. and Yao that it solves their problems.
An even better path forward is to switch to three channel normal maps so we can also use the magnitude for specular filtering plus use same texture inspector configuration for both object and tangent space normal maps but as a quick fix this makes sense to me.

mmikk:
I have just tried this fix but I am still getting artifacts.

EvgeniiG:
As I said in the description, the issue is with the data. It has been corrupted by the compressor. It will not dramatically improve the visuals.

mmikk:
yeah unfortunately the only fix that seems to work for me and still works with compression is using three channel normal map compression (ie. BC7 using type Default in the inspector with sRGB disabled). This requires using shader graph ofc. since it involves custom decoding.

EvgeniiG:
Test 2308 required an updated reference image.

mmikk:
Im wondering if listing 4 —> https://mmikk.github.io/papers3d/sgp.pdf
might be a cheaper way to achieve the same thing (no extra rsqrt).
The factor 128 becomes a limit on the slope and corresponds to 89.55 degrees roughly.
Ie tan(angularThresholdInDegrees * (pi/180))

EvgeniiG:
I have a better idea. Stay tuned.

EvgeniiG:
I submitted another solution. Better, but not perfect.

PR #5923, opened 2020-02-19
Removed legacy VR code from HDRP

fabien-unity:
### Purpose of this PR
Builtin VR code is being removed from Unity Engine with 2020.x and replaced by xr sdk plugins.
This PR removes the now unreachable code path within HDRP.
Also cleanup of the API by removing now unused parameters (no more deps on renderContext, yeah!)
**Should not be backported to 7.x !**

---
### Testing status

- [ ] Built a player

---
### Comments to reviewers
- the one feature we lose with the move to xr sdk is the ability to select left/right eye on the camera. I don't believe it's an issue for HDRP but it might be for URP.
- @thomas-zeng : once URP also move to pure xrsdk implementation, we'll need to remove XRUtils.cs and XRGraphics.cs from the SRP Core package.
- @seblagarde : there's a public API change in StartSinglePass() and StopSinglePass() -- 2 parameters are removed from the signature. Is there a process to handle this ?
- @DennisDeRykeUnity : for testing, this should not affect the current behavior at all as it's only removing dead code. But let's make sure all providers are still working as expected.


JasonCostanza:
Is there going to be a subsequent update to the HDRP template to be preset into XR SDK?

sebastienlagarde:
Note: we indeed need to check what is happening with StopSinglePass as they are public API, we need to add new method that fallback on it

PR #5928, opened 2020-02-19
[Shader Graph] Fix Bug with Enum Keyword Reference Suffix

alindmanUnity:
### Purpose of this PR 
Fixes a bug where having a custom reference suffix on an enum keyword would not compile properly on a material. [internal bug](https://fogbugz.unity3d.com/f/cases/1214802/)
Changes behavior to accommodate the material keyword system. 
Previously:
- Both the entry display name and the reference suffix could be edited.
- The display name and the reference suffix did not have to match. 
- Having mismatching display name + reference suffix would cause keywords not to work on materials.
- The display name could have special characters such as `( ) !` etc, which could cause the material to error.

New behavior:
- Only the entry "display" name can be edited, so it is now renamed "Entry Name". 
- The reference suffix is greyed out, and simply displays the sanitized version of the display name that the shader will be using in generation. Updates every time the entry name is changed.
- The display name can no longer contain special characters. Any special characters found will be converted to an underscore `_` to match the behavior of other reference fields on the blackboard.
- Documentation has been updated to reflect the new expected behavior. 
keywords_enum

---
### Testing status

- [x] Built a player
- Other: 
- Created new enum keywords on a graph. Changed the display names. Observed that the reference name updates to the sanitized version whenever the display name is changed.
- All special characters are converted to an underscore when present. 
- Using the keyword in the graph and changing the entry on the material displays the correct branch in the shader. 

---
### Comments to reviewers
Whitespace changes from my IDE, hope it doesn't clutter the changeset too much. 

note for Martin: If there's a better place to put the GUIStyle definition for the greyed out label that I'm missing please let me know. 

note for QA: When doing my own manual testing I found that currently changing the display name is not registered in the Undo stack, so changing the name and then pressing undo will not undo the change name action. This is also in master, and should be filed as a separate bug please. 
Also, currently any existing reference name is not overridden on load. It will only change to follow the new convention if the entry name is changed. Would you prefer for them to upgrade immediately/automatically?


LandonTownsendUnity:
"Also, currently any existing reference name is not overridden on load. It will only change to follow the new convention if the entry name is changed. Would you prefer for them to upgrade immediately/automatically?"

Instinctively I would say no, it seems less bug prone the way you did it. I'll ask Grant what he thinks.

PR #5948, opened 2020-02-21
Fixed for case 1209292

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR

Fixed double sRGB correction for quest:
https://fogbugz.unity3d.com/f/cases/1209292/

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
In XR, the final target is eye texture instead of back buffer. Because of that, the logic for URP sRGB handling needs special care for the XR case.
QA: This PR should also fix no sRGB correction on WMR. I didn't find fogbugz for that, could you please help verify this?


com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -336,8 +336,15 @@ int GetDestination()
...
-                if (Display.main.requiresSrgbBlitToBackbuffer && m_EnableSRGBConversionIfNeeded)
+
+                bool requiresSRGBConvertion = Display.main.requiresSrgbBlitToBackbuffer;
+                // For stereo case, eye texture always want color data in sRGB space.
+                // If eye texture color format is linear, we do explicit sRGB convertion
+#if ENABLE_VR && ENABLE_VR_MODULE
+            if (cameraData.isStereoEnabled)
+                requiresSRGBConvertion = !XRGraphics.eyeTextureDesc.sRGB;
+#endif
+                if (requiresSRGBConvertion && m_EnableSRGBConversionIfNeeded)
```
phi-lira: Minor. Seems like to avoid code duplication/improve maintanance you could apply this logic to m_EnableSRGBConvertionIfNeeded in Setup() (only once)

thomas-zeng: Addressed in 835d1b7.
- Added helper function RequireSRGBConversionBlitToBackBuffer to avoid code duplication.
  - Reasoning about adding helper function instead of doing the logic in Setup is that postFXPass has Setup and SetupFinalPass, and doing logic in Setup still requires duplicated logic in SetupFinalPass.

thomas-zeng:
Moved to #6036

PR #5950, opened 2020-02-21
Improve Android (OpenGL ES) Universal RP tests

tomzig16:
### Purpose of this PR
This PR solves two problems.
-  With newer version of Unity editor SDK is searched in default path, but as internal Unity downloader does not download SDK it was resulting in an null sdk path error. This PR fixes that.
-  This PR also adds split build and run functionality.

---
### Testing status

-  Manually running tests on bokken agent
-  Running "Build" pipeline for Android on yamato (expected to only build project that we can reuse later)
-  Running "Test" pipeline for Android on yamato (expected to reuse built projects)
-  Running tests locally

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/android%252Ftomaszi%252Ffix-null-sdk-path/.yamato%252Fupm-ci-universal.yml%2523Universal_Android_OpenGLES_fast-trunk/1385327/job

The same UniversalGraphicsTest project as for other platforms

---
### Comments to reviewers

This PR also filters out several scenes (only for Android) as they were causing several problems that need more investigation. Right now our aim is to get Android tests into automation.
Split build and run was broken for Android in test-framework packages so packages had to be updated too.


sophiaaar:
should we also add an android test to the nightlies in this PR?

tomzig16:
> should we also add an android test to the nightlies in this PR?

That would be great!

PR #5952, opened 2020-02-21
Fixed resource lifecycle issue. (#5941)

phi-lira:
Backport of #5941


phi-lira:
Tests failing due to timeout.

PR #5956, opened 2020-02-21
Lighting/probevolumes

thefranke:
---
### Purpose of this PR
This PR introduces ProbeVolumes, a new type of volume handled in the LightLoop, which encodes regular 3d grids of Light Probes. ProbeVolumes are used to add GI by covering a part of the scene with a volume and baking it. Multiple volumes can be added to a scene, as each volume manages its data individually. Volumes that overlap are blended in special blending regions. On top of that, a blending mode can be selected for each volume.

---
### Testing status

A test project can be found here:
https://drive.google.com/file/d/1e_rUbBFDvvhe4KzSZganEv4y-RFqGbvI/view?usp=sharing

The scene contains a temple ruin. The floor uses Lightmaps for GI, whereas the rest is set to Light Probes, but all objects are marked as Contribute GI. Because the current branch does not have layer support for baking ProbeVolumes, the way this scene was prepared was by enabling and disabling groups of objects/lights and then baking them independently via the Bake button for individual volumes.

The first volume called "Ruins Volume" is set to "Normal" blending mode, which means it only adds GI to those parts of the scene that are not lightmapped. The other two volumes called "Emissive volume" where baked using the scene with the emissive object inside each group and only contain a volume with the emissive light baked into the probe. Both of these are set to "Additive" blending mode, which means their GI portion gets added to the scene everywhere, even the lightmapped parts. These latter two can be turned on and off or faded in and out without breaking the overall result, as they simply add additional light.

In the project settings you will also find a volume override for Probe Volumes called "Probe Volume Controller". This section contains special filter modes for the probes to avoid leaks. ProbeVolumes are already baked with validity filtering included (meaning probes inside of geometry are filled with light from neighboring probes), which will suppress invalid/dark shadow splotches leaking into the scene. Further leak mitigation options of light through objects are available in the Probe Volume Controller.

ProbeVolume baking can be done either via the standard "Generate Lighting" button in the Lighting window or - for individual volumes - via the Bake button in the baking section.

**Manual Tests**: 
- [x] Built a player
- Other: 

**Automated Tests**:
There are currently no automated tests.

---
### Comments to reviewers

This PR introduces the first batch of code, but disabled. To enable ProbeVolumes, they have to be enabled in the HDRP configuration package, the HDRP pipeline asset and per camera.


com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
```diff
@@ -0,0 +1,106 @@
...
+        mean2 += sampleCurrentDepth * sampleCurrentDepth * sampleCurrentWeight;
+        weightTotal += sampleCurrentWeight;
+    }
+    float filterNormalization = rcp(weightTotal);
+    mean *= filterNormalization;
+    mean2 *= filterNormalization;
+    float variance = max(0.0f, mean2 - mean * mean); // avoid small negative values due to precision.
+
+    // Sync group before writing to ensure we dont create race conditions for neighboring pixels that might still be resolving.
+    GroupMemoryBarrierWithGroupSync();
```
pmavridis: Unless I'm missing something, you will still get race conditions when reading across thread-groups. GroupMemoryBarrierWithGroupSync will wait for all threads in a group, but not for other thread-groups that might be executing in parallel.

In practice I would not expect any major visual artifacts from this type of race condition (some times a thread will read the mean depth instead of the actual depth which might be "close-enough"). The biggest issue perhaps is that, due to the race condition, the results might be different every time the user bakes the lighting. 

Aside from the race condition, the memory barrier is introducing stalls that might affect the performance negatively. I would suggest instead of doing the filtering in-place, write the results in a new buffer.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
```diff
@@ -0,0 +1,106 @@
...
+    // In order to generate a {mean, variance} distribution, we will convolve a weighted neighborhood of samples
+    // via fibonacci spiral spherical gaussian importance sampling.
+    // Really we are just importance sampling a spherical cap, whos solid angle is approximately equal to the
+    // solid angle of our spherical gaussian filter.
+    float3 filterCenterWS = UnpackNormalOctRectEncode(octahedralDepthUV * 2.0f - 1.0f);
+
+    // Generate full basis for our spherical gaussian filter to sample within.
+    // TODO: Dither rotate basis per pixel around filterCenterWS to reduce bias at the cost of noise.
+    float3 basisNormal = filterCenterWS;
+    float3 basisUp = (basisNormal.y < 0.999f) ? float3(0.0f, 1.0f, 0.0f) : float3(1.0f, 0.0f, 0.0f);
```
pmavridis: This is rather minor, but for the orthonormal generation, have you considered using one of the methods in Pixar's paper? They have copy-paste code:  https://graphics.pixar.com/library/OrthonormalB/paper.pdf and even the basic Frisvad's method should be faster.

Aside from that, I think the filter can be further optimized by pre-computing the points in the fibonacci spiral.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasBlit.compute
```diff
@@ -0,0 +1,75 @@
...
+float3 _ProbeVolumeResolutionInverse;
+float4 _ProbeVolumeAtlasScaleBias;
+float4 _ProbeVolumeAtlasResolutionAndInverse;
+
+uint2 ComputeWriteIndexFromReadIndex(uint readIndex, float3 resolution, float3 resolutionInverse, float4 scaleBias, float4 atlasResolutionAndInverse)
+{
+    // return uint2(readIndex % 1024u, readIndex >> 10);
+
+    // _ProbeVolumeAtlasReadBuffer[z * resolutionY * resolutionX + y * resolutionX + x]
+    // TODO: Could implement as floating point operations, which is likely faster.
```
FrancescoC-unity: Yup, integer module is horribly slow on GCN and it is well worth getting rid of. uint is faster than int, but not that much.  A function that does

X res = a % b
return res;

if the type is uint it is 25 VALU, 12 if float, 28 if int. All assuming a and b are scalar, so likely will be worse.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
pastasfuture: Thanks for keeping an eye out for this. Indeed in the general case there would be a race condition here. However in this specific case there is not because the group size is exactly the size of the pano it is filtering in the atlas (8x8 pixels), so it reads from and writes to the entire tiny pano here.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
pastasfuture: Totally. I've read that paper and agree its an improvement. I'll upgrade the code to that next time I'm working with this. This octahedral depth code is very WIP. Thanks

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
pmavridis: Thanks a lot for the clarification.

PR #5957, opened 2020-02-21
[Yamato] Editor Priming jobs

sophiaaar:
### Purpose of this PR
Often when we start tests on Yamato, we are bottlenecked by waiting for Katana to build Unity. This can lead to a playmode or editmode test, that should normally take 15-30 mins, taking multiple hours. And we use up the machines available during this time, meaning the queue builds up and everyone ends up waiting for machines to free up.

This PR moves the Katana build portion to a separate "editor priming" job, and the builds in the pipeline all depend on this. Now we won't be holding up all the devices in the build farm while we wait for them to perform an unrelated task.

image
^ This diagram is just an example of how the pipeline looks - the OSX failure has since been fixed :)
(Also note the timings in the above pipeline!)

---
### Testing status

**Yamato**: 
Note-worthy yamato tests:
Packages https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/ci%252Feditor-priming/.yamato%252Fupm-ci-packages.yml%2523all_package_ci_fast-2020.1/1397635/job/pipeline

Mac Metal (this one was important because of how different the mac devices are yamato-wise) https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/ci%252Feditor-priming/.yamato%252Fupm-ci-universal.yml%2523Universal_OSX_Metal_playmode_fast-trunk/1396555/job/pipeline

Will trigger abv with bors once PR is created


sophiaaar:
bors try

pbbastian:
bors try

sophiaaar:
had to do some conflict resolution - running abv to check everything runs smooth

.yamato/upm-ci-abv.yml
```diff
@@ -111,13 +106,16 @@ smoke_test_{{ testplatform.name }}_{{ editor.version }}:
   - git clone git@github.cds.internal.unity3d.com:unity/utr.git TestProjects/SRP_SmokeTest/utr
   - npm install upm-ci-utils -g --registry https://api.bintray.com/npm/unity/unity-npm
   - pip install unity-downloader-cli --extra-index-url https://artifactory.eu-cph-1.unityops.net/api/pypi/common-python/simple
-  - cd TestProjects/SRP_SmokeTest && unity-downloader-cli {{ editor.cmd }} -c editor --wait --published    
+  - cd TestProjects/SRP_SmokeTest && unity-downloader-cli --source-file ../../unity_revision.txt {% for component in platform.components %}-c {{ component }} {% endfor %}--wait --published-only
```
sophiaaar: i actually think this part is wrong but abv started fine anyway?

.yamato/z_editor.yml
```diff
@@ -0,0 +1,67 @@
...
+    PATH: /home/bokken/bin:/home/bokken/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/sbin:/home/bokken/.npm-global/bin
+    DISPLAY: ":0"
+    {% for v in template_environment_variables %}
+    {{ v[0] }}: {{ v[1]}}
+    {% endfor %}
+    {% if editor.version == 'CUSTOM-REVISION' %}
+    CUSTOM_REVISION: custom_revision_not_set
+    {% endif %}
+  commands:
+    - pip install unity-downloader-cli --user --upgrade --extra-index-url https://artifactory.eu-cph-1.unityops.net/api/pypi/common-python/simple
```
HenrikPoulsen: Since you are adding new things. Can you also replace `https://artifactory.eu-cph-1.unityops.net/api/pypi/common-python/simple` with `https://artifactory.internal.unity3d.com/api/pypi/common-python/simple`. The other domain name is deprecated and at some point IT will start poking people hard to stop using it. So might as well dodge that bullet

.yamato/z_editor.yml
```diff
@@ -0,0 +1,67 @@
...
+      --wait
+      --skip-download
+{% for component in platform.components -%}
+      -c {{ component }}
+{% endfor -%}
+      > unity_revision.txt
+  artifacts:
+    unity_revision.zip:
+      paths:
+        - "**/**/unity_revision.txt"
```
HenrikPoulsen: Since the file is generated in the root you don't really need the recursiveness of it. `- "unity_revision.txt"` should do just fine

.yamato/z_editor.yml
```diff
@@ -0,0 +1,67 @@
...
+  - os: windows
+    components:
+      - editor
+      - il2cpp
+  - os: linux
+    components:
+      - editor
+      - il2cpp
+
+template_environment_variables:
```
HenrikPoulsen: You can remove this section and all
```
{% for v in template_environment_variables %}
    {{ v[0] }}: {{ v[1]}}
{% endfor %}
```
It's not needed for anything and is just something that comes from the original yamato ci templates, from which this is diverging anyway so no point in keeping it :)

.yamato/z_editor.yml
```diff
@@ -0,0 +1,67 @@
...
+    type: Unity::VM
+    image: 'cds-ops/ubuntu-16.04-base:stable'
+    flavor: b1.small
+  variables:
+    PATH: /home/bokken/bin:/home/bokken/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/sbin:/home/bokken/.npm-global/bin
+    DISPLAY: ":0"
+    {% for v in template_environment_variables %}
+    {{ v[0] }}: {{ v[1]}}
+    {% endfor %}
+    {% if editor.version == 'CUSTOM-REVISION' %}
```
HenrikPoulsen: I assume that in theory you could get rid of this if statement and just have 
`CUSTOM_REVISION: {{ editor.version }}` everywhere
And invoke the unity downloader with `-u $CUSTOM_REVISION`
and also adding an `{%if editor.fast %} --fast `{%` endif %}`
The benefit would be that it generates less jobs and CUSTOM_REVISION could be set on any job.
On the negative side I guess it would be harder to spot which jobs has had a custom revision set. Though to me that feels like a Yamato UI issue mostly in that it should probably highlight these variable changes more clearly.
This isn't something I would say is a huge problem so you might consider that for future investigation perhaps

.yamato/upm-ci-abv.yml
HenrikPoulsen: Checking the linter it does not generate any `-c` for these calls because this file is missing the `platforms:` section with a list of components. So it will generate the jobs just fine but the jobs themselves will fail when run because the unity downloader will complain

.yamato/z_editor.yml
sophiaaar: got it, thanks

.yamato/z_editor.yml
sophiaaar: ah cool i did wonder but i wasnt sure if it was one of those sneaky variables that do something behind the scenes! (like the promotion one)

.yamato/z_editor.yml
sophiaaar: thank you - yeah i'll keep that in mind for future but i won't address it in this PR

sophiaaar:
The VFX failures match current master

.yamato/upm-ci-packages.yml
```diff
@@ -180,9 +179,10 @@ all_package_ci_{{ editor.version }}:
 {% endfor %}
 {% endfor %}
 
+{% for editor in editors %}
 {% for package in packages %}
-publish_{{ package.id }}:
-  name: z_(do not use) Publish {{ package.name }}
+publish_{{ package.id }}_{{ editor.version }}:
```
HenrikPoulsen: This seems a bit odd. Why does publish need to be unity version dependent?
I would expect it to just either hard code a specific editor version's tests as a dependency or that you do an exhausive one with a bunch of different editor versions to block the single publishing job per package

.yamato/upm-ci-vfx_lwrp.yml
```diff
@@ -99,15 +106,13 @@ Build_{{ project.name }}_Win_{{ win_api.name }}_Player_{{ editor.version }}:
   {% endif %}
   commands:
     - git clone git@github.cds.internal.unity3d.com:unity/utr.git  TestProjects/{{ project.folder }}/utr
-    - pip install unity-downloader-cli --extra-index-url https://artifactory.eu-cph-1.unityops.net/api/pypi/common-python/simple --pre
+    - pip install unity-downloader-cli --extra-index-url https://artifactory.internal.unity3d.com/api/pypi/common-python/simple --pre
```
HenrikPoulsen: Please remove --pre from the install unless you really really need it. I dont expect there is anything in --pre currently that is needed

PR #5958, opened 2020-02-21
Forward Port - yamato fix

phi-lira:
Forward port of yamato fix. Fix already tested in 8.x.x/release branch #5952 5952


phi-lira:
bors +

PR #5963, opened 2020-02-21
Hybrid Renderer V2 [skip ci]

sebbbi:
### Purpose of this PR
This PR adds support for Hybrid Renderer V2. All new C# code is inside #if ENABLE_HYBRID_RENDERER_V2 brackets. Our new codew should not be active unless this define is set in a user project and Hybrid Renderer V2 DOTS package is installed. Shader changes are made with macros that do nothing, except in new DOTS_INSTANCING_ON shader variant. This shader variant is stripped by default. Build time should not regress. If hybrid.renderer package is installed, this shader variant is NOT stripped. 

We added a dedicated shader variant for Hybrid Renderer V2 (DOTS_INSTANCING_ON). The old Hybrid Renderer V1 was modifying existing instanced shader variant, which caused incompatibilities with GameObject rendering. MaterialPropertyBlocks didn't work anymore. Also with the old hybrid renderer, the instanced variants were stripped in some cases causing standalone build image corruption. We have discussed with @joravainen and @stramit and others, validating that our new shader stripping approach is the best possible solution.

### Testing status
[X] Opened test project + Run graphic tests locally
[X] Built a player
[X] C# and shader warnings (supress shader cache to see them)

We added two new test projects to DOTS repository: HybridHDRPSamples and HybridURPSamples. These projects include unit tests and graphics tests (image comparison based). We can't land these test projects before this PR lands to SRP and a new SRP package is released. After this we can configure our test project manifest files to use this new SRP package and land it in DOTS master. After this happens @cinight can integrate that to CI. @cinight has been doing our testing. Please ask her for more info how to setup the test projects and for detailed test reports.

Link to DOTS PR:
https://github.com/Unity-Technologies/dots/pull/4301

We have validated that both old Hybrid Renderer V1 and the new Hybrid Renderer V2 render correctly with these changes. We haven't noticed any issues with default GameObject based URP and HDRP rendering. The changes should have zero effect on these.


com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalPBRSubShader.cs
```diff
@@ -80,7 +80,7 @@ class UniversalPBRSubShader : IPBRSubShader
             {
                 "prefer_hlslcc gles",
                 "exclude_renderers d3d11_9x",
-                "target 2.0",
+                "target 4.5",
```
sebbbi: @stramit this line shouldn't be here? Did your merge your changes to our branch?

com.unity.render-pipelines.universal/Shaders/Lit.shader
```diff
@@ -239,9 +244,190 @@ Shader "Universal Render Pipeline/Lit"
...
+            #pragma fragment DepthOnlyFragment
+
+            // -------------------------------------
+            // Material Keywords
+            #pragma shader_feature _ALPHATEST_ON
+            #pragma shader_feature _SMOOTHNESS_TEXTURE_ALBEDO_CHANNEL_A
+
+            //--------------------------------------
+            // GPU Instancing
+            #pragma multi_compile_instancing
```
stramit: not needed

com.unity.render-pipelines.universal/Shaders/Utils/FallbackError.shader
```diff
@@ -8,7 +8,7 @@ Shader "Hidden/Universal Render Pipeline/FallbackError"
             CGPROGRAM
             #pragma vertex vert
             #pragma fragment frag
-            #pragma target 2.0
+            #pragma target 4.5
```
sebbbi: @stramit target 2.0 shader seems to be missing?

com.unity.shadergraph/Editor/Resources/Shaders/FallbackError.shader
```diff
@@ -7,7 +7,7 @@
             CGPROGRAM
             #pragma vertex vert
             #pragma fragment frag
-            #pragma target 2.0
+            #pragma target 4.5
```
sebbbi: @stramit target 2.0 shader seems to be missing?

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -28,7 +28,7 @@ class ShaderGraphImporter : ScriptedImporter
             CGPROGRAM
             #pragma vertex vert
             #pragma fragment frag
-            #pragma target 2.0
+            #pragma target 4.5
```
sebbbi: @stramit target 2.0 shader seems to be missing?

com.unity.render-pipelines.universal/Shaders/Utils/FallbackError.shader
sebbbi: @stramit target 2.0 shader seems to be missing?

com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
```diff
@@ -462,6 +462,7 @@ Shader "HDRP/LayeredLit"
 
     //enable GPU instancing support
     #pragma multi_compile_instancing
+    #pragma multi_compile _ DOTS_INSTANCING_ON
```
JussiKnuuttila: I think LayeredLit is not supported for Hybrid Renderer V2 right now (at least it doesn't have any DOTS instanced properties currently). We should remove the keyword from it and other unsupported shaders.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaderUtilities.cs
```diff
@@ -780,26 +779,30 @@ public static bool GenerateShaderPass(AbstractMaterialNode masterNode, Pass pass
             ShaderSpliceUtil.BuildType(typeof(HDRPShaderStructs.VertexDescriptionInputs), activeFields, vertexGraphInputs, debugOutput);
 
 
-            int instancedCount = sharedProperties.GetDotsInstancingPropertiesCount(mode);
+            int instanceCount = sharedProperties.GetDotsInstancingPropertiesCount(mode);
```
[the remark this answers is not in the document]
stramit: fixed

com.unity.render-pipelines.universal/Shaders/Utils/FallbackError.shader
stramit: fixed

com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -34,11 +34,14 @@
 #if defined(UNITY_SUPPORT_INSTANCING) && defined(PROCEDURAL_INSTANCING_ON)
     #define UNITY_PROCEDURAL_INSTANCING_ENABLED
 #endif
+#if defined(UNITY_SUPPORT_INSTANCING) && defined(DOTS_INSTANCING_ON)
```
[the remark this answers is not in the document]
sebbbi: We could leave it like this. Neither option is any better ,and this current one compiles.

com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl
```diff
@@ -327,21 +351,44 @@
             #define unity_ProbesOcclusion UNITY_ACCESS_INSTANCED_PROP(unity_Builtins2, unity_ProbesOcclusionArray)
         #endif
     UNITY_INSTANCING_BUFFER_END(unity_Builtins2)
+    #endif
+
+    // TODO: What about UNITY_DONT_INSTANCE_OBJECT_MATRICES for DOTS?
```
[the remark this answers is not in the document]
sebbbi: Somebody needs to search UNITY_DONT_INSTANCE_OBJECT_MATRICES in the C++ and SRP code base. Most likely this doesn't affect hybrid rendering at all, but we should confirm and remove this TODO.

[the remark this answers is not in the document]
JussiKnuuttila: This seems to correspond to `#pragma instancing_options nomatrices`, which is used by some Unity built-in shaders (not supported by Hybrid V2), and terrain shaders in both URP and HDRP (not supported yet, but will be supported later when necessary).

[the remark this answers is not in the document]
sebbbi: This is fine then!

LandonTownsendUnity:
Is there anything I should test here besides basic smoke tests?

sebbbi:
> Is there anything I should test here besides basic smoke tests?

This PR should not break anything. Everything should in #if brackets or affect only DOTS_INSTANCING_ON shader variant. 

Running smoke tests on URP with a device that only supports "target = 2.0" shader model, would be preferable. We don't have such a device for testing here.

LandonTownsendUnity:
> Running smoke tests on URP with a device that only supports "target = 2.0" shader model, would be preferable. We don't have such a device for testing here.

We don't really have one here either. I don't even know for sure if that low of a spec is even still supported by Unity or Shadergraph. That would be something like OpenGLES2, right? If I just switched the API to that, would that work for testing? I already know that there's a huge amount of features that are not supported in GLES2. Last time I checked even the basic Shadergraph UI didn't work in GLES2 (but this may have since been fixed).

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -483,13 +483,55 @@ float4x4 GetRawUnityWorldToObject() { return unity_WorldToObject; }
...
-
 // This define allow to tell to unity instancing that we will use our camera relative functions (ApplyCameraTranslationToMatrix and  ApplyCameraTranslationToInverseMatrix) for the model view matrix
 #define MODIFY_MATRIX_FOR_CAMERA_RELATIVE_RENDERING
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl"
 
+#undef unity_ObjectToWorld
+#undef unity_WorldToObject
+#ifdef UNITY_DOTS_INSTANCING_ENABLED
+// TODO: This might not work correctly in all cases, double check!
+UNITY_DOTS_INSTANCING_START(BuiltinPropertyMetadata)
```
FrancescoC-unity: Could we the changes here akin to what the URP counterpart does? (A separate file included here)

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
```diff
@@ -483,13 +483,55 @@ float4x4 GetRawUnityWorldToObject() { return unity_WorldToObject; }
...
 #define unity_WorldToObject Use_Macro_UNITY_MATRIX_I_M_instead_of_unity_WorldToObject
 
-// Define View/Projection matrix macro
-#include "Packages/com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesMatrixDefsHDCamera.hlsl"
-
 // This define allow to tell to unity instancing that we will use our camera relative functions (ApplyCameraTranslationToMatrix and  ApplyCameraTranslationToInverseMatrix) for the model view matrix
 #define MODIFY_MATRIX_FOR_CAMERA_RELATIVE_RENDERING
 #include "Packages/com.unity.render-pipelines.core/ShaderLibrary/UnityInstancing.hlsl"
 
+#undef unity_ObjectToWorld
```
FrancescoC-unity: So, above we define these two matrix with an error as an error, just to double check, is the error obsolete (hence we should remove the above definition) or is this undef essentially doing the same thing? (forcing an error when using unity_ObjectToWorld etc.) ?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
JussiKnuuttila: We could change it like that.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariables.hlsl
JussiKnuuttila: This undef here is required, because otherwise the `UNITY_DOTS_INSTANCED_PROP(float4x4, unity_ObjectToWorld)` macro does not work.

I think we could move the undefs under the ifdef, so we will still get the error checking when the DOTS variant is not compiled.

JussiKnuuttila: I have changed the code here slightly according to this conversation.

FrancescoC-unity: :+1: Great thank you!

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
```diff
@@ -121,6 +134,51 @@ static class HDPragmas
                 new FieldCondition(HDFields.DotsInstancing, false),
                 new FieldCondition(HDFields.DotsProperties, false),
             } },
+            #endif
+        };
+
+        public static PragmaCollection DotsInstancedInV2Only = new PragmaCollection
```
alindmanUnity: if these are v2 only, should they be entirely wrapped in the v2 if? why is only half of the "v2 only" pragma collection wrapped?

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
```diff
@@ -121,6 +134,51 @@ static class HDPragmas
...
+            { Pragma.DOTSInstancing },
+            { Pragma.InstancingOptions(InstancingOptions.NoLodFade) },
+            #endif
+        };
+
+        public static PragmaCollection DotsInstancedInV2OnlyEditorSync = new PragmaCollection
+        {
+            { Basic },
+            { Pragma.MultiCompileInstancing },
+            { Pragma.EditorSyncCompilation },
```
alindmanUnity: if the editor sync pragma is only needed for scene selection pass, we can make it conditional on scene selection and avoid some duplication.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
```diff
@@ -121,6 +134,51 @@ static class HDPragmas
...
+            { Pragma.DOTSInstancing },
+            { Pragma.InstancingOptions(InstancingOptions.NoLodFade) },
+            #endif
+        };
+
+        public static PragmaCollection DotsInstancedInV2OnlyRenderingLayer = new PragmaCollection
+        {
+            { Basic },
+            { Pragma.MultiCompileInstancing },
+            { Pragma.InstancingOptions(InstancingOptions.RenderingLayer) },
```
alindmanUnity: if the rendering layer pragma is only needed for stack lit subshader, we can make it conditional and delete the bottom three duplicated collections

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
```diff
@@ -25,25 +25,31 @@ static class HDPragmas
...
         {
             { Basic },
             { Pragma.MultiCompileInstancing },
+            // Hybrid Renderer V2 requires a completely different set of pragmas from Hybrid V1
+            #if ENABLE_HYBRID_RENDERER_V2
+            { Pragma.DOTSInstancing },
+            { Pragma.InstancingOptions(InstancingOptions.NoLodFade) },
+            { Pragma.InstancingOptions(InstancingOptions.RenderingLayer) },
+            #else
             { Pragma.InstancingOptions(InstancingOptions.NoLightProbe), new FieldCondition[]
```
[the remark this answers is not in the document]
JussiKnuuttila: Yes, and RenderingLayer needs to be disabled as well, because Hybrid V1 supports none of these features.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
JussiKnuuttila: We could do it like that. I initially chose this approach so that all ifdefs would be in the pragma collection definitions, and no ifdefs would be in the use sites (HDPasses.cs).

JussiKnuuttila: The intention in this approach is that the same collection is unconditionally used, but the contents of the collection depend on whether V2 is enabled or not.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
JussiKnuuttila: I am unsure about this, as I don't know what it does. I attempted to retain the semantics that already existed in the file previously.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
JussiKnuuttila: I'm unsure about this also, we should confirm with someone who knows exactly what it does and when it's needed / unneeded.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
alindmanUnity: that approach seems better, I would personally rename it to be `DotsInstancedV2Supported` or something similar that doesn't insinuate the whole thing only works in v2.

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDPragmas.cs
alindmanUnity: it's fine if you don't have the answer -- as long as it's working that's important, it can be cleaned up later. no reason to block the PR.

PR #5965, opened 2020-02-22
Adaptive sampling for SSS [skip ci]

EvgeniiG:
### Purpose of this PR
- Added support for Quality Levels to Subsurface Scattering.
- The number of samples is now automatically adjusted at runtime, so there is a chance to use fewer samples than what is budgeted for.
- Also fixed a weird bug with very bright lights (the new test images look more correct).
- Removed a ton of code and shader variables.

---
### Testing status

**Manual Tests**: Opened test project + Run graphic tests locally.
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Fadaptive-sss/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_trunk

---
### Comments to reviewers
I also touched the diffusion profile preview code. The visuals are now better (I think?).


EvgeniiG:
Updated ref images.

EvgeniiG:
All tests pass except for 1351 and 1710.

sebastienlagarde:
@EvgeniiG any thought about performance? does it vary compare to previous version?
@Unity-Technologies/gfx-qa-hdrp Please check the migration with "high quality SSS" as well. thanks

EvgeniiG:
I fixed a broken test. Tests should now pass.
I added a migration step for High Quality SSS.
Perf-wise, we'll have to see. I suspect it will be slower as long as the algorithm needs all the samples (num_samples_required = sample_budget) due to the extra ALU. It should be faster for objects at distance (when adaptive sampling kicks in). I will run some tests.

EvgeniiG:
image

EvgeniiG:
I did a little bit of testing on the PS4. I compared the current version with the version where I manually pass the constants to the shader and tell the compiler to unroll the loop (so it can importance-sample at the compile time).

The shader is heavily memory-bound, so the extra ALU for importance sampling only makes around 3-5% difference.

I should note that it is still very easy to destroy the perf by using a large scattering distance (or simply jamming the camera into the character's face), which will result in trashing the texture cache and cost of 10 ms+ even on low settings.

EvgeniiG:
It's 6 wavefronts per SIMD, used to be 7.

image

EvgeniiG:
I will see if I can win another wavefront by massaging it in the SCUI.

EvgeniiG:
7x wavefronts now, not really faster, though.

image

EvgeniiG:
Razor's data for the SSS pass looks pretty healthy, so not sure what else we can do here except reduce the number of samples.

EvgeniiG:
This is 2.5 ms.

image

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
```diff
@@ -132,24 +132,18 @@ float3 EvaluateTransmittance_Punctual(LightLoopContext lightLoopContext,
...
-    float3 S = _ShapeParams[bsdfData.diffusionProfileIndex].rgb;
-
-#if 0
-    float3 expOneThird = exp(((-1.0 / 3.0) * thicknessDelta) * S);
-#else
-    // Help the compiler. S is premultiplied by ((-1.0 / 3.0) * LOG2_E) on the CPU.
-    float3 p = thicknessDelta * S;
-    float3 expOneThird = exp2(p);
-#endif
+    float3 exp_13 = exp2(((LOG2_E * (-1.0/3.0)) * dt) * S); // Exp[-S * r / 3]
```
FrancescoC-unity: Should this comment be // Exp[-S * dt / 3] ?

com.unity.render-pipelines.high-definition/Runtime/Material/DiffusionProfile/DiffusionProfile.hlsl
```diff
@@ -27,14 +12,47 @@ float3 ComputeTransmittanceDisney(float3 S, float3 volumeAlbedo, float thickness
...
+    // Premultiply & optimize: T = (1/4 * A) * (e^(-S * t) + 3 * e^(-S * t / 3))
+    return volumeAlbedo * (exp_13 * (exp_13 * exp_13 + 3));
+}
+
+// Performs sampling of the Normalized Burley diffusion profile in polar coordinates.
+// The result must be multiplied by the albedo.
+float3 EvalBurleyDiffusionProfile(float r, float3 S)
+{
+    float3 exp_13 = exp2(((LOG2_E * (-1.0/3.0)) * r) * S); // Exp[-S * r / 3]
+    float3 expSum = exp_13 * (1 + exp_13 + exp_13);        // Exp[-S * r / 3] + Exp[-S * r]
```
FrancescoC-unity: To get what the comment says I think it should be exp_13 * (1 + exp_13 ***** exp_13); and not plus, is the comment or code correct?

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
```diff
@@ -7,21 +7,19 @@
 // #pragma enable_d3d11_debug_symbols
 #pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
 
-#pragma kernel SubsurfaceScatteringMQ SubsurfaceScattering=SubsurfaceScatteringMQ SSS_ENABLE_NEAR_FIELD=0
-#pragma kernel SubsurfaceScatteringHQ SubsurfaceScattering=SubsurfaceScatteringHQ SSS_ENABLE_NEAR_FIELD=1
-#pragma kernel SubsurfaceScatteringMQ_MSAA SubsurfaceScattering=SubsurfaceScatteringMQ_MSAA SSS_ENABLE_NEAR_FIELD=0 ENABLE_MSAA
-#pragma kernel SubsurfaceScatteringHQ_MSAA SubsurfaceScattering=SubsurfaceScatteringHQ_MSAA SSS_ENABLE_NEAR_FIELD=1 ENABLE_MSAA
+#pragma kernel SubsurfaceScattering SubsurfaceScattering=SubsurfaceScattering
+#pragma kernel SubsurfaceScattering_MSAA SubsurfaceScattering=SubsurfaceScattering_MSAA ENABLE_MSAA
```
FrancescoC-unity: Small heads up here, since very recently (I think PR got merged today), we support multi_compile in compute. 

I had ported the old SSS to it so it has the multi_compile for ENABLE_MSAA already defined, I guess though that some conflict will come when merging due to the SSS_ENABLE_NEAR_FIELD removal

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
```diff
@@ -113,17 +110,23 @@ float4 LoadSample(int2 pixelCoord, int2 cacheOffset)
     int2 cacheCoord = pixelCoord - cacheOffset;
     bool isInCache  = max((uint)cacheCoord.x, (uint)cacheCoord.y) < TEXTURE_CACHE_SIZE_1D;
 
-    UNITY_BRANCH if (isInCache)
```
FrancescoC-unity: Are all the UNITY_BRANCH  removed because we are good with the compiler choosing for us or because ineffective (compiler not taking the hint)?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.Migration.cs
```diff
@@ -416,5 +416,16 @@ internal static void MigrateToSeparateColorGradingAndTonemapping(ref FrameSettin
         {
             cameraFrameSettings.SetEnabled(FrameSettingsField.Tonemapping, true);
         }
+
+        internal static void MigrateSubsurfaceParams(ref FrameSettings fs, bool previouslyHighQuality)
+        {
+            // SSS moved from 25 to 46.
+            fs.SetEnabled((FrameSettingsField)25, false);
```
FrancescoC-unity: I wonder what happens if we reuse 25 before this package is released or if a user upgrade to when we use 25 again - downgrade - reupgrade.

To be on the safe side we could probably reserve out the slot 25 (adding a comment about it), though I am not 100% sure if this would be ever a problem.

com.unity.render-pipelines.high-definition/Runtime/Lighting/SurfaceShading.hlsl
EvgeniiG: Indeed, although there's no 'r' in this code.

com.unity.render-pipelines.high-definition/Runtime/Material/DiffusionProfile/DiffusionProfile.hlsl
EvgeniiG: Oh crap, that's a typo. Good catch.

com.unity.render-pipelines.high-definition/Runtime/Material/SubsurfaceScattering/SubsurfaceScattering.compute
EvgeniiG: The compiler will figure it out. Let's not pollute the code.

sebastienlagarde:
Note: for the conflict on screenshots, the screenshots need to be move as path have change, there is now an extra "None" folder since the PR from Sophia

EvgeniiG:
> Favro card: http://go/ihglm
> Two issues so far:
> 
> * Sample budget is being taken from the default HDRP asset only. Switching different quality settings has no effect. (detailed video in favro)
> * In Frame Settings SSS Quality Mode has a toggle check box by default. Should only be visible when you override global SSS toggle.
>   image

Fixed both issues.

EvgeniiG:
Tests pass locally.

PR #5974, opened 2020-02-24
Misc fixes for the sky

JulienIgnace-Unity:
### Purpose of this PR
Misc fixes for the sky
- Fixed a material leak in PBR sky renderer
- Added a tooltip to the Ambient Mode parameter in the Visual Envionment volume component.
- Static lighting sky now takes the default volume into account (this fixes discrepancies between baked and realtime lighting).
- Fixed an issue where having ambient mode set to dynamic might prevent the sky from being included in baking.
- Fixed an issue where rendering might not be updated when changing static lighting sky until user hovers the mouse over the scene view.

Fixes:
https://fogbugz.unity3d.com/f/cases/1220615/
https://fogbugz.unity3d.com/f/cases/1222439/
https://fogbugz.unity3d.com/f/cases/1222810/

Note to QA for 1220615: 
The issue spotted by users is not that baking does not take the sky into account but just an exposure problem: The sky setup for baking in the lighting environment panel does not override the exposure parameter so it takes the default value of 0 to bake lightmaps. The problem is that there is a default volume profile (project settings => default HDRP settings) that has an HDRI sky with an exposure of 11 that is used for the regular sky rendering. This causes a discrepancy between the visual sky, rendered with an exposure of 11 so super bright and the baking that used an exposure of 0 and therefore looks black). Taking the default volume into account for baking will fix this issue.
Also: Be aware that at high exposure, baking seems to be wrong (https://fogbugz.unity3d.com/f/cases/1222492/ , repro also on legacy) so the result will still look darker than the screenshots in the original case.

---
### Testing status

- [ ] Built a player


iM0ve:
The bugs are fixed and indeed the baked result is a bit darker. Tested with sky Exposure, Multiplier, and Lux overrides. Tooltip added.

Found a new issue, changes in the Static Lighting settings dont update the game view until you hover the mouse over game view and r mouse click it.

updateBug

PR #5980, opened 2020-02-24
[Shader Graph] Fix Normal Slot Connection Bug

alindmanUnity:
### Purpose of this PR 
Fixes a bug where making any change to a master node setting on PBR or Lit (HDRP) that redraws slots would lose connection to the normal slot, whether the normal slot was changed or not. [internal link only](https://fogbugz.unity3d.com/f/cases/1221215/)
Fixes an unreported side effect where connected normal slots would lose connection when coordinate space is changed. 
before:
normalSlotBefore
after:
normalSlotAfter

---
### Testing status

- [x] Built a player
- Other: 
Created a PBR and Lit master node. Made connections to both the normal slot and other slots. Made changes to the settings that would update slots (such as material mode to subsurface, or shading model to specular).  Observed that if the slot is already connected, the connection remains. 
Removed the slot connection and observed that changing the normal coordinate space still updates the default flag input on the normal slot. 

---
### Comments to reviewers


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a bug where the error `Output value 'vert' is not initialized` displayed on all PBR graphs in Universal. [1210710](https://issuetracker.unity3d.com/issues/output-value-vert-is-not-completely-initialized-error-is-thrown-when-pbr-graph-is-created-using-urp)
 - Fixed a bug where PBR and Unlit master nodes in Universal had Alpha Clipping enabled by default.
 - Fixed an issue in where analytics wasn't always working.
+- Fixed a bug where any change to the PBR master node settings would lose connection to the normal slot.
```
LandonTownsendUnity: It's fixed for both PBR and and HD Lit shaders, should that be indicated here?

alindmanUnity: I'll ask docs, but the HD Lit masternode is in HD package code, not ours, which is why I left it out.

PR #5989, opened 2020-02-25
Fix issue with AO using wrong RT Handle scale to sample history

FrancescoC-unity:
This now uses proper RT Handle scale. 

This will fix this https://fogbugz.unity3d.com/f/cases/1222421/

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/AmbientOcclusion.RenderGraph.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/AmbientOcclusion.RenderGraph.cs
@@ -22,10 +22,15 @@ namespace UnityEngine.Rendering.HighDefinition
                 {
                     EnsureRTSize(settings, hdCamera);
 
-                    var aoParameters = PrepareRenderAOParameters(hdCamera, renderGraph.rtHandleProperties, frameCount);
-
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.high-definition/Runtime/Lighting/ScreenSpaceLighting/AmbientOcclusion.RenderGraph.cs
```diff
@@ -22,11 +22,16 @@ public RenderGraphResource Render(RenderGraph renderGraph, HDCamera hdCamera, Re
                 {
                     EnsureRTSize(settings, hdCamera);
 
-                    var aoParameters = PrepareRenderAOParameters(hdCamera, renderGraph.rtHandleProperties, frameCount);
-
-                    var currentHistory = renderGraph.ImportTexture(hdCamera.GetCurrentFrameRT((int)HDCameraFrameHistoryType.AmbientOcclusion));
+                    var historyRT = hdCamera.GetCurrentFrameRT((int)HDCameraFrameHistoryType.AmbientOcclusion);
```
JulienIgnace-Unity: Could we put all this in PrepareRenderAOParameters to avoid duplication?

FrancescoC-unity: Yeah I guess we can, though that means going through the overhead of GetCurrentFrameRT twice (inside the prep and outside), unless we put the history itself in the aoParameters struct.

PR #5993, opened 2020-02-26
[Yamato] Fixing Mac issue

sophiaaar:
### Purpose of this PR

Yamato changed where the repo folder gets cloned to, and this broke our Mac Metal jobs that use scp commands. This fix makes use of an environment variable to reference the source directory.

---
### Testing status

**Yamato**: Only tested one job, but the same fix was made in each file, so I'm confident that this job was all that was needed to verify the fix

---
### Comments to reviewers
Only affects the mac minis because these are devices rather than VMs


phi-lira:
While looking at this do we have any other relative or hardcoded path that we could use env variables and prevent future issues?

phi-lira:
bors try+

sophiaaar:
the issue that was fixed here doesnt seem to affect 8.x.x/release and 7.x.x/release, so removing the labels for now

PR #6008, opened 2020-02-27
[SG] Fix fog density returning zero with an Unlit Master Node

Nightmask3:
---
### Purpose of this PR

Bug details: https://fogbugz.unity3d.com/f/cases/1215767/

Current behavior:

3obBkcdTES

Fixed behavior:

gT9Ehubc4w

Previously the density output on the fog node would always return a 0 when connected to an unlit master node. This turned out to be an issue with certain shader constants being not properly included in the unlit shaders for the Unlit master node, **meaning that any unlit shader graph would not have access to fog density information.** 

This only matters to URP projects as the fog node is by design not intended to work in HDRP and only return a value of 0 if used an HDRP project.
---
### Testing status

- [x] Opened test project
- [x] Ran GFX tests from here: https://github.cds.internal.unity3d.com/unity/Graphics-QA-FTP-GraphicsGeneral
    - Image comparison tests
    - Performance tests
- [x] Checked for C# and shader warnings 

**Yamato**:

---
### Comments to reviewers

I reached out to the URP team and asked if this solution looks appropriate, and Felipe, the team lead, responded that it looks good.


com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalUnlitSubShader.cs
```diff
@@ -106,6 +107,7 @@ class UniversalUnlitSubShader : IUnlitSubShader
                 "prefer_hlslcc gles",
                 "exclude_renderers d3d11_9x",
                 "target 2.0",
+                "multi_compile_fog",
```
phi-lira: Should only be done for the unlit pass. Depth only and shadow caster don't need it.

com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalUnlitSubShader.cs
```diff
@@ -156,6 +158,7 @@ class UniversalUnlitSubShader : IUnlitSubShader
...
                 "multi_compile_instancing",
```
phi-lira: Seems like this is shadow caster pass. For variants should not be added.

com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalUnlitSubShader.cs
Nightmask3: Ah right I was fairly certain that this was the case but I wasn't sure, thanks for confirming!

com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalUnlitSubShader.cs
Nightmask3: 👍

phi-lira:
@alindmanUnity let me know if I should merge this or wait until your PR lands.

alindmanUnity:
> @alindmanUnity let me know if I should merge this or wait until your PR lands.

I can roll this in to targets if we'd rather just land in one PR -- it's a one line change in the pass definition and we want the targets PR to land within the week.

phi-lira:
Can you confirm we should close this PR?

phi-lira:
Closing as it's merged in ShaderGraph targets.

PR #6009, opened 2020-02-27
[Shader Graph][7.x.x] Minor Caching Improvements to Searcher

alindmanUnity:
### Purpose of this PR
Backport of #5997 
Previously, the `Create Node` menu/searcher window regenerated an instance of every node on every load call. This makes minor improvements to the caching of entries to speed up the load of the searcher for the majority of calls. 
The searcher window will now only generate new entries when changes to properties or keywords are detected, or when the shader graph window is refocused (to account for subgraph changes). Only the first invoke after these changes are detected will lag, and every call after that will use a cache and load much faster.
More improvements can be made later as well but this speeds up 80% of the use of the searcher and is probably worth putting in for UX improvements. 
Issue link: [public](https://issuetracker.unity3d.com/product/unity/issues/guid/1209567/) [internal](https://fogbugz.unity3d.com/f/cases/1209567/)

---
### Testing status

- [x] Built a player
- Other: 
- Rapidly opened and closed the searcher (using spacebar) and observed rapidly improved load times. 
- Added many properties, keywords, and nodes from the searcher entries to the graph to ensure there are no guid errors in the copy system. 
- Checked that when refocusing the window, adding, removing, or changing a property or keyword the next load did lag, but reflected the changes in the search entries. Every load after that remained fast. 

---
### Comments to reviewers


alindmanUnity:
closing because the optimizations do not help in this version

PR #6011, opened 2020-02-27
Scalable volumetrics

EvgeniiG:
### Purpose of this PR
Added support for arbitrary resolution scaling of Volumetric Lighting based on the Volume Framework. This means you can control the cost and quality of volumetrics by placing volumes in your scene.

---
### Testing status

**Manual Tests**: Yes. Need to test VR (have not tested).

---
### Comments to reviewers
The RTHandle system does not allow to control the resolution of the buffers using the Volume Framework, so I had to write my own system based on RenderTexture. Unfortunately, RenderGraph does not appear to support RenderTexture, so I couldn't get RenderGraph to work. I have indicated in the code where the issue is.


EvgeniiG:
I have a weird issue while switching the scenes. I will investigate.

EvgeniiG:
Tests should now pass except for instabilities in URP/VFX that have nothing to do with my changes.

EvgeniiG:
Nope, that was easy to add.

What does not work is controlling the resolution of the buffers using the Volume Framework (see `ComputeVolumetricViewportSize`). You can only auto-resize an RTHandle using a lambda, and that would not work since you would need to capture the camera to get the Volume Stack, and we share textures across multiple cameras.

The resulting code is efficient and clean (and far simpler than the RTHandle code), and works pretty well.

JulienIgnace-Unity:
How about this part of my comment:

That being said, it's preferable that you use RTHandle anyway, you can still alloc RTHandles with explicit width height and depth so definitely don't need to switch to regular RenderTexture and this way it will be compatible with RenderGraph.

You can still use RTHandles instead of RenderTexture. Just use the allocation functions with explicit sizes. They won't go through the auto rescale system. This way the render graph version can import them.

Also, thinking about it, I have a major concern about the method you choseas it's going to completely break down with dynamic resolution as the system is going to reallocate render textures potentially every frame. With the RTHandle system and the fact that we don't reallocate, we don't have this problem.

EvgeniiG:
> Also, thinking about it, I have a major concern about the method you choseas it's going to completely break down with dynamic resolution as the system is going to reallocate render textures potentially every frame. 

Umm, why? I already replicate the sub-allocation functionality of RTHandles.
The resolution of the volumetric buffer is derived from the resolution of the camera, so it should dynamically scale just fine.

EvgeniiG:
>You can still use RTHandles instead of RenderTexture. Just use the allocation functions with explicit sizes. They won't go through the auto rescale system. This way the render graph version can import them.

I can try. I recall it not working too well, but I can't remember why.

EvgeniiG:
I encourage you to look at my existing code, though, it's clean and dead simple.

JulienIgnace-Unity:
> Umm, why? I already replicate the sub-allocation functionality of RTHandles.
My bad, did not read properly the realloc condition!

EvgeniiG:
@JulienIgnace-Unity I switched to RTHandles per your request. RenderGraph should now work. Please verify.

EvgeniiG:
@fabien-unity Please make sure I did not break XR with my changes. :-)

EvgeniiG:
Tests pass locally, running Yamato.

PR #5944, opened 2020-02-20
Fixed issue that caused materials to reimport if clearing player prefs.

phi-lira:
### Purpose of this PR
 Fixed [case 1192259](https://issuetracker.unity3d.com/issues/lwrp-clearing-playerprefs-through-a-script-or-editor-causes-delay-and-console-errors-to-appear-when-entering-the-play-mode)

### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
 Used the same approach as HDRP to store project settings file with upgrade versions.

PR #6014, opened 2020-02-27
Fix VisualEffect URP test

PaulDemeulenaere:
### Purpose of this PR
*WIP* revert of this PR https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5944

It seems to time out after a few frame running VisualEffect tests.

*DO NOT MERGE* (without finding the real root cause)


phi-lira:
I will merge this to have yamato green and will work on a fix asap to land the bug.

phi-lira:
Fix might be as simple as the URPProjectSettings file missing for new test projects. Will check this.

PR #6015, opened 2020-02-27
Fix texture curve dispose

alelievr:
### Purpose of this PR
Fixes The texture curve being destroyed from another thread than main: https://fogbugz.unity3d.com/f/cases/1211754

I ran the URP tests locally and no errors


sambickley:
Will test this for platforms.

PR #6018, opened 2020-02-27
Handled targetTexture dimension not compatible with XR case. [case 1198052]

thomas-zeng:
### Checklist for PR maker

---
### Purpose of this PR

https://fogbugz.unity3d.com/f/cases/1198052/

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
Note: this hides SPI mode rendering things overlapped in URP when rendering to texture2D: we disable XR in that case. SPI requires targetTexture to be texArr type.

Although the current logic makes sense, it **DOES NOT** have parity with the legacy renderer. Legacy renderer disables XR if targetTexture is being used. We can't do it in URP as URP uses targetTexture in cameraStack and XR needs to support cameraStack in SPI/Multiview.

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
+++ b/com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
@@ -183,8 +183,7 @@ namespace UnityEngine.Rendering.Universal
             bool isGameCamera = IsGameCamera(camera);
             bool isCompatWithXRDimension = true;
 #if ENABLE_VR && ENABLE_VR_MODULE
-            isCompatWithXRDimension = camera.targetTexture ?
-                camera.targetTexture.dimension == UnityEngine.XR.XRSettings.deviceEyeTextureDimension : true;
+            isCompatWithXRDimension &= (camera.targetTexture ? camera.targetTexture.dimension == UnityEngine.XR.XRSettings.deviceEyeTextureDimension : true);
 #endif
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipelineCore.cs
```diff
@@ -181,7 +181,12 @@ public static bool IsStereoEnabled(Camera camera)
                 throw new ArgumentNullException("camera");
 
             bool isGameCamera = IsGameCamera(camera);
-            return XRGraphics.enabled && isGameCamera && (camera.stereoTargetEye == StereoTargetEyeMask.Both);
+            bool isCompatWithXRDimension = true;
+#if ENABLE_VR && ENABLE_VR_MODULE
+            isCompatWithXRDimension = camera.targetTexture ?
+                camera.targetTexture.dimension == UnityEngine.XR.XRSettings.deviceEyeTextureDimension : true;
```
grayson-unity: Minor suggestion:  this might be clearer as simply
`isCompatWithXRDimension &= camera.targetTexture.dimension == UnityEngine.XR.XRSettings.deviceEyeTextureDimension;`

thomas-zeng: Yes, this is better and also handles isCompatWithXRDimension false case!
Addressed in 64054cc

thomas-zeng:
Moved to #6036

PR #6031, opened 2020-02-28
Shadergraph/first pass integration tests

elizabeth-legros:
### Checklist for PR maker

---
### Purpose of this PR
Adding a bunch of integration tests to simulate more userlike graph creation
---
### Testing status

- Other: 

Any shader graphs or subgraphs placed in TestProjects/ShaderGraph/Assets/CommonAssets/Graphs/BuildGraphTests/
will get smoke tested in creation. There is also an undo-redo test that runs on one of the larger graphs in that folder (test case source attribute)

The repo test project
---
### Comments to reviewers


TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+                window.UpdateAsset();
+                window.Close();
+                Assert.IsNotNull(AssetDatabase.LoadAssetAtPath<Shader>(outputPath));
+            }
+            
+        }
+
+        private static void TrySaveWindows(MaterialGraphEditWindow copyWindow, MaterialGraphEditWindow testGraphWindow)
+        {
+
```
Nightmask3: Nit: whitespace

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+        {
+            var window = EditorWindow.CreateWindow<MaterialGraphEditWindow>(typeof(MaterialGraphEditWindow), typeof(SceneView));
+            window.Initialize(AssetDatabase.AssetPathToGUID(assetPath));
+            return window;
+        }
+
+        public static void CreateEmptyTestGraph(string basedOnGraph)
+        {
+            var window = OpenShaderGraphWindowForAsset(basedOnGraph);
+            GraphObject graphObject = window.GetPrivateProperty<GraphObject>("graphObject");
```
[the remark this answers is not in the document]
zadi0648: Also note that this is an class (GraphObject) is one that Peter is going to delete in the future with the serialization changes.

I think it's on his hands to make this change when that change comes.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+                                  + '.' + ShaderSubGraphImporter.Extension;
+                FileUtilities.WriteShaderGraphToDisk(outputPath, graphData);
+                AssetDatabase.Refresh();
+                window.UpdateAsset();
+                window.Close();
+            }
+            else
+            {
+                graphData.AddNode(rootNode);
+                graphData.path = "Shader Graphs";
```
zadi0648: This shouldn't matter for tests reasons, but it might be best to have the test keep the original path and also for the path to match conventions of the project (unless if there is some test reason for that to differ), more info coming in another comment...

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+                    Debug.LogError(cantSaveOrCloseWindowsException);
+                }
+                throw cantOpenWindowsOrAccessPrivatePropertyException;
+            }
+
+            yield return new WaitForSecondsRealtime(userTime);
+
+            var nodeLookup = new Dictionary<AbstractMaterialNode, AbstractMaterialNode>();
+            var temporaryMarks = ListPool<(AbstractMaterialNode, SlotReference?, SlotReference?)>.Get();
+            var permanentMarks = ListPool<(AbstractMaterialNode, SlotReference?, SlotReference?)>.Get();
```
zadi0648: Excuse me for taking the lazy approach, but could we get a very brief rundown of the temporary versus permanent thing and how the algorithm works? (Imo all algorithms could use a single line comment explanation)

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+            nodeLookup.Add(graphObjectToCopy.graph.outputNode, testGraphObject.graph.outputNode);
+
+
+            foreach (var node in graphObjectToCopy.graph.GetNodes<AbstractMaterialNode>())
+            {
+                stack.Push((node, null, null));
+            }
+            while (stack.Count > 0)
+            {
+                (AbstractMaterialNode node, SlotReference? to, SlotReference? from) = stack.Pop();
```
zadi0648: Nit-pick: could put everything inside this loop into a method (for in general we want to avoid having ginormous method, even in tests / single-method-single-purpose rule)

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+            Assert.IsNotNull(property);
+
+            //ToConcrete does not usually copy over drawstate, so create template node ourselves 
+            AbstractMaterialNode concrete = property.ToConcreteNode();
+            concrete.drawState = propertyNode.drawState;
+            concrete.owner = propertyNode.owner;
+            yield return UserlikeAddNodeUsingSearcherAndCopyValues(concrete, graphObject, graphEditWindow, graphEditorView);
+            Assert.IsNotNull(searcherAddedNode);
+
+            yield return new WaitForSecondsRealtime(userTime);
```
zadi0648: We may have talked about this, why real time? (seems like we want a certain number of editor updates to happen or something)

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
```diff
@@ -0,0 +1,536 @@
...
+    {
+
+        private static readonly string testGraphLocation = "Assets/Testing/CreatedTestGraphs/";
+        private static readonly string testPrefix = "_Test_";
+
+        public static void CloseAllOpenShaderGraphWindows()
+        {
+            foreach (MaterialGraphEditWindow graphWindow in Resources.FindObjectsOfTypeAll<MaterialGraphEditWindow>())
+            {
+                graphWindow.Close();
```
zadi0648: You may want to close-without-saving (at least, for nothing else, for local runs), which you can do by setting the graphObject to null before closing. Or, leave existing SGs windows alone and instead have this system use a fresh SG instead, as I'll mention in my other comment.

zadi0648:
Adding Landon: anything else we should test?

zadi0648:
Notes from the meeting:

- Changing userTime to 1.0: fixed Unity crashing, wait until the compiled shader is finished? Also perhaps a missing try-catch? (failure sure, why the Unity crash though?)
- Adding slots to Sub Graph output node to have validation / wait after each slot (as users don't add all the slot suddenly at once and there-have-been-bugs depending on a sub graph's outputs)
- Setting the path causes the SG to immediately open dirty: look into that (Zack will assign this to himself unless someone else wants to take a look)
- try yield return null versus try yeild return 0seconds: are they really the same? Find out, if not try the former instead as that should guarantee that a UI refresh has happened.

TestProjects/ShaderGraph/Assets/CommonAssets/Editor/GraphCreationTests.cs
elizabeth-legros: changed to waiting until graph is done compiling

PR #6042, opened 2020-03-02
Analytics added to Universal

martint-unity:
### Purpose of this PR
Adding analytics for Universal users. This helps the Universal team focus our efforts on the most common scenarios, and better understand the needs of our customers.


martint-unity:
Need to test building projects using this branch. 
This will send data after successful builds.

com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
```diff
@@ -0,0 +1,99 @@
...
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.Universal;
+using UnityEditor.Build;
+using UnityEditor.Build.Reporting;
+
+namespace UnityEditor.Rendering.Universal
+{
+    class UniversalAnalytics : IPostprocessBuildWithReport
+    {
+        static bool s_EventRegistered = false;
```
ellioman: Do we need this variable? 
Could EnableAnalytics() just return (result == AnalyticsResult.Ok) or do we need to cache it?

com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        struct AnalyticsData
+        {
+            public string[] renderer_data;
+            public int renderer_data_amount;
+            public string[] renderer_features;
+            public int renderer_features_amount;
+            public string main_light_rendering_mode;
+            public string additional_light_rendering_mode;
```
ellioman: Do we not want to keep using the Camel Case standard rather than using Snake Case?

martint-unity: yes that is for the database.

com.unity.render-pipelines.universal/Editor/UniversalAnalytics.cs
martint-unity: Could probably work.

PR #6058, opened 2020-03-03
Hdrp/volume debug

adrien-de-tocqueville:

### Checklist for PR maker
- [x] Have you updated or added the documentation for your PR? When you add a new feature, change a property name, or change the behavior of a feature, it's best practice to include related documentation changes in the same PR.

---
### Purpose of this PR
This PR is adding a new debug panel for volumes as well as a new table widget for the debug UI.

debug panel

---
### Testing status

- [x] Built a player
- Other: 

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/HDRP%252Fvolume-debug

---
### Comments to reviewers


com.unity.render-pipelines.core/Runtime/Volume/Volume.cs
```diff
@@ -82,7 +82,7 @@ public VolumeProfile profile
             set => m_InternalProfile = value;
         }
 
-        internal VolumeProfile profileRef => m_InternalProfile == null ? sharedProfile : m_InternalProfile;
+        public VolumeProfile profileRef => m_InternalProfile == null ? sharedProfile : m_InternalProfile;
```
JulienIgnace-Unity: I'd like to have @Chman opinion on this. Is that dangerous?

iM0ve:
Test status: in progress

Issues:
1) Deprecated fog components are present and should be removed.

com.unity.render-pipelines.core/Runtime/Volume/Volume.cs
Chman: Not necessarily dangerous but honestly I'd rather avoid it. The volume API currently follows the same pattern as the Material API (with auto-instantiated profiles etc), and exposing this field will only make it more confusing, which is why we have `HasInstantiatedProfile()` in the public API. It's also missing documentation.

adrien-de-tocqueville: Ok, I'll use `HasInstantiatedProfile`.

sebastienlagarde:
Poking @iM0ve for an update on QA side after all those fix commit :)

PR #6060, opened 2020-03-03
Adding all shaders to the AlwaysIncludedShaders list 

ellioman:
Testing whether this will speed up the time to make our test project builds.


ellioman:
bors try

phi-lira:
did we see improvement on this?

PR #6066, opened 2020-03-04
Use 2-component edge texture for SMAA on Mali

AlbinBernhardssonARM:
---
### Purpose of this PR

The SMAA edge texture only requires 2 components, but using 4 may be
faster on some vendors (SMAA authors mention it is faster on Nvidia).
2 components is faster on Mali, so make sure 2 components is used
when running on a Mali device but keep 4 components for other vendors.

---
### Testing status

Measurements were done in the URP default template scene with SMAA on
medium settings and all other post-processing effects turned off on a
Mali G-76.

| Counter | Change (error) |
| --- | --- |
| Mali Core Internal Bus Write Beats: Tile buffer internal | -12.6% (+-2.3%) |
| Mali Core Tiles (CRC): Killed tiles | +16.1% (+-2.3%) |
| Mali External Bus Accesses: Write transaction | -10.9% (+-2.3%) |
| Mali External Bus Beats: Write beat | -10.8% (+-2.3%) |
| Mali External Bus Outstanding Writes: 0-25% | -14.8% (+-2.1%) |
| Mali L2 Cache Lookups: Write lookup | -10.2% (+-3.2%) |

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
@@ -49,6 +49,7 @@ namespace UnityEngine.Rendering.Universal.Internal
         const int k_MaxPyramidSize = 16;
         readonly GraphicsFormat m_DefaultHDRFormat;
         bool m_UseRGBM;
+        readonly GraphicsFormat m_SMAAEdgeFormat;
         readonly GraphicsFormat m_GaussianCoCFormat;
         Matrix4x4 m_PrevViewProjM = Matrix4x4.identity;
         bool m_ResetHistory;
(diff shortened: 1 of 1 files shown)


com.unity.render-pipelines.universal/Runtime/Passes/PostProcessPass.cs
```diff
@@ -92,6 +93,12 @@ public PostProcessPass(RenderPassEvent evt, PostProcessData data, Material blitM
                 m_UseRGBM = true;
             }
 
+            // Only two components are needed for edge render texture, but on some vendors four components may be faster.
+            if (SystemInfo.graphicsDeviceVendor.ToLowerInvariant().Contains("arm"))
```
phi-lira: Maybe also check if format is supported:
```suggestion
            if (SystemInfo.IsFormatSupported(GraphicsFormat.R8G8_UNorm) && SystemInfo.graphicsDeviceVendor.ToLowerInvariant().Contains("arm"))
```

AlbinBernhardssonARM: Done

phi-lira:
LGTM.
Changed base so we run ABV test before merging to master.

PR #6069, opened 2020-03-04
[CI] Turn off Bors and turn on PR expression triggers

sophiaaar:
### Purpose of this PR
* Turn off Bors
* Turn on Yamato expression triggers
* Update the PR template to explain the new expression triggers and the different accepted PR workflows

---
### Testing status

Created 2 dummy PRs to test the different types of trigger #6063 #6064  (I don't think the commit history shows very well for some reason)
* Verified that creating a draft PR **does not** start ABV
* Verified that ABV **does** start when the PR is marked as ready for review
* Verified that ABV **does not** start when the change is only docs

---
### Comments to reviewers
* See the PR template (changed in this PR) for more info
* Cancelled ABV since no changes were made to pipelines or dependencies


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,10 @@
-**DONT FORGET TO ADD A CHANGELOG**
+### **Please read**
```
phi-lira: We should add this info in index.md of the repo not to PR template. PR template is already too verbose now and many people don't remove these sections...

PR #6094, opened 2020-03-05
Importance Sampling Pipeline

skhiat:
Result A:
image
image
Result B:
image
image

### **Please read**
**PR workflow guidelines**
* SRP ABV will start automatically when you open your PR

### Checklist for PR maker

---
### Purpose of this PR
Allow Importance Sampling

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Lighting/AtmosphericScattering/ShaderVariablesAtmosphericScattering.hlsl
```diff
@@ -3,11 +3,15 @@
 #else
     TEXTURE3D(_VBufferLighting);
     TEXTURECUBE_ARRAY(_SkyTexture);
+    bool            _SkyTextureImportanceSamplerReady;
```
JulienIgnace-Unity: Two things about these constants  (_SkyTextureImportanceSamplerReady and _SkyTextureSizeInfos)
- They should not be included in the #else because this way they are going to be in a default global CB which is bad (one more CB to bind) so they should be in the generated part.
- I don't see them used anywhere in this PR, did I miss something?

skhiat: 1. [Fixed] Moved
2. Will be used on incoming PR

skhiat:
> A general question on the framework: Is it supposed to be used realtime? (even not every frame but frequently?) Or in a standalone player?
> Asking this because I see a ton of usage of temporary materials and RTHandles created on the fly for different purpose and this is gonna cause a lot of garbage and potentially bad perf.

True, TBD, coudl need deeper refactoring.

PR #6103, opened 2020-03-06
Update the GC alloc test message [skip CI]

alelievr:
Update the GC alloc test message


sebastienlagarde:
Please target master to not block us for HDRP/Staging.

PR #6106, opened 2020-03-06
Fixed a bug in debug light volumes

anisunity:
A bug was introduced in this commit: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/0e604b48b19ab8d54e58036a8850b2bb126642e5#diff-841af5d658cecc486b4a2dfc265faba6L131

The size of the volume was never overriden  forr point and area lights.

Fixes this bug: https://fogbugz.unity3d.com/f/cases/1224323/

image


JMargevics:
All light volumes look correct now. They react to range/angle/ration adjustments correspondingly  ✔️

PR #6108, opened 2020-03-06
Fixed ray traced point and spot lights shadows not rejecting history when semi-transparent or colored.

anisunity:
History rejection was working for point spot lights, the issue was with semi-transparent and colored shadows. This PR fixes this bug: https://fogbugz.unity3d.com/f/cases/1225435/


anisunity:
Weird, I am going to check it

PR #6111, opened 2020-03-06
Hdrp/fix/particle light shadows

alelievr:
### Purpose of this PR
Half fixes the shadow Shuriken for particle lights.

We still have issues when there is more than one particle in the particle system that emits light but at least this one is good.

Case postponed because it will be solved by a future PR: https://fogbugz.unity3d.com/f/cases/1223054/


alelievr:
Not sure if the changelog is really relevant here but i added it

PR #6112, opened 2020-03-06
[Draft] RenderPass Deferred initial NOT FINAL PR

jonuuukas:
will update this next week, as i'm in a hurry

if you want to test - check it on windows vulkan editor or android


com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -33,6 +33,28 @@
 #define BUMP_SCALE_NOT_SUPPORTED !SHADER_HINT_NICE_QUALITY
 #endif
 
+#ifndef API_SUPPORTS_RENDER_PASS
+
+//Should be implemented properly per-platform
+#define TEXTURE2D_FLOAT_MS(textureName) Texture2DMS<float4> textureName
+#define TEXTURE2D_HALF_MS(textureName) Texture2DMS<half4> textureName
+#define TEXTURE2D_INT_MS(textureName) Texture2DMS<int4> textureName
+#define TEXTURE2D_UINT_MS(textureName) Texture2DMS<uint4> textureName
```
hkr: You could probably get rid of those and just change UNITY_DECLARE_FRAMEBUFFER_INPUT_MS below.

com.unity.render-pipelines.universal/Shaders/Utils/TileDeferred.shader
```diff
@@ -180,18 +180,19 @@ Shader "Hidden/Universal Render Pipeline/TileDeferred"
 
     #endif
 
-    TEXTURE2D_X(_CameraDepthTexture);
-    TEXTURE2D_X_HALF(_GBuffer0);
-    TEXTURE2D_X_HALF(_GBuffer1);
-    TEXTURE2D_X_HALF(_GBuffer2);
+    UNITY_DECLARE_FRAMEBUFFER_INPUT(0, HALF);
```
hkr: We should probably change these macros so that we don't have to use all caps type names

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -441,18 +460,137 @@ internal void Clear(CameraRenderType cameraType)
...
+                    ExecuteRenderPass(context, renderPass, ref renderingData, eyeIndex);
+                }
+                else
+                {
+                    int depthAttachmentIdx = -1;
+                    var desc = renderPass.renderPassDescriptor;
+                    var colorCount = RenderingUtils.GetValidColorBufferCount(renderPass.colorAttachments);
+
+                    NativeArray<int> indices =
+                        new NativeArray<int>((int)colorCount, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
```
hkr: I'd make it initialized. Easier to catch errors and not really worth optimizing.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -441,18 +460,137 @@ internal void Clear(CameraRenderType cameraType)
...
-            m_CameraDepthTarget = BuiltinRenderTextureType.CameraTarget;
+            m_CameraColorTarget = RenderTargetHandle.CameraTarget;
+            m_CameraDepthTarget = RenderTargetHandle.CameraTarget;
         }
 
         void ExecuteBlock(int blockIndex, NativeArray<int> blockRanges,
             ScriptableRenderContext context, ref RenderingData renderingData, int eyeIndex = 0, bool submit = false)
         {
             int endIndex = blockRanges[blockIndex + 1];
+            bool renderPassStarted = false;
```
hkr: Why do we need all of this?
I thought only deferred renderer should use RenderPass. Why do we have to touch the base class?

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -21,6 +21,7 @@ struct FragmentOutput
     half4 GBuffer1 : SV_Target1; // maps to GBufferPass.m_GBufferAttachments[1] on C# side
     half4 GBuffer2 : SV_Target2; // maps to GBufferPass.m_GBufferAttachments[2] on C# side
     half4 GBuffer3 : SV_Target3; // maps to DeferredPass.m_CameraColorAttachment on C# side
+    float4 Depth : SV_Target4;
```
hkr: why 4 components for depth?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -465,7 +603,7 @@ void ExecuteRenderPass(ScriptableRenderContext context, ScriptableRenderPass ren
             Camera camera = cameraData.camera;
             bool firstTimeStereo = false;
 
-            CommandBuffer cmd = CommandBufferPool.Get(k_SetRenderTarget);
+            CommandBuffer cmd = CommandBufferPool.Get(k_RenderPass + " " + renderPass.GetType().ToString());
```
kaychang-unity: Will this cause dynamic allocation/release ?

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
kaychang-unity: Is it a renderpass limitation to force us to output depth as a separate color RT? From talking with ARM, it may be faster for Mali GPUs but I don't think it would be most likely slower on other platforms, so if we need it for renderpass, I would like to have it optional.

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
```diff
@@ -1,176 +1,178 @@
...
-    TEXTURE2D_X_HALF(_GBuffer0);
-    TEXTURE2D_X_HALF(_GBuffer1);
-    TEXTURE2D_X_HALF(_GBuffer2);
-    float4x4 _ScreenToWorld;
-
-    float3 _LightPosWS;
-    float3 _LightColor;
+    }
+
+    UNITY_DECLARE_FRAMEBUFFER_INPUT(0, HALF);
```
kaychang-unity: We decided to not support XR for this preview release, but we may support it later. Will the macro works in XR mode ( the meaning of _X_) ?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -31,25 +31,36 @@ namespace UnityEngine.Rendering.Universal
...
         }
 
-        public RenderTargetIdentifier colorAttachment
+        public RenderTargetHandle colorAttachment
         {
             get => m_ColorAttachments[0];
         }
 
-        public RenderTargetIdentifier depthAttachment
+        public List<RenderTargetHandle> inputAttachments
```
phi-lira: There's an API inconsistency here that for colorAttachments we return an array an [] and the input we return List<T>. We should be consistent, If we want to add something other than array we should probably use `NativeArray` instead of List<T>, but then we need to have overloads for [] anyway as we already have it.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -65,31 +76,48 @@ public Color clearColor
         internal bool overrideCameraTarget { get; set; }
         internal bool isBlitRenderPass { get; set; }
 
-        RenderTargetIdentifier[] m_ColorAttachments = new RenderTargetIdentifier[]{BuiltinRenderTextureType.CameraTarget};
-        RenderTargetIdentifier m_DepthAttachment = BuiltinRenderTextureType.CameraTarget;
+        internal bool hasInputAttachment;
```
phi-lira: should we make this a property that just checks if the input list is null or empty ?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -203,6 +275,11 @@ public virtual void OnFinishCameraStackRendering(CommandBuffer cmd)
         /// <param name="material">Material to use.</param>
         /// <param name="passIndex">Shader pass to use. Default is 0.</param>
         /// <seealso cref="ScriptableRenderer"/>
+        public void Blit(CommandBuffer cmd, RenderTargetHandle source, RenderTargetHandle destination, Material material = null, int passIndex = 0)
```
phi-lira: This API seems not needed. If you have `RenderTargetHandle`, then you can use RenderTargetHandle.Identifier().

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -33,6 +33,28 @@
 #define BUMP_SCALE_NOT_SUPPORTED !SHADER_HINT_NICE_QUALITY
 #endif
 
+#ifndef API_SUPPORTS_RENDER_PASS
```
phi-lira: This should be in CoreRP ShaderLibrary so custom render pipelines also benefit from this. 
You could add to Common.hlsl. URP Core.hlsl includes Common.hlsl so it should be fine.

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -33,6 +33,28 @@
...
+
+#define UNITY_DECLARE_FRAMEBUFFER_INPUT_MS(idx, type) TEXTURE2D_##type##_MS(_UnityFBInput##idx); float4 _UnityFBInput##idx##_TexelSize
+#define UNITY_READ_FRAMEBUFFER_INPUT_MS(idx, sampleIdx, v2fvertexname) LOAD_TEXTURE2D_MSAA(_UnityFBInput##idx, v2fvertexname.xy, sampleIdx)
+#define UNITY_READ_FRAMEBUFFER_SCREENSPACE_INPUT_MS(idx, sampleIdx, v2fvertexname) LOAD_TEXTURE2D_MSAA(_UnityFBInput##idx, v2fvertexname.xy * _UnityFBInput##idx##_TexelSize.zw, sampleIdx)
+
+#undef TEXTURE2D_FLOAT_MS
+#undef TEXTURE2D_HALF_MS
+#undef TEXTURE2D_INT_MS
+#undef TEXTURE2D_UINT_MS
+#endif
```
phi-lira: Put the `#else` here and define the render pass macros instead of defining in Vulkan.hlsl and Metal.hlsl

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
```diff
@@ -160,21 +182,21 @@ float4 ComputeScreenPos(float4 positionCS)
...
-    float fogFactor = saturate(clipZ_01 * unity_FogParams.z + unity_FogParams.w);
-    return real(fogFactor);
-#elif defined(FOG_EXP) || defined(FOG_EXP2)
-    // factor = exp(-(density*z)^2)
-    // -density * z computed at vertex
-    return real(unity_FogParams.x * clipZ_01);
-#else
-    return 0.0h;
-#endif
+real ComputeFogFactor(float z)
```
phi-lira: I can't seem to find any changes here, are there any I'm missing?

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
```diff
@@ -34,7 +35,7 @@ uint UnpackMaterialFlags(float packedMaterialFlags)
 }
 
 // This will encode SurfaceData into GBuffer
-FragmentOutput SurfaceDataToGbuffer(SurfaceData surfaceData, InputData inputData, half3 globalIllumination, int lightingMode)
+FragmentOutput SurfaceDataToGbuffer(SurfaceData surfaceData, InputData inputData, half3 globalIllumination, int lightingMode, float4 clipPos)
```
phi-lira: Maybe clipPos should be addded in `InputData` instead?

com.unity.render-pipelines.universal/Shaders/Utils/TileDeferred.shader
phi-lira: Can we have template for HALF or FLOAT, 
Something like DeclareFrameBufferInput<float4> , DeclareFrameBufferInput<half4> ?

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -73,12 +75,12 @@ void SetShaderTimeValues(float time, float deltaTime, float smoothDeltaTime, Com
             }
         }
 
-        public RenderTargetIdentifier cameraColorTarget
+        public RenderTargetHandle cameraColorTarget
```
phi-lira: Breaking API change.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
```diff
@@ -187,7 +191,7 @@ protected virtual void Dispose(bool disposing)
         /// </summary>
         /// <param name="colorTarget">Camera color target. Pass BuiltinRenderTextureType.CameraTarget if rendering to backbuffer.</param>
         /// <param name="depthTarget">Camera depth target. Pass BuiltinRenderTextureType.CameraTarget if color has depth or rendering to backbuffer.</param>
-        public void ConfigureCameraTarget(RenderTargetIdentifier colorTarget, RenderTargetIdentifier depthTarget)
+        public void ConfigureCameraTarget(RenderTargetHandle colorTarget, RenderTargetHandle depthTarget)
```
phi-lira: Breaking API change.

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
phi-lira: Yes, good catch. I have a PR that will allocate GC.Alloc tests that will make catching things like this automatic.

com.unity.render-pipelines.universal/Runtime/ScriptableRendererFeature.cs
```diff
@@ -31,5 +32,9 @@ void OnValidate()
         {
             Create();
         }
+
+        public abstract void SetFeatureColorTarget(RenderTargetHandle color);
```
phi-lira: This will cause all existing Features to fail to compile as they will need to implement this.

phi-lira: It's not really clear to me why we need this. Checking RenderObjects feature it's only used to call ConfigeTarget for the RenderPass?

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -65,31 +76,48 @@ public Color clearColor
...
+        internal bool hasInputAttachment;
+        internal bool useNativeRenderPass;
+
+        internal RenderTargetHandle[] m_ColorAttachments = new RenderTargetHandle[] { RenderTargetHandle.CameraTarget };
+        internal List<RenderTargetHandle> m_InputAttachments = new List<RenderTargetHandle>(0);
+        internal RenderTargetHandle m_DepthAttachment;// = RenderTargetHandle.CameraTarget;
+        RenderPassDescriptor m_RenderPassDescriptor;
         ClearFlag m_ClearFlag = ClearFlag.None;
         Color m_ClearColor = Color.black;
+        internal RenderBufferLoadAction m_LoadAction = RenderBufferLoadAction.DontCare;
```
phi-lira: I'm not sure if configuring Load and Store actions for render pass makes sense. 
It should be configured per-attachment instead, using internally `AttachmentDescriptor`.

com.unity.render-pipelines.universal/Editor/ShaderGraph/SubShaders/UniversalPBRSubShader.cs
```diff
@@ -232,6 +232,7 @@ class UniversalPBRSubShader : IPBRSubShader
             requiredVaryings = new List<string>()
             {
                 "Varyings.positionWS",
+                "Varyings.positionCS",
```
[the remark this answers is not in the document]
alindmanUnity: this file doesn't exist anymore. the changes can be updated fine into the targets but this file won't merge. you can ping me for assistance on upgrading.

com.unity.render-pipelines.universal/Shaders/Utils/Sampling.shader
```diff
@@ -79,5 +79,30 @@ Shader "Hidden/Universal Render Pipeline/Sampling"
             }
             ENDHLSL
         }
+        // 1 - Read Framebuffer
```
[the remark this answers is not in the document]
jonuuukas: this is used in CopyColor pass which is not necessarily used in post processing, but the idea is just to copy on-tile data instead of blitting into separate render texture

com.unity.render-pipelines.universal/Runtime/ScriptableRenderer.cs
jonuuukas: Deferred renderer is just the first renderer to get RenderPass-ed

com.unity.render-pipelines.universal/ShaderLibrary/UnityGBuffer.hlsl
jonuuukas: The limitation is that depth buffer cannot be read on iOS and on fallback platforms (so basically only vulkan can avoid this). Currently, fallback platforms get a depth copy engine-side if used as input, but that creates bad behaviour as we cannot guarantee same depth attachment will be used in multiple render passes. Do you want a keyword added to disable this behaviour, @kaychang-unity ?

com.unity.render-pipelines.universal/Shaders/Utils/StencilDeferred.shader
jonuuukas: i'll try to test this as i haven't checked XR

com.unity.render-pipelines.universal/ShaderLibrary/Core.hlsl
jonuuukas: probably whitespace issues as my IDE messed that up pretty bad

com.unity.render-pipelines.universal/Runtime/ScriptableRendererFeature.cs
jonuuukas: this was temporary until i figured out a better way, already done

PR #6114, opened 2020-03-06
Recording of accumulation motion blur and path tracing

pmavridis:
### Purpose of this PR
This PR adds the ability to properly record scenes with path tracing and/or accumulation motion blur using the Unity recorder.

In particular :

- The behaviour of HDRP did not change when not recording a sequence of frames
- When recording a scene with path tracing, the recorder will now wait and output only converged frames (no additional action is required from the users).
- When recording, the user now has the option to activate motion blur.

This PR includes the required API changes from the HDRP side, but the recorder also needs some minimal changes.

---
### Code changes
- the accumulation was moved from the path tracing shader to a separate compute shader. The same shader is used for accumulation motion blur without path tracing.
- a subframe manager was introduced to manage sub-frames (and time-scale) when recording. It should not affect HDRP's behavior when recording is not active.
- New public API to activate accumulation and motion blur settings (shutter profile). 

---
### Testing status

I have tested by recording image sequences and movies in this test scene:
https://drive.google.com/file/d/1lmmlgr-DB6QgsbS8_fkdrRjKzuUwl041/view?usp=sharing

You will need a modified recorder for the testing:
https://drive.google.com/file/d/1UsnvxieExTfq3BoZoWhfR8t9vP-mmIXp/view?usp=sharing

And here is how shutter profiles affect the resulting motion blur on a blue sphere moving fast from left to right:

image

### Known limitation
- 360 recording / RenderToCubemap is not supported (because all cubemap faces will use the same accumulation buffer)


eturquin:
There are a few changes from the fog support to be merged, otherwise all good.

PR #6119, opened 2020-03-08
Make public and document HDCamera.SetupGlobalParams

karl-:
### Purpose of this PR

Expose the `HDCamera.SetupGlobalParams` method. This is necessary for custom passes that modify the camera frustum or viewport.

### Testing status

**Manual Tests**: Tested with the `com.unity.picking` package.

### Comments to reviewers

I could use some help documenting the `frameCount` parameter, as well as validation that my description of the function is correct and complete.


JulienIgnace-Unity:
We are not going to be able to merge this for several reasons:
- We are currently refactoring global constant management to use constant buffers so this PR would conflict with it.
- Until HDRP has really matured, we don't want to expose public APIs for this kind of internal management or we won't be able to change and fix things later on.
- In the meantime, as a workaround we propose that you use the SetGlobalXXX APIs with the names copied from the HDShaderIDs class. As we are aware that this is not robust, we won't change the names there until we have a proper public API to fill that role so this should be pretty safe for now.

PR #6135, opened 2020-03-09
Fix Pragma locales

Kink3d:
### Purpose of this PR
Fixes an issue where when `#pragma target X` is generated, the generated code is invalid when system locale uses commas instead of periods.

We decided to switch the code to being enum based instead of adding proper globalization as it is less error prone in general.

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers
QA:
- I was unable to test this as it requires a non-english system locale...


com.unity.shadergraph/CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a bug where any change to the PBR master node settings would lose connection to the normal slot. 
 - Fixed a bug where the user couldn't open up HDRP Master Node Shader Graphs without the Render Pipeline set to HDRP.
 - Fixed a bug where adding a HDRP Master Node to a Shader Graph would softlock the Shader Graph.
+- Fixed a bug where shaders fail to compile due to `#pragma target` generation when your system locale uses commas instead of periods.
```
marctem: No changelog necessary as we never shipped this, it's in master only.

marctem: i was wrong, we treat master as released so it's fine.

com.unity.shadergraph/Editor/Generation/Enumerations/ShaderModel.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+
+namespace UnityEditor.ShaderGraph
+{
+    [GenerationAPI]
+    internal enum ShaderModel
```
phi-lira: The downside of this is that you have to be dilligent with it to keep upgrading to shader models. F.ex, what about SM 6.0?
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/hlsl-shader-model-6-0-features-for-direct3d-12

This is supports with DXC for couple of platforms, @EvgeniiG to confirm.

phi-lira: Also this seems to be something core that users would need to create master nodes. Why can't it be public?

Kink3d: Yes, we have decided that, at least at this time, code gen safety is more important having to implement new SMs.

Also, we are not ready yet to make this stuff public. It will happen in due time.

EvgeniiG: SM 6.0 (and later versions) are indeed supported by DXC. We already use DXC for ray tracing, and may actually more broadly use it (remove FXC) in the near future. I suggest adding 6.0 (and later) as described here: https://github.com/Microsoft/DirectXShaderCompiler/wiki/Shader-Model

PR #6142, opened 2020-03-10
Remove tempId

pbbastian:
### Purpose of this PR
Removes `tempId` from `AbstractMaterialNode` and replaced all uses with `guid`. Historically, we've seen a number of bugs with `tempId` due to its design not being completely sound.

It was originally introduced for performance reasons (by me), but I have very big doubts about whether it actually does something. Removing it reduces the complexity of our code. So I think it's a good path forward to remove it, and if this turns out to be a real performance problem in the future, we can figure out a proper solution to that.

---
### Testing status

- Ran all tests in Shader Graph project
- Opened a graph
- Deleted a node
- Performed undo on the deletion
- Added a node by dragging an edge out from a port
- Performed undo on the addition

**Automated Tests**: Updated tests that broke after refactoring

---
### Comments to reviewers
Given that these changes are completely insignificant to users, should I update the change log?


pbbastian:
`HDRP on Win_DX11_playmode XR on version fast-2020.1` is failing due to the timeout issue we're seeing on many branches, so CI result should be good regardless :)

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
```diff
@@ -448,13 +443,14 @@ void UpdateShaders()
                 if (!node.hasPreview && !(node is SubGraphOutputNode || node is VfxMasterNode))
                     continue;
 
-                var renderData = GetRenderData(node.tempId);
+                Guid id = node.guid;
```
marctem: Why the copy?

LandonTownsendUnity:
Unfortunately there DOES seem to be a noticeable perf regression.

Before:
image
After: 
image

LandonTownsendUnity:
Master:
image
This branch:
image

Doesn't look like there are any performance regressions.

com.unity.shadergraph/Editor/Drawing/PreviewManager.cs
pbbastian: ahh this must have happened when I inlined `GetRenderData`

PR #6155, opened 2020-03-11
[8.x.x backport]

phi-lira:
Backports:
  - #5513 
 - #5671 
 - #5870 
 - #6088 
 - #6036


phi-lira:
InputNodes test in shader graph is failing but issue is fixed in master.

PR #6161, opened 2020-03-11
[Draft] Initial implementation of SSGI

anisunity:
This PR includes the structure and a first version of SSGI. Is is accessible through the volume component *Global Illumination* and is dependent on an hdrp asset setting and a frame setting.
Ping me for more details and discussions.


anisunity:
Need to add a test for this.

anisunity:
There is a solution to avoid the ghosting when changing from fullscreen to halfres. @sebastienlagarde we can discuss it. I can take a closer look at the other presets see if i can do a bit better

PR #6171, opened 2020-03-12
Fix for blank screen on Android 8+ when Depth Texture is enabled

tomzig16:
### Checklist for PR maker
- [ ] Have you added a backport label (if needed)?
- [x] Have you updated the changelog?
- [ ] Have you updated or added the documentation for your PR? 
- [ ] Have you added a graphic test for your PR (if needed)? 

---
### Purpose of this PR
This PR fixes an issue that causes blank screen on Android GLES when depth texture is selected without opaque texture.
[Case 1219325](https://issuetracker.unity3d.com/issues/scene-is-not-rendered-on-android-8-and-9-when-depth-texture-is-enabled-in-urp-asset)

---
### Testing status

**Manual Tests**: 
- Opened test project + Run graphic tests locally
- Built a player
- Other: 
  - Tested on several physical devices (Android 6, Android 7, Android 8 (affected), Android 9 (affected)
  - Tested with GLES 2, 3 and Vulkan on OnePlus 6T (Android 9)

**Automated Tests**: 
- Ran Android OpenGLES and Vulkan tests in Yamato

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fbugfix%252F1219325

---
### Comments to reviewers
In aa24dd7 commit Vulkan has failed due to missing test.


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -459,6 +459,18 @@ bool RequiresIntermediateColorTexture(ref CameraData cameraData)
             if (isOffscreenRender)
                 return requiresBlitForOffscreenCamera;
 
+#if UNITY_ANDROID
+            if (SystemInfo.graphicsDeviceType != GraphicsDeviceType.Vulkan)
+            {
+                if (!cameraData.requiresOpaqueTexture && cameraData.requiresDepthTexture)
```
phi-lira: This doesn't cover all cases. Looking at ForwardRenderer we have the following:
 https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs#L175-L176

`bool createDepthTexture = cameraData.requiresDepthTexture && !requiresDepthPrepass;
            createDepthTexture |= (cameraData.renderType == CameraRenderType.Base && !cameraData.resolveFinalTarget);` 

We should probably change the logic in forward renderer to do somethign like:

```
bool createDepthTexture = RequiresDepthTexture();
bool createColorTexture = RequiresColorTexture();
if (AndroidOpenGL)
   createColorTexture |= createDepthtTexture.
```

This way if someone changes the logic to create depth texture for some reason (we did it recently because of camera stacking), the logic still holds.

com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -175,6 +175,15 @@ public override void Setup(ScriptableRenderContext context, ref RenderingData re
             bool createDepthTexture = cameraData.requiresDepthTexture && !requiresDepthPrepass;
             createDepthTexture |= (cameraData.renderType == CameraRenderType.Base && !cameraData.resolveFinalTarget);
 
+#if UNITY_ANDROID
```
phi-lira: Do we need the same check for 2D renderer?

tomzig16: Checked and I think that no, we don't need that. I tried in different ways playing around with the settings to trigger 2 separate textures, but in the result this is solved by RP doing final blit or by rendering directly to backbuffer by already existing code.

PR #6174, opened 2020-03-12
Fixed an issue with emissive light meshes not being in the RAS

anisunity:
@eturquin 
A change in the way emissive meshes are attached to light sources broke the behavior for ray tracing. This makes the path tracing work again


sebastienlagarde:
@remi-chapelain if you can valdiated.

PR #6178, opened 2020-03-12
Implemented ray traced reflections for transparent objects.

anisunity:
As a follow up to ssr transparent, here is an implementation for ray traced reflections on transparent objects (for windows, water, etc). It is accessible through the same options than ssr transparent. Updates the tests scenes (DXR) to cover this.


com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaders.cs
```diff
@@ -69,7 +69,15 @@ static class HDSubShaders
                 { HDPasses.HDLit.MotionVectors },
                 { HDPasses.HDLit.DistortionVectors, new FieldCondition(HDFields.TransparentDistortion, true) },
                 { HDPasses.HDLit.TransparentBackface, new FieldCondition(HDFields.TransparentBackFace, true) },
-                { HDPasses.HDLit.TransparentDepthPrepass, new FieldCondition(HDFields.TransparentDepthPrePass, true) },
+                { HDPasses.HDLit.TransparentDepthPrepass, new FieldCondition[]{
+                                                            new FieldCondition(HDFields.TransparentDepthPrePass, true),
+                                                            new FieldCondition(HDFields.DisableSSRTransparent, true) }},
+                { HDPasses.HDLit.TransparentDepthPrepass, new FieldCondition[]{
```
sebastienlagarde: I don't understand how this code work here, they all drive the same TransparentDepthPrepass

anisunity: let's chat to clarify why i don't see an other way of doing this

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaders.cs
```diff
@@ -119,7 +127,15 @@ static class HDSubShaders
                 { HDPasses.Hair.DepthForwardOnly },
                 { HDPasses.Hair.MotionVectors },
                 { HDPasses.Hair.TransparentBackface, new FieldCondition(HDFields.TransparentBackFace, true) },
-                { HDPasses.Hair.TransparentDepthPrepass, new FieldCondition(HDFields.TransparentDepthPrePass, true) },
+                { HDPasses.HDLit.TransparentDepthPrepass, new FieldCondition[]{
```
anisunity: Mistake

anisunity:
There is an HDRP Asset transparent SSR option and a frame setting for it before being able to enable the effect

anisunity:
Ok, needs to be fixed for SG

sebastienlagarde:
Waiting for to adress Remi issue.

PR #6192, opened 2020-03-14
Override ShaderGUI per ShaderGraph [skip ci]

zadi0648:
**Summary:**

- Back end changes such that the entire generated shaders now only contains a single CustomEditor "ShaderGUI" line, as what we did prior (per SubShader) was not supported by Unity.
- Added a Toggle & TextField to each Master Node allowing the user to override the ShaderGUI per Shader Graph:
ShaderGUIOverridePresentation0
Toggle to open up the Text Field, which automatically gets populated with the SRP's default, and if an invalid ShaderGUI is given an warning badge is thrown.

While a lot of simple changes in the Master Nodes, a lot is needed changes for the future of ShaderGraphShaderGUI and will carry over into Stacks and the Internal Inspector.

---
**Implementation Notes & Notes to Reviewers:**

- I would say this copy & pasted code in each master node would belongs in a shared MasterNode class, but since Targets removed the old MasterNode class as a step towards the master node-less future I did what I did. That is to say all MasterNodes now extend `ICanChangeShaderGUI `and have the same code starting with `[SerializeField] private string m_ShaderGUIOverride` and ending with `set => m_OverrideEnabled = value`. This is except for the Sprite MasterNodes, which now have a shared super simple settings plane class.
- First tooltip in a Node Settings pane used here. This should be easily transferable into the Internal Inspector.
- ITargetImplementation now has a new method for getting the sub descriptors depending on the master node, was needed.
- `MasterNodeSettingsView` has the front end changes.

---
**Notes to QA:**

- _Known Issue:_ If a Master Node has a Red Error Badge (for example, no SRP assigned), and then you change the ShaderGUI to something incorrect (such that a White Warning Badge spawns), then fix the issue, the Red Error doesn't reappear. Ideally the White Warning Badge doesn't override the Error anyways. I am look into this, but so far it seems like a wider bug with the MessageManager.
- If you don't know what ShaderGUI is, this is useful: https://docs.unity3d.com/Manual/SL-CustomShaderGUI.html, recommended to at least get surface level familiar with it as more ShaderGUI will be coming in the future! Basically a single line is inserted inside the generated shader that determines which ShaderGUI is used when selecting a Material that uses the Shader Graph.
- We are only inserting a single line inside the generated shader, as such the complexity of the ShaderGUI used is irrelevant to these changes.
- If you do not want to write your own custom ShaderGUI, I can give you some really crappy ShaderGUIs I tested with during development. Again, nothing fancy needed for the testing of this specific feature.
- In order to use the default trunk ShaderGUI, type in BaseShaderGUI into the field.
- Remember that HDRP & URP use different ShaderGUIs for the same Master Node, which will come up in testing. For exmample, if you set Override ShaderGUI, does it first display the pipeline's default?
- You need to save the SG for the material to use the new ShaderGUI.

---
**Manually Tested:**

- Undo & Redo all new actions (i.e. changing of the new fields).
- Type in all sorts of junk into the new text field, confirm only allowed characters allowed (ones that could be in a C# class).
- Closing and reopening Unity and confirming that these two fields are in still in the same state.
- Bigger shaders.
- Confirmed that the master nodes and the ShaderGUI textfield (once Override ShaderGUI is checked) spawn with the right ShaderGUI AFTER switching between different SRPs.
- That every single Master Node still spawns with the correct ShaderGUI with OverrideEnabled is unchecked.
- Checking the generated code after switching active masters and undo & redo operations; confirming that the CustomEditor override is what it should be.
- Confirmed that there is only one CustomEditor line ever regardless of the # of implementations (current # of different SRPs to choose from).
- Confirmed that Master Nodes that aren't compatible with the current SRP don't spawn the CustomEditor line in the generated shader, in addition the ShaderGUI TextField doesn't have any text when Override ShaderGUI is checked.
- Confirmed that this works with custom user namespaces, such as Testing.Bull.doubleTime, ect.

---
**Technical Risk:** 3/4
**Halo Effect:** 1/4, affects nothing else, aside for a minor change in the generator.

---
**Notes to Documentation:**
https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6192/commits/f1cbef8b20ced2ea6dbbde903ace8d845333cc22

All Master Nodes with documentation have this:

> | **Override ShaderGUI**        | Enable overriding of the [https://docs.unity3d.com/ScriptReference/ShaderGUI.html](ShaderGUI) that this Shader Graph uses, spawning the "ShaderGUI" field below. |
> | **ShaderGUI**                 | The full name of the ShaderGUI class (including the class path) to use. A ShaderGUI class inside the UnityEditor namespace may omit UnityEditor. |

Meanwhile HD will have this documentation under Material Inspector:

>  These section does not apply if **Override ShaderGUI** is enabled.

---
**Yamato:**
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fshadergui-override-per-sg


com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
```diff
@@ -38,15 +38,15 @@ internal static void GenerateSubShaderTags(IMasterNode masterNode, SubShaderDesc
                     builder.AppendLine("// RenderPipeline: <None>");
 
                 // Render Type
-                string renderType = !string.IsNullOrEmpty(descriptor.renderTypeOverride) ? 
+                string renderType = !string.IsNullOrEmpty(descriptor.renderTypeOverride) ?
```
zadi0648: Can skip most of the changes in this file: there is just a new method at the end.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalMasterNode.cs
```diff
@@ -112,6 +112,20 @@ bool MaterialTypeUsesSlotMask(SlotMask mask)
...
+            set => m_ShaderGUIOverride = value;
+        }
+
+        [SerializeField] private bool m_OverrideEnabled;
+        public bool OverrideEnabled
+        {
+            get => m_OverrideEnabled;
+            set => m_OverrideEnabled = value;
+        }
+
```
zadi0648: The above is basically what got copied and pasted inside of every applicable master node

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSettingsView.cs
```diff
@@ -81,6 +81,7 @@ public DecalSettingsView(DecalMasterNode node)
             });
 
             Add(ps);
+            Add(GetShaderGUIOverridePropertySheet());
```
zadi0648: Similarly, all Settings Views also do this now: extend MasterNodeSettingsView & add the property sheet that contains the new Toggle & TextField

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDDecalTarget.cs
```diff
@@ -26,11 +27,19 @@ public void SetupTarget(ref TargetSetupContext context)
...
+        {
+            switch (masterNode)
             {
-                case DecalMasterNode decalMasterNode:
-                    context.SetupSubShader(HDSubShaders.Decal);
-                    break;
+                case DecalMasterNode _:
+                    return HDSubShaders.Decal;
+                default:
+                    return null;
```
zadi0648: As mentioned in the description, all targets now have this new method and thus have these sort of changes, and the code has be cleaned a little bit to avoid duplicate are-these-master-nodes-okay, to avoid inconsistencies/bugs

zadi0648: Another example is right below in HDMeshTarget

com.unity.shadergraph/Editor/Data/Interfaces/ICanChangeShaderGUI.cs
```diff
@@ -0,0 +1,19 @@
...
+            set;
+        }
+
+        bool OverrideEnabled
+        {
+            get;
+            set;
+        }
+    }
+}
```
zadi0648: All applicable Master Nodes implement this

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
```diff
@@ -0,0 +1,183 @@
+using System;
+using System.Text.RegularExpressions;
```
zadi0648: ❗️ This is the new work horse class that should be inspected more closely.

com.unity.shadergraph/Editor/Data/Util/GraphUtil.cs
```diff
@@ -348,5 +349,21 @@ public static void OpenFile(string path)
...
+                {
+                    var context = implementation.GetSubShaderDescriptorFromMasterNode(masterNode);
+                    if (context != null)
+                        return context.Value.customEditorOverride;
+                }
+            }
+
+            return null;
+        }
+
```
zadi0648: I wasn't sure Util class this method best fits in

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDSubShaders.cs
```diff
@@ -36,6 +37,7 @@ static class HDSubShaders
                 { HDPasses.PBR.MotionVectors, new FieldCondition(Fields.SurfaceOpaque, true) },
                 { HDPasses.PBR.Forward },
             },
+            customEditorOverride = "Rendering.HighDefinition.HDPBRLitGUI",
```
zadi0648: For the SubShaders classes (HD & URP), we forgot these two HD ShaderGUIs when Targets landed,
Also `@"CustomEditor` was removed from all of them (out of necessity as opposed to convenience, I don't what the other stuff to be in the string in the text field for example), Also removed UnityEditor from all of them as that is not needed by Unity.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSettingsView.cs
Nightmask3: This makes sense with how this works currently, in the future we'd be able to mark up the stacks directly and gather the inspector data needed from them.

I'm thinking that eventually you'd want to make the ShaderGUIOverride and OverrideEnabled fields be part of the graph data or the graph object? 

Because I dont think that the MasterNodes are directly equivalent to the stacks, as the stacks encapsulate individual shader features that a graph can opt into, the shaderGUI stuff needs to be associated with the graph itself. Or so I'm thinking.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
```diff
@@ -0,0 +1,183 @@
...
+            if (string.IsNullOrEmpty(customEditorName))
+            {
+                if (HasPreferredGUI())
+                {
+                    return false;
+                }
+                return true; // No default, so this is valid.
+            }
+
+            var unityEditorFullName = $"UnityEditor.{customEditorName}"; // For convenience: adding UnityEditor namespace is not needed in the shader
```
marctem: But what if they want a custom gui that's not in the UnityEditor namespace? Anyone can override this; I think we need to take exactly the namespace they give us an put UnityEditor back on all the MNs. Otherwise it's kind of magic that people will be confused about.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
```diff
@@ -0,0 +1,183 @@
...
+using UnityEngine;
+using UnityEditor.Graphing.Util;
+using UnityEditor.Rendering;
+using UnityEngine.UIElements;
+
+namespace UnityEditor.ShaderGraph.Drawing
+{
+    class MasterNodeSettingsView : VisualElement
+    {
+        private const string k_InvalidShaderGUI = " is not a valid ShaderGUI";
```
marctem: " is not a valid ShaderGUI class in this project."

com.unity.render-pipelines.high-definition/Editor/ShaderGraph/HDMeshTarget.cs
```diff
@@ -33,32 +26,33 @@ public void SetupTarget(ref TargetSetupContext context)
...
-                case FabricMasterNode fabricMasterNode:
-                    context.SetupSubShader(HDSubShaders.Fabric);
-                    break;
-                case HairMasterNode hairMasterNode:
-                    context.SetupSubShader(HDSubShaders.Hair);
-                    break;
-                case StackLitMasterNode stackLitMasterNode:
-                    context.SetupSubShader(HDSubShaders.StackLit);
-                    break;
+                case PBRMasterNode _:
```
Nightmask3: This is cool! Didn't know discards were a feature C# supported. TIL!

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
```diff
@@ -941,5 +941,25 @@ internal static string GetDefaultSharedTemplateDirectory()
         {
             return "Packages/com.unity.shadergraph/Editor/Generation/Templates";
         }
+
+        // Returns null if no 'CustomEditor "___"' line should be added, otherwise the name of the ShaderGUI class.
+        // Note that it's okay to add an "invalid" ShaderGUI (no class found) as Unity will simply take no action if that's the case, unless if its BaseShaderGUI.
+        public static string FinalCustomEditorString(ICanChangeShaderGUI canChangeShaderGUI)
```
marctem: I don't like how this function needs to use 2 different interfaces of the MN; the casting is icky. I'm not sure how to work around it, but give it a thought and we can discuss if you don't come up with anything.

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalMasterNode.cs
Nightmask3: So one thing I don't like is how these values have to be copy-pasted inside every MasterNode. Has a code smell...

Have you considered making these public fields in the ICanChangeShaderGUI and then using this trick to still let them have the SerializeField capability?

```
[field: SerializeField]
public float Speed { get; set; }
```

Nightmask3: Comes from this forum post: https://forum.unity.com/threads/serialize-c-properties-how-to-with-code.506027/
(third comment down)

Nightmask3: That way you only have to do it once, and then you basically opt into this system by inheriting from ICanChangeShaderGUI.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
zadi0648: Hmmm. So to be clear as the comment mentions this is exactly what trunk does, as such this method shouldn't be changed: it should return true if the ShaderGUI will work. However we could add the UnityEditor back to all the MN defaults, to avoid possible user confusion about all name spaces being okay. However I don't like that: in that the TextField is already really small and we can't make it bigger, as such the user wouldn't be able to see the important part of the default strings (the end). Instead I would rather just have this documented, which it should be anyways as it's how trunks works.

com.unity.shadergraph/Editor/Generation/Processors/Generator.cs
```diff
@@ -110,11 +112,24 @@ void BuildShader()
...
                 }
 
+                // Either grab the pipeline default for the active node or the user override
+                if (m_OutputNode is ICanChangeShaderGUI canChangeShaderGui)
+                {
+                    string customEditor = GenerationUtils.FinalCustomEditorString(canChangeShaderGui);
+
+                    if (customEditor != null)
+                    {
+                        m_Builder.AppendLine("CustomEditor \"" + customEditor + "\"");
```
Nightmask3: Nice, yeah given that all of them have CustomEditor appended to the front of them, makes sense to do it here instead

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
```diff
@@ -0,0 +1,183 @@
...
+                }
+            }
+
+            m_ShaderGUITextField.value = m_CanChangeShaderGUI.ShaderGUIOverride;
+
+            AddWarningIfNeeded();
+        }
+
+        // Add a warning to the node if the ShaderGUI is not found by Unity.
+        private void AddWarningIfNeeded()
```
Nightmask3: Good stuff 👍

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
zadi0648: Yeah that's clearer. Should we use something more along the lines of "found" instead of "valid"?

" is not a ShaderGUI found in this project."

Also should warnings have periods?

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalSettingsView.cs
zadi0648: So that's a really good question that I've brought up with Matt. Basically all of HDRP will have a single ShaderGUI and all of URP will have a single ShaderGUI regardless of the stack situation, as such I suppose this will be graph wide as opposed to... per master node. So yeah, inside of GraphData (as GraphObject is getting the kill from Peter)

com.unity.render-pipelines.high-definition/Editor/Material/Decal/ShaderGraph/DecalMasterNode.cs
zadi0648: Fyi to other reviews: Sai and I talked about this and decided it makes the serialized names ugly

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
marctem: That's a good point about being able to see the class name in the text box. Makes sense to not add "UnityEditor" back to the MN's if possible. If "UnityEditor" is not necessary in the editor, then we don't need to add it in the shader, right? My concern is that a gamedev will want to put their custom ShaderGUI class in the `MyGame.Tools.Editor` namespace, and tacking `UnityEditor` on the front makes it invalid.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
marctem: What about "No class named {0} which derives from ShaderGUI was found in this project."

I'd say yes on the period.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
zadi0648: So I tested custom weird ShaderGUIs that are in custom namespaces (for example all shorts of crappy ShaderGUIs I made such as Testing.Bull.doubleTime) and it works (I forgot to add that to the manual testing until now).

> If "UnityEditor" is not necessary in the editor, then we don't need to add it in the shader, right

That is correct and I we are not tacking it on so custom user name spaces are good. Also I just tried UnityEditor.Testing.Bull.doubleTime and that correctly throws and error.

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
zadi0648: That sounds good.

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
zadi0648: Will at least take a look at this.

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
```diff
@@ -55,12 +55,14 @@ To view these properties, click the **Cog** in the top right of the Master Node.
 | **Affects Ambient Occlusion** | Enable or disable the effect of the **Ambient Occlusion** property. |
 | **Affects Smoothness**        | Enable or disable the effect of the **Smoothness** property. |
 | **Affects Emission**          | Enable or disable the effect of the **Emission** property.   |
+| **Override ShaderGUI**        | Enable overriding of the [https://docs.unity3d.com/ScriptReference/ShaderGUI.html](ShaderGUI) that this Shader Graph uses, spawning the "ShaderGUI" field below. |
```
sharlenet: Link should be formatted as:
[ShaderGUI](https://docs.unity3d.com/ScriptReference/ShaderGUI.html)

JordanL8: Full content should be:

Allows you to override the [ShaderGUI](https://docs.unity3d.com/ScriptReference/ShaderGUI.html) this Shader Graph uses.

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
```diff
@@ -55,12 +55,14 @@ To view these properties, click the **Cog** in the top right of the Master Node.
...
+| **ShaderGUI**                 | The full name of the ShaderGUI class (including the class path) to use. A ShaderGUI class inside the UnityEditor namespace may omit UnityEditor. |
```
JordanL8: Is this property indented? If so, change the property name in the doc to `- **ShaderGUI**`.

The first sentence is okay, but I don't quite understand the second one. Could you give me a quick explanation?

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
```diff
@@ -55,12 +55,14 @@ To view these properties, click the **Cog** in the top right of the Master Node.
...
 
 <a name="MaterialProperties"></a>
 
 ### Material Properties
 
-These properties are in the **Exposed Properties** section of the Inspector, alongside the properties that you exposed in the Shader Graph's Blackboard.
+These properties are in the **Exposed Properties** section of the Inspector, alongside the properties that you exposed in the Shader Graph's Blackboard. These section does not apply if **Override ShaderGUI** is enabled.
```
JordanL8: What do you mean by `does not apply`? Is it displayed in the editor, but not used? Or not displayed at all?

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
sharlenet: I wrote:
`Lets you override the [ShaderGUI](https://docs.unity3d.com/ScriptReference/ShaderGUI.html) that this Shader Graph uses. If `true`, the **ShaderGUI** property appears, which lets you specify the ShaderGUI to use.`

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
sharlenet: Per discussion with Zack, I added indentation and removed the second sentence: "A ShaderGUI class inside the UnityEditor namespace may omit UnityEditor."

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
marctem: I was confused, this is fine. =)

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Decal.md
sharlenet: Suggested the following:
`If you set **Override ShaderGUI** to `true`, the Material Properties section does not appear, and the ShaderGUI you specified appears instead.`

com.unity.shadergraph/Editor/Drawing/Views/MasterNodeSettingsView.cs
```diff
@@ -0,0 +1,188 @@
...
+        {
+            if (m_CanChangeShaderGUI.OverrideEnabled && m_CanChangeShaderGUI.ShaderGUIOverride != null && !ValidCustomEditorType(m_CanChangeShaderGUI.ShaderGUIOverride))
+            {
+                m_MasterNode.owner.messageManager?.ClearNodesFromProvider(this, m_MasterNode.ToEnumerable());
+                m_MasterNode.owner.messageManager?.AddOrAppendError(this, m_MasterNode.tempId,
```
rainsing: This line no longer compiles after you've merged the latest master. It's due to this change: https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/14092cd0671a79e002ec3b7257c6ec7fd39c3f08

com.unity.shadergraph/Editor/Generation/Processors/GenerationUtils.cs
marctem: Discussed with Zack and we agreed to land this for now since it will be removed with the Stacks refactor. We can do it better then.

PR #6194, opened 2020-03-16
[Bugfix]1069375 metallic gamma tag

Verasl:
### Purpose of this PR
Universal RenderPipeline previously used the same Gamma curve on metallic that builtin did, this is incorrect when it comes to what Metallic is and thus should be linear, also this now makes it fit in line with shadergraph shaders.

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


phi-lira:
We should add an upgrade guide for this IMO.

phi-lira:
Rerunning failed test. Seems instability. Should we add upgrade guide on this PR?

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -133,6 +133,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed an issue where XR multipass throws doesn't support camera stack spamming.
 - Fixed an issue with shadows not appearing on terrains when no cascades were selected [case 1226530](https://issuetracker.unity3d.com/issues/urp-no-shadows-on-terrain-when-cascades-is-set-to-no-cascades-in-render-pipeline-asset-settings)
 - Fixed a shader issue that caused the Color in Sprite Shape to work improperly.
+- Fixed an issue with URP switching such that every available URP makes a total set of supported features such that all URPs are taken into consideration. [case 1157420](https://issuetracker.unity3d.com/issues/lwrp-srp-switching-doesnt-work-even-with-manually-adding-shadervariants-per-scene)
+- Metallic slider on the Lit shader is now linear meaning correct values are used for PBR.
```
olko-unity-work: To make is easier for users to see the changes that might affect the look of their game, I suggest to add the following line in the Changed section, preferably on top (to make the change more visible):
 - The Metallic property value of a Material is now linear, which is the correct behavior for the PBR approach. In the previous URP package version, those values were interpreted as gamma values. This change might affect the look of Materials that use this property after upgrading from the previous package version.

PR #6213, opened 2020-03-16
Fix z-fighting in sceneview when scene lighting is off

pmavridis:
### Purpose of this PR
This is a bugfix for https://fogbugz.unity3d.com/f/cases/1203927/#BugEvent.1072099375

### What was causing the problem
The problem was due to z-fighting, because transparent co-planar UI elements were rendered with  depth writes set to ON, even when the material was setting the corresponding state to OFF.

This was happening because when we turn off the scene lighting, HDRP is  using RenderDebugViewMaterial to render the scene and this call was overriding the depth state for both opaque and transparent objects. After this PR, it only overrides opaque objects.

### Testing status
I have opened the HDRP template scene, changed the "stud" material to transparent and then switched on the debug view and selected to show the albedo. I got the expected result (the albedo was correctly displayed on the transparent surface).


sebastienlagarde:
@Unity-Technologies/gfx-qa-hdrp Please, check that debug view mode for transparent Material is still working correctly after this fix.

PR #6214, opened 2020-03-16
Bitangent computation per fragment

phi-lira:
### Purpose of this PR
 This PR changes bitangent computation for all URP stock shaders to be `per-fragment` and consistent between HDRP and ShaderGraph. This way developers will have a unified normal map generation workflows for all the new render pipelines when baking normal maps.

Before:
`bi-tangent when computing normal maps per-pixel`
image (1)

Now:
`bi-tangent when computing normal maps per-pixel`
image (2)

---
### Testing status

**Manual Tests**: Tested myself.

- [ ] Built a player
- Other: 

**Automated Tests**: Test scene 128_TBN added. However there's a refactor for reference images in progress so I'll wait until the refactor lands to upload reference images/enable test scene.

---
### Comments to reviewers
 We need to evaluate performance impact on low end platforms.


phi-lira:
@Unity-Technologies/mobile-qa we need to validate this doesn't regress performance in mobile.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -86,48 +86,73 @@ float4 TransformWViewToHClip(float3 positionVS)
     return mul(GetViewToHClipMatrix(), float4(positionVS, 1.0));
 }
 
-real3 TransformObjectToWorldDir(real3 dirOS)
+// Normalize to support uniform scaling
```
EvgeniiG: You mean non-uniform?

EvgeniiG: Or you mean non-uniform scaling is not supported?

EvgeniiG: Assuming dir is a polar (rather than axial) vector, it can be transformed just like a regular point (without translation, of course).

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -86,48 +86,73 @@ float4 TransformWViewToHClip(float3 positionVS)
...
+    return dirHCS;
 }
 
 // Transforms normal from object to world space
-float3 TransformObjectToWorldNormal(float3 normalOS)
+float3 TransformObjectToWorldNormal(float3 normalOS, bool doNormalize = true)
 {
 #ifdef UNITY_ASSUME_UNIFORM_SCALING
-    return TransformObjectToWorldDir(normalOS);
+    return TransformObjectToWorldDir(normalOS, doNormalize);
```
EvgeniiG: OK, I see, the comment comes from here. We should probably elaborate on the comment above.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -146,41 +171,48 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
     return mul(dirTS, tangentToWorld);
 }
 
+// This function does the exact inverse of TransformTangentToWorld() and is
+// also decribed within comments in mikktspace.h and it follows implicitly
+// from the scalar triple product (google it).
```
EvgeniiG: Best add a URL.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -146,41 +171,48 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
...
+    float3 col0 = cross(row1, row2);
+    float3 col1 = cross(row2, row0);
+    float3 col2 = cross(row0, row1);
+    
+    float determinant = dot(row0, col0);
+    float sgn = determinant<0.0 ? (-1.0) : 1.0;
+    
+    // inverse transposed but scaled by determinant
+    // Will remove transpose part by using matrix as the first arg in the mul() below
+    // this makes it the exact inverse of what TransformTangentToWorld() does.
```
EvgeniiG: Up to the scale of the determinant. :-)

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -146,41 +171,48 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
     return mul(dirTS, tangentToWorld);
 }
 
+// This function does the exact inverse of TransformTangentToWorld() and is
+// also decribed within comments in mikktspace.h and it follows implicitly
+// from the scalar triple product (google it).
 real3 TransformWorldToTangent(real3 dirWS, real3x3 tangentToWorld)
```
EvgeniiG: This function is for normals (axial vectors), I suppose. "dir" is rather ambiguous.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -146,41 +171,48 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
...
+    float3 col1 = cross(row2, row0);
+    float3 col2 = cross(row0, row1);
+    
+    float determinant = dot(row0, col0);
+    float sgn = determinant<0.0 ? (-1.0) : 1.0;
+    
+    // inverse transposed but scaled by determinant
+    // Will remove transpose part by using matrix as the first arg in the mul() below
+    // this makes it the exact inverse of what TransformTangentToWorld() does.
+    real3x3 matTBN_I_T = real3x3(col0, col1, col2);
```
EvgeniiG: Is our math even correct? This is what I get:

N_world = transpose(inverse(TangentToWorld)) * N_tangent
N_world = N_tangent * inverse(TangentToWorld)
N_world * TangentToWorld = N_tangent * (inverse(TangentToWorld) * TangentToWorld)
transpose(TangentToWorld) * N_world = N_tangent

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
```diff
@@ -146,41 +171,48 @@ real3 TransformTangentToWorld(real3 dirTS, real3x3 tangentToWorld)
...
+    return TransformWorldToObjectNormal(normalWS);
 }
 
 real3 TransformObjectToTangent(real3 dirOS, real3x3 tangentToWorld)
 {
     // Note matrix is in row major convention with left multiplication as it is build on the fly
-	float3 normalWS = TransformObjectToWorldNormal(dirOS);
-	return TransformWorldToTangent(normalWS, tangentToWorld);
+
+    // don't normalize, as normalWS will be normalized after TransformWorldToTangent
```
EvgeniiG: I wonder if we should call some of our functions TransformAndNormalize()... The way we do this with comments here is rather confusing and error-prone.

com.unity.render-pipelines.universal/Shaders/LitForwardPass.hlsl
```diff
@@ -50,17 +47,16 @@ void InitializeInputData(Varyings input, half3 normalTS, out InputData inputData
...
 
-#ifdef _NORMALMAP
-    half3 viewDirWS = half3(input.normalWS.w, input.tangentWS.w, input.bitangentWS.w);
-    inputData.normalWS = TransformTangentToWorld(normalTS,
-        half3x3(input.tangentWS.xyz, input.bitangentWS.xyz, input.normalWS.xyz));
+    half3 viewDirWS = SafeNormalize(input.viewDirWS);
+#ifdef _NORMALMAP 
+    float sgn = input.tangentWS.w;      // should be either +1 or -1
+    float3 bitangent = sgn * cross(input.normalWS.xyz, input.tangentWS.xyz);
+    inputData.normalWS = TransformTangentToWorld(normalTS, half3x3(input.tangentWS.xyz, bitangent.xyz, input.normalWS.xyz));
```
EvgeniiG: Just a suggestion, but this piece of code is being repeated several times. Perhaps add a function called BuildWorldSpaceTBN() or something like that.

com.unity.render-pipelines.universal/Shaders/LitForwardPass.hlsl
```diff
@@ -90,20 +86,23 @@ Varyings LitPassVertex(Attributes input)
     UNITY_INITIALIZE_VERTEX_OUTPUT_STEREO(output);
 
     VertexPositionInputs vertexInput = GetVertexPositionInputs(input.positionOS.xyz);
+    
+    // normalWS and tangentWS already normalize.
+    // this is required to avoid skewing the direction during interpolation
+    // also required for per-vertex lighting and SH evaluation
     VertexNormalInputs normalInput = GetVertexNormalInputs(input.normalOS, input.tangentOS);
-    half3 viewDirWS = GetCameraPositionWS() - vertexInput.positionWS;
+    float3 viewDirWS = GetCameraPositionWS() - vertexInput.positionWS;
```
EvgeniiG: I think you can still use half3 if you normalize() before quantization.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
mmikk: Keep in mind by "mikkts/traditional real-time normal mapping" definition TangentToWorld() is understood to be the transformation that transforms **a normal** (not a direction) from tangent space to world space.
So the inverse of this transformation for a normal is then simply just Inverse(TangentToWorld()) of that. There's no additional Transpose() involved here. The only reason we end up having to remove an unintended Transpose() in this function is a result of how the real3x3() syntax happens to work to initialize the matrix.

The visual also confirms it since this test case is designed to be extremely vulnerable to a mismatch. But beyond that it's been in use for almost a decade now.

mmikk: When I say "this case' I'm referring to some tests @alindmanUnity and myself did with this asset where we use TransformWorldToTangent() in various scenarios such as dropping off an object space normal in tangent space to the master node and a few additional scenarios.

mmikk: Also note TransformWorldToTangent() has not actually changed functionally in this PR. It's just changes to comments and whitespace. The graphics tests in the original PR where it was changed confirmed it works.

EvgeniiG: So the tangent-to-world transformation you pass in is actually already the inverse-transpose? That needs to be indicated in the comments.

EvgeniiG: You have a function called TransformWorldToTangent that takes a direction (axial vector? polar vector? it does not say) and a transformation (with a built-in inverse-transpose you need to remove? it does not say).

mmikk: Technically it is not but it is an implied assumption/approximation to pretend that it is :) Since everyone has transformed as T*m.x + B*m.y + N*m.z since 1997.
Regardless the main point is whatever TangentToWorld() does the transformation WorldToTangent() must do the exact inverse (not including the uniform scale since we normalize the normal at the end).
Whether we think of it as a normal or a direction it is the only way we get back the same normal we started out with if we transform first by one transform followed by the other (not including positive magnitude).

mmikk: Ultimately that is really all that matters here. Ensuring we get: 
V = (ExactInv(M) * M) * V
          = ExactInv(M) * (M * V)

Whether or not we think of V as a normal or a vector here is not so much the point really. All that matters is that the two transformations cancel each other out.

mmikk: 1. Bake Step does:    m = ExactInv(M) * Nhp
2. Real-time App using the normal map reconstructs the original using:   Nhp = M * m

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
phi-lira: I'm confused about the comment as well. I just moved places.

com.unity.render-pipelines.universal/Shaders/LitForwardPass.hlsl
phi-lira: Good suggestion. Will do in a follow up PR.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
EvgeniiG: For normals (axial vectors), you don't need to use the inverse-transpose (as it's done a couple of lines below) and can use the transformation matrix directly, only if the transformation matrix does NOT have non-uniforml scaling. For regular (polar) vectors, you can always use the matrix directly, as is.

com.unity.render-pipelines.core/ShaderLibrary/SpaceTransforms.hlsl
EvgeniiG: Ah, right, you assume mikkt-space normals that are transformed into the world space s.t.

N_world = TangentToWorld * **N_mikkt**
inverse(TangentToWorld) * N_world = **N_mikkt**

In contrast to the regular tangent-space normals **N_tangent**

N_world = transpose(inverse(TangentToWorld)) * **N_tangent**
N_world = **N_tangent** * inverse(TangentToWorld)
N_world * TangentToWorld = **N_tangent** * (inverse(TangentToWorld) * TangentToWorld)
transpose(TangentToWorld) * N_world = **N_tangent**

I forgot, my bad.

So that does indeed imply that this function takes the regular TBN matrix and can work with both axial and polar vectors (normals and non-normals), with the caveat that _polar vectors end up in the tangent space_ and _axial vectors end up in the mikkt space_ (which is a **different space**).

Not sure what to do with this function, but this should at least be documented.

EvgeniiG: @sebastienlagarde

EvgeniiG: Note that I don't think that's a serious limitation in practice assuming we _never_ do any computations in the "tangent" space that involve both polar and axial vectors.

sebastienlagarde:
Hey,

I have concern with the change in UnpackNormalAG as people assume when they use this function that it return a normalize normal. See for example the Terrain lit code in hdrp.

float3 SampleNormalGrad(TEXTURE2D_PARAM(textureName, samplerName), float2 uv, float2 dxuv, float2 dyuv, float scale)
{
    float4 nrm = SAMPLE_TEXTURE2D_GRAD(textureName, samplerName, uv, dxuv, dyuv);
#ifdef SURFACE_GRADIENT
    #ifdef UNITY_NO_DXT5nm
        return float3(UnpackDerivativeNormalRGB(nrm, scale), 0);
    #else
        return float3(UnpackDerivativeNormalRGorAG(nrm, scale), 0);
    #endif

ifdef _NORMALMAP
    #define SampleNormal(i) SampleNormalGrad(_Normal##i, sampler_Splat0, splatuv, splatdxuv, splatdyuv, _NormalScale##i)
#else
    #define SampleNormal(i) float3(0, 0, 0)
#endif
#(...)
#define SampleResults(i, mask)                                                                                  \
    UNITY_BRANCH if (mask > 0)                                                                                  \
    {                                                                                                           \
        float2 splatuv = splatBaseUV * _Splat##i##_ST.xy + _Splat##i##_ST.zw;                                   \
        float2 splatdxuv = dxuv * _Splat##i##_ST.x;                                                             \
        float2 splatdyuv = dyuv * _Splat##i##_ST.y;                                                             \
        albedo[i] = SAMPLE_TEXTURE2D_GRAD(_Splat##i, sampler_Splat0, splatuv, splatdxuv, splatdyuv);            \
        albedo[i].rgb *= _DiffuseRemapScale##i.xyz;                                                             \
        normal[i] = SampleNormal(i);                                                                            \
        masks[i] = SampleMasks(i, mask);                                                                        \
    }                             

surfaceData.normalData += normal[i].rgb * weights[i]; // no need to normalize

(cc @zeroyao ), so we will need to inspect and clean the code before we merge it. + Don't know what will happen on users side. I am not confident.

Regarding the comment of Evgenii on TransformTangentToWorld, could we update the comment with his remark? Looks like we should assume that we always use Mikktspace if I get it. But good to explain what we do. However, 99.99% will not understand the comment. This is not really a big deal from my point of view as the function is mainly use in two cases: POM and object space normal. So not widely spread usage.

To not block the PR I will suggest to put the scaling change part in another PR so we can discuss about it, and same thing for the comment. This PR for URP contain an important fix and we need to move forward :)

mmikk:
Hey @seblagarde regarding the change to UnpackNormalAG().

Notice that the way it is in master is wrong:
normal.xy *= scale;	
normal.z = sqrt(1.0 - saturate(dot(normal.xy, normal.xy)));

this will cause bad results when scale is increased beyond 1.0 due to sqrt(1-sat(x)). Clearly correct results require completing the unpack before applying the scale. Also notice that the way I changed it actually matches how it works when UNITY_NO_DXT5nm is enabled in which case we use three channel normal maps and in this case there is also no normalization of the unpacked normal. So with my change there's actually parity between when UNITY_NO_DXT5nm  is set and when it is not set (and parity with derivatives). 
That being said I don't know if we are actually using the scale feature with UnpackNormal() anywhere. It appears to be disabled on mobile in URP.
Whether or not we need renormalization I am not so sure since we renormalize anyway after TBN transformation. For derivatives and surface gradients we definitely do not nor are we impacted by this. It doesn't matter what the magnitude of a tspace normal is since the derivative is (-m.x/m.z, -m.y/m.z).

I don't think there's anything to add regarding TransformTangentToWorld(). It is just the standard TBN transformation and as the comment already says and the name indicates TransformWorldToTangent() does the exact inverse transformation of what TransformTangentToWorld() does (but with a normalize). There isn't a specific dependency on mikkt. The point is the dynamic between the two is designed to support baked normal maps regardless of which standard is being used for tangent space by making one cancel out the other.

mmikk:
Another way of saying it there are only three assumptions being made in the implementation of TransformWorldToTangent():

1. We assume both TransformWorldToTangent() and TransformTangentToWorld() receive the same 3x3 matrix tangentToWorld.
2. We assume that TransformTangentToWorld() transforms as mul(dirTS, tangentToWorld);
3. We assume it's ok to renormalize at the end since tangentToWorld is generally not orthogonal.

zeroyao:
> surfaceData.normalData += normal[i].rgb * weights[i]; // no need to normalize

Hey Seb TerrainLit in reality always assumes it works with derivative data instead of real normal vectors. It doesn't assume it gets normalized normal vectors from the unpack function.

EvgeniiG:
Simple solution. We keep this variant of the function normalized (we add another normalize). We add another variant of the function with a different name that explicitly claims that the resulting vector is not normalized. We replace certain call sides with the new function where we know we do not have to normalize.

Generally speaking, normals should almost always be normalized, that's a conventional expectation. If we keep them normalized somewhere for optimization purposes, we should clearly label them a such.

mmikk:
Though I agree with what you're saying Evgenii I think it would entail a refactor which imo is beyond the scope of this PR. It's not just UnpackNormal() where we do not renormalize today (when UNITY_NO_DXT5nm  is enabled). The same is true for shaders that use TransformTangentToWorld(). This doesn't renormalize either since URP allows very low-end platforms to skip it.
I imagine in both cases the code was structured this way to avoid the risk of injecting redundant normalizes since they aren't cheap.

EvgeniiG:
Short term solution is to add a normalize to your existing code (along with a TODO) to fix the bug, and optimize in a subsequent PR.

sebastienlagarde:
It is true that the scale parameter isn't used in ShaderGraph at all and we use derivative in lit.shder and other, so I will approve this PR, but agree with Evgenii that we should make this more clear for users.

PR #6219, opened 2020-03-17
Deprecate Utilities namespace

fredericv-unity3d:
### Purpose of this PR
* Made Utilities namespace obsolete.
* Copied its content to UnityEngine.Rendering namespace.

It partially fixes [case 1204677](https://fogbugz.unity3d.com/f/cases/1204677/):
* We keep the obsolete API to avoid to break compatibility.
* We will need to remove the breaking API on next major update (breaking API will be allowed).

---
### Testing status

**Manual Tests**: 
- Deleted the obsolete code and checked that the HDRP is still executed.

**Automated Tests**:
- Will execute on HDRP/staging

**Yamato**: 

---
### Comments to reviewers
This is open for discussion: how to handle removal of an API?

  * Here are the steps I followed:
    1. I tagged as obsolete the API so the compiler will provide hints for our users that this API will be removed. (But the question is which major version will remove this api?)
    1. I added a comment _// TODO: Remove when breaking changes (introduced in 8.x.x)_
         So we can track where are the api that need to be removed, and when they were introduced.

So when upgrading a major version we should (for instance):
* Remove all obsolete api since n versions. (like in package 10.0.0, remove all obsolete 8.x.x apis)

#### Note 2
HDRP/Staging targets 9;x cycle, so we can just remove the API.
Although, this PR can be backported to 8.x cycle


sebastienlagarde:
we don't need QA pass for this one. Merging

PR #6227, opened 2020-03-17
Clearer naming of tile and cluster categories for environment lights

FrancescoC-unity:
Now looks like this

image

Will perform back port when QA validated.


sebastienlagarde:
guess we need to update documentation as well of debug menu?

FrancescoC-unity:
> guess we need to update documentation as well of debug menu?

They were never mentioned explicitly (the doc only refers to Light type)

PR #6228, opened 2020-03-17
GC fix (#1227490)

Chman:
### Purpose of this PR
Fixed GC pressure in `VolumeParameter<T>.GetHashCode()` in CoreRP. `ReferenceEquals()` will box value types so it's better to use the default `EqualityComparer` as it's properly typed.

This fixes fogbugz#1227490 and affects both URP and HDRP (in the sense that it will reduce GC pressure for both of these pipelines).

Not sure which QA to assign to this PR so I'll let the reviewers choose.

---
### Testing status

Issue only reproduces in standalone player. Editor is fine.

How to reproduce:
    - Open URP GraphicsTestScene.
    - Build Project containing scene 104 BokehDepthOfField
    - Build with Deep Profiling enabled.
    - Check profiler, see that VolumeParameter.GetHashCode is not allocating memory anymore.


com.unity.render-pipelines.core/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix LookDev used Cubemap when asset is deleted
 - Fixed the definition of `rcp()` for GLES2.
 - Fixed copy/pasting of Volume Components when loading a new scene
+- Fixed GC pressure in `VolumeParameter<T>.GetHashCode()`
```
phi-lira:
```suggestion
- Fixed GC pressure caused by `VolumeParameter<T>.GetHashCode()`
```

PR #6231, opened 2020-03-17
Fix cubemap inspector so that thumbnail for cubemap work

FrancescoC-unity:
This was quite a long investigation, but thankfully the fix is simple. Apparently, in order to render a thumbnail custom inspectors need to also implement the function RenderStaticPreview so that a thumbnail can be generated at import time. 

The new thumbnail look like this: 
image

I tested it by: 
 - Importing new texture asset, changed it to cubemap and verified that the thumbnail appeared. 
 - Reimporting an old asset and verified that the thumbnail appeared. 

I will perform the backport once the PR will be QA validated.


sebastienlagarde:
I have backported this PR on 7.x and 8.x

PR #6241, opened 2020-03-17
Hdrp xr public layout api cluster display

etienne-p:
### Purpose of this PR

Add Cluster Display support to HDRP,
we introduce conversions from device to cluster space,
please check out https://docs.google.com/document/d/1tN5LBbXIMb8UXSLSMuZ6ImWcuepewO_WV9qrOI5i_F0/edit?usp=sharing for details about the changes

### Testing status

Tests are still to be written, as part of the Cluster Display Graphics package, we will need some unit tests but also some graphics tests, i'll look into how packages may support graphics tests.

### Note to reviewers

No final decision as been made at this point regarding how this PR may be integrated to HDRP itself, in collaboration with the HDRP team. It is still worthwhile to go through the changes in order to be in the best place possible when the integration conversation starts.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRLayout.cs
```diff
@@ -2,21 +2,21 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
-    internal struct XRLayout
+    public struct XRLayout
```
sebastienlagarde: any public interface must have documentation. cc @fabien-unity  (and be aware that anything public, will be almost forever, so be sure about what you are making public! :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRPass.cs
```diff
@@ -13,7 +13,7 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
-    internal struct XRPassCreateInfo
+    public struct XRPassCreateInfo
```
sebastienlagarde: same as above, think about the documentation

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRPass.cs
```diff
@@ -35,14 +36,16 @@ internal struct XRView
         internal readonly Matrix4x4 projMatrix;
         internal readonly Matrix4x4 viewMatrix;
         internal readonly Rect viewport;
+        internal readonly Matrix4x4 clusterDisplayParams;
         internal readonly Mesh occlusionMesh;
         internal readonly int textureArraySlice;
 
-        internal XRView(Matrix4x4 proj, Matrix4x4 view, Rect vp, int dstSlice)
+        internal XRView(Matrix4x4 proj, Matrix4x4 view, Rect vp, Matrix4x4 clusterDisplayParams_, int dstSlice)
```
sebastienlagarde: avoid weird name like "clusterDisplayParams_" follow C# convention (i.e previate member should be m_

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRPass.cs
```diff
@@ -99,7 +104,7 @@ class XRPass
         Material occlusionMeshMaterial = null;
 
         // Ability to override mirror view behavior for each pass
-        internal delegate void CustomMirrorView(XRPass pass, CommandBuffer cmd, RenderTexture rt, Rect viewport);
+        public delegate void CustomMirrorView(XRPass pass, CommandBuffer cmd, RenderTexture rt, Rect viewport);
```
sebastienlagarde: documentation, do we really want to make this public? have in mind that you can change it anymore after this.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace UnityEngine.Rendering.HighDefinition
 {
-    internal partial class XRSystem
+    public partial class XRSystem
```
sebastienlagarde: same as above

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
```diff
@@ -109,10 +109,10 @@ internal List<(Camera, XRPass)> SetupFrame(Camera[] cameras, bool singlePassAllo
                 ReleaseFrame();
             }
 
-            if ((singlePassTestModeActive || automatedTestRunning) && testModeEnabled)
-                SetCustomLayout(LayoutSinglePassTestMode);
-            else
-                SetCustomLayout(null);
+            //if ((singlePassTestModeActive || automatedTestRunning) && testModeEnabled)
```
sebastienlagarde: if commented, please delete it :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRPass.cs
etienne-p: fixed in 3c638bf, thx

fabien-unity:
The XR team is not ready to move this code to public yet. We first need to solidify the API and move it to SRP Core (for URP).
As such, we need to rework the implementation to not be dependant on XRLayout.
I've talked with @etienne-p and I can recommend a few pointers:
- modify the camera in user script
- add a HDRP custom pass to handle the compositing
- push the macro changes to HDRP for UV offset in postfx

PR #6245, opened 2020-03-17
Bugfix: uncached reflection probe cameras were reseting the debug mode

pmavridis:
### Purpose of this PR
https://fogbugz.unity3d.com/f/cases/1224601/#BugEvent.1072099765

### What was the problem
- When a camera is created, we register the frame settings history if it is not a reflection camera.
- However, when a new reflection probe camera was created, its type was set by default to "Game" camera, thus registering itself in the debug frame settings history. 
- The behavior was also affected by the caching of the cameras. To trigger the bug, you have to look away from the reflection probe, so it is not in the frustum for 2 frames (this is the threshold when we delete old unused cameras), and then look back again.


fredericv-unity3d:
> Looks good but I think there was some issue at some point with CameraType and reflection probe where the setup needed to go through CameraType.Game first or something like that. @fredericv-unity3d might remember.

There were a lot of issues with _CameraType_, but those seemed to be fixed on Unity side. But @pmavridis change make sense, so I approve ^^.

PR #6251, opened 2020-03-18
Rejecting history for ray traced reflections based on a threshold evaluated on the neighborhood of the sampled history.

anisunity:
DXR tests are passing locally. This is a fix for the bug: https://fogbugz.unity3d.com/f/cases/1217537/


anisunity:
@sebastienlagarde i am waiting for @remi-chapelain 's validation before backporting it

PR #6253, opened 2020-03-18
Fix: Nothing rendering on GLES2 when PostProcessing is turned on

ellioman:
### Purpose of this PR
We noticed that nothing was rendering on GLES2 when Post-Processing was enabled. The problem is that we were missing a check for one of the PP variables. PP is suppose to be disabled on GLES2.

---

### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -330,6 +330,7 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
             List<Camera> cameraStack = (supportsCameraStacking) ? baseCameraAdditionalData?.cameraStack : null;
 
             bool anyPostProcessingEnabled = baseCameraAdditionalData != null && baseCameraAdditionalData.renderPostProcessing;
+            anyPostProcessingEnabled &= SystemInfo.graphicsDeviceType != GraphicsDeviceType.OpenGLES2;
```
phi-lira: This fix is fine for now. for future we should probably do this while building `renderingData.postProcessingEnabled` as it's more generic.

ellioman: Yes, most definitely!
Just needed a quick fix for the release.

PR #6257, opened 2020-03-18
[9.x.x ForwardPort] Bugfix: Nothing rendering on GLES2 when PostProcessing is turned on #6256

ellioman:
### Purpose of this PR
ForwardPort of #6253 & #6256 

---
### Testing status

- [ ] Built a player
- Other: 


com.unity.render-pipelines.universal/Runtime/UniversalRenderPipeline.cs
```diff
@@ -260,6 +260,7 @@ static void RenderCameraStack(ScriptableRenderContext context, Camera baseCamera
             List<Camera> cameraStack = (supportsCameraStacking) ? baseCameraAdditionalData?.cameraStack : null;
 
             bool anyPostProcessingEnabled = baseCameraAdditionalData != null && baseCameraAdditionalData.renderPostProcessing;
+            anyPostProcessingEnabled &= SystemInfo.graphicsDeviceType != GraphicsDeviceType.OpenGLES2;
```
phi-lira: We should probabably do this while computeing RenderingData.anyPostProcessingEnabled.

ellioman: Yeah, you're right. 
I shall look into that when I get the chance. I'm going to merge this so the same code is in all branches and then make new PRs with the update.

PR #6259, opened 2020-03-18
Add XR setting to control camera jitter for temporal effects

fabien-unity:
### Purpose of this PR
- expose setting on the HDRP asset to control camera jitter in XR
- tweak XR autotests so that the first view (of the 2 instanced views) has a different viewpoint and projection to catch more issues
- disabled AOV test for XR

---
### Testing status

**Manual Tests**: run HDRP tests in XR mode


sebastienlagarde:
Hey, can you run test on yamato, thanks

PR #6265, opened 2020-03-18
Fix issue on thumbnail generation so that no error on new project and thumbnail are generated at first import

FrancescoC-unity:
This PR fixes two issues: 

 - An error appearing when opening a new project with empty library. 
 - Thumbnail not being rendered if the asset is imported at load time. 

The issue was related to the fact that the rendering of thumbnails happen before HDRP is fully initialized when loading a project. The fix is using a shader that comes bundled with the editor and then forcing legacy pipeline for thumbnail generation. 

For QA: 

 - To test this start from an HDRP project with at least one cubemap.
 - Delete the library.
 - Reopen the project.

Verify that: 
  - The thumbnail for the cubemap is generated
 - No error related to Cubemap inspector are in the console. 

I tried to do as above, but worth giving another try on your end. 

**Note** Have not added a changelog

I am going to do backport right after QA validation.


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/HDUtils.cs
```diff
@@ -449,6 +449,42 @@ internal static string GetHDRenderPipelinePath()
...
+            assetWasFromQuality = false;
+            if (graphicSettingAsset != null)
+            {
+                // Check if the currently used pipeline is the one from graphics settings or is from quality settings
+                if (GraphicsSettings.currentRenderPipeline == graphicSettingAsset)
+                {
+                    GraphicsSettings.renderPipelineAsset = null;
+                    return graphicSettingAsset;
+                }
+
```
FrancescoC-unity: Oops gonna remove this line

iM0ve:
Currently, there is an issue when baking reflection probes. Developer reproduced it and working on a fix.

FrancescoC-unity:
Ok gave me a massive headache but I found the issue... so apparently the currentPipeline in the RenderPipeline manager is set only when camera.Render() is called.

The baked reflection probe import cycle imports several times each probe (to go around asset database issue I assume).
 Before importing it pulls something from the RenderPipelineManager.currentPipeline so what happens is
 - Import 1: RenderPipelineManager.currentPipeline is HD, the settings are pulled OK, Preview is rendered with legacy, RenderPipelineManager.currentPipeline is now null since last render call was with legacy and we didn't have any other camera.Render() in between.
 - Import 2: RenderPipelineManager.currentPipeline is now null! The settings are not pulled ok and a crashes happen.


The fix calls a dummy camera.Render() to reset the pipeline. 

Phew.

PR #6267, opened 2020-03-18
Fixed an issue with refraction model and ray traced recursive rendering (case 1198578).

anisunity:
There was an issue (using rasterization parametrisation for things that can be compltely resolved with ray tracing.

Waiting for the validation of @iM0ve to backport


anisunity:
No other model that is currently supported by ray tracing supports refraction, so the fix is only required for Lit

anisunity:
DXR tests are still passing locally

anisunity:
This one is a bit risky, i'd only backport it to 8.x.x

PR #6272, opened 2020-03-18
Fix case in PBR sky where a camera below ground might invalidate result of a camera above

FrancescoC-unity:
This was a fun one given that I knew nothing about the sky system! 
First of all but that it fixes: https://fogbugz.unity3d.com/f/cases/1226588/ 

The issue was that a planar reflection was placed so that it was below ground. Because the sky convolution texture is shared among various cameras, when the planar reflection triggered an update of the sky, it would cause the PBR sky to render black given that it is below ground.
Because the texture was shared with cameras above ground, the result for main camera was broken. 

This PR should solve two issues: 

 - Fix for the case above: if a camera is above ground and one is below, they will use different sky context in case of PBR sky. 
 - Planar reflections will use the position of the parent camera to perform PBR sky computations (so we don't reflect black sky if the planar probe is below ground but main camera is above). 

Note, as mentioned I never ventured in the sky code, so please tell me if I am missing something here. 

For QA: The bug case above is a good test scene, I would like though if you were to do a pass with a PBR sky and planar reflections in the scene.


EvgeniiG:
Thanks for looking into this.
I can see how the probe generation issue Is fixed you just use the parent camera that is above ground). Not sure about the texture look-up, though.

image

The correct (but expensive) solution is to sum the contributions along paths A and B. The simplest approximation is to neglect the contribution of the path A. What you're doing is... D? Use the position of one camera and the orientation of another?

FrancescoC-unity:
> Thanks for looking into this.
> I can see how the probe generation issue is fixed: you just use the parent camera that is above ground. Not sure about the texture look-up, though.
> 
> image
> 
> The correct (but expensive) solution is to sum the contributions along paths A and B. The simplest approximation is to neglect the contribution of the path A (assuming that path is short and thus its contribution is small). What you're doing is... D? Use the position of one camera and the orientation of another?

Hey, as I said not familiar with what the system is doing so I am not surprised I missed something here :-)  Also things have slipped a bit my mind, it has been a busy rest of the week! 

So I am assuming you refer to this change here  https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6272/files#diff-3243f57ff4b4361ecaad378ee33c7a7bR702 ? 

I can see by your diagram how this is indeed odd, the issue was trying to solve was for when the planar reflection position is below ground level. Then the Sky computations will be black and I guess the sky reflected would look odd if camera is above ground? 

Again don't know much details here, so I am gonna take this over slack so I can better understand!

EvgeniiG:
Note: regular opaque geometry should have the contribution A applied to it during the opaque atmospheric scattering pass (but AFAIK it's disabled now). So if the planar reflection pass receives the additional contribution B, we can get the desired A+B if it is also affected by the subsequent opaque atmospheric scattering pass.

JulienIgnace-Unity:
@EvgeniiG 
> I can see how the probe generation issue is fixed: you just use the parent camera that is above ground. Not sure about the texture look-up, though.
I think we can safely assume that planar probe position will always be close to the parent camera so it's fair to use this position for rendering the sky in the planar probe. The result should be very close (without the under the ground issue of course :) )

EvgeniiG:
While I agree (except for the case of a flight sim, perhaps), we should try to avoid adding hacks and approximations to the system, as once a failure case does show up, it may be hard to diagnose.

JulienIgnace-Unity:
> While I agree (except for the case of a flight sim, perhaps), we should try to avoid adding hacks and approximations to the system (if possible), as once a failure case does show up, it may be hard to diagnose.

I would argue that flight sims is the 1% and you most likely would not use (or even see) a planar probe on the ground. So even in this case this change won't make much of a difference IMO.

EvgeniiG:
Can you explain how the texture lookup is performed for the planar probe? Do you actually use the position of the original camera and the orientation of the reflected camera? Or something else? Thanks.

EvgeniiG:
To put it simply, I understand how the precomputation step works (use the position of the original camera). I don't understand how the runtime step works (sample the precomputed textures while rendering with the reflected camera).

EvgeniiG:
Note that the precomputation step only depends on the properties of the atmosphere and does not depend on the camera or any of the lights. So planar reflection should use the same precomputed textures. At runtime, we sample the precomputed textures, and that depends on both the position and the orientation of the camera.

I assume what is happening here is that we use one set of precomputed textures, and you use the position of the original camera and the orientation of the parent camera to sample these precomputed textures.

Can you please confirm or otherwise clarify how it works? It is not clear to me. Also, add a comment in the code. Thanks.

JulienIgnace-Unity:
> Can you please confirm or otherwise clarify how it works? It is not clear to me. Also, add a comment in the code. Thanks.

It does not change the precomputation, only the sampling of data for the rendering part. The precomputing hash used by the PBR sky renderer is independent from this one and hasn't been changed.

FrancescoC-unity:
> Note that the precomputation step only depends on the properties of the atmosphere and does not depend on the camera or any of the lights. So planar reflection should use the same precomputed textures. At runtime, we sample the precomputed textures, and that depends on both the position and the orientation of the camera.
> 
> I assume what is happening here is that we use a single set of precomputed textures, and we use the position of the original (parent) camera and the orientation of the reflected camera to sample these precomputed textures.
> 
> Can you please confirm or otherwise clarify how it works? It is not clear to me. Also, add a comment in the code. Thanks.

So yes, it is how you assumed above. The precomputed textures are not-affected. 
And at runtime when rendering the sky (m_SkyManager.RenderSky) we  use the _parent_ camera position in place of the probe camera. Everything else remains unchanged.  

(aaand I've been ninja-d by Julien)

EvgeniiG:
Thanks for the confirmation. I would prefer that we compute B rather than D (refer to the figure), but that is still better than the current situation.

PR #6280, opened 2020-03-18
Fix for bug 1227659 : NullReferenceException when saving a shader graph in a URP project with the HDRP package also present

Nightmask3:
### Checklist for PR maker
  -> Don't think this needs a changelog because it doesn't repro in 8.x.x according to bug details, it was introduced within work on 9.0.0. Will still add one if needed. 
---
### Purpose of this PR

Fixes this bug: https://fogbugz.unity3d.com/f/cases/1227659/

---
### Testing status

- [x] Opened URP test project and ran tests locally
- [x] Opened HDRP test project and ran tests locally
- [x] Opened Shader graph test project and ran tests locally


Nightmask3:
The exception was thrown because an HDRP callback that gets called when the graph is saved was accessing a possibly undefined symbol if the HDRP package is present in a URP project, but isn't properly configured.

Nightmask3:
Just moved the offending line of code past a check to see if the HDRP pipeline was properly defined for this project, and that returns out if it isn't.

sebastienlagarde:
Does this fix need to be backport on 8.x / 7.x ?

marctem:
> Does this fix need to be backport on 8.x / 7.x ?

@sebastienlagarde  It only needs a backport if you backport the bug. =) The offending line was merged to master last Friday in the staging merge.

PR #6286, opened 2020-03-19
Added an alpha blend option for recursive rendering

anisunity:
Previously if a transparent mesh was not set refractive, it behaved like it was opaque. With this PR, if the refraction keyword is not defined, we operate Alpha blend (like we would in rasterization).
I will add a test for it in a second commit.


anisunity:
Test scene added

PR #6291, opened 2020-03-19
Partial fix LookDev opened when CoreRP package reimported

RSlysz:
### Purpose of this PR
Partly fix https://fogbugz.unity3d.com/f/cases/1219692/
Remaining issue should be fixed in https://fogbugz.unity3d.com/f/cases/1228706/

The issue happens when the LookDev window is open and user upgrade the CoreRP/HDRP package to a newer version.
Currently, doing this, will call OnEnable of the LookDev window as soon as the script is reimported and this happens at a time where the Stylesheet have not been reimported yet. So importing it in order to use it will cause issue later as you will have the one in the former version that will be garbage collected soon.
The first OnGUI is called at a time that the stylesheet is reimported but if we load the stylesheet at that time, the first frame of the window will not have any stylesheet (so no layout and wrong colors)

As this case is sufficiently a rare occurence, if it happens, the present fix will use the wrong stylesheet in OnEnable. This will cause potentially few issue in the console. But we now detect the stylesheet have been garbage collected and thus close and reopen the LookDev window in first OnGUI to stop the error at editor frame, and fully fix the window. 

The "remaining" issue linked need to be fixed on UIElement side to not try to use null Stylesheet. Once it is landed, it will be able to finish the fix.

---
### Testing status

**Manual Tests**:
- Other: Tested manually while migrating from package 7.1.8 to 7.2.1

---


com.unity.render-pipelines.core/Editor/LookDev/DisplayWindow.cs
```diff
@@ -200,8 +200,30 @@ SidePanel sidePanel
             remove => OnUpdateRequestedInternal -= value;
         }
 
+        StyleSheet styleSheet = null;
+        StyleSheet styleSheetLight = null;
+
         void OnEnable()
         {
+            Debug.Log("OnEnable");
```
TomasKiniulis: Debug.Log left out

PR #6302, opened 2020-03-20
Hdrp/custom pass aov

pmavridis:
[Draft PR to get some feedback on the API]

### Purpose of this PR
This PR exposes the results of custom passes to the AOV API. Since the desired output might be the result of multiple custom passes, AOVs are recorded after all passes of an injection point are executed. The API allows you to output either the custom pass buffer or the camera buffer.

An example use-case for this is to generate Object IDs with a custom pass, output them with the AOV API and record them with the unity recorder.

image

---
### Testing status
For testing I have used the standard HDRP template scene, I setup a custom pass to generate Object IDs and then I have used a slightly modified AOV recorder to record the results.

You can find the test scene (with the Object ID pass) here:
https://drive.google.com/open?id=1jE1S_Uqor8-gNK-wSrzPFPEoyTwZKq5P
(you will have to change the paths in packages/manifest.json) 

And the modified AOV recorder here:
https://drive.google.com/open?id=1zOYm8BSj-Az6KrcBclNWM-qqP6TvVf1r

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -1988,7 +1988,7 @@ namespace UnityEngine.Rendering.HighDefinition
             RenderCustomPass(renderContext, cmd, hdCamera, customPassCullingResults, CustomPassInjectionPoint.BeforeRendering);
 
             // Push the custom pass buffer, in case it was requested in the AOVs
-            aovRequest.PushCameraTexture(cmd, AOVBuffers.CustomPassBufferBeforeRendering, hdCamera, m_CustomPassColorBuffer, aovBuffers);
+            aovRequest.PushCustomPassTexture(cmd, CustomPassInjectionPoint.BeforeRendering, m_CameraColorBuffer, m_CustomPassColorBuffer, aovBuffers);
 
             // This is always false in forward and if it is true, is equivalent of saying we have a partial depth prepass.
             bool shouldRenderMotionVectorAfterGBuffer = RenderDepthPrepass(cullingResults, hdCamera, renderContext, cmd);
@@ -2024,7 +2024,7 @@ namespace UnityEngine.Rendering.HighDefinition
             RenderCustomPass(renderContext, cmd, hdCamera, customPassCullingResults, CustomPassInjectionPoint.AfterOpaqueDepthAndNormal);
 
             // Push the custom pass buffer, in case it was requested in the AOVs
-            aovRequest.PushCameraTexture(cmd, AOVBuffers.CustomPassBufferAfterOpaqueDepthAndNormal, hdCamera, m_CustomPassColorBuffer, aovBuffers);
+            aovRequest.PushCustomPassTexture(cmd, CustomPassInjectionPoint.AfterOpaqueDepthAndNormal, m_CameraColorBuffer, m_CustomPassColorBuffer, aovBuffers);
 
             // In both forward and deferred, everything opaque should have been rendered at this point so we can safely copy the depth buffer for later processing.
             GenerateDepthPyramid(hdCamera, cmd, FullScreenDebugMode.DepthPyramid);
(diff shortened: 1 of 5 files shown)


alelievr:
So correct me if i'm wrong but with the current design it's not possible to output more than one custom target per injection point ?
For example if a user allocate one or more RTHandle(s) in his custom pass, then he won't be able to output it in the AOV ?

pmavridis:
> So correct me if i'm wrong but with the current design it's not possible to output more than one custom target per injection point ?
> For example if a user allocate one or more RTHandle(s) in his custom pass, then he won't be able to output it in the AOV ?

That's correct, the user can output either the custom target or the camera color buffer. I did not thought about exposing user allocated RTHandles.Do we have a use-case that this is necessary?

alelievr:
>Do we have a use-case that this is necessary?

Not yet :) but i can imagine user asking for it. That being said it's maybe a very rare usecase not sure if we need to add it now or wait for the feature request since it may not be trivial to implement.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1987,6 +1988,9 @@ AOVRequestData aovRequest
 
             RenderCustomPass(renderContext, cmd, hdCamera, customPassCullingResults, CustomPassInjectionPoint.BeforeRendering);
 
+            // Push the custom pass buffer, in case it was requested in the AOVs
+            aovRequest.PushCustomPassTexture(cmd, CustomPassInjectionPoint.BeforeRendering, m_CameraColorBuffer, m_CustomPassColorBuffer, aovCustomPassBuffers);
```
alelievr: Maybe all the aovRequest.PushCustomPassTexture() calls can be factorized into one that we put in the RenderCustomPass function ?
It would avoid to duplicate the injection point code each time (and it also can be forgotten if we add a new injection point in the future)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVBuffers.cs
```diff
@@ -14,4 +14,23 @@ public enum AOVBuffers
         /// <summary>Motion vectors buffer at the end of the frame.</summary>
         MotionVectors
     }
+
+
+    public struct CustomPassAOVBuffers
```
alelievr: API Documentation :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequestData.cs
```diff
@@ -41,12 +52,15 @@ public struct AOVRequestData
 
         private AOVRequest m_Settings;
         private AOVBuffers[] m_RequestedAOVBuffers;
+        private CustomPassAOVBuffers[] m_CustomPassAOVBuffers;
         private FramePassCallback m_Callback;
+        private FramePassCallbackEx m_CallbackEx;
```
alelievr: Why do we need to have add a new calback here ? Can't we just add another parameter `FramePassCallback` ?

alelievr:
As this feature was initialy done to export the object id buffer I think the feature is good as it is but we may need to consider these points:
- Like i said before we may need to add an API in the custom pass themselve to export an arbitrary buffer to AOV, but that can be done after
- When I'll add support for multiple custom pass volume to be executed at the same injection point, we'll have to be careful of how we use the Custom buffer and the execution order of the volumes. We don't want our object id to be cleared because there is another custom pass executed just before we capture the buffer.

sebastienlagarde:
Hey, so Antoine PR have been merged, need a refreshed on this one.
Regarding your question Antoine, we have currently no idea :). 

>Like i said before we may need to add an API in the custom pass themselve to export an arbitrary >buffer to AOV, but that can be done after

But what I know for sure is that it will be a never ending story. Users will always do crazy request for everything :). For what we want to cover it is fine, so let's start with it. Let's not overenginer too much.

sebastienlagarde:
So Pavlos, we are good to go with this version, and Remi can test it.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequestData.cs
pmavridis: FramePassCallback is used in a public API (see AOVRequestBuilder), so adding an extra parameter in the callback will break existing user scripts, as far as I can see.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVBuffers.cs
pmavridis: Added (GitHub has not updated the code-preview due to a force-push).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/AOV/AOVRequestData.cs
alelievr: ah yes, forgot about it

PR #6304, opened 2020-03-20
Dxc compatibility

Dredhog:
---
### Purpose of this PR
- To make SRP shaders compatible with the DirectX shader compiler.
- Fix a usage of `atan2` which was relying on undefined behavior (in this PR only because it was looking differently with DXC).

---
### Testing status

**Manual Tests**: 
- QA inspected HDRP_Tests project for differences/errors with and w/o DXC

---
### Comments to reviewers

- All shaders should produce the same results as they did before this PR with the existing compilers  ( DebugFullScreen.shader might be an exception as now it's explicitly assigned a black output color when motion vectors are zero so test images could potentially have minor differences)
- As far as we know only 1 shader has incorrect output with DXC `com.unity.render-pipelines.high-definition\Runtime\Lighting\LightLoop\Deferred.compute` which compiles but has no planar reflections due to a bug in DXC itself (reported here https://github.com/microsoft/DirectXShaderCompiler/issues/2772)

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -663,6 +663,45 @@ uint GetMipCount(Texture2D tex)
 // ----------------------------------------------------------------------------
 // Texture format sampling
 // ----------------------------------------------------------------------------
+#if defined(UNITY_COMPILER_DXC) && !defined(DXC_SAMPLER_COMPATIBILITY)
+#define DXC_SAMPLER_COMPATIBILITY 1
+struct sampler1D            { Texture1D t; SamplerState s; };
+struct sampler2D            { Texture2D t; SamplerState s; };
+struct sampler3D            { Texture3D t; SamplerState s; };
+struct samplerCUBE          { TextureCube t; SamplerState s; };
+
+float4 tex1D(sampler1D x, float v)              { return x.t.Sample(x.s, v); }
+float4 tex2D(sampler2D x, float2 v)             { return x.t.Sample(x.s, v); }
+float4 tex3D(sampler3D x, float3 v)             { return x.t.Sample(x.s, v); }
+float4 texCUBE(samplerCUBE x, float3 v)         { return x.t.Sample(x.s, v); }
+
+float4 tex1Dbias(sampler1D x, in float4 t)              { return x.t.SampleBias(x.s, t.x, t.w); }
+float4 tex2Dbias(sampler2D x, in float4 t)              { return x.t.SampleBias(x.s, t.xy, t.w); }
+float4 tex3Dbias(sampler3D x, in float4 t)              { return x.t.SampleBias(x.s, t.xyz, t.w); }
+float4 texCUBEbias(samplerCUBE x, in float4 t)          { return x.t.SampleBias(x.s, t.xyz, t.w); }
+
+float4 tex1Dlod(sampler1D x, in float4 t)           { return x.t.SampleLevel(x.s, t.x, t.w); }
+float4 tex2Dlod(sampler2D x, in float4 t)           { return x.t.SampleLevel(x.s, t.xy, t.w); }
+float4 tex3Dlod(sampler3D x, in float4 t)           { return x.t.SampleLevel(x.s, t.xyz, t.w); }
+float4 texCUBElod(samplerCUBE x, in float4 t)       { return x.t.SampleLevel(x.s, t.xyz, t.w); }
+
+float4 tex1Dgrad(sampler1D x, float t, float dx, float dy)              { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 tex2Dgrad(sampler2D x, float2 t, float2 dx, float2 dy)           { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 tex3Dgrad(sampler3D x, float3 t, float3 dx, float3 dy)           { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 texCUBEgrad(samplerCUBE x, float3 t, float3 dx, float3 dy)       { return x.t.SampleGrad(x.s, t, dx, dy); }
+
+float4 tex1D(sampler1D x, float t, float dx, float dy)              { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 tex2D(sampler2D x, float2 t, float2 dx, float2 dy)           { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 tex3D(sampler3D x, float3 t, float3 dx, float3 dy)           { return x.t.SampleGrad(x.s, t, dx, dy); }
+float4 texCUBE(samplerCUBE x, float3 t, float3 dx, float3 dy)       { return x.t.SampleGrad(x.s, t, dx, dy); }
+
+float4 tex1Dproj(sampler1D s, in float2 t)              { return tex1D(s, t.x / t.y); }
+float4 tex1Dproj(sampler1D s, in float4 t)              { return tex1D(s, t.x / t.w); }
+float4 tex2Dproj(sampler2D s, in float3 t)              { return tex2D(s, t.xy / t.z); }
+float4 tex2Dproj(sampler2D s, in float4 t)              { return tex2D(s, t.xy / t.w); }
+float4 tex3Dproj(sampler3D s, in float4 t)              { return tex3D(s, t.xyz / t.w); }
+float4 texCUBEproj(samplerCUBE s, in float4 t)          { return texCUBE(s, t.xyz / t.w); }
+#endif
 
 float2 DirectionToLatLongCoordinate(float3 unDir)
 {
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -197,8 +197,8 @@
 #define LODDitheringTransition ERROR_ON_UNSUPPORTED_FUNCTION(LODDitheringTransition)
 #endif
 
-// On everything but GCN consoles we error on cross-lane operations
-#ifndef PLATFORM_SUPPORTS_WAVE_INTRINSICS
+// On everything but GCN consoles or DXC compiled shaders we error on cross-lane operations
+#if !defined(PLATFORM_SUPPORTS_WAVE_INTRINSICS) && !defined(UNITY_COMPILER_DXC)
 #define WaveActiveAllTrue ERROR_ON_UNSUPPORTED_FUNCTION(WaveActiveAllTrue)
 #define WaveActiveAnyTrue ERROR_ON_UNSUPPORTED_FUNCTION(WaveActiveAnyTrue)
 #define WaveGetLaneIndex ERROR_ON_UNSUPPORTED_FUNCTION(WaveGetLaneIndex)
--- a/com.unity.render-pipelines.high-definition/Runtime/Debug/DebugFullScreen.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Debug/DebugFullScreen.shader
@@ -241,6 +241,11 @@ Shader "Hidden/HDRP/DebugFullScreen"
                         d = 1.0 - saturate(d);
                     }
 
+                    // Explicitly handling the case where mv == float2(0, 0) as atan2(mv.x, mv.y) above would be atan2(0,0) which
+                    // is undefined and in practice can be incosistent between compilers (e.g. NaN on FXC and ~pi/2 on DXC)
+                    if(!any(mv))
+                        color = float3(0, 0, 0);
+
                     return float4(color + d.xxx, 1.0);
                 }
                 if (_FullScreenDebugMode == FULLSCREENDEBUGMODE_COLOR_LOG)
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
@@ -10,7 +10,7 @@ TEXTURE2D(_DecalAtlas2D);
 SAMPLER(_trilinear_clamp_sampler_DecalAtlas2D);
 
 #ifdef PLATFORM_SUPPORTS_BUFFER_ATOMICS_IN_PIXEL_SHADER
-RWStructuredBuffer<uint> _DecalPropertyMaskBuffer;
+RWStructuredBuffer<uint> _DecalPropertyMaskBuffer : register(u4);
 StructuredBuffer<uint> _DecalPropertyMaskBufferSRV;
 #endif
 
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
@@ -1773,9 +1773,9 @@ IndirectLighting EvaluateBSDF_ScreenspaceRefraction(LightLoopContext lightLoopCo
     refractionOffsetMultiplier *= (hitLinearDepth > posInput.linearDepth);
 
     float2 samplingPositionNDC = lerp(posInput.positionNDC, hit.positionNDC, refractionOffsetMultiplier);
-    float3 preLD = SAMPLE_TEXTURE2D_X_LOD(_ColorPyramidTexture, s_trilinear_clamp_sampler,
-                                        // Offset by half a texel to properly interpolate between this pixel and its mips
+    float3 preLD = SAMPLE_TEXTURE2D_X_LOD(_ColorPyramidTexture, s_trilinear_clamp_sampler, \
                                         samplingPositionNDC * _ColorPyramidScale.xy, preLightData.transparentSSMipLevel).rgb;
+                                        // Offset by half a texel to properly interpolate between this pixel and its mips
 
     // Inverse pre-exposure
     preLD *= GetInverseCurrentExposureMultiplier();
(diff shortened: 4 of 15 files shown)


com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
```diff
@@ -10,7 +10,7 @@ TEXTURE2D(_DecalAtlas2D);
 SAMPLER(_trilinear_clamp_sampler_DecalAtlas2D);
 
 #ifdef PLATFORM_SUPPORTS_BUFFER_ATOMICS_IN_PIXEL_SHADER
-RWStructuredBuffer<uint> _DecalPropertyMaskBuffer;
+RWStructuredBuffer<uint> _DecalPropertyMaskBuffer : register(u4);
```
sebastienlagarde: why do we need to provide an explicity number here? looks dangerous as any change on C# side will break.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
```diff
@@ -63,7 +63,7 @@
     #define TEXTURE2D_X_UINT(textureName)                                    Texture2DArray<uint> textureName
     #define TEXTURE2D_X_UINT2(textureName)                                   Texture2DArray<uint2> textureName
     #define TEXTURE2D_X_UINT4(textureName)                                   Texture2DArray<uint4> textureName
-    #define TEXTURE2D_X_MSAA(type, textureName)                              Texture2DMSArray<type> textureName
+    #define TEXTURE2D_X_MSAA(type, textureName)                              Texture2DMSArray<type, 1> textureName
```
sebastienlagarde: We should add a comment for this one to remember why it is here :)

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
```diff
@@ -1773,9 +1773,9 @@ IndirectLighting EvaluateBSDF_ScreenspaceRefraction(LightLoopContext lightLoopCo
     refractionOffsetMultiplier *= (hitLinearDepth > posInput.linearDepth);
 
     float2 samplingPositionNDC = lerp(posInput.positionNDC, hit.positionNDC, refractionOffsetMultiplier);
-    float3 preLD = SAMPLE_TEXTURE2D_X_LOD(_ColorPyramidTexture, s_trilinear_clamp_sampler,
-                                        // Offset by half a texel to properly interpolate between this pixel and its mips
+    float3 preLD = SAMPLE_TEXTURE2D_X_LOD(_ColorPyramidTexture, s_trilinear_clamp_sampler, \
```
sebastienlagarde: DXC don't support comment like this ? :), wow. 
it don't support just end of line?  "\" seems useless

sebastienlagarde:
Note for myself, backport some fix in 8.x / 7.x

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
aras-p: Because in theory that was already always the case -- if you use UAV output in a pixel shader, you need to manually bind it to the register (render target index) that the C# side uses in SetRenderTargets.

With fxc (shader model 5.0) it was lucky that fxc was most of the time automatically assigning the register that was "number of pixel shader color outputs + 1".

But with dxc (and even fxc with shader model 5.1 -- which we don't use so we never noticed) that's no longer the case.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -663,6 +663,45 @@ uint GetMipCount(Texture2D tex)
 // ----------------------------------------------------------------------------
 // Texture format sampling
 // ----------------------------------------------------------------------------
+#if defined(UNITY_COMPILER_DXC) && !defined(DXC_SAMPLER_COMPATIBILITY)
```
aras-p: I'd like a similar comment as what you had in `HLSLSupport.cginc` for trunk PR, explaining why this is all needed.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
Dredhog: So, UAV's should probably have explicit locations assigned anyway, since the C# side uses indices to reference them (from what I've seen). This was the only location that had omitted that. The reason this is needed for DXC is that S.M. 5.0. (which we target exclusively with FXC) binds UAVs to the same space as PS outputs (see http://shader-playground.timjones.io/e5051934adb37db53d208583c3032991), whereas S.M. 6.0 and up (which is what DXC targets) binds UAVs to a separate space which  different (see http://shader-playground.timjones.io/56668cd7b4659389b11634e7218d0726) resulting in incompatible indices

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -663,6 +663,45 @@ uint GetMipCount(Texture2D tex)
...
+#define DXC_SAMPLER_COMPATIBILITY 1
+struct sampler1D            { Texture1D t; SamplerState s; };
+struct sampler2D            { Texture2D t; SamplerState s; };
+struct sampler3D            { Texture3D t; SamplerState s; };
+struct samplerCUBE          { TextureCube t; SamplerState s; };
```
EvgeniiG: What's the point of these? Why can't we just use two independent objects?
Also, misleading name.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
EvgeniiG: I would also very strongly favor an automated approach. Is this only for pixel shaders?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
EvgeniiG: What we are doing here is a bit fishy, so it's hard for me to tell whether this should be supported or not. Here's the macro we expand:

`#define SAMPLE_TEXTURE2D_ARRAY_LOD(textureName, samplerName, coord2, index, lod)         textureName.SampleLevel(samplerName, float3(coord2, index), lod)`

I suppose we could ask MS.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
Dredhog: This is just a hack for backwards compatibility with the legacy sampler* types, which are no longer available with DXC. This makes it so that places that use e.g. sampler2D would not have to be rewritten. Using independent objects is the best way to do things, this is just intended to ease the transition. As @aras-p mentioned above I'll add an explanation for this section in a comment.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
EvgeniiG: So, why is it here?

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
Dredhog: Yes, this is only for PS. But do we even have an API to set UAVs based on reflection info .i.e. by querying the index by the name. I could only find https://docs.unity3d.com/ScriptReference/Graphics.SetRandomWriteTarget.html

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
Dredhog: Ok, so If there's no explicit second template parameter with a valid sample count for DXC -reflection will report D3D12_SHADER_INPUT_BIND_DESC.NumSamples == 0xffffffff which means not multi-sampled. Our code doesn't actually care about how many actual samples there are, only if it's multisampled or not, and it doesn't seem that hardware discriminates between the counts either. So this seems to be an option to get the reflection to actually tell us if it's MS or not.

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
aras-p: @EvgeniiG we don't have a way of "automatically" doing UAV bindings and making code resolve that by reflection or whatever, and we never had (maybe we should! but someone would have to make it). All the cases in existing shaders where RWTexture in pixel shaders worked without an explicit register binding are just "happens to work by a lucky accident" if the compiler assigns the same binding index as what the code is doing.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
aras-p: @EvgeniiG these are "legacy DX9 style HLSL syntax". HDRP does not use them, but URP does (since URP also needs to work on OpenGL / OpenGL ES, and this legacy syntax is the only thing that works there).

DXC does not support this legacy syntax anymore, so it needs to be "manually emulated". This was a very similar case on PS4/PSSL forever already, where they do a very similar "trick".

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
Dredhog: Hmm, just checked, it seems I'm no longer able to reproduce the problem this was solving... This could have probably been an issue with the optional new preprocessor which I was using exclusively with DXC at the time of changing this. DXC compilation no longer relies on it so I could undo this change. Should I do so?

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
Dredhog: Added comment in 82e6a4d

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
Dredhog: Explanation added in 82e6a4d

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugFullScreen.shader
```diff
@@ -241,6 +241,11 @@ Shader "Hidden/HDRP/DebugFullScreen"
                         d = 1.0 - saturate(d);
                     }
 
+                    // Explicitly handling the case where mv == float2(0, 0) as atan2(mv.x, mv.y) above would be atan2(0,0) which
+                    // is undefined and in practice can be incosistent between compilers (e.g. NaN on FXC and ~pi/2 on DXC)
+                    if(!any(mv))
+                        color = float3(0, 0, 0);
+
```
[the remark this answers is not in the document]
Dredhog: Let's hope the test images remain identical 🤞

Dredhog:
> Was there any change needed for template or generated code from the VisualEffectGraph ? (There isn't any VFXGraph test in HDRP_Tests).
> 
> Our templates are stored :
> 
> * _com.unity.render-pipelines.high-definition\Editor\VFXGraph\Shaders\Templates\Mesh_ (for HDRP)
> * _com.unity.visualeffectgraph\Shaders\RenderPipeline\Universal_ (for URP)
>   There are two test projects _VisualEffectGraph_ (for HDRP) & _VisualEffectGraph_LWRP_ (for URP, still wrongly named).
> 
> Some details :
> 
> * I don't think we are using the output semantic "COLOR".
> * There isn't any usage of UAV at pixel shader stage (specifically from the VFX).
> * I'm not sure we are free of warning issue like this :
>   `LOAD_TEXTURE2D(texture, int2(input.xy)); )> LOAD_TEXTURE2D(texture, int2(input.xy));`
> * We already had got some platform issue on complex compute shader like this one : https://github.com/Unity-Technologies/ScriptableRenderPipeline/blob/master/com.unity.visualeffectgraph/Shaders/Sort.compute
> * We are using explicitly `#pragma target 4.5` for generated shader.
> 
> But in the end, for now, DXC is only available on DX12, right ?
> 
> Let me know if you encounter an unexpected issue.

@PaulDemeulenaere thanks for the details on VFXGraph! I just asked our team QA too go over those projects with DXC.
So far I haven't seen any issues in shader templates from just grepping through the source, but I could have missed something.
Yes, the version that's about to go into trunk is for DX12 only, but these changes should carry over to other DXC targets (such as Vulkan which is being worked on ATM, and Metal)

com.unity.render-pipelines.high-definition/Runtime/Debug/DebugFullScreen.shader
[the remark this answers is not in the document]
PaulDemeulenaere: I'm pretty confident for this, you can already check this with this job : https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/dxc_compatibility/.yamato%252Fupm-ci-vfxmain.yml%2523VFXMain_Win_DX11_playmode_trunk/1685983/job (it will not use DXC but it will confirm this change doesn't change the outputted result)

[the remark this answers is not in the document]
Dredhog: Thanks, good to know!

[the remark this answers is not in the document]
PaulDemeulenaere: It's green 🍏 !

com.unity.render-pipelines.high-definition/Runtime/Material/Decal/ShaderVariablesDecal.hlsl
EvgeniiG: I am not saying that is the fault of this PR, but it is nonetheless a concern. :-)

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
EvgeniiG: OK, thanks.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
EvgeniiG: > //Using explicit sample count of 1 to force DXC to actually reflect the texture as MS. The actual count appears to be irrelevant and any 2D MS texture array should bind to it

That is very weird. Is MS planning to fix this?

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
```diff
@@ -90,7 +91,8 @@
     #define TEXTURE2D_X_UINT(textureName)                                    Texture2D<uint> textureName
     #define TEXTURE2D_X_UINT2(textureName)                                   Texture2D<uint2> textureName
     #define TEXTURE2D_X_UINT4(textureName)                                   Texture2D<uint4> textureName
-    #define TEXTURE2D_X_MSAA(type, textureName)                              Texture2DMS<type> textureName
+    //Using explicit sample count of 1 to force DXC to actually reflect the texture as MS. The actual count appears to be irrelevant and any 2D MS texture should bind to it
+    #define TEXTURE2D_X_MSAA(type, textureName)                              Texture2DMS<type, 1> textureName
```
EvgeniiG: That is very weird. Is MS planning to fix this?

com.unity.shadergraph/Editor/Resources/Shaders/Checkerboard.shader
```diff
@@ -16,7 +16,7 @@ Shader "Hidden/Checkerboard"
             static const float rows = 24;
             static const float columns = 24;
 
-            float4 frag(v2f_img i) : COLOR
+            float4 frag(v2f_img i) : SV_Target
```
EvgeniiG: Do we want some kind of API abstraction for this?

com.unity.render-pipelines.high-definition/Runtime/Material/Lit/Lit.hlsl
EvgeniiG: It's fine.

com.unity.shadergraph/Editor/Resources/Shaders/Checkerboard.shader
aras-p: I don't think so. 99% of shader code uses SV_Target, just a very small uses COLOR semantic for pixel shader output. DXC no longer supports COLOR, but SV_Target works fine everywhere.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/TextureXR.hlsl
Dredhog: Good point, I haven't reported this to Microsoft yet but am planning to.

Dredhog: DXC issue for Microsoft https://github.com/microsoft/DirectXShaderCompiler/issues/2788

EvgeniiG: Thanks!

Dredhog:
@sebastienlagarde should I run some tests on Yamato or will they get run automatically when everyone has approved the PR? If I need to do it myself, which ones should I run?

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
phi-lira: out of curiosity, we only use these for gles2 and that was because when gles2 API file was added hlsl2glsl was not supported for it. isn't it something we can update now the gles2 API file to remove these?

Dredhog: @phi-lira If we are no longer required to use sampler2D and the other DX9 types on any platforms e.g. gles2 then sure, it would be preferred to remove those usages! This would also simplify adding Vulkan support for DXC since this compatibility trick doesn't yet work on Vulkan due to a bug in the DXC SPIR-V backend

Dredhog: Hmm, but thinking more about this, I don't think we can really remove this trick just yet (not until we stop supporting GL as a target) since in glsl there is no way to actually use samplers and textures independently (there aren't separate sampler and texture type, only combined image samplers)

PR #6312, opened 2020-03-20
Added support for stack lit for ray tracing effects.

anisunity:
Added the support for stack lit with a simple fallback for the performance mode. Added stack lit shaders in some tests.


anisunity:
i probably missed something i'll check

sebastienlagarde:
@remi-chapelain 
>Missing support for "Geometric Specular AA" in Stacklit SG Properties
what do you mean? not related to this PR

Anis if you can merge with HDRP/staging :)

sebastienlagarde:
>I mean when this checkbox is checked in StackLit SG Master Node, the reflection in RT SSR become black.

ok so it is an issue already present. the geometryic AA use ddx/ddy instruction, so it shouldn't compile. OR I guess we replace it with "something", maybe 0 and then it could explain why there is no spec as we multiply by 0. Anis could you check? but it isn't related to this PR.

anisunity:
I fixed the specular lobe and the geometric AA issue

PR #6315, opened 2020-03-20
Fix "Wrong RP for Master Node" message handling so it clears properly.

marctem:
This is wrong under the hood but caused no actual bug for users until we added the warning about invalid ShaderGUI in a different PR.

### Checklist for PR maker
- [X] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.

---
### Purpose of this PR
The Master Node UI was adding an error badge if the active RP was not supported (or there was no active RP configured). But the badge was not using underlying MessageManager system, which means it didn't get cleared and prioritized properly.

---
### Testing status

Verified with Zack's branch that this resolves the badge's lifespan issues.

### Comments to reviewers
Two variables to play with - Pipeline (valid for MN, invalid for MN, none), and ShaderGUI Override (none, valid, invalid)
Expect the "Wrong RP" Error to take precedence over the "Invalid ShaderGUI" warning, and the badges to only appear when there is an actual error state.


LandonTownsendUnity:
Can you merge master so I can test this with the ShaderGUI override?

PR #6327, opened 2020-03-23
Added support for hair SG for ray tracing effects.

anisunity:
Added the support for hair with a simple fallback for the performance mode. Added stack lit shaders in some tests.

The PR is done on top of this one: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6312


anisunity:
Was not able to repro the black issue, and the other issue is by design (transparent is in shadow map)

PR #6328, opened 2020-03-23
Deferred Shadow Quality as multi_compile

FrancescoC-unity:
Based on https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6322 for simplicity. The commits that are specific to this are https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/6aba273610ca6c4cac70dca5a95c5823acf883d2 and https://github.com/Unity-Technologies/ScriptableRenderPipeline/commit/56586d6f0020f52a7a4fa91d1eca087a0dc22caf

Now that we have multi-compile for compute shaders we can have any quality for shadows. 

This is for the moment a draft to see if we'll get variant stripping for compute soon, if we do we should probably wait a bit to incorporate it here. 

If we don't, we should profile the implications of this change and decide if merge it for the time being or not.


sebastienlagarde:
Don't forget to update documentation :)

PR #6329, opened 2020-03-23
Implement constant buffers for HDRP global variables.

JulienIgnace-Unity:
### **Please read**
**PR workflow guidelines**

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.

---
### Purpose of this PR

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.core/Runtime/Common/ConstantBuffer.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// Constant Buffer management class.
+    /// </summary>
+    /// <typeparam name="CBType">Type of the structure that represent the constant buffer data.</typeparam>
+    public class ConstantBuffer<CBType> where CBType : struct
+    {
+        CBType[]                        m_Data = new CBType[1]; // Array is required by the ComputeBuffer SetData API
+        ComputeBuffer                   m_GPUConstantBuffer = null;
+        static ConstantBuffer<CBType>   m_TypedConstantBuffer = new ConstantBuffer<CBType>();
+
+        internal ConstantBuffer()
```
FrancescoC-unity, line 15: Do we even want it internal? I guess is safer for ourselves if this is private, or do you have a use case in mind in which we don't want to go through the (static) API below and handle manually?

JulienIgnace-Unity: I think you're right. It can be private. It's probably a leftover of a previous version. I will try to make it private and see if I forgot about something requiring it being internal.

com.unity.render-pipelines.high-definition/Editor/Material/UIBlocks/AxfSurfaceInputsUIBlock.cs
```diff
@@ -426,7 +426,6 @@ void DrawAxfSurfaceOptionsGUI()
                     BRDFVariants |= ((uint)wardVariant) << 2;
                     BRDFVariants |= ((uint)blinnVariant) << 4;
 
-//                    cmd.SetGlobalFloat( HDShaderIDs._TexturingModeFlags, *(float*) &texturingModeFlags );
```
FrancescoC-unity, line 429: Leftover comment?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -3042,7 +3010,7 @@ static bool DeferredUseComputeAsPixel(FrameSettings frameSettings)
             return frameSettings.IsEnabled(FrameSettingsField.DeferredTile) && (!frameSettings.IsEnabled(FrameSettingsField.ComputeLightEvaluation) || k_PreferFragment);
         }
 
-        BuildGPULightListParameters PrepareBuildGPULightListParameters(HDCamera hdCamera)
+        unsafe BuildGPULightListParameters PrepareBuildGPULightListParameters(HDCamera hdCamera)
```
FrancescoC-unity, line 3013: Why is this unsafe now? Does the array manip down count as ptr manip?

JulienIgnace-Unity: Yes it does.

FrancescoC-unity: Then to be more explicit I'd scope those in unsafe rather than the whole function (which is quite big)

JulienIgnace-Unity: Not sure I like that. Having a scope just for the unsafe bit adds visual noise to the code IMO. As a modifier to the whole function at least it's not "in the way" of reading the content of the function.

FrancescoC-unity: Well it add noise, but it is also explicit where the "unsafe stuff" is happening. 
One could make the counter argument that putting on the whole function it is not clear that unsafe stuff is going on (as one could skip reading it) if there is a lot in the body.

The question here is more: do we care we are doing unsafe stuff? If so I think should be explicit. 
If we don't care we are doing unsafe work, but we need to do it, then I guess I am happy with as a whole function modifier

com.unity.render-pipelines.high-definition/Runtime/Lighting/VolumetricLighting/VolumetricLighting.cs
```diff
@@ -709,23 +736,18 @@ static void GetHexagonalClosePackedSpheres7(Vector2[] coords)
...
-            public float            anisotropy;
-            public Vector4          xySeqOffset;
-            public bool             enableReprojection;
-            public bool             historyIsValid;
-            public int              viewCount;
-            public bool             filterVolume;
-
-            public Matrix4x4[]      pixelCoordToViewDirWS;
+            public ComputeShader                volumetricLightingCS;
+			public ComputeShader    			volumetricLightingFilteringCS;
```
FrancescoC-unity, line 740: indent

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
JulienIgnace-Unity: I'm more on the "don't care" side myself :p
Maybe someone else could  chime in?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Camera/HDCamera.cs
```diff
@@ -566,57 +566,63 @@ internal static void CleanUnused()
...
+            cb._XRViewCount = (uint)viewCount;
 
             float exposureMultiplierForProbes = 1.0f / Mathf.Max(probeRangeCompressionFactor, 1e-6f);
-            cmd.SetGlobalFloat(HDShaderIDs._ProbeExposureScale, exposureMultiplierForProbes);
+            cb._ProbeExposureScale  = exposureMultiplierForProbes;
+        }
+
 
+        // Set up UnityPerView CBuffer.
+        internal void SetupGlobalParams(CommandBuffer cmd, int frameCount)
```
FrancescoC-unity, line 621: Why is the following not going through cbuffer api too? Shader side looks already encapsulated in a global one?

JulienIgnace-Unity: The XR stuff down there is coming in the next PR.

com.unity.render-pipelines.high-definition/Runtime/ShaderLibrary/ShaderVariablesGlobal.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        public float    _DirectionalTransmissionMultiplier;
+        public float    _ProbeExposureScale;
+        public float    _ContactShadowOpacity;
+        public float    _ColorPyramidLodCount;
+
+        public Vector4 _AmbientOcclusionParam; // xyz occlusion color, w directLightStrenght
+        public Vector4 _IndirectLightingMultiplier; // .x indirect diffuse multiplier (use with indirect lighting volume controler)
+
+        public float _SSRefractionInvScreenWeightDistance; // Distance for screen space smoothstep with fallback
```
FrancescoC-unity, line 172: This one is used close to _EnableSSRefraction, we could probably swap this with _MicroShadowOpacity.
Similarly, I think _ColorPyramidLodCount could be in the same float4 (accessed by SSRefraction too and I think less often with the other floats it is with), maybe replaced with the _ReplaceDiffuseForIndirect or whichever makes sense.

PR #6345, opened 2020-03-23
[Shader Graph][7.x.x] Fix Position Node upgrader

alindmanUnity:
### Purpose of this PR
backport #6343 
Fixed an issue where new Position nodes added to graphs would always be "upgraded", causing the coordinate space drop down to switch from World to Absolute World when entering play mode or restarting the project. 

[public link](https://issuetracker.unity3d.com/product/unity/issues/guid/1184617/)
[internal link](https://fogbugz.unity3d.com/f/cases/1184617/)

---
### Testing status

- [x] Built a player
- Other: 
Previous repro behavior: 
- Create a new graph, add a Position node. leave the coordinate space on default World. Save graph.
- Enter and exit playmode to force shader compilation. 
- Open the graph, observe that World has been automatically set to Absolute World. 

Fix behavior: 
- Create a new graph, add a Position node. leave the coordinate space on default World. Save graph.
- Enter and exit playmode to force shader compilation. 
- Open the graph, observe that the coordinate space has not changed.  

---
### Comments to reviewers
This was initially intended to be an upgrader for the position node when Absolute World space was added. Because HDRP uses different values, the intent was that any existing graphs would maintain the same output even though the entry changed. The node version was never set properly and it caused the Position node to always run through the upgrader no matter what. 
Any graphs with the Position node already in use should not see any change in behavior. This only applies to newly created Position nodes after this fix is merged.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -6,8 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-Version Updated
-The version number for this package has increased due to a version update of a related graphics package.
+### Fixed
+- Fixed a bug where the `Position` node would change coordinate spaces from `World` to `Absolute World` when shaders recompile. [1184617](https://issuetracker.unity3d.com/product/unity/issues/guid/1184617/)
 
 ## [7.3.0] - 2020-03-11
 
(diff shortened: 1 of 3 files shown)


alindmanUnity:
after changing the behavior, new manual testing was:

Create a new graph, add a Position node. leave the coordinate space on default World. Save graph.
Enter and exit playmode to force shader compilation.
Open the graph, observe that the coordinate space has not changed.
Added a new function in the upgrader for Position node to change coordinate space to view on upgrade to version number 2. set compiled node version to 2.
Reloaded graph, observed that existing Position nodes were changed to view space.
added new position, observed that coordinate space of new nodes were not changed.

PR #6351, opened 2020-03-24
Pospow and SG triplanar fix

slunity:
### **Please read**
**PR workflow guidelines**

### Checklist for PR maker

---
### Purpose of this PR
Fixes a NaN issue in triplanar SG node, and make core PositivePow generally safer.
---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers

### DIFF
--- a/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
+++ b/com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
@@ -495,11 +495,58 @@ uint FastLog2(uint x)
 }
 #endif
 
-// Using pow often result to a warning like this
+// PositivePow: Same as pow(x,y) but considers x always positive and never exactly 0 such that
+// PositivePow(0,y) will numerically converge to 1 as y -> 0, including PositivePow(0,0) returning 1.
+//
+// Using pow often results in a warning like this
 // "pow(f, e) will not work for negative f, use abs(f) or conditionally handle negative values if you expect them"
-// PositivePow remove this warning when you know the value is positive or 0 and avoid inf/NAN.
-// Note: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509636(v=vs.85).aspx pow(0, >0) == 0
-TEMPLATE_2_REAL(PositivePow, base, power, return pow(abs(base), power))
+//
+// First, PositivePow removes this warning for when you know the x value is positive or 0 and you know you avoid a NaN:
+// ie you know that x == 0 and y > 0, such that pow(x,y) == pow(0, >0) == 0
+// PositivePow(0, y) will however return close to 1 as y -> 0, see below.
+//
+// Also, pow(x,y) is most probably approximated as exp2(log2(x) * y), so pow(0,0) will give exp2(-inf * 0) == exp2(NaN) == NaN.
+//
+// PositivePow avoids NaN in allowing PositivePow(x,y) where (x,y) == (0,y) for any y including 0 by clamping x to a
+// minimum of FLT_EPS. The consequences are:
+//
+// -As a replacement for pow(0,y) where y >= 1, the result of PositivePow(x,y) should be close enough to 0.
+// -For cases where we substitute for pow(0,y) where 0 < y < 1, PositivePow(x,y) will quickly reach 1 as y -> 0, while
+// normally pow(0,y) would give 0 instead of 1 for all 0 < y.
+// eg: if we #define FLT_EPS  5.960464478e-8 (for fp32), 
+// PositivePow(0, 0.1)   = 0.1894646
+// PositivePow(0, 0.01)  = 0.8467453
+// PositivePow(0, 0.001) = 0.9835021
+//
+// Depending on the intended usage of pow(), this difference in behavior might be a moot point since:
+// 1) by leaving "y" free to get to 0, we get a NaNs
+// 2) the behavior of PositivePow() has more continuity when both x and y get closer together to 0, since
+// when x is assured to be positive non-zero, pow(x,x) -> 1 as x -> 0.
+//
+// TL;DR: PositivePow(x,y) avoids NaN and is safe for positive (x,y) including (x,y) == (0,0),
+//        but PositivePow(0, y) will return close to 1 as y -> 0, instead of 0, so watch out
+//        for behavior depending on pow(0, y) giving always 0, especially for 0 < y < 1.
+//
+// Ref: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509636(v=vs.85).aspx
+float PositivePow(float base, float power)
+{
+    return pow(max(abs(base), float(FLT_EPS)), power);
+}
+
+float2 PositivePow(float2 base, float2 power)
+{
+    return pow(max(abs(base), float2(FLT_EPS, FLT_EPS)), power);
+}
+
+float3 PositivePow(float3 base, float3 power)
+{
+    return pow(max(abs(base), float3(FLT_EPS, FLT_EPS, FLT_EPS)), power);
+}
+
+float4 PositivePow(float4 base, float4 power)
+{
+    return pow(max(abs(base), float4(FLT_EPS, FLT_EPS, FLT_EPS, FLT_EPS)), power);
+}
 
 // Composes a floating point value with the magnitude of 'x' and the sign of 's'.
 // See the comment about FastSign() below.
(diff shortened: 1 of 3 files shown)


com.unity.shadergraph/Editor/Data/Nodes/UV/TriplanarNode.cs
```diff
@@ -91,7 +91,7 @@ public virtual void GenerateNodeCode(ShaderStringBuilder sb, GenerationMode gene
                 // Whiteout blend method
                 // https://medium.com/@bgolus/normal-mapping-for-a-triplanar-shader-10bf39dca05a
                 case TextureType.Normal:
-                    sb.AppendLine("$precision3 {0}_Blend = max(pow(abs({1}), {2}), 0);"
```
slunity: @sebastienlagarde Note how this is a false NaN guard, the compiler will ditch this because we compile with fast math and thus pow() is assumed to always return >= 0, ie no need for max( ? , 0)

alindmanUnity: I see the function is in core, so I don't think it's an issue but to be safe did you check this against Universal as well? This node uses the same code for both pipelines so you need to test that it works safely in both.

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -495,11 +495,58 @@ uint FastLog2(uint x)
...
 
-// Using pow often result to a warning like this
+// PositivePow: Same as pow(x,y) but considers x always positive and never exactly 0 such that
+// PositivePow(0,y) will numerically converge to 1 as y -> 0, including PositivePow(0,0) returning 1.
+//
+// Using pow often results in a warning like this
 // "pow(f, e) will not work for negative f, use abs(f) or conditionally handle negative values if you expect them"
-// PositivePow remove this warning when you know the value is positive or 0 and avoid inf/NAN.
-// Note: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509636(v=vs.85).aspx pow(0, >0) == 0
-TEMPLATE_2_REAL(PositivePow, base, power, return pow(abs(base), power))
```
LandonTownsendUnity: This function is used in around 30 different files, giving this PR a large halo affect. If this is not meant to fix bugs with any other functionality using PositivePow, are we sure we want to make such a large change to fix Triplanar, instead of some other way?

com.unity.shadergraph/Editor/Data/Nodes/UV/TriplanarNode.cs
slunity: Good point, forgot to add the list of touched files also in my analysis in my commit comment.
In URP, /ShaderLibrary/MetaInput.hlsl uses it for an albedo boost from LightmapSettings (quite like a gamma) the rest is for gamma curve also (LutBuilderHdr and SMAA)

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
slunity: Indeed it is used a bit everywhere, mostly for gamma curves, I've reviewed all cases and ran all HDRP_Tests and HDRP_RuntimeTests, passing. The additional max( ) should be extremely cheap. All that said, definitely has halo, I understand the reluctance. It just made

LandonTownsendUnity: Upon further inspection, this function is used in the majority of color conversion functions in color.hlsl (SRGB to linear, linear to gamma, etc) which means the halo effect is HUGE. I would strongly suggest not making a change to this function.

LandonTownsendUnity: Tests don't cover all edge cases. There are a good deal of edge cases I can think of where this may cause different behavior, especially in regards to black colors (0, 0, 0) being converted between color modes. Remember that with positivepow, when you have a base of 0 and a power of 1, you want the result to be 0, and this may not be the case anymore.
I think doing this as a function called SafePositivePow and only using that function in Triplanar and any other place that specifically needs NAN values corrected would be a much safer alternative.

slunity: Agree it is safer, will amend.

slunity: Done

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -501,6 +501,57 @@ uint FastLog2(uint x)
...
+// 1) by leaving "y" free to get to 0, we get a NaNs
+// 2) the behavior of SafePositivePow() has more continuity when both x and y get closer together to 0, since
+// when x is assured to be positive non-zero, pow(x,x) -> 1 as x -> 0.
+//
+// TL;DR: SafePositivePow(x,y) avoids NaN and is safe for positive (x,y) including (x,y) == (0,0),
+//        but SafePositivePow(0, y) will return close to 1 as y -> 0, instead of 0, so watch out
+//        for behavior depending on pow(0, y) giving always 0, especially for 0 < y < 1.
+//
+// Ref: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509636(v=vs.85).aspx
+float SafePositivePow(float base, float power)
```
sebastienlagarde: cant you use a macro function for this case?

TEMPLATE_2_FLOAT(SafePositivePow, base, power, return pow(max(abs(base), float(FLT_EPS)), power))

it is mean for this :)

sebastienlagarde: also why not a REAL version ?

TEMPLATE_2_REAL

slunity: I've fixed it, I had this sitting for a while but I think originally I mechanically did that for the constant vec type, which is useless obviously since scalars smear to larger types. (float I think it was to be consistent with FLT_EPS, and in the worst case if the platform/driver forces fp16 eg mobile, it reverted to the old function since FLT_EPS would become 0. The function is separate now anyway)

slunity:
> Edit: I'm entering large values into the "blend" variable, I assume that this is causing divide by 0 
> errors instead of pow(0, 0) errors. Is this under the category of things we wanted to fix for this >node?
Ah yes, this is expected. 
I think just the name blend is not every informative and our documentation is not better (at all in fact) at explaining it to the user either. I'm thinking I should update the doc too. 
Blend is a "contrast enhancement" of the "normal selection" that is the basis of triplanar mapping.
In short, the more the blend exponent is high, the more the blending is biased towards the plane towards which the normal is point "more towards". 
In the limit, it should become a kind of "argmax", except like you noticed, it can quickly underflow the float precision and the normalization will make a divide to zero.

We can patch it and/or just warn the user. But I definitely think we need to update the doc too!

com.unity.render-pipelines.core/ShaderLibrary/Common.hlsl
```diff
@@ -501,6 +501,51 @@ uint FastLog2(uint x)
...
+// when x is assured to be positive non-zero, pow(x,x) -> 1 as x -> 0.
+//
+// TL;DR: SafePositivePow(x,y) avoids NaN and is safe for positive (x,y) including (x,y) == (0,0),
+//        but SafePositivePow(0, y) will return close to 1 as y -> 0, instead of 0, so watch out
+//        for behavior depending on pow(0, y) giving always 0, especially for 0 < y < 1.
+//
+// Ref: https://msdn.microsoft.com/en-us/library/windows/desktop/bb509636(v=vs.85).aspx
+TEMPLATE_2_REAL(SafePositivePow, base, power, return pow(max(abs(base), real(REAL_EPS)), power))
+
+// Helpers for making shadergraph functions consider precision spec through the same $precision token used for variable types
```
slunity: @sebastienlagarde Do you mind I added those here, so the triplanar node can seamlessly support the $precision without any new splicing keywords / system to get min/maxes etc.

sebastienlagarde: hey, yes you can do it, however you must chose, 
either:
TEMPLATE_2_REAL(SafePositivePow, base, power, return pow(max(abs(base), real(REAL_EPS)), power))

or both:
TEMPLATE_2_FLT(SafePositivePow_float, base, power, return pow(max(abs(base), float(FLT_EPS)), power))
TEMPLATE_2_HALF(SafePositivePow_half, base, power, return pow(max(abs(base), half(HALF_EPS)), power))

otherwise you will conflict. Haven't you get a compile error with function being redefined?

sebastienlagarde: Ah sorry, misread :), missed the _float and _half.
ok then

PR #6355, opened 2020-03-24
Execute all custom pass volumes

alelievr:
### Purpose of this PR
- Added the priority on custom pass volumes, it's used to determine the execution order when you have multiple custom pass volumes using the same injection point in a scene.
- All custom pass volumes are now executed for each injection point instead of the first one before.

### Testing status

Tested locally in a scene with multiple custom pass volumes and DrawRenderers passes at the same injection point (Before transparent). I setup a material override with a Unlit of a different color for each custom pass volume and play with the priority. The volume with higher priority must be executed before.
CustomPassPriority


com.unity.render-pipelines.high-definition/Editor/RenderPipeline/CustomPass/CustomPassVolumeEditor.cs
```diff
@@ -26,6 +26,8 @@ static class Styles
             public static readonly GUIContent isGlobal = new GUIContent("Mode", "A global volume is applied to the whole scene.");
             public static readonly GUIContent fadeRadius = new GUIContent("Fade Radius", "Radius from where your effect will be rendered, the _FadeValue in shaders will be updated using this radius");
             public static readonly GUIContent injectionPoint = new GUIContent("Injection Point", "Where the pass is going to be executed in the pipeline.");
+            public static readonly GUIContent priority = new GUIContent("Priority", "Determine the execution order when multiple Custom Pass Volume overlaps with the same injection point.");
```
pmavridis: Typo: multiple Custom Pass Volume overlaps => multiple Custom Pass Volumes overlap

com.unity.render-pipelines.high-definition/Documentation~/Custom-Pass.md
```diff
@@ -15,7 +15,8 @@ Custom Passes have been implemented through a volume system, but note that it's
 - The data of the custom passes are saved in the volume GameObject in itself, not in an asset in the project
 
 Like in volumes, there is two modes for the custom pass volume: `Local` and `Global`. The `Local` mode uses colliders attached to the GameObject where the custom pass is to define a zone where the effect will be executed. `Global` volumes are executed everywhere in your scene.  
-Additionally you have a `fade` system that allow you to smooth the transition between your normal rendering and the custom custom pass. The control over the distance of the fade is done by the **Fade Radius** field in the UI of the Custom Pass Volume Component, the radius is exposed in meter and is not scaled with the object transform.  
+The priority is used to determine the execution order when you have multiple custom pass volume in your scene that shares the same injection point.
```
pmavridis: Typo: 
volume => volumes
shares => share

pmavridis:
Looks good from my side and I like the fact that, if I understand correctly, old code will continue to work without any modifications (volumes will get a priority of zero by default).

However, my only concern is that this will change the old behaviour and some users might not like that (even if the new behavior makes more sense). If we don't want this, you might have to add a flag in  RenderCustomPass to select between the old and new behavior. But honestly, I would keep it as it is :)

alelievr:
I think it's fine to change the old behavior, as it wasn't said in the documentation, it's considered to be an undefined behavior.
When you had multiple custom pass volumes with the same injection point, only the first one is executed but there is no definition of "first" it's literally the first that registers to the system and it can be completely random.

PR #6356, opened 2020-03-24
Add Performance framework package and HDRP performance tests

alelievr:
### Purpose of this PR
- Add the graphic performance framework package, based on the performance testing package: https://docs.unity3d.com/Packages/com.unity.test-framework.performance@1.0/manual/index.html
- Add the implementation of HDRP tests for graphic performance.

### Testing status
This framework helps you to setup a performance testing which includes: 
- Frame timings (based on profilingSampler)
- Memory profiling (based on asset profiling)
- Build time (shader compilation, total build time, variant count after stripping per platform)
- Build size (compiled and compressed shader size, Compute shader size ...)
- VGPR / SGPR for shader, ComputeShader, ShaderGraphs and materials

Here's the result of a performance test in the editor:

Also made yamato pipeline so performance tests are automatically run every day

Here's the performance dashboard that we made for HDRP, other team that want to use it will need to build their own one to visualize the performance data.

https://grafana.internal.unity3d.com/d/cLLln6UWk/hdrp-performances?orgId=1&from=now-7d&to=now


com.unity.testing.graphics-performance/Editor/Common/EditorLogWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+    const string editorLogFileName = "Editor.log";
+
+    public delegate void OnLogWriteCallback(string newLines);
+
+    OnLogWriteCallback  logWriteCallback;
+    FileStream          logStream;
+
+    public EditorLogWatcher(OnLogWriteCallback callback)
+    {
+        logWriteCallback = callback;
```
fredericv-unity3d: You should throw an exception if the callback is `null`.

```cs
logWriteCallback = callback ?? throw new ArgumentNullException(nameof(callback));
```

It does not make sense to use this class with a null callback.

com.unity.testing.graphics-performance/Editor/Common/EditorLogWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.IO;
+using UnityEngine;
+
+public class EditorLogWatcher : IDisposable
+{
+    const string editorLogFileName = "Editor.log";
+
+    public delegate void OnLogWriteCallback(string newLines);
+
+    OnLogWriteCallback  logWriteCallback;
```
fredericv-unity3d: `logWriteCallback` and `logStream` don't follow private field convention.

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Text.RegularExpressions;
+using UnityEngine.Rendering;
+using static PerformanceTestUtils;
+using static PerformanceMetricNames;
+
+using Object = UnityEngine.Object;
+
+public class EditorPerformanceTests
+{
+    const int BuildTimeout = 10 * 60 * 1000; // 10 min for each build test
```
fredericv-unity3d: Unused field

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+    const string buildLocation = "TmpBuild";
+
+    // TODO
+    public const string testSceneResourcePath = "TestScenes";
+    static TestSceneAsset testScenesAsset = PerformanceTestSettings.GetTestSceneDescriptionAsset();
```
fredericv-unity3d: Unused field

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using static PerformanceMetricNames;
+
+using Object = UnityEngine.Object;
+
+public class EditorPerformanceTests
+{
+    const int BuildTimeout = 10 * 60 * 1000; // 10 min for each build test
+    const string buildLocation = "TmpBuild";
+
+    // TODO
```
fredericv-unity3d: Describe your todo

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+    public const string testSceneResourcePath = "TestScenes";
+    static TestSceneAsset testScenesAsset = PerformanceTestSettings.GetTestSceneDescriptionAsset();
+
+    protected BuildReport BuildPlayer(string scenePath)
+    {
+        BuildPlayerOptions buildPlayerOptions = new BuildPlayerOptions();
+        buildPlayerOptions.scenes = new[] { scenePath };
+        buildPlayerOptions.locationPathName = buildLocation;
+        buildPlayerOptions.target = BuildTarget.StandaloneWindows64;
+        buildPlayerOptions.options = BuildOptions.Development; // TODO: remove dev build test
```
fredericv-unity3d: Remove the dev build

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        BuildSummary summary = report.summary;
+        Measure.Custom(FormatSampleGroupName(kSize, kTotal).ToSampleGroup(SampleUnit.Byte), summary.totalSize);
+        Measure.Custom(FormatSampleGroupName(kSize, kShader).ToSampleGroup(SampleUnit.Byte), GetAssetSizeInBuild(report, typeof(Shader)));
+        Measure.Custom(FormatSampleGroupName(kSize, kComputeShader).ToSampleGroup(SampleUnit.Byte), GetAssetSizeInBuild(report, typeof(ComputeShader)));
+        Measure.Custom(FormatSampleGroupName(kTime, kTotal).ToSampleGroup(SampleUnit.Millisecond), summary.totalTime.TotalMilliseconds);
+        Measure.Custom(FormatSampleGroupName(kBuild, kWarnings).ToSampleGroup(), summary.totalWarnings);
+        Measure.Custom(FormatSampleGroupName(kBuild, kSuccess).ToSampleGroup(), summary.result == BuildResult.Succeeded ? 1 : 0);
+    }
+
+    protected ulong GetAssetSizeInBuild(BuildReport report, Type assetType)
```
fredericv-unity3d: Can be private

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            foreach (var content in packedAsset.contents)
+                if (content.type == assetType)
+                    assetSize += content.packedSize;
+        }
+
+        return assetSize;
+    }
+
+    protected void ReportShaderSize(BuildReport report, string shaderNameFilter)
```
fredericv-unity3d: Should be static

com.unity.testing.graphics-performance/Editor/Common/EditorLogWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    string GetEditorLogFolderPath() => Path.GetDirectoryName(GetEditorLogPath());
+
+    public void Dispose()
+    {
+        using (var s = new StreamReader(logStream))
+        {
+            while (!s.EndOfStream)
+                logWriteCallback?.Invoke(s.ReadLine());
```
fredericv-unity3d: You can remove the null conditional operator (?.) as we ensure the callback is not null. (see above comment)

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                if (content.type == typeof(ComputeShader))
+                {
+                    var computeShader = AssetDatabase.LoadAssetAtPath<ComputeShader>(content.sourceAssetPath);
+                    Measure.Custom(FormatSampleGroupName(kSize, kComputeShader, computeShader.name).ToSampleGroup(SampleUnit.Byte), content.packedSize);
+                }
+                else if (content.type == typeof(Shader))
+                {
+                    var shader = AssetDatabase.LoadAssetAtPath<Shader>(content.sourceAssetPath);
+
+                    if (shader?.name?.Contains(shaderNameFilter) ?? false)
```
fredericv-unity3d: A few notes here:
```cs
shader?.name?.Contains(shaderNameFilter) ?? false
```
Is quite complicated to understand to evaluate to a boolean.

Also, null conditional operator works badly with `UnityEngine.Object`. It does not use the overload of the operator `==` by `UnityEngine.Object`, so the C# object may be non-null while the related C++ object is.

Although, in that particular case it would work as you just load the reference from the AssetDatabase, but, we should not propagate bad practice in the code

I suggest to use instead
```cs
if (shader != null && !shader.Equals(null) && shader.name.Contains(shaderNameFilter))
```

com.unity.testing.graphics-performance/Editor/Common/SettingsProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+                Debug.LogError("You must choose an asset within a Resources folder");
+            }
+        }
+    }
+}
+
+// Create PerformanceTestsProvider by deriving from SettingsProvider:
+class PerformanceTestsProvider : SettingsProvider
+{
+    private SerializedObject m_CustomSettings;
```
fredericv-unity3d: Field is not used, only written to.

com.unity.testing.graphics-performance/Editor/Common/SettingsProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+    public override void OnActivate(string searchContext, VisualElement rootElement)
+    {
+        // This function is called when the user clicks on the MyCustom element in the Settings window.
+        m_CustomSettings = PerformanceTestSettings.GetSerializedSettings();
+    }
+
+    public override void OnGUI(string searchContext)
+    {
+        // Use IMGUI to display UI:
+        // EditorGUILayout.PropertyField(m_CustomSettings.FindProperty("m_Number"), Styles.number);
```
fredericv-unity3d: This does not seems to be finished here. You should probably clean this code.

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+using UnityEditor.Compilation;
+using System;
+using System.Linq;
+using static PerformanceMetricNames;
+using Object = UnityEngine.Object;
+
+[CustomEditor(typeof(TestSceneAsset))]
+class TestSceneAssetEditor : Editor
+{
+    ReorderableList counterSceneList;
```
fredericv-unity3d: Fields do not follow private field naming convention

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+            var scenePath = elem.FindPropertyRelative(nameof(TestSceneAsset.SceneData.scenePath));
+            var sceneLabels = elem.FindPropertyRelative(nameof(TestSceneAsset.SceneData.sceneLabels));
+            var enabled = elem.FindPropertyRelative(nameof(TestSceneAsset.SceneData.enabled));
+            rect.height = EditorGUIUtility.singleLineHeight;
+
+            // Scene field
+            var sceneGUID = AssetDatabase.FindAssets($"t:Scene {sceneName.stringValue}", new [] {"Assets"}).FirstOrDefault();
+            var sceneAsset = String.IsNullOrEmpty(sceneGUID) ? null : AssetDatabase.LoadAssetAtPath<SceneAsset>(AssetDatabase.GUIDToAssetPath(sceneGUID));
+            sceneAsset = EditorGUI.ObjectField(rect, "Test Scene", sceneAsset, typeof(SceneAsset), false) as SceneAsset;
+            sceneName.stringValue = sceneAsset?.name;
```
fredericv-unity3d: Same remark for null conditional here.

You should use 
```cs
sceneAsset != null && !sceneAsset.Equals(null) ? sceneAsset.name : null
```

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+                serializedObject.ApplyModifiedProperties();
+                serializedObject.Update();
+            }
+        };
+        list.onAddCallback = DefaultListAdd;
+        list.onRemoveCallback = DefaultListDelete;
+        list.elementHeight = fieldHeight * 2;
+    }
+
+    string GetLabelForAsset(Object asset)
```
fredericv-unity3d: Can be static

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+    {
+        ReorderableList.defaultBehaviours.DoAddButton(list);
+
+        // Enable the scene by default
+        var element = list.serializedProperty.GetArrayElementAtIndex(list.count - 1);
+        var enable = element.FindPropertyRelative(nameof(TestSceneAsset.SceneData.enabled));
+        if (enable != null)
+            enable.boolValue = true;
+
+        serializedObject.ApplyModifiedProperties();
```
fredericv-unity3d: You should call `serializedObject.Update()` before editing the serialized object.

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    void DefaultListDelete(ReorderableList list)
+    {
+        list.serializedProperty.DeleteArrayElementAtIndex(list.index);
```
fredericv-unity3d: You should call `serializedObject.Update()` before editing the serialized object.

com.unity.testing.graphics-performance/Editor/Common/TestSceneAssetEditor.cs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    public override void OnInspectorGUI()
+    {
+        if (Resources.Load(PerformanceTestUtils.testSceneResourcePath) == null)
+            EditorGUILayout.HelpBox($"Test Scene Asset have been moved from it's expected location, please move it back to Resources/{PerformanceTestUtils.testSceneResourcePath}", MessageType.Error);
+
+        EditorGUIUtility.labelWidth = 100;
+
+        DrawTestBlock(counterSceneList, counterSRPAssets, "Performance Counters Tests");
```
fredericv-unity3d: Move the texts in a private Style class

com.unity.testing.graphics-performance/Runtime/PerformanceMetricNames.cs
```diff
@@ -0,0 +1,23 @@
+public static class PerformanceMetricNames
+{
+    public readonly static string kNA = "NA";
```
fredericv-unity3d: It does not follow naming convention for public constants. (camelCase)

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
```diff
@@ -0,0 +1,86 @@
+﻿using System.Collections;
```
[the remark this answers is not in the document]
alelievr: Yes i know, I just find that it's clearer like this and clarity should prevail a naming convention when necessary. Also `HDRPEditorPerformanceTests` is not a valid name either if we follow the Unity/Microsoft naming convention `HdrpEditorPerformanceTests` is.
That being said i don't mind changing it.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    const int BuildTimeout = 10 * 60 * 1000;    // 10 min for each build test
+    const string shaderNameFilter = "HDRP";
+    
+    public static IEnumerable<BuildTestDescription> GetBuildTests()
+    {
+        if (testScenesAsset == null)
+            yield break;
+
+        // testName is hardcoded for now
```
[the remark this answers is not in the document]
alelievr: Yes, it's just that i didn't knew how to put it in the config asset. And also i didn't wanted to bloat the UI with useless info, we already have the tags in the scene asset to filter by test category.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    [Timeout(BuildTimeout), Version("1"), UnityTest, Performance]
+    public IEnumerator Build([ValueSource(nameof(GetBuildTests))] BuildTestDescription testDescription)
+    {
+        SetupTest(testDescription.sceneData.scene, testDescription.assetData.asset);
+
+        HDRPreprocessShaders.reportShaderStrippingData += ReportShaderStrippingData;
+
+        var match = new Regex(@"^\s*Compiled shader '(.*)' in (\d{1,}.\d{1,})s$").Match("Compiled shader 'HDRP/Lit' in 103.43s");
```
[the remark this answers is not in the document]
alelievr: I guess it's dead code, it should be here

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: It's part of the API when we extend the `EditorPerformanceTests` so i'll keep it protected

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: Why should it be static ?
It's part of the API when we extend `EditorPerformanceTests` making it static just forces users to write the base class name before the name of the function in this case.

com.unity.testing.graphics-performance/Editor/Common/SettingsProvider.cs
alelievr: yep, not yet finished, will clean

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
```diff
@@ -0,0 +1,12 @@
+# About the Graphics Test Framework
+
+TODO
```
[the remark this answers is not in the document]
alelievr: Agree, I was just waiting to have the final workflow for implementing new test and managing test assets to write a getting started section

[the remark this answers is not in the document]
fredericv-unity3d: I think you have pretty much material to write it, or at least to start it. Making a good getting started section requires several drafts.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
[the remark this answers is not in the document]
fredericv-unity3d: So, it is only used for filtering? When do we need this filtering?

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
fredericv-unity3d: > making it static just forces users to write the base class name before the name of the function in this case.

That is incorrect, users can use the method without specifying the base class.

However, making it non-static implies the execution of the methods relies on its internal state. 
Most bugs comes from invalid internal state reading and writing, so avoiding to rely on internal state is a best practice.

In short, you will prevent bugs by making static methods that can be static easily. This is a bigger deal than the syntactic sugar issue aforementioned.

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
fredericv-unity3d: I usually avoid to provide extensibility features without a use case or an example.
It can be made private and when the extensibility use case occurs we can more accurately work on such API.

Doing so, you will avoid to support an API that is not mature.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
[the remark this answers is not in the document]
alelievr: Yes it's for filtering the test on grafana, right now it's not used and i'm not sure it's useful either.
I guess we sould set it to "NA" for now ?

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRP_RuntimePerformanceTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using UnityEngine.Rendering;
+using Unity.PerformanceTesting;
+using NUnit.Framework;
+using UnityEngine.TestTools;
+using static PerformanceTestUtils;
+using static PerformanceMetricNames;
+
+public class HDRP_RuntimePerformanceTests : PerformanceTests
+{
+    const int WarmupCount = 10;
```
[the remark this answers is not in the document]
alelievr: Forgot to use it, thanks

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: It was about the Resource.Load that was using a hardcoded path, it have been fixed

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: Seeing the current design of the Build test:
```csharp
var buildReport = BuildPlayer(testScenesAsset.GetScenePath(testDescription.sceneData.scene));

ReportBuildData(buildReport);
ReportShaderSize(buildReport, k_ShaderNameFilter);
// Add your custom data report ex: Measure.Custom(s, GetAssetSizeInBuild(buildReport, typeof(Material)));
```
I feel like it's a pretty common use case

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: Ok

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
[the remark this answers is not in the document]
fredericv-unity3d: If it is not useful then it is fine.
However, insert a comment to describe the intent of testName

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
fredericv-unity3d: It is fine as an example.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Editor/HDRP_EditorPerformanceTests.cs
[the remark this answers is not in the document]
alelievr: Ok

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: 👍

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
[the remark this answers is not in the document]
alelievr: I've done the first iteration of the doc, we still need to write 2 more pages (one for the performance description test asset and one for the static analysis asset).
I'll let you write the static analysis one since you know better how it works

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
```diff
@@ -0,0 +1,323 @@
...
++-- Scenes
+```
+
+To create the Editor and Runtime test folders you can use the **Create > Testing > Test Assembly Folder** menu, it will automatically create a folder and configure an assembly definition file in test mode.
+
+Now in the assembly definition file, we need to add all the references we will use to write our tests. A fast way to to it is to paste this in your asmdef file.  
+Runtime:
+```
+    "references": [
+        "GUID:91836b14885b8a34196f4aa8303d7793",
```
fredericv-unity3d: It seems that there are more assemblies than describe below. Maybe you can provide only required assembly references?

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
```diff
@@ -0,0 +1,323 @@
...
+using UnityEngine.Rendering;
+using NUnit.Framework;
+using UnityEngine.TestTools;
+using Unity.PerformanceTesting;
+using static PerformanceTestUtils;
+
+public class MyRuntimePerformanceTests : PerformanceTests
+{
+    // Number of frames before we start recording the profiling samplers
+    const int WarmupCount = 20;
```
fredericv-unity3d: `WarmupCount` is not used, maybe you miss an attribute in the tests?

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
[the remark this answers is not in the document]
fredericv-unity3d: Nice first iteration of the documentation! Can you add the doc team to review the documentation?

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
```diff
@@ -0,0 +1,323 @@
+# Graphics Performance Test Framework
```
fredericv-unity3d: I think it misses a section about the naming standard of measurements. (Asset alias, asset category, test name, ...).

In my opinion, this is quite important to describe this standard.

alelievr: Ah yes, you can add it if you want :)

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
alelievr: It' used in the last line: `        yield return MeasureProfilingSamplers(GetAllMarkers(), WarmupCount, MeasurementCount);
`

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
alelievr: Yes I added core as this framework is only intended to be used by SRP projects (and every project in SRP depend on Core).

Plus it's used in the example below (ProfilingSampler is in Core) so it wouldn't compile without it

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
fredericv-unity3d: Sorry, misread my search tool apparently.

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
fredericv-unity3d: Then add a line to explain the added assemblies. (Otherwise it can be confusing)

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using UnityEngine.Rendering;
+using NUnit.Framework;
+using UnityEngine.TestTools;
+using static PerformanceTestUtils;
+using Unity.PerformanceTesting;
+
+public class HDRPRuntimePerformanceTests : PerformanceTests
+{
+    const int WarmupCount = 20;
+    const int MeasurementCount = 30;  // Number of frames to measure
```
FrancescoC-unity: Can we expose this to test? I can see tests that might want to have more than 30 frames to measure

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            yield return ProfilingSampler.Get((HDProfileId)val);
+    }
+
+    [Timeout(GlobalTimeout), Version("1"), UnityTest, Performance]
+    public IEnumerator Counters([ValueSource(nameof(GetCounterTests))] CounterTestDescription testDescription)
+    {
+        yield return SetupTest(testDescription.sceneData.scene, testDescription.assetData.asset);
+
+        var camera = GameObject.FindObjectOfType<Camera>();
+        var hdCamera = HDCamera.GetOrCreate(camera, 0); // We don't support XR for now
```
FrancescoC-unity: What happen if a scene has multiple cameras? 

Imagine we have a game-like perf test and we want to profile different locations (without having to copy the whole map for each view point especially if it is heavy), we cannot do that?

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
```diff
@@ -292,6 +298,10 @@ public void OnProcessShader(Shader shader, ShaderSnippetData snippet, IList<Shad
                     }
                 }
             }
+
+            Debug.Log(shader + " -> " + shaderStripingWatch.Elapsed.TotalMilliseconds);
```
FrancescoC-unity: My understanding is that the line below (shaderPreprocessed) is what sends the data to the perf framework. Do we want the Debug.Log here? 

I guess it is useful information to have, though I wonder if it won't pollute the log too much on big projects?
I am not necessarily for removing it, but mostly checking if this is wanted or is a debug leftover.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
alelievr: Do you mean in a per-scene configuration ?

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    protected BuildReport BuildPlayer(string scenePath)
+    {
+        BuildPlayerOptions buildPlayerOptions = new BuildPlayerOptions();
+        buildPlayerOptions.scenes = new[] { scenePath };
+        buildPlayerOptions.locationPathName = buildLocation;
+        buildPlayerOptions.target = BuildTarget.StandaloneWindows64;
+
+        // Make sure we compile the shaders when we build:
+        ClearShaderCache();
```
FrancescoC-unity: What happen if the folder is locked by another process? I assume this will never happen on yamato, but if run locally it'd be nice to display an error in that case.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
FrancescoC-unity: Yup

com.unity.testing.graphics-performance/Editor/StaticAnalysis/NDA_PS4/EditorStaticAnalysisTestPS4.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using NUnit.Framework;
+using Unity.PerformanceTesting;
+using UnityEditor;
+
+public class EditorStaticAnalysisTestPS4
```
FrancescoC-unity: Should these file be in the public one repo? (I am gonna add something on slack :) )

com.unity.testing.graphics-performance/Editor/StaticAnalysis/Common/EditorStaticAnalysisTests.cs
```diff
@@ -0,0 +1,221 @@
...
+        SampleGroup CUOccupancyCount,
+        SampleGroup CUOccupancyMax,
+        SampleGroup LDSSize,
+        SampleGroup SIMDOccupancyCount,
+        SampleGroup SIMDOccupancyMax
+    ) GetReportSampleGroups(ShaderBuildReport.PerformanceUnit unit)
+    {
+        var name = string.Join(";", unit.compileUnit.defines);
+        var fullName = $"{unit.program.name}:{name}";
+        return (
```
FrancescoC-unity: Note: some of these fields are not reported for all platforms, we could I guess filter on grafana side, but for non-PS4 platforms these might have bogus values.

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
alelievr: If we have a test like this I guess it would be in another project ? since it could be quiet big and so we could use another code for the test.

And also when we gather the markers we don't know which view it was. So we need to take in account this while repporting the data.

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
alelievr: It's debug i forgot to remove it :p

com.unity.testing.graphics-performance/Editor/StaticAnalysis/NDA_PS4/EditorStaticAnalysisTestPS4.cs
alelievr: @fredericv-unity3d

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
alelievr: done

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
alelievr: Okay so right now we don't have anything to do per-scene config, so i need to add a monobehavior that we attach to a camera so we can configure stuff per-scene

com.unity.testing.graphics-performance/Documentation~/com.unity.testing.graphics-performance.md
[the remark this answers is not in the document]
alelievr: Lewis will review to next week when we finish all the pages

com.unity.testing.graphics-performance/Runtime/PerformanceTests.cs
```diff
@@ -0,0 +1,125 @@
...
+        foreach (var item in data)
+        {
+            string name = String.IsNullOrEmpty(item.name) ? item.GetType().Name : item.name;
+            long currSize = Profiler.GetRuntimeMemorySizeLong(item);
+
+            // There are too many items here so we only keep the one that have a minimun of weight
+            if (currSize > minMemoryReportSize)
+            {
+                results.Add((name, currSize));
+                totalMemory += currSize;
```
FrancescoC-unity: I would argue that we should probably increment the totalMemory regarding if we add the object to the list or not. 

If we have for whatever reasons thousands of just sub-min memory objects, they still add up to something non insignificant (though indeed we don't wnat them separately shown in the report)

com.unity.render-pipelines.high-definition/Editor/BuildProcessors/HDRPPreprocessShaders.cs
alelievr: done

com.unity.testing.graphics-performance/Runtime/PerformanceTests.cs
alelievr: agree, i fixed it

com.unity.testing.graphics-performance/Editor/Common/EditorPerformanceTests.cs
alelievr: Didn't encountered any issue yet with this which is surprising because unity is running and probably using some files in the cache but i can add a try-catch

alelievr: I've add a debug log error in case something wrong happens during the delete

TestProjects/HDRP_PerformanceTests/Assets/PerformanceTests/Runtime/HDRPRuntimePerformanceTests.cs
alelievr: I added this component on the scene, the test setup function will automatically retrieve it and setup the camera settings from the component

image

FrancescoC-unity: Great stuff!

com.unity.testing.graphics-performance/Editor/StaticAnalysis/Common/EditorStaticAnalysisTests.cs
alelievr: @fredericv-unity3d

fredericv-unity3d: I think this should be part of an improvement of the static analysis framework, not in this performance framework.
We will only run the PS4 tests, so the values will be correct.

If we want to run on other platforms, we should probably do the static analysis improvement first. That way we will be able to send the appropriate metrics per platform.

FrancescoC-unity: There are some informations that we'll never get from other platforms. The static analysis can only get whatever the compiler informs us about, so there is no way we get some of those values, I am not sure what improvement of the static analysis will make those appear. 

Also, we'll run PS4 tests _now_, but it'd be shortsighted thinking of running only on that, the "common parts" of the framework shouldn't need updating. If that is the case then the report format should be in platform specific files and not in a "Common" folder. 

The way I see it is that the format of the data sent over should be either platform specific or a susbset that we guarantee is reported for all. 
I would also argue that the one that are missing from other platforms are fairly useless and mostly noise.

FrancescoC-unity: I am fine with doing the filtering on grafana, but this is something to be aware regardless.

PR #6358, opened 2020-03-24
Check if R16G16B16A16_SFloat is supported before using it

hkr:
### Purpose of this PR
Fix regression in URP that broke fallback path when using HDR rendering on a device that does not support it. (case 1230184)

### Testing status
manual, QA


phi-lira:
Oh boy, I merged it and just realized I forgot about changelog. Will add in another PR.

PR #6363, opened 2020-03-25
[8.x.x Backport] Fix QualitySettings changing without user interaction (#6093)

phi-lira:
Backport of #6093


com.unity.render-pipelines.universal/Runtime/ForwardRenderer.cs
```diff
@@ -432,8 +432,6 @@ void SetupBackbufferFormat(int msaaSamples, bool stereo)
 
             if (stereo && msaaSampleCountHasChanged)
                 XR.XRDevice.UpdateEyeTextureMSAASetting();
-#else
-            QualitySettings.antiAliasing = msaaSamples;
 #endif
```
ellioman: Just to be safe. 

In the original PR you also removed setting antiAliasing in the 
`#if ENABLE_VR && ENABLE_VR_MODULE` part but you're not doing that here. Is that intended?

phi-lira: Yes. That introduced an issue in XR, so in master I made a follow up PR to revert the XR change.

phi-lira:
Failures not caused by this PR.

PR #6366, opened 2020-03-25
7.x.x/universal/fix yflip

phi-lira:
Backports:

#5227
#5913


thomas-zeng:
Do we really need to backport this to 19.3?
19.3 has to support legacyXR and y-flip changes are high risk for legacy XR.

PR #6367, opened 2020-03-25
First pass of light loop optimizations (still single threaded)

FrancescoC-unity:
Some profiling guided optimizations here.

**Profiling methodology**
Using mostly dotTrace (after verifying that results are coherent with Unity Profiler). Accumulating at least 3 minutes worth then normalizing cumulative timings to a common baseline to get win amount.

**Profiling scene**
 Very artificial scene with 100 visible lights (mostly point lights, but not only those) and _no shadows_. I will profile shadow case later on, however it is worth noting that shadow cost is additive so whatever was optimized here will carry in all cases.

**Notes on functions targeted**: 
Lightloop is _PrepareLightsForGPU_ , this function is split into two main parts _PrepareGPULightData_ and _PreprocessVisibleLights_ . In this pass I focused mostly on the first part. 

**Results:**

Percentage change for the totality of _PrepareLightsForGPU_ :   **-27%**

Percentage change for _PrepareGPULightData_ :  **-48%** (basically twice as fast) 

Main takeaway here: matrix matrix multiply is _SLOW_ and a lot 😄 

Adding anis for visibility on what he's doing now. 

For QA: 

 - Essentially the testing here is to play a bit with lights and see if there is something off compared to staging.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1462,12 +1462,14 @@ internal int shadowPrecision
         [System.NonSerialized] internal int shadowIndex = -1;
 
         // Runtime datas used to compute light intensity
-        Light m_Light;
+        Light m_Light = null;
```
JulienIgnace-Unity: This should not be needed and is contrary to code convention.
Default for references is always null (yeah, coming from C++ it's hard to NOT initialize with null :) )

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -2188,6 +2215,15 @@ void PrepareGPULightdata(CommandBuffer cmd, HDCamera hdCamera, CullingResults cu
             // The lightLoop is in charge, not the shadow pass.
             // For now we will still apply the maximum of shadow here but we don't apply the sorting by priority + slot allocation yet
 
+            // Determine data that is light independent
+            int viewCount = hdCamera.viewCount;
+            List<Matrix4x4> worldToMatrices = new List<Matrix4x4>(viewCount);
```
JulienIgnace-Unity: This will cause garbage and needs to be cached.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
FrancescoC-unity: Yes :D Feels dirty otherwise, but will remove the line!

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
FrancescoC-unity: Ah yes 👍 
Will make the change

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1467,7 +1467,9 @@ internal Light legacyLight
         {
             get
             {
-                TryGetComponent<Light>(out m_Light);
+                if(m_Light == null)
```
sebastienlagarde: maybe add a comment to say that calling TryGetComponent slow thing and thus we test for null case before? as TryGetComponent corectly handle to return null

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1633,7 +1635,7 @@ internal void EvaluateShadowState(HDCamera hdCamera, in ProcessedLightData proce
             // When creating a new light, at the first frame, there is no AdditionalShadowData so we can't really render shadows
             m_WillRenderShadowMap &= shadowDimmer > 0;
             // If the shadow is too far away, we don't render it
-            m_WillRenderShadowMap &= type == HDLightType.Directional || processedLight.distanceToCamera < shadowFadeDistance;
+            m_WillRenderShadowMap &= processedLight.lightType == HDLightType.Directional || processedLight.distanceToCamera < shadowFadeDistance;
```
sebastienlagarde: is it a fix or an optimization :) ? if it is a fix, we will backport it

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
FrancescoC-unity: TryGetComponent  is slower yeah, will add comment

com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
FrancescoC-unity: Optimization

sebastienlagarde:
Looks like we should backport this PR for 8.x and 7.x, don't think there is conflict and seems to fix issue. Right / wrong ?

FrancescoC-unity:
> Looks like we should backport this PR for 8.x and 7.x, don't think there is conflict and seems to fix issue. Right / wrong ?

This one I think we can safely backport, is a huge win and it is nothing that requires newer engine versions.

FrancescoC-unity:
Also I need a changelog

PR #6368, opened 2020-03-25
Fixed an issue where the wrong instance of the debug setting was used in some places.

JulienIgnace-Unity:
### Purpose of this PR
In some places in the code, the _debugDisplaySettings_ property was accessed instead of _m_CurrentDebugDisplaySettings_. The consequence is that there may have been debug options active when rendering previews or reflection probes. Thankfully, the places where the mistake was could not have been visible anyway.
This PR also removes useless DebugDisplaySettings parameters to GetXXXData functions in the light loop.

@Unity-Technologies/gfx-qa-hdrp the change is pretty trivial but playing a bit with the light filter debug mode could be needed.

---
### Testing status

- [ ] Built a player


sebastienlagarde:
Add a changelog

JulienIgnace-Unity:
> Add a changelog

I did not add a changelog because this is not a user facing change and there is no currently logged public fogbugz about this. So it would just be noise in the changelog.

PR #6371, opened 2020-03-25
Fix custom pass gc alloc

alelievr:
### Purpose of this PR
Fix custom pass gc alloc


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/RenderPass/CustomPass/CustomPassVolume.cs
```diff
@@ -259,19 +258,25 @@ internal static void Cleanup()
...
+            return volumes.FirstOrDefault();
+        }
 
         /// <summary>
         /// Gets the currently active Custom Pass Volume for a given injection point.
         /// </summary>
         /// <param name="injectionPoint">The injection point to get the currently active Custom Pass Volume for.</param>
         /// <returns>Returns the Custom Pass Volume instance associated with the injection point.</returns>
-        public static IEnumerable<CustomPassVolume> GetActivePassVolumes(CustomPassInjectionPoint injectionPoint)
+        public static void GetActivePassVolumes(CustomPassInjectionPoint injectionPoint, List<CustomPassVolume> volumes)
```
Chman: API change, xmldoc needs updating.
The previous method should stay there and be marked as deprecated with a message to let users know they should use the new one, we can't break the code like this anymore, unless this method hasn't been in a released version yet?

alelievr: The previous method was added today by another PR, it's not backported / in master yet so it's fine but i need to update the xml, good catch

sebastienlagarde:
Already discuss with antoine about the false positive breaking change, merging.

PR #6376, opened 2020-03-25
[9.x.x] Fixing defines and calculations for shadow coordinates [Skip CI]

ellioman:
### Purpose of this PR
We recently (7.2.0) updated the defines to use when passing and calculating the shadow coordinates. Those defines weren't updated in all of our shaders and this PR fixes that.

---

### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 

**Yamato**:
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/universal%252Fbugfix%252Fshadow-coords-fixes/.yamato%252Fupm-ci-abv.yml%2523all_project_ci_fast-2020.1/1760457/job


com.unity.render-pipelines.universal/Editor/ShaderGraph/Includes/PBRForwardPass.hlsl
```diff
@@ -13,14 +13,16 @@
...
-    
+
 #else
     inputData.normalWS = input.normalWS;
 #endif
     inputData.normalWS = NormalizeNormalPerPixel(inputData.normalWS);
     inputData.viewDirectionWS = SafeNormalize(input.viewDirectionWS);
 
-#if defined(MAIN_LIGHT_CALCULATE_SHADOWS)
+#if defined(REQUIRES_VERTEX_SHADOW_COORD_INTERPOLATOR)
```
alindmanUnity: Does this `REQUIRES_VERTEX_SHADOW_COORD_INTERPOLATOR` ever need to be defined in a ShaderGraph? I'm not seeing any code that would declare into a final generated shader.

ellioman: This is exactly why I needed a review from you :)

This is used now in all other URP shaders so I figure ShaderGraph should probably be using it as well. But I wasn't sure. If it should be part of the ShaderGraph shaders then I need help to change how we generate code with ShaderGraph to include it.

We create these defines like this in Shadows.hlsl
```
#if !defined(_RECEIVE_SHADOWS_OFF)
    #if defined(_MAIN_LIGHT_SHADOWS)
        #define MAIN_LIGHT_CALCULATE_SHADOWS

        #if !defined(_MAIN_LIGHT_SHADOWS_CASCADE)
            #define REQUIRES_VERTEX_SHADOW_COORD_INTERPOLATOR
        #endif
    #endif

    #if defined(_ADDITIONAL_LIGHT_SHADOWS)
        #define ADDITIONAL_LIGHT_CALCULATE_SHADOWS
    #endif
#endif

#if defined(_ADDITIONAL_LIGHTS) || defined(_MAIN_LIGHT_SHADOWS_CASCADE)
    #define REQUIRES_WORLD_SPACE_POS_INTERPOLATOR
#endif
```

ellioman: But if it is not needed in ShaderGraph then I'll simply remove it from Varyings.hlsl and PBRForwardPass.hlsl.

alindmanUnity: Hmm, looking at that, seems like those defines should just work TM because of some coincidences in how the defines are being set. I'd be curious if it's possible to run into a configuration where, because of Shadows.hlsl, `REQUIRES_VERTEX_SHADOW_COORD_INTERPOLATOR` gets defined but the `shadowCoord` attribute of the struct is not, which would produce an error state. In current master, that error state is pretty specific to get into (trying to do custom shadow calculations from an unlit shader graph, which never declares shadowCoord in generated structs). 
It's probably fine as it is now, but @LandonTownsendUnity if you could use your copy of the UniversalGraphicsTests that uses all Shader Graph shaders to check any shadow-related tests aren't failing, that would be awesome.

ellioman: @alindmanUnity Landon has tested and approved. Are we good with these changes from the ShaderGraph side?

alindmanUnity: yep, with no regressions this should be fine as is

PR #6381, opened 2020-03-26
Revert RenderQueue '+50' offset on all URP shaders

Verasl:
### Purpose of this PR
Reverts the offset we applied to all shaders in URP for the priority slider.
Added a material upgrader to make sure all materials that did have a priority slider are upgraded with their correct render queue.

Fixes:
https://fogbugz.unity3d.com/f/cases/1219877/
https://fogbugz.unity3d.com/f/cases/1222762/

---
### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.universal/Editor/ShaderGUI/BaseShaderGUI.cs
```diff
@@ -425,63 +428,68 @@ public static void SetupMaterialBlendMode(Material material)
 
             var queueOffset = 0; // queueOffsetRange;
             if(material.HasProperty("_QueueOffset"))
-                queueOffset = queueOffsetRange - (int) material.GetFloat("_QueueOffset");
```
phi-lira: should also delete the variable `queueOffsetRange`?

Verasl: yes we should, will update

Verasl:
> I'm not sure I understand how we fix the offset to be 0 for existing materials in the upgrader.

We simple are not adding the offset when we validate the keywords, so in the upgrader we just call the update keywords, it is the same function that gets called when a user simply looks at the material inspector, which reverts the +50 offset.
This wont change their priority level, as this was stored before, so if they had a material with priority 10 = render queue 2060 it will now be render queue 2010.

PR #6397, opened 2020-03-26
Avoid negative input to sqrt()

ChristianF-ARM:
This fixes a precision bug in ACES.hlsl resulting in rendering artifacts on Mali drivers newer than Bifrost/Valhall R21.

Before: 
stock

After: fix

Summary:
- The calculation of `k` is not numerically robust when done in FP16 and might generate a negative value.
- This can be observed as rendering artifacts on Mali drivers newer than Bifrost/Valhall R21.
- This commit fixes this by by clamping `k` to [0, inf] before the `sqrt()`.
- On Mali-G77 the clamp compiles down to a free output modifier on the previous instruction, so there's no performance impact.


sebastienlagarde:
switch to hdrp/Staging to be able to run Yamato before merging in master.

PR #6401, opened 2020-03-26
- Avoid building the mip chain a second time for SSR for transparent objects.

anisunity:
Currently for transparentSSR, we are rebuilding the mip-chain which is expensive (and not useful). By using the post transparentDepthPrepass depth buffer for building the ray and the previous mip chain (without transparents) to ray marche we get a better result and cheaper.

The transparent SSR test is still green. Transparent SSR was not working for render graph, so I did not make it work.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
@@ -241,6 +241,7 @@ namespace UnityEngine.Rendering.HighDefinition
         class RenderSSRPassData
         {
             public RenderSSRParameters parameters;
+            public TextureHandle depthBuffer;
             public TextureHandle depthPyramid;
             public TextureHandle colorPyramid;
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.LightLoop.cs
```diff
@@ -282,7 +284,8 @@ class RenderSSRPassData
 
                     var colorPyramid = renderGraph.ImportTexture(hdCamera.GetPreviousFrameRT((int)HDCameraFrameHistoryType.ColorBufferMipChain));
 
-                    passData.parameters = PrepareSSRParameters(hdCamera);
+                    passData.parameters = PrepareSSRParameters(hdCamera, true);
+                    passData.depthBuffer = builder.ReadTexture(depthBuffer);
```
JulienIgnace-Unity, line 288: This should be tested with parameters.transparentSSR.
Otherwise you tell the system that the texture is actually used even though it's not, making the resource's lifetime extend wrongly.

PR #6402, opened 2020-03-26
Sg/export dependencies

elizabeth-legros:
### **Please read**
**PR workflow guidelines**

### Checklist for PR maker

---
### Purpose of this PR
When exporting a shadergraph or subgraph with dependencies, the depended assets were not grabbed correctly. 
---
### Testing status

- Other: 
Reimported all files in the test project and tried exporting a few shadergraphs

---
### Comments to reviewers

### DIFF
--- a/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
+++ b/com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
@@ -17,12 +17,6 @@ namespace UnityEditor.ShaderGraph
     [ScriptedImporter(32, Extension, 3)]
     class ShaderGraphImporter : ScriptedImporter
     {
-        [Serializable]
-        private class ShaderGraphExportDependencies : ScriptableObject
-        {
-            public List<UnityEngine.Object> assetDependencies = new List<UnityEngine.Object>();
-        }
-
         public const string Extension = "shadergraph";
 
         public const string k_ErrorShader = @"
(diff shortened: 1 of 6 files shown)


LandonTownsendUnity:
Two potential concerns:
1. If the dependency is part of a package (such as VFX graph) it will not be added as a dependency.
2. Diffusion profiles attached to the HD Lit master node when subsurface scattering is enabled will not be added as dependencies.

Any opinions on whether these should be fixed / whether they can be fixed?

elizabeth-legros:
The VFX graph one I do not see any sort of nice solution to, without modifying their code directly. If anything includes a shadergraph, it should include all the shadergraph dependencies but we cant make other things that use a shadergraph include the shadergraph through any meaningful system
As for diffusion profiles, that gets a bit trickier but I would tend to log that as a separate bug since this grabs all asset dependencies, so if hdrp assets are not included in export then that shadergraph doesnt seem them as a dependent asset. That said, make sure you are re-importing shadergraphs and subgraphs and the like to be sure that things are set up correctly

elizabeth-legros:
https://fogbugz.unity3d.com/f/cases/1210945#BugEvent.1072132109

com.unity.shadergraph/Editor/Importers/ShaderGraphImporter.cs
```diff
@@ -135,6 +135,13 @@ public override void OnImportAsset(AssetImportContext ctx)
             {
                 metadata.outputNodeTypeName = graph.outputNode.GetType().FullName;
             }
+            metadata.assetDependencies = new List<UnityEngine.Object>();
+            var deps = GatherDependenciesFromSourceFile(ctx.assetPath);
+            foreach(string dependency in deps)
```
zadi0648: I meant to include this super nit pick in my review; but as a team it would be good to be consistent on spaces before parenthesis as I've noticed it varying a lot in our code base (this `foreach(string dependency in deps)` has no space but the one below does)

elizabeth-legros: Woof I mean part of that depends on IDE formatting preferences and we dont all use the same IDEs either. I can make the change here but if thats something we want to have standard then we really need a common set of formatting preferences

LandonTownsendUnity:
I think there was a misunderstanding about my previous comment.

What I meant was, if you, for instance, use a texture called "Noise" which is a texture that is a part of the VFX Graph package, and you reference it in the shadergraph and then export package, that texture will not be added to the exported package as a dependency.

However, it looks like this is a problem with the way Unity tracks dependencies in general, not with this PR. So there's not much you can do about it.

PR #6403, opened 2020-03-26
[Shader Graph] Update Error Messaging for Sub Graph Previews

alindmanUnity:
### Purpose of this PR
The preview renderer cannot handle certain data types implicitly. The Sub Graph output node allows these data types as output ports, but enters an error state if the first port (used for rendering previews) is an invalid data type. Additionally, the existing validation for empty ports actually hit a null ref before presenting the error message. 
This fixes the existing validation for empty port lists and display the correct error message. 
This also adds an extra validation path for invalid slot types. If the preview renderer is failing because of an incompatible data type, the node will validate and tell the user why. Additionally, docs have been updated to reflect this preview state and explicitly state supported data types.
It is expected that both of these states make the sub graph asset invalid. If the user tries to save while in this invalid state, errors will log in the console. If the user tries to use the subgraph in a main graph while in this state, the referenced Sub Graph node will throw an Invalid Asset error. 

[internal bug link only](https://fogbugz.unity3d.com/f/cases/1205973/)

---
### Testing status

- [x] Built a player
- Other: 
- Created a new sub graph asset. Put the slot list into several states of error (all invalid slot types as first input, no inputs at all)
- Checked the asset in main graphs while in error states to confirm "Invalid Asset" error is shown. 
- Put the graph back in a valid state and observed errors clearing in both the sub graph asset and any referenced graphs. 

subGraphErrors

---
### Comments to reviewers
Note to @Unity-Technologies/gfx-qa-bellevue , would you like this backported?


alindmanUnity:
@LandonTownsendUnity confirm at 1700bd8508 for adding boolean c:

com.unity.shadergraph/Editor/Data/SubGraph/SubGraphOutputNode.cs
```diff
@@ -62,17 +74,31 @@ void ValidateSlotName()
...
-            if (!this.GetInputSlots<MaterialSlot>().Any())
+            if (!slots.Any())
+            {
+                owner.AddValidationError(guid, s_MissingOutputSlot, ShaderCompilerMessageSeverity.Error);
+            }
+            else if (!s_ValidSlotTypes.Contains(slots.FirstOrDefault().concreteValueType))
             {
-                owner.AddValidationError(guid, s_MissingOutputSlot, ShaderCompilerMessageSeverity.Warning);
+                IsFirstSlotValid = false;
+                owner.AddValidationError(guid, "Preview can only compile if the first output slot is a Vector or Matrix type. Please adjust slot types.", ShaderCompilerMessageSeverity.Error);
```
LandonTownsendUnity: This line still needs to be updated with boolean.

alindmanUnity: confirm at 
f762348

PR #6411, opened 2020-03-27
backport #6388 a null check in Analytics file

martint-unity:
Added a null check for the renderer feature. #6388


ellioman:
Is the analytics also in 8.x.x and 7.x.x? Then we should add the backport labels...

PR #6413, opened 2020-03-27
Fix URP Unlit, SimpleLit and BakedLit to use float colors with Hybrid V2

JussiKnuuttila:
URP shaders are using half for colors, but half is not implemented in V2 yet. V2 variants of URP shaders will use float colors until half is fully implemented.

---
### Purpose of this PR
Without this PR, some URP shaders will not work with Hybrid V2.

---
### Testing status

- [ ] Built a player
- Other: 

HybridURPSamples in dots repo


JussiKnuuttila:
> LGTM. Do we have a plan when to support half?

I think we're trying to get it in by 2020.2. @sebbbi ?

sebbbi:
> LGTM. Do we have a plan when to support half?

ETA = before 2020.2 ships. This includes half storage in ComputeBuffer. Burst team just added F16C intrinsic codegen, so we are hoping to exploit that on CPU side.

JussiKnuuttila:
@sebbbi @phi-lira I think someone authorized must merge this?

PR #6414, opened 2020-03-27
Corrected typo and changed links to .md

JordanL8:
### Purpose of this PR
Corrects a typo and changes the link format on the fabric master node page.


sebastienlagarde:
@JordanL8 : At last publishing, we have @remi-chapelain that have revet the various .md to .html as it wasn't working. What should be the right way for extention? Maybe we should not put any extension ?

com.unity.render-pipelines.high-definition/Documentation~/Master-Node-Fabric.md
```diff
@@ -73,7 +73,7 @@ To view these properties, click the **Cog** in the top right of the Master Node.
...
 | **Energy Conserving Specular** | Enable the checkbox to make HDRP reduce the diffuse color of the Material if the specular effect is more intense. This makes the lighting of the Material more consistent, which makes the Material look more physically accurate. |
-| **Material Type**              | Specifies the type of Material the fabric Shader represents. The options are:<br/>&#8226; **Cotton Wool**: A  variant inspired by the Imageworks’s cloth model.<br/>&#8226; **Silk**: A fabric variant based on an anisotropic DinseyGGX BRDF Model. |
+| **Material Type**              | Specifies the type of Material the fabric Shader represents. The options are:<br/>&#8226; **Cotton Wool**: A  variant inspired by the Imageworks’s cloth model.<br/>&#8226; **Silk**: A fabric variant based on an anisotropic DisneyGGX BRDF Model. |
```
sebastienlagarde: repalce DisneyGGX  by GGX

JordanL8: I had a chat with Remi and most of the issues found were from link extensions not being present. Both html and md links work fine now because the doctools convert the extension for links to local md files to .html (URP and LWRP mainly use .md extension).

Using md for internal package links has one advantage as you can follow the links locally on your PC. That's why I've started to convert the links to .md when I touch a page. However, I don't mind which we use as they both work when the package docs are built.

PR #6203, opened 2020-03-16
[Draft] Stack Master Node

Kink3d:
### Purpose of this PR
**Under Construction**

#### Checklist
- [ ] Update changelogs
- [ ] Update documentation
- [ ] Add tests

---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in devloper mode, you have a button at end of resources that check the pathes)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

---
### Comments to reviewers
**Under Construction**

PR #6419, opened 2020-03-27
[Shader Graph] Target API Inversion Refactor

Kink3d:
### Checklist
- [ ] New graphics tests?
- [ ] Changelog?

---

### Purpose of this PR
This PR refactors the Shader Graph Target framework as a first step to solving the following problems:
- `Target` layer has become redundant.
- HDRP requires "plug and play" material types.
- Render Pipelines require a way to share serialised data between different material types.

NOTE: This PR is a prerequisite to #6203. It should, in itself, not change any behaviour inside the graph or with the output shader. It does, however, mean the following PR will be more manageable. 

We do this by:
Previously `Target` represented a Unity rendering feature (Mesh/Decal/VFX) and `TargetImplementation` represented a Render Pipeline's implementation of that Target (although this became somewhat muddied by HD Raytracing and Hyrbid V2 being implemented as separate `TargetImplementation`). The top layer became redundant as pipeline agnostic implementations are being avoided at all costs.

In this PR, the `Target` layer represents a Render Pipeline and a `Target` may choose to implement any number of `SubTarget` types. These, at least for now, are 1:1 mirrors of the master nodes (per pipeline). The `SubTarget` usage is not enforced and should be manually decided by the `Target`.

This also removes the redundant `TargetImplementation` layer when a `Target` is actually pipeline agnostic (`PreviewTarget` and `VFXTarget`). `VFXTarget` will later be removed and re-implemented in a different way. Forward questions about this to #project-shadergraph.

Notes:
- This PR does not reintroduce the ability to strip material types in HDRP. This is because we still require a way to lookup a `SubTarget` based on the master node. This lookup table will be removed by #6203 (as the active `SubTarget` will be selected by the user and not derived) and then `SubTarget` types will be strippable.
- This PR assumes that both render pipeline teams would prefer maximum code stripping when stripping `SubTarget` types. This means all code (keywords, pragmas, render states etc) that is only referenced by a single `SubTarget` type is local to that `SubTarget`. If you wish to move any of this to your pipeline's `CoreX` classes then please comment.
- All "core" shader pass code (keywords, pragmas, render states etc) are found within root classes inside each pipeline's `Target` file. This has resulted in fairly large files (especially `HDTarget`). If you would like these moved please comment.
- HD Raytracing and URP Hyrbid V2 implementations have been merged back into their matching `SubTarget` types (HDUnlit, HDLit, HDFabric and UniversalLit, UniversalUnlit respectively). This was achieved by allowing each `Target` and/or `SubTarget` to provide multiple subshaders.
- Per-target level overrides for pass templates have been removed. This avoids added complications with the new, manual approach, to `SubTarget` implementations. Pass templates must now be assigned per-pass.

---

### Testing status
What I've done:
- Checked all master nodes on all available pipelines. Tested a wide range of master node settings (but not exhaustive).
- Checked most configurations of geometry input nodes (but not all)
- Tested with each pipeline individually, as well as both pipelines in the same project.

I'm hoping the CI will cover the rest...

Recommend QA tests as many master node and geometry input node combinations as possible.

Nothing. This should be covered?

---
### Comments to reviewers
See Notes above.


LandonTownsendUnity:
Pretty much every node in Universal RP has an error on it. The shader still compiles with them but the node previews don't work.
image

PR #6380, opened 2020-03-26
Platform/playstation

timcannellunity3d:
### Purpose of this PR
Use playstation keyword insted of ps4 for shader

For shader graph team there is a change in shader graph file here:
com.unity.shadergraph/Editor/Generation/Enumerations/Platform.cs

Verified in generated code that it work correctly.
---
### Testing status

**Manual Tests**: What did you do?
- [ ] Opened test project + Run graphic tests locally
- [ ] Built a player
- [ ] Checked new UI names with UX convention
- [ ] Tested UI multi-edition + Undo/Redo + Prefab overrides + Alignment in Preset
- [ ] C# and shader warnings (supress shader cache to see them)
- [ ] Checked new resources path for the reloader (in developer mode, you have a button at end of resources that check the paths)
- Other: 

**Automated Tests**: What did you setup? (Add a screenshot or the reference image of the test please)

**Yamato**: (Select your branch):
https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline

Any test projects to go with this to help reviewers?

---
### Comments to reviewers
Notes for the reviewers you have assigned.

PR #6428, opened 2020-03-27
Lighting/probevolumes [skip CI]

thefranke:
---
### Purpose of this PR
This PR introduces ProbeVolumes, a new type of volume handled in the LightLoop, which encodes regular 3d grids of Light Probes. ProbeVolumes are used to add GI by covering a part of the scene with a volume and baking it. Multiple volumes can be added to a scene, as each volume manages its data individually. Volumes that overlap are blended in special blending regions. On top of that, a blending mode can be selected for each volume.

---
### Testing status

A test project can be found here:
https://drive.google.com/file/d/1rSf5clQQD_458gRwanhxwiSwDwQRhPHr/view?usp=sharing

The scene contains a temple ruin. The floor uses Lightmaps for GI, whereas the rest is set to Light Probes, but all objects are marked as Contribute GI. Because the current branch does not have layer support for baking ProbeVolumes, the way this scene was prepared was by enabling and disabling groups of objects/lights and then baking them independently via the Bake button for individual volumes.

The first volume called "Ruins Volume" is set to "Normal" blending mode, which means it only adds GI to those parts of the scene that are not lightmapped. The other two volumes called "Emissive volume" where baked using the scene with the emissive object inside each group and only contain a volume with the emissive light baked into the probe. Both of these are set to "Additive" blending mode, which means their GI portion gets added to the scene everywhere, even the lightmapped parts. These latter two can be turned on and off or faded in and out without breaking the overall result, as they simply add additional light.

In the project settings you will also find a volume override for Probe Volumes called "Probe Volume Controller". This section contains special filter modes for the probes to avoid leaks. ProbeVolumes are already baked with validity filtering included (meaning probes inside of geometry are filled with light from neighboring probes), which will suppress invalid/dark shadow splotches leaking into the scene. Further leak mitigation options of light through objects are available in the Probe Volume Controller.

ProbeVolume baking can be done either via the standard "Generate Lighting" button in the Lighting window or - for individual volumes - via the Bake button in the baking section.

**Manual Tests**: 
- [x] Built a player
- Other: 

**Automated Tests**:
There are currently no automated tests.

---
### Comments to reviewers

This PR introduces the first batch of code, but disabled. To enable ProbeVolumes, they have to be enabled in the HDRP configuration package, the HDRP pipeline asset and per camera.


com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
```diff
@@ -24,7 +24,7 @@
 
 #endif
 
-#if (SHADERPASS == SHADERPASS_VOLUMETRIC_LIGHTING || SHADERPASS == SHADERPASS_VOLUME_VOXELIZATION)
+#if (SHADERPASS == SHADERPASS_VOLUMETRIC_LIGHTING || SHADERPASS == SHADERPASS_VOLUME_VOXELIZATION || SHADERPASS == SHADERPASS_GBUFFER)
```
sebastienlagarde: >|| SHADERPASS == SHADERPASS_GBUFFER)
Why do we have this definition here? not expected, it present users from configuring the shadow mode in GBuffer (+ now we are adding a way to configure shadow in GBuffer)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
```diff
@@ -410,6 +421,17 @@ void LightLoop( float3 V, PositionInputs posInput, PreLightData preLightData, BS
     }
 #endif
 
+#if defined(SHADEROPTIONS_PROBE_VOLUMES_EVALUATION_MODE)
+#if SHADEROPTIONS_PROBE_VOLUMES_EVALUATION_MODE == PROBEVOLUMESEVALUATIONMODES_LIGHT_LOOP
+    if (featureFlags & LIGHTFEATUREFLAGS_PROBE_VOLUME)
+    {
+        float3 probeVolumeDiffuseLighting = EvaluateProbeVolumesLightLoop(posInput, bsdfData, builtinData);
+        builtinData.bakeDiffuseLighting = probeVolumeDiffuseLighting;
+        builtinData.backBakeDiffuseLighting = float3(0, 0, 0);
```
sebastienlagarde: Just as a note but you are aware of this : this code don't apply code happening in ModifyBakedDiffuseLighting() (i.e it don't handle transmission nor preLightData.diffuseFGD). This can have impact for foliage for example.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeAtlasOctahedralDepthConvolve.compute
```diff
@@ -0,0 +1,106 @@
+#pragma kernel ProbeVolumeAtlasOctahedralDepthConvolveKernel PROBE_VOLUME_ATLAS_OCTAHEDRAL_DEPTH_CONVOLVE_KERNEL=ProbeVolumeAtlasOctahedralDepthConvolveKernel
+
+#ifdef SHADER_API_PSSL
+#   pragma argument( scheduler=minpressure ) // instruct the shader compiler to prefer minimizing vgpr usage
+#endif
+
+#pragma only_renderers d3d11 ps4 xboxone vulkan metal switch
```
sebastienlagarde: note for myself: we need to convet all PS4 keyword by playstation, once this PR is merge: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6380

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeLighting.cs
```diff
@@ -0,0 +1,814 @@
...
+            cmd.SetGlobalInt(HDShaderIDs._ProbeVolumeCount, 0);
+            cmd.SetGlobalTexture(HDShaderIDs._ProbeVolumeAtlasSH, TextureXR.GetBlackTexture());
+            cmd.SetGlobalTexture(HDShaderIDs._ProbeVolumeAtlasOctahedralDepth, TextureXR.GetBlackTexture());
+            cmd.SetGlobalInt(HDShaderIDs._ProbeVolumeLeakMitigationMode, (int)LeakMitigationMode.NormalBias);
+            cmd.SetGlobalFloat(HDShaderIDs._ProbeVolumeNormalBiasWS, 0.0f);
+            cmd.SetGlobalFloat(HDShaderIDs._ProbeVolumeBilateralFilterWeightMin, 0.0f);
+            cmd.SetGlobalFloat(HDShaderIDs._ProbeVolumeBilateralFilterWeight, 0.0f);
+        }
+
+        public void ReleaseProbeVolumeFromAtlas(ProbeVolume volume)
```
sebastienlagarde: Be sure everything is internal , we shouldn't have anything public so we can keep changing API

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeLighting.cs.hlsl
```diff
@@ -0,0 +1,124 @@
...
+    float4 octahedralDepthScaleBias;
+    float3 resolution;
+    uint lightLayers;
+    float3 resolutionInverse;
+    float unused;
+};
+
+// Generated from UnityEngine.Rendering.HighDefinition.SphericalHarmonicsL1
+// PackingRules = Exact
+struct SphericalHarmonicsL1
```
sebastienlagarde: this kind of struct should be in a function with other SH in Core/shaders

com.unity.render-pipelines.high-definition/Runtime/Material/Eye/Eye.hlsl
```diff
@@ -817,6 +817,16 @@ IndirectLighting EvaluateBSDF_Env(  LightLoopContext lightLoopContext,
...
 }
 
+//-----------------------------------------------------------------------------
+// EvaluateBSDF_LightProbeL1 for ProbeVolumes
+// ----------------------------------------------------------------------------
+
+float3 EvaluateBSDF_LightProbeL1(BuiltinData builtinData, BSDFData bsdfData,
+                                 float4 shAr, float4 shAg, float4 shAb)
+{
+    return ShadeSurface_LightProbeL1(builtinData, bsdfData, shAr, shAg, shAb) * GetAmbientOcclusionForMicroShadowing(bsdfData);
```
sebastienlagarde: You don't need to call GetAmbientOcclusionForMicroShadowing() in all functions of material, this is a function define for all our material, so it can be call outside the function

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDProfileId.cs
```diff
@@ -189,5 +189,9 @@ internal enum HDProfileId
         CustomPostProcessAfterPP,
         CustomPostProcessAfterOpaqueAndSky,
         ContrastAdaptiveSharpen,
+		PrepareProbeVolumeList,
```
sebastienlagarde: indent

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -2021,6 +2030,40 @@ AOVRequestData aovRequest
 
             RenderRayTracingPrepass(cullingResults, hdCamera, renderContext, cmd, false);
 
+            // When evaluating probe volumes in material pass, we build a custom probe volume light list.
+            // When evaluating probe volumes in light loop, probe volumes are folded into the standard light loop data.
+            // Build probe volumes light list async during depth prepass.
+            // TODO: (Nick): Take a look carefully at data dependancies - could this be moved even earlier? Directly after PrepareVisibleProbeVolumeList?
+            // The probe volume light lists do not depend on any of the framebuffer RTs being cleared - do they depend on anything in PushGlobalParams()?
+            // Do they depend on hdCamera.xr.StartSinglePass()?
```
sebastienlagarde: you must test more if the system is disabled. We shouldn't have code execute if volumeti clightprobe are disable.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
```diff
@@ -101,6 +101,8 @@ internal static RenderPipelineSettings NewDefault() => new RenderPipelineSetting
             lightLayerName5 = "Light Layer 5",
             lightLayerName6 = "Light Layer 6",
             lightLayerName7 = "Light Layer 7",
+			supportProbeVolume = false,
```
sebastienlagarde: indent

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
```diff
@@ -185,10 +187,14 @@ public struct LightSettings
         public bool supportDitheringCrossFade;
         /// <summary>Support terrain holes.</summary>
         public bool supportTerrainHole;
-        /// <summary>Support ray tracing.</summary>
+        /// <summary>Support Probe Volumes.</summary>
+        public bool supportProbeVolume;
+		/// <summary>Support ray tracing.</summary>
```
sebastienlagarde: indent

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/Texture2DAtlas.cs
```diff
@@ -134,6 +134,7 @@ public void Reset()
     class Texture2DAtlas
     {
         protected RTHandle m_AtlasTexture = null;
+		protected bool m_IsAtlasTextureOwner = false;
```
sebastienlagarde: intent

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
```diff
@@ -267,12 +267,11 @@ internal void RenderMirrorView(CommandBuffer cmd)
...
                 cmd.SetRenderTarget(BuiltinRenderTextureType.CameraTarget);
 
-                int mirrorBlitMode = display.GetPreferredMirrorBlitMode();
-                if (display.GetMirrorViewBlitDesc(null, out var blitDesc, mirrorBlitMode))
+                if (display.GetMirrorViewBlitDesc(null, out var blitDesc))
                 {
                     if (blitDesc.nativeBlitAvailable)
                     {
-                        display.AddGraphicsThreadMirrorViewBlit(cmd, blitDesc.nativeBlitInvalidStates, mirrorBlitMode);
+                        display.AddGraphicsThreadMirrorViewBlit(cmd, blitDesc.nativeBlitInvalidStates);
```
sebastienlagarde: ?

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
sebastienlagarde: ok, I see that this code must happen in EvaluateBSDF_LightProbeL1() (just not done yet).

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
pastasfuture: This can be removed - it was from the change set that required getting the regular light loop code working in the GBuffer - before we committed to treating the probe volume light list as it's own data structure with it's own accessors.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
pastasfuture: looks like buildProbeVolumeLightListTask should just be constructed inside of the if case at 2053 to handle this. Good catch.

com.unity.render-pipelines.high-definition/Runtime/Lighting/Shadow/HDShadowAlgorithms.hlsl
pastasfuture: Just committed this removal.

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeLighting.cs.hlsl
pastasfuture: I'm not seeing it in core, but are you referring to the file in: high-definition/Runtime/Lighting/SphericalHarmonics.cs

It contains the struct definition for ZonalHarmonicsL2, and the static class SphericalHarmonicMath

I'll move it to there for now. It seems like the right place. Though I think this whole file should be moved to Core at some point since Core will care about these types as well.

sebastienlagarde: The files don't exist in core, this is what I suggest. Move SH function into Core/shaders/  name of a file .hlsls, where name of a file you can chose, coulb be shpericalharmonic.hlsl
and yes, do the same for C# side.

pastasfuture: I just committed this cleanup

com.unity.render-pipelines.high-definition/Runtime/Material/Eye/Eye.hlsl
pastasfuture: I can go through and remove these calls to GetAmbientOcclusionForMicroShadowing() but I want to confirm my understanding of the problem: This function is only used for the lightloop evaluation mode. In that mode, and in deferred, we do not actually have access to the bsdfData.ambientOcclusion (it is always set to 1.0). So this call is pointless. But in forward, this value should contain real data.

Is the request that we should remove this completely, for consistency across deferred and forward, or is the request that we should move the GetAmbientOcclusionForMicroShadowing() call into the call site of EvaluateBSDF_LightProbeL1() and do the multiplication there?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
pastasfuture: I can move buildProbeVolumeLightListTask = new HDGPUAsyncTask("Build probe volume light list", ComputeQueueType.Background); inside of the conditional at line 2058. However, the declaration needs to be outside of these conditionals, since it is accessed at the sync point on line 2097.

I don't feel like this is a great change however, because it means that buildProbeVolumeLightListTask will sit around as an uninitialized struct, polluting the scope. I would rather have it initialized to the correct value, and not used in the case where probe volumes is disabled. And given that HDGPUAsyncTask is just a single struct, it doesn't seem like we should be worried about optimizing this initial value assignment out. I could move all of this code into the LightLoop file, and buildProbeVolumeLightListTask as a member variable there, but it would functionally be identical, and would diverge from the pattern for other HDGPUAsyncTasks in HDRenderPipeline.cs

the asyncParams = new HDGPUAsyncTaskParams declaration on line 2042 needs to stay outside, as it's the same struct that's also used for the other async graphics tasks (just declared earlier so probe volumes can use it).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDProfileId.cs
pastasfuture: just pushed up the fix.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
pastasfuture: just pushed up the fix.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/RenderPipelineSettings.cs
pastasfuture: just pushed up the fix.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/XR/XRSystem.cs
pastasfuture: Just pushed the fix. Thanks for catching this. It was introduced by accident during a merge.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Utility/Texture2DAtlas.cs
pastasfuture: just pushed the fix.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
pastasfuture: Yeah - can take a look building out EvaluateBSDF_LightProbeL1() a little more for the lightloop evaluation mode (especially so that when profiling we are comparing the same thing).

Thinking maybe I should add a comment for now inside that function, and then create a follow up PR for this work (shouldn't be a lot - I just don't want to hold up the PR too much) - what do you think?

com.unity.render-pipelines.high-definition/Runtime/Lighting/ProbeVolume/ProbeVolumeLighting.cs
pastasfuture: I just pushed up this fix. Went through the change set and ensured everything was internal or private.

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
```diff
@@ -3209,6 +3586,12 @@ void UpdateDataBuffers()
             // These two buffers have been set in Rebuild(). At this point, view 0 contains combined data from all views
             m_TileAndClusterData.convexBoundsBuffer.SetData(m_lightList.lightsPerView[0].bounds);
             m_TileAndClusterData.lightVolumeDataBuffer.SetData(m_lightList.lightsPerView[0].lightVolumes);
+
+            if (ShaderConfig.s_ProbeVolumesEvaluationMode == ProbeVolumesEvaluationModes.MaterialPass)
+            {
+                m_TileAndClusterData.probeVolumesConvexBoundsBuffer.SetData(m_lightList.lightsPerView[0].probeVolumesBounds);
+                m_TileAndClusterData.probeVolumesLightVolumeDataBuffer.SetData(m_lightList.lightsPerView[0].probeVolumesLightVolumes);
```
sebastienlagarde: indent

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
sebastienlagarde: yes. let the code as it is now, we will update this later. 
Just one thing, can you enclose all definition of EvaluateBSDF_LightProbeL1 with the defiend 
SHADEROPTIONS_PROBE_VOLUMES_EVALUATION_MODE == PROBEVOLUMESEVALUATIONMODES_LIGHT_LOOP ?
Otherwise nayone that have a custom shader will fail to compile, and I want to be sure about our final solution before moving to a new API.

com.unity.render-pipelines.high-definition/Runtime/Material/Eye/Eye.hlsl
sebastienlagarde: Oh sorry, I have misunderstand what was happening here, you have currently use GetAmbientOcclusionForMicroShadowing in an unexpected way :), you use it to get the baked AO in case you are in forward :). 
This is not how we should do it. 
Right now AO is already apply (in forward)and without double occlusion with SSAO.
See code in PostEvaluateBSDF():
  GetScreenSpaceAmbientOcclusion(posInput.positionSS, preLightData.NdotV, bsdfData.perceptualRoughness, bsdfData.ambientOcclusion, bsdfData.specularOcclusion, aoFactor);
    ApplyAmbientOcclusionFactor(aoFactor, builtinData, lighting);

so just remove the " * GetAmbientOcclusionForMicroShadowing(bsdfData)", it is counter productive :)

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
sebastienlagarde: ok, let's not move it :).

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.hlsl
pastasfuture: Thanks - I just pushed this fix.

com.unity.render-pipelines.high-definition/Runtime/Material/Eye/Eye.hlsl
pastasfuture: Thanks - I just pushed this fix. Deleted * GetAmbientOcclusionForMicroShadowing(bsdfData) from inside of all implementations of EvaluateBSDF_LightProbeL1().

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
pastasfuture: Thanks :)

com.unity.render-pipelines.high-definition/Runtime/Lighting/LightLoop/LightLoop.cs
pastasfuture: Sorry, I'm not actually seeing the indentation issue here. Looks like the correct 4x space indentation:

image

PR #6431, opened 2020-03-27
Hdrp /fix framesettingsoverridemask api

RSlysz:
### Purpose of this PR
Currently the public API make us unable to expend the FrameSettings size if we run out of space.
The modification in this PR deprecate the access to the visible element BitArray128. It introduce two new accessor for user to use instead:
    public struct FrameSettingsOverrideMask
    {
        public bool GetBitFor(FrameSettingsField field);
        public void SetBitFor(FrameSettingsField field, bool value);
    }
This also add utility to directly compare struct FrameSettingsOverrideMask to another one.
And finally, the IBitArray utility is completed to be able to perform the masking operation when aggregating regardless the king of BitArray used. Sadly we cannot yet add operator in Interface (will be available in C#8 so we must add the implementation for each case inside all the BitArrays.

Current project will have an warning regarding that they use an obsolete field. So this is only for HDRP/staging for now.
When the obsolete periode is over (for 9.x.x ?) we should only change the visibility of the FrameSettingsOverrideMask to be private, and remove the pragma removing the warning inside the FrameSettingsOverrideMask struct.

---
### Testing status

- [ ] Built a player
- Other: 
     - regarding the change, either they totally broke the pipeline, or it will works

---
### Comments to reviewers
Note: Not sure it is at no risk regarding the performance though.

### DIFF
--- a/com.unity.render-pipelines.core/Runtime/Utilities/BitArray.cs
+++ b/com.unity.render-pipelines.core/Runtime/Utilities/BitArray.cs
@@ -41,6 +41,12 @@ namespace UnityEngine.Rendering
         /// </summary>
         /// <returns></returns>
         IBitArray BitNot();
+
+        /// <summary>
+        /// Affectation utility to deal with underneath type
+        /// </summary>
+        /// <param name="other">the IBitArray to copy content into this one. Must have same size.</param>
+        void AssignFrom(IBitArray other);
     }
 
     // /!\ Important for serialization:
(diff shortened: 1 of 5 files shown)


com.unity.render-pipelines.core/Runtime/Utilities/BitArray.cs
```diff
@@ -163,6 +169,19 @@ public BitArray8(IEnumerable<uint> bitIndexTrue)
...
+
+        /// <summary>
+        /// Helper to convert from IBitArray 
+        /// </summary>
+        /// <param name="other">The IBitArray to copy data from</param>
+        /// <returns></returns>
+        public void AssignFrom(IBitArray other)
+        {
+            if (!(other.capacity == 8u))
+                throw new InvalidCastException();
```
fredericv-unity3d: I think this should be an `ArgumentException` with a message explaining the error

fredericv-unity3d:
For a future work, we may need to rewrite the interface to avoid the use of `IBitArray` in parameter and return type to avoid boxing with value types.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/Settings/FrameSettings.cs
```diff
@@ -321,8 +321,85 @@ public enum FrameSettingsField
     public struct FrameSettingsOverrideMask
     {
         /// <summary>Mask of overridden values.</summary>
+        [Obsolete("This will be removed in future version. Please use IsOverriding/SetOverriding instead.")]
```
JulienIgnace-Unity, line 324: Comment does not match the new API

PR #6432, opened 2020-03-27
Reducing the number of variants for ray tracing when building a player

anisunity:
- Replaced the DIFFUSE_LIGHTING_ONLY multicompile by a uniform.
- Removed the dynamic lightmap multicompile (was un-used, but still induced a lot of shader compilation)
- Removed the LOD cross fade multi compile for ray tracing (was un-used, but still induced a lot of shader compilation)
- Added a ray tracing mode option in the HDRP asset that allows to override and shader stripping.

When in something else than both, volume components only display the current configuration parameters for indirect diffuse and reflection.

image

DXR tests are still green locally.

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
@@ -890,7 +890,6 @@ Shader "HDRP/LayeredLit"
             #define SHADERPASS SHADERPASS_RAYTRACING_INDIRECT
 
             // multi compile that allows us to
-            #pragma multi_compile _ DIFFUSE_LIGHTING_ONLY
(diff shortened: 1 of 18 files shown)


com.unity.render-pipelines.high-definition/Runtime/Material/LayeredLit/LayeredLit.shader
```diff
@@ -890,7 +890,6 @@ Shader "HDRP/LayeredLit"
             #define SHADERPASS SHADERPASS_RAYTRACING_INDIRECT
```
sebastienlagarde, line 890: you also need to move the #pragma multi_compile _ LOD_FADE_CROSSFADE for layered lit to the respective pass

sebastienlagarde:
Need to move CROSS_LOD_FADE for layredlit as well before merging

PR #6436, opened 2020-03-27
[ShaderGraph] Fix: View Direction Node: World Space to be Normalized in URP

zadi0648:
**Summary:**

Fix for https://fogbugz.unity3d.com/f/cases/1217639/

---
**Manually Tested:**

* That the bug no longer reproduces
* That the other spaces seem fine in URP

---

(Always)
**Technical Risk:** 0.1/4
**Halo Effect:** 2/4 - I looked around but didn't see this effecting anything else

---
**Automation:**

* Test should fail, I'll grab those generated shaders and put them in test project (as I can't trust my own home machine to produce usable images)

---
**Yamato:**

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Ffix-view-direction-for-urp


LandonTownsendUnity:
We need to prepare an upgrade guide or something similar; even though this is fixing incorrect behavior with correct behavior, this is going to break a lot of shadergraphs.

phi-lira:
now that this is normalized, do you see any need to have a view vector node?

PR #6437, opened 2020-03-27
[Shader Graph] Fix: Flipbook Node

zadi0648:
**Summary:**

* Fix for https://fogbugz.unity3d.com/f/cases/1219169/ (see automation for before & after!)
* Originally I had looked at a more complex solution involving dynamic solutions, I didn't go that way however I kept an optimization from it, which will counteract the performance hit the fix requires so that no one can complain about this fix (and overall I performance win).
* Many materials for test scene: import time of these materials was less than 1 second on my machine.

---
**Manually Tested:**

Working5x5Flipbook0
Various node set ups, including:
* 5x5, 6x6, 5x7, 7x7, 7x5
* Using time as an input as real world use case would be
* Half & Float precision
* Negative values

---
**Technical Risk:** 3/4 - Only affects this node, but in a big way
**Halo Effect:** 0/4 - Only affects this node

---
**Notes to Documentation:**

* The user no longer has to floor the `Tile` input before putting inserting it into the node. From the documentation, this kind of seems expected/implied anyways, so I didn't change anything. Thoughts?

---
**Notes to QA:**

* If you want to be more sure of this fix & have the time, I would use a real world use case on one of the grid sizes where the bug used to reproduces that isn't mentioned in my manual testing (ex: did it use to reproduce with 11x11? idk, but if it did you could try that. But that's all just an suggestion.)

---
**Automation:**

❌ Before (notice the left edge columns where the values are wrong / repeated):
FlipbookNode
✔️ After:
FlipbookNode expected
FlipbookNode diff

* Since I basically already had this scene set up for my own testing, I decided might as well make it it's own scene! Having a scene for a specific node seemingly might be overkill, but I think we can make some exceptions. I tested the import time of the new materials, and it's less than one second on my machine.
* The old Flipbook test was removed from the UV scene, as it was testing the node incorrectly anyhow.
* The scene test a couple of various things, including: 5x5, 7x7, half precision & full precision, a SG that generates all the different functions.

---
**Yamato:**

https://yamato.prd.cds.internal.unity3d.com/jobs/78-ScriptableRenderPipeline/tree/sg%252Fflipbook-fix


com.unity.shadergraph/Editor/Data/Nodes/UV/FlipbookNode.cs
```diff
@@ -134,15 +148,48 @@ public void GenerateNodeFunction(FunctionRegistry registry, GenerationMode gener
...
-                        s.AppendLine("Tile = fmod(Tile, Width*Height);");
+                        string offset;
+                        if (concretePrecision == ConcretePrecision.Half)
+                        {
+                            offset = "0.0001h";
+                        }
+                        else
+                        {
+                            offset = "0.00001";
+                        }
```
alindmanUnity: not a big deal, just a formatting this -- since these are all one line you can remove the brackets. condenses the code a little -- we do this a lot in the generator but we also have inconsistent formatting so /shrug 

otherwise it looks good to me

LandonTownsendUnity:
BTW I tested 11x11 and it works fine so if you fix the problem with the node preview I will approve.

PR #6443, opened 2020-03-30
 Fix Editor Gizmos/Grid not working.

phi-lira:
### Purpose of this PR
 Fix regression that caused the gizmos and grid to now display correctly. [case 1228544](https://issuetracker.unity3d.com/issues/universal-rp-geometry-does-not-properly-obscure-grid-or-gizmos)

---
### Testing status

**Manual Tests**: Tested with the editor camera. Grid and gizmos work correctly.

**Automated Tests**: I'm not sure we can add automation test to scene view.

---
### Comments to reviewers

- The regression was introduced in this PR: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/5913
- The logic to do copy depth changed, in that PR I changed the CopyDepth pass but didn't know there's a specific pass to copy depth for the scene view at the end of the pipeline. We should possibly merge these two passes, but for now just fixing the issue by doing the same changes I did in CopyDepth.cs to the scene view copy depth pass.

-  Changelog is not required as this issue is not in any public package.

-  Made the ShaderProperty API internal, this API is new in 9.0.0, not introduced in any public package.
 This is due to recent dicussion in slack to don't allow shader public API as we want to be able to move them to cbuffer (more efficiency)

### DIFF
--- a/com.unity.render-pipelines.universal/Runtime/Passes/SceneViewDepthCopy.cs
+++ b/com.unity.render-pipelines.universal/Runtime/Passes/SceneViewDepthCopy.cs
@@ -6,6 +6,8 @@ namespace UnityEngine.Rendering.Universal
 
         Material m_CopyDepthMaterial;
         const string m_ProfilerTag = "Copy Depth for Scene View";
+        int m_ScaleBiasId = Shader.PropertyToID("_ScaleBiasRT");
+
 
         public SceneViewDepthCopyPass(RenderPassEvent evt, Material copyDepthMaterial)
         {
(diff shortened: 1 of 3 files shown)


com.unity.render-pipelines.universal/Runtime/Passes/SceneViewDepthCopy.cs
```diff
@@ -6,6 +6,8 @@ internal class SceneViewDepthCopyPass : ScriptableRenderPass
 
         Material m_CopyDepthMaterial;
         const string m_ProfilerTag = "Copy Depth for Scene View";
+        int m_ScaleBiasId = Shader.PropertyToID("_ScaleBiasRT");
```
ellioman: We seem to be using _ScaleBias and _ScaleBiasRT in our code.
Maybe rename this this variable to `m_ScaleBiasRTId` ?

phi-lira:
HDRP testing failure is unrelated to this PR.

PR #6468, opened 2020-03-31
Hdrp/combine material samples with shader samples

SeanPuller:
### Purpose of this PR
Merge material samples and shader samples so the similar assets are in a single place reducing unneeded sample folders. Align scene visual style and information with material samples to give users more information about how to use the materials/shaders provided.
Adding fabric materials and textures to give users many examples to cover all different types of fabric.

image

- Merged material samples and shader samples
- Update material samples scene visuals
- Added scenes for hair and fabric and decals with material samples
- Added fabric materials and textures
- Added information for fabric materials in fabric scene

- Fix MaterialBalls having same guid issue
- Fix spelling and grammatical errors in material samples


iM0ve:
REVIEW IN PROGRESS (Will update this comment)

1) The package.json file needs to be edited (located in com.unity.render-pipelines.high-definition). To remove the old sample. Not certain if the only reference, clarify with devs.
Screen Shot 2020-03-31 at 17 20 13

PR #6469, opened 2020-03-31
Reduced planar reflection atlas, texture resolutions. Upgraded to 202…

JMargevics:
PR touches 001-HDTemplate runtime test

- Planar reflection reduced from 8k to 1k
- Bust textures reduced from 8k to 1k
- Upgraded to 2020.2.0a5.695


sebastienlagarde:
adding @sambickley for validation. thanks

sambickley:
@JMargevics is the update to 2020.2 strictly necessary? 9.x.x should support 2020.1, and I'm concerned this could cause errors on opening the project in older versions.

sebastienlagarde:
We should be able to do another PR for 2020.1. Our new issue is that 2020.1 and 2020.2 don't have the same material.

PR #6470, opened 2020-03-31
[9.x.x] Obsoleting FrameCleanup(), Adding OnCameraSetup() and OnCameraCleanup()

ellioman:
### Purpose of this PR
**This PR does four things:**
1. Adds ScriptableRenderPass.OnCameraSetup()
2. Makes the ScriptableRenderPass.FrameCleanup() obsolete
3. Adds ScriptableRenderPass.OnCameraCleanup() as its replacement.
4. Updates the render passes URP uses to the new naming

**There are two reasons for this PR:**
**First**, there's an issue in VR with how the render passes are getting temporary render textures and releasing them. Currently they are being created in ScriptableRenderPass.Configure() function and released in ScriptableRenderPass.FrameCleanup(). The problem is that Configure() is called **once per eye for each camera** but FrameCleanup() **once for each camera.** ScriptableRenderPass therefore needed a function that gets called once per camera before calling Configure().

**Second,** the FrameCleanup() is an incorrect name for the function as it is called per camera. So it made sense to include the function name change here as we created OnCameraSetup().

---
### Checklist for PR maker

---

### Testing status

- [ ] Built a player
- Other: 

---
### Comments to reviewers


com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
```diff
@@ -152,6 +152,20 @@ public void ConfigureClear(ClearFlag clearFlag, Color clearColor)
...
+        /// Override this method if you need to to configure render targets and their clear state, and to create temporary render target textures.
+        /// If a render pass doesn't override this method, this render pass renders to the active Camera's render target.
+        /// You should never call CommandBuffer.SetRenderTarget. Instead call <c>ConfigureTarget</c> and <c>ConfigureClear</c>.
+        /// </summary>
+        /// <param name="cmd">CommandBuffer to enqueue rendering commands. This will be executed by the pipeline.</param>
+        /// <param name="cameraTextureDescriptor">Render texture descriptor of the camera render target.</param>
+        /// <param name="renderingData">Current rendering state information</param>
+        /// <seealso cref="ConfigureTarget"/>
+        /// <seealso cref="ConfigureClear"/>
+        public virtual void FrameSetup(CommandBuffer cmd, RenderTextureDescriptor cameraTextureDescriptor, ref RenderingData renderingData)
```
phi-lira: I think in RenderingData.CameraData we already have cameraTargetDescriptor, so cameraTextureDescriptor is redundant if we have renderingData.

com.unity.render-pipelines.universal/CHANGELOG.md
```diff
@@ -40,6 +41,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Renderer Feature list is now redesigned to fit more closely to the Volume Profile UI, this vastly improves UX and reliability of the Renderer Features List.
 - Default color values for Lit and SimpleLit shaders changed to white due to issues with texture based workflows.
 - You can now subclass ForwardRenderer to create a custom renderer based on it.
+- DepthOnlyPass, CopyDepthPass and CopyColorPass now use FrameSetup() instead of Configure() to set up their passes before executing
```
phi-lira: Maybe we should say why is this useful/ what issue we are fixing with this.

com.unity.render-pipelines.universal/Runtime/Passes/ScriptableRenderPass.cs
ellioman: Correct! I have now removed it.

com.unity.render-pipelines.universal/CHANGELOG.md
ellioman: I have now updated the changelog. Please take another look.

PR #6472, opened 2020-03-31
[Fix] case 1227087: Fix issue where LBS node could not be converted to Sub Graph [skip ci]

HedvigAxelsson:
### Purpose of this PR
Make LBS node convert to Sub Graph and added documentation.

---
### Testing status
- [x] Opened test project
- [ ] Built a player


HedvigAxelsson:
I've updated the node description.
Example projects can be found here: https://github.com/Unity-Technologies/Unity.Animation/tree/master/UnityAnimationHDRPExamples

PR #6407, opened 2020-03-27
Optimizing light loop part 2 (Dependent on PR 6367)

FrancescoC-unity:
**First up, this is based on https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6367 branch. Please review the code by diffing with that or after that PR is merged.**

**TL;DR** :  PrepareLightsForGPU  

Note this is different from previous PR in that it focused on the optimization of the case where there are LOTS of shadow casting lights. 

**Profiling methodology**
Using mostly dotTrace (after verifying that results are coherent with Unity Profiler). Accumulating at least 3 minutes worth then normalizing cumulative timings to a common baseline to get win amount.

**Profiling scene**
- 26 point lights of which 13 casting shadows
- 26 spot lights of which 13 casting shadows
- 26 area lights of which 13 casting shadows

Total : 78 lights, 36 casting shadows.

Notes on functions targeted:
When shadows are a considerable amount, the dominating function is Update shadow request 

**Results**

**SHADOW HEAVY SCENE**

|          Function name      |% of cost decrease vs previous PR                          |% of cost decrease vs previous staging                         |
|----------------|-------------------------------|-----------------------------|
|PrepareLightsForGPU (parent function)|    **20%**    |**35%**            |
|PrepareGPULightdata          |**30%**|**43%**            |
|UpdateShadowRequest          |**47%**|**49%** (twice as fast)            |

**NO SHADOW SCENE**

|          Function name      |% of cost decrease vs previous PR                          |% of cost decrease vs previous staging                         |
|----------------|-------------------------------|-----------------------------|
|PrepareLightsForGPU (parent function)|    **14%**    |**68%** (3x faster)            |
|PrepareGPULightdata          |**17%**|**78.5%** (more than 4x faster)            |

PR #6477, opened 2020-03-31
Fix shadow tesselation culling

alelievr:
### Purpose of this PR
Fix the tesselation triangle culling in shadows. Before this fix, the culling planes we sent to the shader to cull the triangles were wrong so no triangles were culled. Now i'm sending the good culling planes for shadows and we should see a slight performance improvement.

I opened the HDRP graphic test scene for tesselation and moved around with the camera. At first there was an issue with the culling planes (they were not in camera relative space) and thus there was visible artifacts in the shadows. 
After i fixed this issue it worked as intended.

You can check out the performance gain in the shadow tessellation test scene on the HDRP performance dahsboard (in comparison mode): https://grafana.internal.unity3d.com/d/isPoRI9Wz/hdrp-branch-compare?orgId=1&from=now-1y&to=now&var-Git_branch_base=master&var-Git_branch_compare=test-auto-perf&var-HDRP_asset_config=Deferred&var-HDRP_asset_config=Forward&var-Selected_counter=All&var-ps4_config=Base&var-base_run_id=27999bd1-e02a-41f9-afb0-76dfeb0ccc28&var-compare_run_id=4a39378d-897b-4396-b9db-1c76299ec8a0&var-counter_test_scenes=LitCube&var-counter_test_scenes=1000_LitCube&var-counter_test_scenes=TesselationShadow


com.unity.render-pipelines.high-definition/Runtime/Lighting/Light/HDAdditionalLightData.cs
```diff
@@ -1964,7 +1964,7 @@ void SetCommonShadowRequestSettings(HDShadowRequest shadowRequest, VisibleLight
             }
 
             // shadow clip planes (used for tessellation clipping)
-            GeometryUtility.CalculateFrustumPlanes(viewProjection, m_ShadowFrustumPlanes);
+            GeometryUtility.CalculateFrustumPlanes(shadowRequest.deviceProjectionYFlip * shadowRequest.view, m_ShadowFrustumPlanes);
```
FrancescoC-unity: Sorry just passing by here and heads up that matrix mul costs a _lot_ . I changed this recently so that it doesn't use the actual matrix full matrix mul here, but I guess it hasn't landed yet. This is the PR in case that lands before this and therefore you'll have conflicts on this: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6407

sebastienlagarde:
optimization on matrix can be done in a separate PR.
Merging without QA as the test currently cover it - it is not a fix in the sense that it was displaying artefact, it is to fix performance issue.

PR #6478, opened 2020-03-31
[8.x.x][Shader Graph] Backport Deleted Asset Dialogue

alindmanUnity:
### Purpose of this PR
Backport #5313 to 8.x.x to resolve a priority 2 regression in the 8.x package [1232246](https://issuetracker.unity3d.com/product/unity/issues/guid/1232246/) [internal link](https://fogbugz.unity3d.com/f/cases/1232246/)

From original PR: 
- Stops users from entering in a still open Shader Graph that is in a buggy soft lock mode; they either ought to close the Shader Graph or last chance to Save As.
- Works with all graphs, allowing users a last chance Save As if they regret the deletion of the Shader Graph
- This popup does not happen until the window has been selected.
- Note that "Save As" is the secondary, not the primary, option. Please feel free to give feedback on that. Although that's abnormal as far as saving stuff goes, I did that because I figured that primary should be the default use case, after all the user will have deleted the file on purpose 96% of the time.

---
### Testing status

- [x] Built a player
Manually Tested (repeated manual testing from original PR):

- Delete -> Don't Save (Close), Delete -> Save As, Delete -> Save As the same file, many times and ways
- Deleting inside of Unity and outside of Unity via File Explorer
- Shader Graphs and Sub Graphs
- Various window / tab arrangements. The popup (should) only popup once the tab or Shader Graph window has been selected, as opposed to immediately when the file has been deleted.

---
### Comments to reviewers
Should behave the same as master.

### DIFF
--- a/com.unity.shadergraph/CHANGELOG.md
+++ b/com.unity.shadergraph/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added a field to the Master Nodes that overrides the generated shader's ShaderGUI.
+- When a Shader Graph or Sub Graph Asset associated with a open window has been deleted, Unity now displays a dialog that asks whether you would like to save the graph as a new Asset or close the window.
 
 ### Fixed
 - Fixed a bug where any change to the PBR master node settings would lose connection to the normal slot.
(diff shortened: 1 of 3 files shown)


LandonTownsendUnity:
Due to changes that made it into 9.x.x but not 8.x.x, the "save as" option defaults to the correct project folder on the original PR, but often defaults to the wrong project folder on this backport. Do we need to do some extra work to make sure that doesn't happen?

alindmanUnity:
> Due to changes that made it into 9.x.x but not 8.x.x, the "save as" option defaults to the correct project folder on the original PR, but often defaults to the wrong project folder on this backport. Do we need to do some extra work to make sure that doesn't happen?

@LandonTownsendUnity can you confirm correct save as locations on 0b7b6ef6f7

PR #6488, opened 2020-04-01
World space UI element reflections

pmavridis:
### Purpose of this PR
World Space UI was not rendered in realtime reflections
https://fogbugz.unity3d.com/f/cases/1155022/

Before:
image

After:
image

And here is the test project:
https://drive.google.com/open?id=1Irjd0WGRoii3VAOnlGEJb62ZCBDmD3zS
(slightly modified from the fogbugz, since we have changed the sky system)


pmavridis:
I'm closing this PR because it does not properly resolve the issue when using stand-alone player. 
To properly resolve the issue we would like is to emit the Canvas UI geometry for non-active cameras and this might require some changes on the C++ side.

PR #6489, opened 2020-04-01
Fixed usage of light size data that are not available at runtime.

anisunity:
Regression introduced by this PR when building a player https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6454


sebastienlagarde:
Merging, Yazmato test will cover it (for QA)

PR #6496, opened 2020-04-01
Expose a debug constant in ShaderLab to make HDRP/Unlit SRP Batcher compatible again.

JussiKnuuttila:
Shaders with UnityPerMaterial constants that are not exposed in ShaderLab are not compatible with the SRP batcher. HDRP/Unlit had one of these in its debug display variant. Expose the constant so the debug variant is also SRP Batcher compatible.

### Checklist for PR maker
- [ ] Have you added a backport label (if needed)? For example, the `need-backport-*` label. After you backport the PR, the label changes to `backported-*`.

---
### Purpose of this PR
Without this PR, Hybrid Renderer V2 cannot render objects that use HDRP/Unlit in the standalone player. Other uses of HDRP/Unlit would still work, but might have worse performance because a slower code path is used compared to the SRP Batcher fast path.

---
### Testing status

- [x] Built a player
- Other: Manually tested using the Hybrid Renderer HDRP test project

---
### Comments to reviewers


com.unity.render-pipelines.high-definition/Runtime/Material/Unlit/Unlit.shader
```diff
@@ -418,7 +422,7 @@ Shader "HDRP/Unlit"
             //enable GPU instancing support
             #pragma multi_compile_instancing
             #pragma multi_compile _ DOTS_INSTANCING_ON
-            
+
```
JussiKnuuttila, line 425: Looks like I had some unintentional whitespace changes because I had whitespace diffing disabled.

sebastienlagarde: fine, no worry

PR #6500, opened 2020-04-01
Fix error in TransformWorldToHClipDir

FrancescoC-unity:
Found this accidentally while compiling unrelated shaders  :-)  A typo, but could potentially lead to proper bugs. 

Note this is in master too, shall I port there too?


FrancescoC-unity:
So this is based on staging branch, I also have opened for master: https://github.com/Unity-Technologies/ScriptableRenderPipeline/pull/6503 

Having both PR to keep also staging working before we merge master into it. (Since I already have this open no point in closing this and porting the one targeting master)

PR #6508, opened 2020-04-01
Hdrp/fix issue with view dependent probes

FrancescoC-unity:
Another interesting one. Fix for: https://fogbugz.unity3d.com/f/cases/1229225/ 

This is a workaround for a design issue more than not.  

The problem is that for planar reflections, we compute the matrices based on the main view. However, the matrices data that will be referenced for rendering later on is set during the execute rendering of the probe and never again. So what happend with two cameras (say A and B) and a planar probe

Planar is rendered for A 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Planar data is set to relative to A
Planar is rendered for B
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Planar data is set to relative to B

Then 

Render Camera A     <<< !!! Planar will have data relative to B! 
Render Camera B

So this PR fixes that issue by storing the data for the probe and setup the probe to the correct state right before rendering the "main" cameras.  

**FOR QA**: I think this needs a bit of testing, The bug itself is fixed, but I think is worth playing a bit with planar probes.

### DIFF
--- a/com.unity.render-pipelines.high-definition/CHANGELOG.md
+++ b/com.unity.render-pipelines.high-definition/CHANGELOG.md
@@ -500,6 +500,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed shadow cascade tooltip when using the metric mode (case 1229232)
 - Fixed how the area light influence volume is computed to match rasterization.
 - Focus on Decal uses the extends of the projectors
+- Fix issue causing wrong planar reflection rendering when more than one camera is present.
 
 ### Changed
 - Color buffer pyramid is not allocated anymore if neither refraction nor distortion are enabled
--- a/com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
@@ -248,7 +248,7 @@ namespace UnityEngine.Rendering.HighDefinition
         /// </summary>
         public RenderData renderData => GetRenderData(mode);
         /// <summary>
-        /// Get the render data of a specific mode
+        /// Get the render data of a specific mode. Note that for real time view dependent probes, the render data is going to be specific to the parent camera.
         /// </summary>
         /// <param name="targetMode">The mode to query</param>
         /// <returns>The requested render data</returns>
@@ -264,7 +264,7 @@ namespace UnityEngine.Rendering.HighDefinition
             }
         }
         /// <summary>
-        /// Set the render data for a specific mode
+        /// Set the render data for a specific mode. Note that in case of real time view dependent probes, this needs to be set with the correct data once per camera.
         /// </summary>
         /// <param name="targetMode">The mode to update</param>
         /// <param name="renderData">The data to set</param>
--- a/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
@@ -1196,6 +1196,7 @@ namespace UnityEngine.Rendering.HighDefinition
             // Indices of render request to render before this one
             public List<int> dependsOnRenderRequestIndices;
             public CameraSettings cameraSettings;
+            public List<(HDProbe.RenderData, HDProbe)> viewDependentProbesData;
         }
         struct HDCullingResults
         {
@@ -1415,7 +1416,8 @@ namespace UnityEngine.Rendering.HighDefinition
                         },
                         dependsOnRenderRequestIndices = ListPool<int>.Get(),
                         index = renderRequests.Count,
-                        cameraSettings = CameraSettings.From(hdCamera)
+                        cameraSettings = CameraSettings.From(hdCamera),
+                        viewDependentProbesData = ListPool<(HDProbe.RenderData, HDProbe)>.Get()
                         // TODO: store DecalCullResult
                     };
                     renderRequests.Add(request);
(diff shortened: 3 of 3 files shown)


com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1499,15 +1501,25 @@ float ComputeVisibility(int visibleInIndex, HDProbe visibleProbe)
 
                             parentCamera = visibleInRenderRequest.hdCamera.camera;
 
+                            var renderDatas = ListPool<HDProbe.RenderData>.Get();
```
fredericv-unity3d, line 1504: Can you use instead the syntax with `using`? It is best to use whenever possible. (It releases the pool if an exception occurs, for instance).

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1522,7 +1534,11 @@ float ComputeVisibility(int visibleInIndex, HDProbe visibleProbe)
                                 visibleInOneViewer = true;
                         }
                         if (visibleInOneViewer)
-                            AddHDProbeRenderRequests(visibleProbe, null, visibilities, 0, parentCamera);
+                        {
+                            var renderDatas = ListPool<HDProbe.RenderData>.Get();
```
fredericv-unity3d, line 1538: Can you use `using` as well here?

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
```diff
@@ -1785,6 +1807,13 @@ ref _cullingResults
                                 target.id = m_TemporaryTargetForCubemaps;
                             }
 
+                            // We need to set the appropriate render data for view dependent probes.
```
fredericv-unity3d, line 1810: Please can you elaborate the comment?
Something like
```cs
// case 1229225
// Currently the HDProbe API store only one RenderData per probe, so when a view dependent probe has multiple viewers
// Only the render data of the last viewer will be used for all the rendering.
// To prevent this, we have stored the RenderData for each viewer and we set the appropriate one right before rendering here.
```

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -248,7 +248,7 @@ public Texture SetTexture(ProbeSettings.Mode targetMode, Texture texture)
         /// </summary>
         public RenderData renderData => GetRenderData(mode);
         /// <summary>
-        /// Get the render data of a specific mode
+        /// Get the render data of a specific mode. Note that for real time view dependent probes, the render data is going to be specific to the parent camera.
```
fredericv-unity3d, line 251: I don't think that describe well the issue you ran into. What do you think of:
```cs
/// Get the render data of a specific mode.
///
/// Note: The HDProbe stores only one RenderData per mode, even for view dependent probes with multiple viewers.
/// In that case, make sure that you have set the RenderData of the expected viewer before rendering.
```

com.unity.render-pipelines.high-definition/Runtime/Lighting/Reflection/HDProbe.cs
```diff
@@ -264,7 +264,7 @@ public RenderData GetRenderData(ProbeSettings.Mode targetMode)
             }
         }
         /// <summary>
-        /// Set the render data for a specific mode
+        /// Set the render data for a specific mode. Note that in case of real time view dependent probes, this needs to be set with the correct data once per camera.
```
fredericv-unity3d, line 267: Same idea:
```cs
/// Set the render data of a specific mode.
///
/// Note: The HDProbe stores only one RenderData per mode, even for view dependent probes with multiple viewers.
/// In that case, make sure that you have set the RenderData of the expected viewer before rendering.
```

com.unity.render-pipelines.high-definition/CHANGELOG.md
```diff
@@ -500,6 +500,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed shadow cascade tooltip when using the metric mode (case 1229232)
 - Fixed how the area light influence volume is computed to match rasterization.
 - Focus on Decal uses the extends of the projectors
+- Fix issue causing wrong planar reflection rendering when more than one camera is present.
```
fredericv-unity3d, line 503: Insert the bug case in the changelog please

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: See comment in the file few lines above

`` // This syntax is awful and hostile to debugging, please don't use it.. ``.

com.unity.render-pipelines.high-definition/Runtime/RenderPipeline/HDRenderPipeline.cs
FrancescoC-unity: See comment in the file few lines above

`` // This syntax is awful and hostile to debugging, please don't use it.. ``.

com.unity.render-pipelines.high-definition/CHANGELOG.md
FrancescoC-unity: The bug is really a small subset of the possible issues that can happen without this fix, I don't think we should reference a specific case here as the fix is wider.

fredericv-unity3d:
@sebastienlagarde The HDProbe API does not support RenderData per view by design.
In this case of view dependent probes with multiple viewers the API is error prone.
We probably should update the API to make it less error prone. (So we can track this as a technical debt task)

fredericv-unity3d:
For QA: only the realtime probe code was changed, so I suggest to test setups with:
* [0;n1] reflection probes
* [0;n2] planar probes
* [0;n3] cameras viewing the probes
[1;n] reflection probe

PR #6519, opened 2020-04-02
Fix MSAA resolve when there is no motion vectors

adrien-de-tocqueville:
https://fogbugz.unity3d.com/f/cases/1230778/

### DIFF
--- a/com.unity.render-pipelines.high-definition/Runtime/Material/SharedRTManager.cs
+++ b/com.unity.render-pipelines.high-definition/Runtime/Material/SharedRTManager.cs
@@ -94,6 +94,8 @@ namespace UnityEngine.Rendering.HighDefinition
                 // Create the required resolve materials
                 m_DepthResolveMaterial = CoreUtils.CreateEngineMaterial(resources.shaders.depthValuesPS);
                 m_ColorResolveMaterial = CoreUtils.CreateEngineMaterial(resources.shaders.colorResolvePS);
+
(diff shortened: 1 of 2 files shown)


com.unity.render-pipelines.high-definition/Runtime/Material/SharedRTManager.cs
```diff
@@ -342,19 +344,36 @@ public void ResolveSharedRT(CommandBuffer cmd, HDCamera hdCamera)
...
+                        m_RTIDs2[0] = m_CameraDepthValuesBuffer.nameID;
+                        m_RTIDs2[1] = m_NormalRT.nameID;
+                        CoreUtils.SetRenderTarget(cmd, m_RTIDs2, m_CameraDepthStencilBuffer);
+
+                        // Set the input textures
+                        Shader.SetGlobalTexture(HDShaderIDs._NormalTextureMS, m_NormalMSAART);
+                        Shader.SetGlobalTexture(HDShaderIDs._DepthTextureMS, m_DepthAsColorMSAART);
+
+                        // Resolve the depth and normal buffers
+                        cmd.DrawProcedural(Matrix4x4.identity, m_DepthResolveMaterial, SampleCountToPassIndex(m_MSAASamples), MeshTopology.Triangles, 3, 1);
```
FrancescoC-unity, line 375: This can be outside the if/else and so can the following two lines 

``                        Shader.SetGlobalTexture(HDShaderIDs._NormalTextureMS, m_NormalMSAART);
                        Shader.SetGlobalTexture(HDShaderIDs._DepthTextureMS, m_DepthAsColorMSAART);``

Also the comment need updating.

